PR #138, opened 2026-06-12
sec: reject unsupported argumentSchema keywords at load time

### LINKED ISSUE #123: fix: argumentSchema silently ignores unknown keywords instead of rejecting at load (SPEC § 3.2.2)
## Problem

SPEC § 3.2.2 defines `argumentSchema` as a closed JSON-Schema subset: a compliant proxy MUST reject at manifest-load time any schema using a keyword outside the supported set (`$ref`, applicators, `const`, `format`, exclusive bounds, …). The implementation instead treats `argumentSchema` contents as opaque:

- `checkManifestKeys` deliberately skips it (`cmd/eunox-mcp/manifest.go` — "argumentSchema follows JSON Schema (a recursive subset) and is treated as opaque; its contents are validated by the enforcement engine, not here"), and
- the typed decode into `capability.ArgumentSchema` (`pkg/capability/constraint.go:125`) silently drops any key the struct doesn't model; the runtime validator (`pkg/enforcement/schema.go`) only evaluates the keywords it knows.

So a manifest author who writes `const`, `exclusiveMinimum`, `allOf`, or `$ref` gets **no error and no enforcement** — the constraint they believe is active is silently absent. That is fail-open relative to author intent, which is exactly the failure mode the unknown-*key* check elsewhere in the manifest exists to prevent (same rationale as the `arguments`-vs-`argument` typo guard).

## Expected (SPEC § 3.2.2, as reconciled in eunolabs/mcp-capability-manifest#3)

Supported keyword set (everything else rejected at load, recursively through `properties` / `items`):

`type` (string or array of strings), `enum`, `properties`, `required`, `additionalProperties`, `items`, `minItems`, `maxItems`, `minLength`, `maxLength`, `pattern`, `minimum`, `maximum`, plus the `description` annotation.

## Also

`ArgumentSchema.Strict` (`pkg/capability/constraint.go:140`) is parsed but never read by the validator — decide whether to implement it or remove the field; as-is it is another silently-ignored author intent.

## Acceptance criteria

- [ ] A recursive load-time keyword check over `argumentSchema` (top level, nested `properties`, nested `items`), rejecting unknown keywords with a path-qualified error (e.g. `capabilities[2].argumentSchema.properties.path: unknown keyword "format"`)
- [ ] Explicit reject cases covered by tests: `$ref`, `allOf`, `const`, `format`, `exclusiveMinimum`, plus an arbitrary typo'd keyword
- [ ] `Strict` either implemented (documented semantics) or removed
- [ ] Table-driven tests per CONTRIBUTING (accept + reject + malformed)
- [ ] `docs/capability-manifest-guide.md` § 1 footnote ("the one exception is argumentSchema") updated once the exception no longer exists

eunolabs:
## Summary

Add recursive validation of `argumentSchema` keywords in capability manifests to reject unsupported JSON-Schema keywords (e.g. `const`, `$ref`, `allOf`, `exclusiveMinimum`, `format`) at load time rather than silently dropping them during typed decode. This prevents a manifest author from unknowingly shipping a weaker policy than intended.

## Why

The `argumentSchema` field accepts a closed subset of JSON-Schema keywords (SPEC § 3.2.2): `type`, `enum`, `properties`, `required`, `additionalProperties`, `items`, `minItems`, `maxItems`, `minLength`, `maxLength`, `pattern`, `minimum`, `maximum`, and `description`. 

Previously, if an author wrote an unsupported keyword like `format: uri` or `const: "frozen"`, the typed struct decode would silently drop it, and the constraint would never be enforced at runtime. This is a fail-open security gap: the policy silently becomes weaker than the author wrote.

This PR adds `checkArgumentSchemaKeywords()` to walk the `argumentSchema` tree recursively (through `properties` and `items`) and reject the first unknown keyword with a path-qualified error message. Unknown keywords now fail at manifest load, with a "did you mean" hint for typos (e.g. `proprties` → `properties`).

## Type of change

- Security hardening

## How it was tested

Added `TestLoadManifest_ArgumentSchemaKeywords` with 8 test cases covering:
- Valid manifest with full supported keyword set nested in properties and items
- Rejection of `const`, `$ref`, `allOf` at top level
- Rejection of `format` nested in properties
- Rejection of `exclusiveMinimum` nested in items
- Typo detection with "did you mean" hint
- Rejection of legacy `strict` keyword

All tests pass; `make test` and `make lint` clean.

## Security considerations

**Fail-closed preserved:** ✓ Unsupported keywords now cause manifest load to fail with a clear error, rather than silently dropping the constraint and leaving it unenforced at runtime.

**Policy enforcement:** ✓ This closes a gap where a manifest author's intent (e.g. "only allow URIs matching this format") could be silently ignored. The policy now matches what the author wrote, or fails to load.

**No audit impact:** The validation happens at load time; audit records are unaffected.

## Checklist

- [x] Docs updated (`docs/capability-manifest-guide.md`) explaining the closed keyword set and rejection behaviour
- [x] Comments updated in `manifest.go` and test file to clarify the change

https://claude.ai/code/session_019Scry3N1WxZYNHdA6C9ga4

Closes #123


eunolabs:
Re: the open question on removing exported `ArgumentSchema.Strict` —

The removal was intentional, and I believe it's the right call here, but flagging the trade-off so it's a deliberate decision rather than an implicit one:

- **The issue asked for exactly this.** Issue #123's acceptance criteria included "`Strict` either implemented (documented semantics) or removed." The field was parsed into the struct but never read by the validator, so it enforced nothing — a silently-ignored author intent, the same class of bug the rest of the PR fixes. Implementing it would mean designing new closed-set semantics; removing it is the cleaner resolution.

- **Project policy favors a clean break pre-1.0.** Per `CONTRIBUTING.md` / `CLAUDE.md`: "Pre-1.0: no backwards-compatibility shims. Change manifest grammar, audit shape, or config keys cleanly and document the migration." I've now documented the removal under a **Removed** heading in the CHANGELOG (ecc5da1).

- **Downstream Go impact is real but narrow.** You're correct that manifest-level rejection works whether or not the field exists — keeping the field would avoid breaking any external struct-literal that sets `Strict:`. But keeping a parsed-but-unenforced field is precisely the fail-open-to-author-intent footgun the issue called out: a downstream author setting `Strict: &true` would believe a constraint is active when none is. Removing it makes that a compile error instead of a silent no-op.

If you'd prefer to keep the field for API stability and only reject the YAML keyword at load (so external Go callers don't break), that's a one-line revert of the struct change — happy to do that instead. Let me know which way you want it.

---

PR #156, opened 2026-06-12
fix: apply upstream timeout to HTTP list handlers

### LINKED ISSUE #146: bug: RATE_LIMITED maps to CAPABILITY_DENIED JSON-RPC code
## Problem

`maxCalls` denials emit symbolic code `RATE_LIMITED`:

- https://github.com/eunolabs/eunox/blob/main/pkg/enforcement/handlers.go#L270-L274

The threat model says this should be returned as JSON-RPC `-32003`:

- https://github.com/eunolabs/eunox/blob/main/docs/threat-model-mcp.md#L269

But `denialToJSONRPCCode` does not handle `ErrCodeRateLimited`, so it falls through to the default `CAPABILITY_DENIED` / `-32002` mapping:

- https://github.com/eunolabs/eunox/blob/main/cmd/eunox-mcp/jsonrpc.go#L166-L176

The existing mapping tests do not include `RATE_LIMITED`:

- https://github.com/eunolabs/eunox/blob/main/cmd/eunox-mcp/jsonrpc_error_codes_test.go#L32-L55

## Impact

Clients and operators see the wrong wire code for rate limits. This is externally visible and contradicts the documented API contract.

## Expected

`RATE_LIMITED` should map to the same integer code family documented for condition failures, currently `-32003`.

## Suggested fix

Add an explicit `capability.ErrCodeRateLimited` case in `denialToJSONRPCCode` and extend the JSON-RPC mapping tests to cover it.

### LINKED ISSUE #147: bug: HTTP list handlers bypass upstream timeout
## Problem

`HTTPProxy.withUpstreamTimeout` says it should be applied to every upstream call:

- https://github.com/eunolabs/eunox/blob/main/cmd/eunox-mcp/http.go#L517-L520

But the HTTP list handlers call upstream directly with the incoming context:

- `tools/list`: https://github.com/eunolabs/eunox/blob/main/cmd/eunox-mcp/http.go#L673-L674
- `resources/list`: https://github.com/eunolabs/eunox/blob/main/cmd/eunox-mcp/http.go#L690-L691
- `prompts/list`: https://github.com/eunolabs/eunox/blob/main/cmd/eunox-mcp/http.go#L708-L709

`tools/call` and other high-risk calls apply the timeout; these list surfaces do not.

## Impact

A hung or slow upstream can hang HTTP list requests despite `--upstream-timeout-ms` being configured. This is a release-quality reliability gap and a trivial DoS vector against discovery/listing endpoints.

## Expected

All HTTP upstream calls should use `withUpstreamTimeout`, including list operations.

## Suggested fix

Wrap these handlers' `sess.callUpstream` calls with:

```go
upCtx, cancel := p.withUpstreamTimeout(ctx)
defer cancel()
upResp, err := sess.callUpstream(upCtx, msg)
```

Add tests with a stuck upstream for `tools/list`, `resources/list`, and `prompts/list` verifying timeout behavior.

eunolabs:
Fixes #147

## Problem

`HTTPProxy.withUpstreamTimeout` is documented as applying to every `callUpstream` invocation, but the HTTP `tools/list`, `resources/list`, and `prompts/list` handlers forwarded with the raw request context. A hung upstream could stall list requests indefinitely even with `--upstream-timeout` configured — a trivial DoS vector against the discovery endpoints.

## Fix

Wrap the three list handlers' `callUpstream` calls with `withUpstreamTimeout`, the same pattern already used by `tools/call`, `resources/read`, `prompts/get`, `resources/subscribe`, and all stdio handlers. When the timeout is unset, `withUpstreamTimeout` is a no-op, so behaviour without the flag is unchanged.

## Tests

`TestRemoteUpstream_ListHandlersApplyUpstreamTimeout` runs each list method through a proxy backed by a stuck upstream (blocks until the proxy aborts the request) with a 100 ms upstream timeout, and asserts the host receives a bounded JSON-RPC upstream error. Verified the test fails against the pre-fix code and passes with the fix; full `go test -race ./...`, `make lint`, and `make check-license` pass.

Closes #146


eunolabs:
Pushed a follow-up commit addressing all 15 findings from the second review round:

**Structural (timeout enforcement):**
- `--upstream-timeout` is now applied once inside the chokepoints (`callSubprocessUpstream`, `callRemoteUpstream`, `StdioProxy.callUpstream`) instead of per-handler — notification forwarding and the HTTP drift check are now bounded, and future handlers cannot forget the wrap.
- The local-mode initialize handshake now honors the session-start deadline: a wedged subprocess is killed and reaped instead of pinning the handler goroutine and leaking an unkillable orphan.

**Audit consistency:**
- `upstreamErrInfo` gains `REQUEST_CANCELED` (host disconnects no longer blame the upstream) and a guard against reporting a misleading duration for inherited deadlines.
- Every upstream-failure path on both transports now classifies + records: HTTP read/subscribe/get (previously hardcoded `UPSTREAM_ERROR`), and all stdio handlers (previously unclassified and unaudited).
- Infrastructure failures are never marked `audit_only` — the error the host receives is enforced even in audit mode.
- `suggest` skips infrastructure denial codes, so transport noise can't surface as draft manifest entries (`prompts/list` no longer misparses as a prompt named `list`).

**Cleanup:** the three identical list handlers per transport are collapsed into one helper parameterised by `ListFilterer` method expression.

**Tests:** audit-record assertions on the timeout test, plus new regression tests for notification bounding, the initialize deadline, error classification, and the stdio audit record. Threat model 5.5/6.1 updated to document the infrastructure codes and the new record shapes.

Full `go test -race ./...`, `make lint`, and `make check-license` pass.

PR #157, opened 2026-06-12
fix: allow manifest sampling opt-in through JWT-wrapped PDPs

### LINKED ISSUE #145: bug: JWT mode breaks HTTP sampling opt-in
## Problem

In HTTP gateway mode, enabling JWT wrapping makes `system:sampling/createMessage` impossible to allow.

`wrapRoutesWithJWT` replaces each route PDP with `*JWTPDP`:

- https://github.com/eunolabs/eunox/blob/main/cmd/eunox-mcp/route.go#L207-L218

But `handleHTTPUpstreamRequest` only allows server-initiated sampling when the route PDP type-asserts directly to `*ManifestPDP`:

- https://github.com/eunolabs/eunox/blob/main/cmd/eunox-mcp/http.go#L843-L846

After JWT wrapping, that assertion fails, so `allowed` stays false and `sampling/createMessage` is denied even when the manifest contains `system:sampling/createMessage` with `allow` or `*`.

## Impact

Operators can configure sampling correctly and still get denied as soon as `--jwks-uri` is enabled. This contradicts the manifest docs and makes JWT + local HTTP upstream sampling unusable.

## Expected

Sampling authorization should work through wrapper PDPs. The JWT/manifest intersection should preserve the manifest's explicit sampling opt-in behavior.

## Suggested fix

Expose sampling authorization through an interface instead of asserting `*ManifestPDP` directly, or add a method on `JWTPDP` that delegates/intersects appropriately with the inner PDP.

Add an HTTP gateway test covering:

- manifest includes `system:sampling/createMessage`
- route is wrapped with JWT
- upstream emits `sampling/createMessage`
- request is broadcast/allowed rather than denied

eunolabs:
Fixes #145.

## Problem

In HTTP gateway mode, enabling `--jwks-uri` wraps every route PDP in a `JWTPDP` (`wrapRoutesWithJWT`), but `handleHTTPUpstreamRequest` only allowed server-initiated sampling when the route PDP type-asserted directly to `*ManifestPDP`. After JWT wrapping the assertion failed, so `sampling/createMessage` was denied even when the manifest contained `system:sampling/createMessage` with `allow` or `*`.

## Fix

- New `SamplingAuthorizer` interface in `pdp.go` (`SamplingAllowed() bool`); the HTTP and stdio handlers assert it instead of the concrete `*ManifestPDP`.
- `JWTPDP.SamplingAllowed()` delegates to its inner PDP. Sampling is initiated by the upstream server, not by an HTTP client, so no bearer token is in scope for the decision — the JWT layer can neither grant nor restrict it, and the manifest's explicit opt-in stays authoritative.
- Fail-closed posture preserved: JWT-only mode (no inner PDP) and unpoliced routes (`alwaysAllowPDP` inner, which does not implement the interface) keep denying sampling.

## Tests

- Gateway handler tests (`transport_unit_test.go`): a route built from a manifest with `system:sampling/createMessage` and wrapped via the real `wrapRoutesWithJWT` wiring broadcasts upstream-emitted `sampling/createMessage`; a JWT-wrapped manifest *without* the opt-in still denies (no widening).
- `JWTPDP.SamplingAllowed` unit tests (`pdp_jwt_test.go`): delegates to inner manifest opt-in; denies with no inner, an inner without the entry, and an `alwaysAllowPDP` inner.

## Docs

`docs/capability-manifest-guide.md` § 2b now states that JWT mode does not change the sampling decision: the manifest `system:` opt-in alone governs, and the exhaustive-allowlist rule (§ 5.2) does not apply to server-initiated sampling.

`make test`, `make lint`, and `make check-license` all pass.

Closes #145


eunolabs:
Addressed the P2 in 7f04dfa by documenting the carve-out where the intersection rule is stated, rather than changing the implementation:

- **README.md** (JWT section) and **docs/architecture.md** (JWTPDP table row): one-line carve-out noting server-initiated sampling is governed solely by the manifest `system:` opt-in.
- **ADR-0001**: new "Scope: server-initiated sampling is outside the intersection" subsection under Decision, plus a note on the final consequence bullet clarifying the carve-out is not an OR path (it never widens the manifest) and that any future exception needs the same justification — no token attributable to the request.
- **capability-manifest-guide.md § 5.2**: explicit note that a present-but-empty `mcp.capabilities` claim does not block manifest-allowed sampling.

Why not evaluate stored per-session/subscriber JWT context: in HTTP mode the upstream-initiated request is broadcast to every SSE subscriber, so there is no single token to intersect — picking one (or requiring all subscribers' tokens to allow it) is a real design decision with its own threat-model implications, and issue #145 explicitly scoped the expected behavior to "preserve the manifest's explicit sampling opt-in". If per-subscriber sampling scoping is wanted, I'd suggest a follow-up issue/ADR rather than folding it in here.

The manifest-as-upper-bound invariant is unchanged: no token can enable sampling the manifest withholds.

eunolabs:
Addressed the P1 in 62f2e1b. The kill-switch gap was actually wider than JWT mode — `SamplingAllowed()` was a pure manifest lookup, so the sampling path never consulted kill state in plain manifest mode (HTTP or stdio) either. Fixed at the PDP layer so every transport gets it:

- `SamplingAuthorizer` now has `DecideSampling(ctx, sessionID) capability.EnforceResponse` instead of `SamplingAllowed() bool`. Both implementations check the kill switch first and fail closed on a kill-store error (`KILL_SWITCH_ERROR`), mirroring `Decide`: `ManifestPDP.DecideSampling` then applies the manifest opt-in; `JWTPDP.DecideSampling` checks its own kill switch, then delegates the capability decision to the inner PDP as before.
- Audit records for sampling denials now carry the real denial code (`KILL_SWITCH` / `KILL_SWITCH_ERROR` / `SAMPLING_DENIED`) instead of hardcoded `SAMPLING_DENIED`.
- The kill-store lookup needs a context, so the upstream reader chain threads one: stdio passes its `Start` context; the HTTP proxy captures its `Serve` context (`baseCtx`) and hands that to session readers — deliberately not the request context, which is cancelled when the initialize request completes while the session lives on (a Redis-backed kill check would otherwise start failing closed mid-session).

Regression tests: killed session, global kill, and kill-store error through `ManifestPDP.DecideSampling` (pdp_manifest_test.go), through the JWT wrapper (pdp_jwt_test.go), and end-to-end through the HTTP gateway handler with the real `wrapRoutesWithJWT` wiring (transport_unit_test.go) — all asserting deny to the upstream and no broadcast. Docs updated (ADR-0001 scope section, architecture.md, guide § 2b kill-switch note).

`make test`, `make lint`, `make check-license` all pass.

eunolabs:
Pushed 5c82273 addressing all eight findings from the extra-high-effort review pass:

**Correctness**
1. *Per-agent kill gap*: the HTTP layer now stamps each session with the `mcp.agent_id` of its most recently validated bearer token (`httpSession.setAgentID`, updated on session creation and every authenticated POST); the upstream-initiated sampling handler attributes that identity to the kill check, so `KillAgent` blocks sampling. Regression tests at the JWTPDP level (`TestJWTPDP_DecideSampling_KilledAgent_Denied`) and through the HTTP handler (`TestHTTPHandleUpstreamRequest_SamplingKilledAgent_Denied`), plus a stamping test through `handleMCPPost`.
2. *Audit-mode sampling denials invisible*: the HTTP handler now records the observed deny (real code, `auditOnly=true`) before broadcasting, mirroring stdio. Verified end-to-end against a real audit sink (`TestHTTPHandleUpstreamRequest_SamplingAuditMode_RecordsDeny`).

**Cleanup**
3. `killCheck(ctx, ks, sessionID)` replaces all six ShouldBlock-then-deny copies. Side effect worth noting for reviewers: `KILL_SWITCH_ERROR` denials now uniformly carry `conditionType: "kill"` (three pre-existing copies omitted it) — pre-1.0, no shim.
4. `decideSampling(ctx, pdp, sessionID)` is the shared transport dispatch (default deny when the PDP has no sampling authorizer).
5. The JWT wrapper skips its own kill check only when the inner ManifestPDP shares the same manager instance (`innerSharesKillSwitch`) — outcome-identical in every wiring, halves kill-store round-trips per sampling message. `JWTPDP.Decide` deliberately keeps its unconditional check so kill denials win over JWT capability denials in the audit log.
6. `denyResponse` moved to pdp.go.
7. `newTestManifestPDPWithKS` fixture removes the inline manifest constructions.

**Docs**
8. Guide § 2b documents agent-kill attribution and that in-flight sampling during shutdown with a remote kill store is denied as `KILL_SWITCH_ERROR` (fail-closed, not an outage); ADR-0001 scope section updated to match.

`make test` (race), `make lint`, `make check-license` all pass.

PR #159, opened 2026-06-13
fix: bound every upstream round-trip and audit infrastructure failures uniformly

eunolabs:
Follow-up to [#156](https://github.com/eunolabs/eunox/pull/156) (issue #147). That PR closed the HTTP `*/list` timeout gap; a deep multi-angle review of the result found the fix was at the wrong altitude and that the new audit records were inconsistent across handlers and transports. This PR addresses all of it.

## Structural — bound every upstream round-trip once

- **Central timeout.** `withUpstreamTimeout` is now applied once inside the three chokepoints (`callSubprocessUpstream`, `callRemoteUpstream`, `StdioProxy.callUpstream`) instead of per-handler. This bounds two paths the per-handler approach kept missing:
  - **Notification forwarding** — `forwardNotification` → `callRemoteUpstream` bypassed every handler wrap; a hung remote upstream could pin one goroutine per notification indefinitely (the exact DoS class #147 set out to close).
  - **The HTTP drift check** — unbounded in non-strict mode (a leaked goroutine per session against a stalled upstream).
- **Bounded initialize handshake.** Local-mode `initUpstream` now honors the caller's session-start deadline: the handshake runs in a goroutine, a deadline expiry kills the subprocess, and `newSession` reaps it. Previously a wedged `initialize` pinned the handler goroutine and left an unkillable orphan process.

## Audit consistency

- `upstreamErrInfo` gains `REQUEST_CANCELED` (host disconnect no longer blamed on the upstream) and a guard against the misleading "within 0 ms" message for inherited deadlines. **Every** upstream-failure path on both transports now routes through it — HTTP read/subscribe/get (previously hardcoded `UPSTREAM_ERROR`) and all stdio handlers (previously unclassified and unaudited).
- Stdio upstream failures now write audit records, matching HTTP.
- Infrastructure failures are never marked `audit_only` — the error the host receives is enforced even in audit mode, so `stats` no longer files timeouts under "observed would-be denials".
- `suggest` skips infrastructure denial codes, so a `prompts/list` timeout no longer misparses as a prompt literally named `list`.

## Cleanup

The three identical list handlers per transport collapse into one `handleHTTPList` / `handleList` helper parameterised by a `ListFilterer` method expression.

## Tests

The list-timeout test now runs in audit mode, wires a real audit sink, and asserts the `deny`/`UPSTREAM_TIMEOUT` record is written and **not** `audit_only` (verified to fail against the pre-fix behavior). New tests cover notification-forwarding bounding, the initialize-handshake deadline, `upstreamErrInfo`/`isInfraDenialCode` classification, the stdio list-failure audit record, and the `suggest` infra-code exclusion. `postMCP` uses a shared client whose 30s backstop sits above the proxy's 20s session-start budget.

Threat model §5.5/§6.1 updated to document the three infrastructure codes, the central timeout, the bounded handshake, and the `audit_only` semantics.

Full `go test -race ./...`, `golangci-lint`, and `make check-license` pass.


eunolabs:
Pushed `553284f` addressing the code-review findings on this branch.

**Confirmed regression — fixed.** Moving the timeout into the `callUpstream` chokepoints also bound the **session-start drift probe** (`fetchHTTPSessionTools`) to `--upstream-timeout`. On `main` that probe ran only under the 20s session-start budget. With a `descriptionHash` pin (where a drift-fetch failure is fatal), an operator who set `--upstream-timeout` below their upstream's cold-start `tools/list` latency would get HTTP 500 on session start where it previously succeeded. The probe's context is now marked `withoutUpstreamTimeout` (a context flag mirroring `enforcement.WithDryRun`), so it's bounded by the session-start deadline, not the per-request knob. Regression test included (returns 500 without the fix, 200 with it).

**Doc overclaims — corrected.** §5.5 had claimed `--upstream-timeout` bounds "every upstream round-trip ... and the startup drift check ... by both transports." That overclaimed: subprocess notification forwarding and the stdio subprocess drift read are *not* bounded by it, and the drift probe is now deliberately on the session-start budget. The paragraph now states accurately what the per-request bound covers vs. what the session-start budget covers.

**Deferred (pre-existing, tracked separately):** the remaining unbounded paths the review flagged — stdio `initUpstream()` not ctx-bounded, the stdio subprocess drift read, and subprocess notification writes — are pre-existing (not introduced here) and need the goroutine+kill machinery to bound safely; I left them out of this round to keep the change low-risk and corrected the doc rather than overclaiming. Also deferred: the cleanup items (the ~11x-duplicated upstream-failure tail, the two-leaf vs. single-dispatch timeout placement, the `forwardToUpstream` dead-code branch). Happy to follow up on any of these.

Full `go test -race ./...`, `golangci-lint`, and `make check-license` pass.

PR #161, opened 2026-06-13
sec!: fail closed on gateway routes with no policy and no audit posture

### LINKED ISSUE #130: security: gateway routes with no policy fail open, contradicting fail-closed invariant
## Problem

[`cmd/eunox-mcp/route.go:138-171`](../blob/main/cmd/eunox-mcp/route.go#L138) (comment labelled `SEC-05`):

```go
// SEC-05 (gateway): a route with no policy allows every call.
fmt.Fprintf(os.Stderr,
    "[eunox-mcp] WARNING: upstream %q has no policy configured — "+
    "ALL calls on /mcp/%s will be ALLOWED without enforcement. ...",
    u.Name, u.Name)
```

```go
// loadUpstreamPDP: with no policy it returns an allow-all PDP.
if len(u.Policy) == 0 {
    return alwaysAllowPDP{}, nil, "", "", nil
}
```

The project's stated invariant (CLAUDE.md, CONTRIBUTING.md, threat model) is **fail closed**:

> On any ambiguity — missing manifest, unmapped method, malformed JWT, unset condition argument — deny.

A gateway route with a missing `policy:` key is not an operator declaration of intent; it is a misconfiguration. The current behavior (warn + allow-all) means a typo in config, a deleted policy file reference, or a partially-applied manifest rollout silently exposes all tools on that route with no enforcement and no blocking.

## Proposed behaviour

Routes with no `policy:` key should **fail closed at startup** unless the operator explicitly declares `enforcement: audit` (which already has a defined "allow-and-observe" semantic). A route that has neither `policy:` nor `enforcement: audit` is ambiguous — deny.

Startup failure is preferred over runtime denial because it surfaces the misconfiguration before any traffic flows, rather than logging warnings that may be missed in production.

## Migration

This is a breaking change for operators relying on policy-less routes as passthrough. Per the pre-1.0 convention in CONTRIBUTING.md, no back-compat shim — document the migration: add `enforcement: audit` to routes that are intentionally unpoliced.

## Acceptance criteria

- [ ] `buildRoutes` returns an error when a route has no `policy:` and no `enforcement: audit`.
- [ ] `enforcement: audit` with no `policy:` continues to work (wiretap mode).
- [ ] Existing tests updated; new test covering the startup-error path.
- [ ] `docs/gateway.md` and `docs/capability-manifest-guide.md` updated.

eunolabs:
## Problem

Fixes #130. A gateway route with a missing `policy:` key was a fail-**open**: `BuildRoutes` warned to stderr and mounted an `alwaysAllowPDP`, so every call on `/mcp/<name>` was forwarded unenforced. A typo, a deleted policy-file reference, or a half-applied rollout silently exposed all tools on that route — contradicting the project's fail-closed invariant.

## Change

- **`BuildRoutes` fails closed at startup** (`route.go`): a route with no `policy:` and not in audit mode returns an error naming the route, instead of warn-and-allow. Startup failure surfaces the misconfiguration before any traffic flows.
- **`enforcement: audit` still works** with no policy (wiretap/observe-only) and now emits a `NOTICE` so the open posture is visible. This is the documented escape hatch for intentionally-unpoliced routes.
- **Consistency in `validate` and `doctor`**: both previously reported a policyless route as "allow-all". `validate` now marks a gateway policyless non-audit route `FAIL` (exit 2); `doctor` flags that it would fail closed at startup. The audit (observe-only) case and the single-upstream **stdio** host passthrough are unchanged — this fix is scoped to gateway mode, per the issue.

## Acceptance criteria (from #130)

- [x] `BuildRoutes` returns an error when a route has no `policy:` and no `enforcement: audit`
- [x] `enforcement: audit` with no `policy:` continues to work (wiretap mode)
- [x] Existing tests updated; new test covering the startup-error path
- [x] Docs updated (`docs/gateway.md` does not exist in the tree — documented in README, `docs/architecture.md`, `docs/capability-manifest-guide.md`, and the gateway config field comments instead)

## Tests

- `TestBuildRoutes_NoPolicyEnforce` → rewritten as `TestBuildRoutes_NoPolicyEnforce_FailsClosed` (asserts the startup error).
- New `TestBuildRoutes_NoPolicyPerRouteAudit` (per-route `enforcement: audit` with no policy builds OK).
- New `TestValidateConfigRoutes_GatewayPolicylessNonAuditFails` (validate exits 2).
- `TestBuildRoutes_GlobalStrictDrift` and `TestBuildRoutes_ExpectVersionWithoutPolicy` updated to use `enforcement: audit` for their (now-illegal) policyless routes, preserving their original intent.

`go test -race ./...` passes; `go vet` and `gofmt` clean.

## Migration (breaking, pre-1.0 — no shim)

Gateway routes relying on an absent `policy:` for unenforced passthrough no longer start. Add `enforcement: audit` to any route that is intentionally unpoliced.

Fixes #130


eunolabs:
Addressed in a49075f. Reworded all four flagged passages — README.md, docs/conformance.md, docs/client-integration.md, and the schema `policy` description — to distinguish three cases: a gateway (`transport: http`) route with no policy and no audit posture **fails closed at startup**; an explicit `enforcement: audit` route is observe-only allow-and-log; the single-upstream stdio host without a policy remains a warn-and-forward passthrough.

A broad rescan turned up one more stale spot — the §1 "No policy (pass-through)" mode row in `docs/threat-model-mcp.md:71`. Since that file is owned by #162 (to avoid a cross-PR merge conflict), I corrected it there rather than here.

PR #162, opened 2026-06-13
docs: map the adversarial eval suite into the threat model (§7.3)

### LINKED ISSUE #132: enhancement: publish adversarial eval suite (prompt injection, tool poisoning, exfiltration, JWT bypass)
## Motivation

eunox's security claims rest on enforcement correctness, but there is no published adversarial test suite that exercises the scenarios the proxy is designed to defeat. Without it:

- There is no regression protection when enforcement logic changes.
- Enterprise evaluators cannot independently verify the claims.
- The threat model's "residual risks" section is untested.

### Prompt injection / indirect injection
- Upstream tool response contains a payload that attempts to override the manifest or issue a new `tools/call`.
- Verify: the proxy forwards the response unmodified and does not re-evaluate it as a policy input.

### Tool poisoning via description
- An allowed tool has a malicious `description` or `inputSchema` annotation designed to steer the model toward a denied tool.
- This is **not** blocked by eunox (it's a model-layer concern), but the test should document the residual and the partial mitigation (hiding denied tools from `tools/list`).

### Read-then-exfiltrate
- Agent calls an allowed `resources/read`, then attempts `tools/call` with the read value as an argument to an exfil tool.
- Verify: if the exfil tool is not in the manifest, the call is denied; if `allowedValues` conditions apply, they fire.

### Overbroad JWT claim
- JWT `capabilities` claim grants `*` on all tools.
- Verify: JWTPDP intersects with the manifest; a `*` JWT claim does not expand beyond what the manifest permits.

### New-method fail-closed
- Client sends a novel MCP method (e.g. `tools/execute`, `resources/patch`).
- Verify: proxy returns `AUTHORIZATION_FAILED`, nothing is forwarded, audit record is written.

### kill-switch / rate-limit bypass
- Repeated calls beyond `maxCalls` with a new session ID each time (in-memory counter scope).
- Verify: when Redis is used, counter is shared across sessions.

## Implementation

A dedicated `cmd/eunox-mcp/adversarial_test.go` file running against a mock upstream, similar to `enforcement_gaps_test.go`. Each scenario should have a pass (blocked as expected) and a fail (would-pass-without-eunox) case.

Results should feed into the threat model's residual-risk table.

eunolabs:
## Context

The adversarial eval suite that #132 asks for — `cmd/eunox-mcp/adversarial_test.go`, scenarios **Adv-1 … Adv-6** — already shipped in [#141](https://github.com/eunolabs/eunox/pull/141) and covers every scenario the issue lists (prompt injection, tool poisoning, read-then-exfiltrate, overbroad JWT claim, novel-method fail-closed, rate-limit session-ID bypass), each with a pass case and, where a residual exists, a baseline case. That PR referenced the issue with a bare URL rather than `Closes`, so #132 stayed open. The remaining ask — *"results should feed into the threat model's residual-risk table"* — is what this PR does.

## Change

- **New §7.3 "Adversarial test coverage"** in `docs/threat-model-mcp.md`: a table mapping each Adv-N scenario to the §3 attack class / residual risk it exercises and exactly what the suite asserts. This makes the suite a published, runnable verification artifact (`go test -run TestAdv ./cmd/eunox-mcp/`) and makes the test file's own "documented in the threat model's residual-risk table" header accurate.
- **Renumbered** "Reporting security issues" §7.3 → §7.4; added a Change Log row (1.7).
- **Tightened** the `adversarial_test.go` header to point back at §7.3 (bidirectional link).
- **Drive-by repair:** three pre-existing **dangling `(§ 7.4)` references** about audit key rotation (lines 244, 491, and Change Log 1.5) — there was never a §7.4 — now point to §3.4, where the `key_id`/rotation mechanism is actually documented. Without this, the renumber above would have silently re-pointed them at "Reporting security issues."

## Verification

`go test -run TestAdv ./cmd/eunox-mcp/` passes; `go vet` and `gofmt` clean. No `(§ 7.4)` references remain except the new §7.4 header.

Fixes #132


eunolabs:
Addressed in b85ecea.

**[P2]** Added `TestAdv3_ReadThenExfil_SequenceBlock_FiresAfterRead` rather than dropping the §3.2.1 mapping — `sequenceBlock` is the canonical sequential-exfiltration control that §3.2.1 documents, so the suite should exercise it. The test allows `read_credentials`, then a `write_external` in the same session is denied with `ConditionType == sequenceBlock` (the read is in session history); a fresh session with no prior read may write. Passes.

**[P3]** Reworded both the §7.3 intro and the `adversarial_test.go` header from "§3 attack classes" to "threat model sections (§2 invariants, §3 attack classes, §4 out-of-scope) and their residual risks", matching the table's actual targets (Adv-1 → §4.1/§2.3, Adv-5 → §2.2).

Also corrected the stale §1 "No policy (pass-through)" mode row flagged on #161 (it lives in this file).

PR #171, opened 2026-06-13
sec: audit drop markers, JWKS transport validation, and redaction fixes

eunolabs:
## Summary

This PR hardens three security boundaries: (1) audit records dropped under back-pressure are now made tamper-evident via signed drop markers in the chain; (2) the JWKS endpoint — the root of trust for every JWT — is validated to use https or loopback http, failing closed on plaintext remote fetches unless explicitly opted into; (3) field redaction now correctly recurses through nested arrays, closing a silent fail-open leak where fields reachable only through array-of-array were forwarded unredacted.

## Why

**Audit drop markers:** When the audit record queue fills under a flood, records are silently dropped. Without an explicit marker, audit-verify cannot distinguish a clean log from one with an excised window — the chain head only advances on successful writes, so a drop leaves no seq gap and no chain break. An attacker who truncates the log at a drop point could hide the loss. Drop markers are normal chained records, so deleting one breaks the chain like any other record, making the loss tamper-evident.

**JWKS transport validation:** The JWKS endpoint is the root of trust for every token. A plaintext http fetch to a remote host lets an on-path attacker substitute the key set and forge capability claims with arbitrary scope. The proxy now enforces https (or http to loopback, where there is no network MITM surface) at startup, failing closed by default. Operators can opt into insecure remote http with `--jwks-allow-insecure-http` for development only.

**Redaction recursion:** The `redactFields` obligation applies a dot-path like `groups.ssn` to filter sensitive fields. The previous implementation stopped at the first array level, so a field reachable only through nested arrays (e.g., `groups[i][j].ssn`) was forwarded unredacted — a silent fail-open leak. Redaction now recurses through arbitrarily nested arrays.

**Additional fixes:**
- Audit log tail HMAC verification on resume: when the proxy restarts, it chains from the last record in the log. The tail's own HMAC is now verified before chaining onto it, so an attacker who truncates or appends a crafted tail cannot silently seed the resumed chain from unverified bytes.
- Redis kill-switch reconciliation loop: pub/sub is at-most-once, so a kill event published during a transient disconnect is lost and would never be observed. A periodic reconciliation loop now re-reads the authoritative Redis state on a fixed interval (default 30s), so a missed kill or fail-open cold start self-heals within one interval.
- X-Forwarded-For parsing: when `--trust-forwarded-for` is set, the proxy now reads the right-most entry (appended by the trusted immediate proxy), not the left-most (which a client can forge). A client sending `"X-Forwarded-For: 10.0.0.1"` to spoof a trusted source for an ipRange condition must not win over the real address the trusted proxy appends.
- Upstream redirect handling: the HTTP client now refuses to follow redirects. An MCP JSON-RPC POST has no legitimate redirect, and Go's default client only strips Authorization/Cookie on cross-host redirects — a custom `upstreamAuthHeader` (e.g., `"X-Api-Key: ..."`) would otherwise be replayed verbatim to whatever host a compromised upstream names in a 30x Location, leaking the credential.
- JWT-only mode capability enforcement: when no manifest PDP is configured, a JWT with an absent `mcp.capabilities` claim is now denied, not allowed. Without a manifest backstop, an absent claim would otherwise grant every target to any validly signed token that simply omits the field, turning capability enforcement into an opt-in the token issuer controls. That contradicts the fail-closed invariant.

## Type of change

- Security hardening
- Bug fix

## How it was tested

```
make test    # go test -race ./...
```

New tests added:
- `TestAuditDropMarker_RecordsDropAndChainStaysValid`: verifies drop markers are written, accounted exactly once, and the chain verifies end-to-

https://claude.ai/code/session_0178SuBF8wjhMAUid5qm6BDi


cmd/eunox-mcp/main.go
```diff
@@ -393,6 +398,46 @@ func validateJWTAudienceConfig(jwksURI, jwtAudience string, allowAnyAudience boo
...
+func validateJWKSURIScheme(jwksURI string, allowInsecure bool) error {
+	if jwksURI == "" {
+		return nil
+	}
+	u, err := url.Parse(jwksURI)
+	if err != nil {
+		return fmt.Errorf("--jwks-uri is not a valid URL: %w", err)
+	}
+	switch strings.ToLower(u.Scheme) {
+	case "https":
```
eunolabs, line 417: This validates only the configured URI. The JWKS fetch itself uses the default client in `newJWKSCache`, which follows redirects, so a valid `https://...` JWKS endpoint can return `302 Location: http://idp.example/jwks` and the subsequent key fetch happens over plaintext remote HTTP without `--jwks-allow-insecure-http`. That reopens the key-substitution path this check is trying to close. Please either disable redirects for JWKS fetches or validate each redirect target with the same scheme/loopback policy, and add an https-to-http redirect regression test.

cmd/eunox-mcp/audit.go
```diff
@@ -199,6 +203,19 @@ func openAuditSink(logPath, keyPath string, rotateSizeBytes int64) (*auditSink,
...
+			// Verify the tail's own HMAC before chaining onto it. An attacker who
+			// truncates the log to a chosen record, or appends a crafted tail,
+			// would otherwise silently seed the resumed chain's seq/prev_hmac from
+			// unverified bytes. A legacy pre-chain record (no _hmac) is exempt: it
+			// predates signing and is resumed as before. We warn rather than abort
+			// so a corrupted local log does not block enforcement (the documented
+			// audit-failure tradeoff); operators should reconcile against an
+			// external sink.
+			if prev.HMAC != "" {
+				if ok, err := s.VerifyRecord([]byte(last)); err != nil || !ok {
```
eunolabs, line 233: This uses `s.VerifyRecord`, but this signer sink has `verifyKeys == nil`, so `keysToTry` falls back to only the active key (`keys[0]`). After a normal key rotation where the existing tail was signed by a retired key still present later in the key file, restart will warn that the tail failed HMAC verification even though `audit-verify` would accept it. Populate the sink's verify keyring from all loaded keys before this check, or verify the tail with a full-keyring verifier, and cover a rotated-tail restart in tests.

eunolabs (commented):
Found two blocking issues in the hardening paths: JWKS redirects can bypass the new transport validation, and audit tail verification does not account for rotated keys.

eunolabs:
Reviewed current head `58fca1f48609fb0c05e7f1ae5880e30e227c1362` again. I don't have additional comments beyond the two unresolved review threads already on this PR:

- JWKS redirects can bypass the new scheme/loopback validation unless redirect targets are blocked or revalidated.
- Audit tail verification on restart needs to verify against the full rotated keyring, not only the active signing key.

No more comments from this pass.

cmd/eunox-mcp/main.go
eunolabs: Fixed in fb191f2. The JWKS HTTP client now sets `CheckRedirect` to re-apply `validateJWKSURIScheme` to every redirect target, so an `https://` endpoint cannot `302` the key fetch onto plaintext remote http (the redirect is refused and the token rejected); legitimate https/loopback redirects still follow, and the 10-hop cap is preserved. Wired via `newJWKSHTTPClient(allowInsecure)` passed into the gateway's `JWTPDPOptions.Client`. Regression tests added: `TestJWKSClient_BlocksHTTPSToHTTPRedirect` (the https→remote-http case) and `TestJWKSClient_AllowsLoopbackRedirect`.

---

cmd/eunox-mcp/audit.go
eunolabs: Fixed in fb191f2. `openAuditSink` now builds the full verification keyring (active + retired keys) from the key file and sets it on the signer sink's `verifyKeys`, so the startup tail check verifies a rotated tail against whichever key signed it rather than the active key alone. Signing still uses the active key (`keys[0]`); `verifyKeys` is consulted only by `VerifyRecord`. Regression test added: `TestAuditKey_RotatedTailVerifiesOnRestart` (writes under key A, rotates to [B, A], reopens, and asserts the A-signed tail verifies).

---

eunolabs:
Re-reviewed current head `46a0a793613698827e6cdbed3196a680cbfc3444`.

The two prior findings are addressed:
- JWKS redirects now re-apply the scheme/loopback policy via `newJWKSHTTPClient`, with regression coverage for remote-http and loopback redirects.
- Audit tail verification now loads the full active+retired keyring before the startup tail check, with rotated-tail restart coverage.

I resolved both prior review threads. No new actionable comments from this pass.

PR #183, opened 2026-06-13
fix: make Redis.Peek use exclusive lower bound for window cutoff

### LINKED ISSUE #175: Redis callcounter Peek includes entries at the exact window cutoff
## Summary

`pkg/callcounter.Redis.Peek` counts a sorted-set entry whose score is exactly equal to the cutoff timestamp, but the mutating Redis path and the in-memory backend both treat that entry as expired.

## Evidence

Relevant code:

- `pkg/callcounter/redis.go`: `IncrementAndGet` removes expired entries with `ZRemRangeByScore(ctx, windowKey, "-inf", cutoff)`, which removes scores `<= cutoff`.
- `pkg/callcounter/redis.go`: `Peek` uses `ZCount(ctx, windowKey, cutoff, "+inf")`, and Redis score ranges are inclusive by default, so it counts scores `>= cutoff`.
- `pkg/callcounter/memory.go`: `Peek` only counts timestamps where `ts.After(cutoff)`, so exact-cutoff timestamps are expired there too.

Focused reproduction used during review:

```go
now := time.Unix(1_700_000_000, 0)
counter := callcounter.NewRedis(client, callcounter.WithRedisTimeFunc(func() time.Time { return now }))
_, _ = counter.IncrementAndGet(ctx, "boundary", 60)
now = now.Add(60 * time.Second)
got, _ := counter.Peek(ctx, "boundary", 60)
// got == 1; expected 0 to match IncrementAndGet and InMemory.Peek semantics
```

## Impact

Any Redis-backed caller that relies on `Peek` can observe stale exact-boundary history. Today that includes `sequenceBlock` session-history checks, where a prior tool call can continue blocking at the exact 24h boundary even though the mutating counter path would have expired it.

This is a narrow boundary bug, but it makes Redis and in-memory behavior disagree and can cause incorrect denials.

## Suggested fix

Make `Redis.Peek` use an exclusive lower bound, e.g. `(" + cutoff)` for `ZCOUNT`, or otherwise share the same expiry predicate as `IncrementAndGet`.

eunolabs:
## Summary

Fix a boundary bug in `Redis.Peek` where entries with a score exactly equal to the window cutoff were incorrectly counted as still-present, while the mutating path (`IncrementAndGet`) and in-memory backend both treated them as expired. The fix changes `ZCOUNT` to use an exclusive lower bound `"("` so all three code paths agree at the window edge.

## Why

`Redis.Peek` was using an inclusive lower bound in its `ZCOUNT` query, counting entries with scores `>= cutoff`. However:
- `IncrementAndGet` drops entries via `ZRemRangeByScore("-inf", cutoff)`, which removes scores `<= cutoff`
- `InMemory.Peek` counts only entries where `ts.After(cutoff)`, excluding the exact cutoff

This inconsistency meant that at the precise window boundary (e.g., exactly 24 hours for a `sequenceBlock` condition), a peek would see a stale entry that the mutating path had already expired, keeping stale history alive and potentially allowing blocked sequences to proceed when they should remain blocked.

## Type of change

- Bug fix

## How it was tested

Added two regression tests:
- `TestRedis_Peek_ExcludesExactCutoff`: verifies that an entry at the exact cutoff timestamp reads as expired (count 0) and is dropped by the next `IncrementAndGet`
- `TestRedis_Peek_IncludesJustInsideWindow`: guards the other boundary—an entry one microsecond newer than the cutoff is still counted

Both tests use miniredis and a controllable time function to precisely position entries at the boundary.

## Security considerations

No security impact. This is a correctness fix that makes the behavior more conservative (entries expire sooner), not less. It ensures that rate-limit and sequence-block conditions enforce their windows consistently across all code paths.

## Checklist

- [x] `make test` passes — new tests added, existing tests unaffected
- [x] `make lint` passes — no new findings
- [x] No new files, no license header needed
- [x] Behavior change documented in code comments explaining the fix

https://claude.ai/code/session_01AiPLKwV9QD6LZr6mnorB4p

Fixes #175


eunolabs (commented):
No issues found.

Reviewed the Redis `Peek` boundary change and the added regression tests. The new exclusive lower bound matches `IncrementAndGet`'s `ZRemRangeByScore("-inf", cutoff)` behavior and the in-memory `ts.After(cutoff)` predicate.

Verified locally against PR head `57565d158a92814eac84edb07cffb9e468a8b5ac`:
- `GOCACHE=/private/tmp/eunox-go-build-cache go test ./pkg/callcounter`
- `GOCACHE=/private/tmp/eunox-go-build-cache go test ./pkg/enforcement`

eunolabs:
No new review comments from this pass.

PR #184, opened 2026-06-13
sec: reject overflowing windowSeconds in call-counter and manifest

### LINKED ISSUE #176: Unbounded maxCalls windowSeconds can overflow callcounter durations and reset quotas
## Summary

`maxCalls.windowSeconds` is only validated as `>= 1`, but both callcounter backends multiply it into `time.Duration`. Very large accepted values can overflow those duration calculations, causing counters to reset instead of accumulating calls.

## Evidence

Relevant code:

- `pkg/callcounter/memory.go`: `IncrementAndGet` computes `time.Duration(windowSec) * time.Second` before pruning timestamps.
- `pkg/callcounter/memory.go`: `Cleanup` computes `time.Duration(e.windowSec) * 2 * time.Second`.
- `pkg/callcounter/redis.go`: `IncrementAndGet` computes both the cutoff duration and `Expire(ctx, windowKey, time.Duration(windowSec)*2*time.Second)`.
- `cmd/eunox-mcp/manifest.go`: `validateMaxCalls` rejects only `windowSeconds < 1`; there is no upper bound.

Focused reproductions used during review:

```go
windowSec := int(math.MaxInt64/int64(time.Second)) + 1
_, _ = inMemory.IncrementAndGet(ctx, "huge", windowSec)
got, _ := inMemory.IncrementAndGet(ctx, "huge", windowSec)
// got == 1; expected 2 because both calls are in an enormous window
```

```go
windowSec := int(math.MaxInt64/(2*int64(time.Second))) + 1
_, _ = redisCounter.IncrementAndGet(ctx, "huge-ttl", windowSec)
got, _ := redisCounter.IncrementAndGet(ctx, "huge-ttl", windowSec)
// got == 1; expected 2 because the key TTL overflows and the key expires immediately
```

## Impact

A malformed or maliciously large `windowSeconds` can make a quota fail open: repeated calls keep returning `1`, so `maxCalls` limits are not enforced. This matters because the manifest validator currently accepts these values, and the public `callcounter.Store` API has no guardrail either.

## Suggested fix

Reject window values that cannot be safely converted before multiplying by `time.Second`, preferably in both places:

1. Manifest / capability validation: enforce a sane upper bound for `windowSeconds`.
2. `callcounter` backends: return an error for invalid or overflowing `windowSec` so enforcement fails closed.

The Redis TTL path needs the stricter `2 * windowSec * time.Second` bound, or a checked helper that separately validates cutoff and cleanup TTL durations.

eunolabs:
Fixes #176

## Summary

Add validation to reject `windowSeconds` values that overflow `time.Duration` arithmetic in the call-counter backends. An out-of-range window would silently reset the quota on every call (fail-open), so reject it at load time in the manifest validator and at runtime in both in-memory and Redis backends.

## Why

Closes #176

The call-counter backends (`memory.go`, `redis.go`) multiply `windowSec` into a `time.Duration` to compute sliding-window cutoffs and key TTLs. A `windowSec` value large enough to overflow this multiplication wraps to a negative or tiny duration, causing the cutoff to land at or after "now" — every call then reads as the first in its window and the `maxCalls` quota silently fails open.

This is a security boundary violation: a manifest with an absurdly large `windowSeconds` (e.g., `99999999999999`) would silently disable rate limiting at runtime rather than enforcing it. The fix:

1. Define `MaxWindowSeconds` as the largest value that doesn't overflow `time.Duration(windowSec) * 2 * time.Second` (the 2x margin used by both backends for TTL/cleanup).
2. Add `checkWindowSec()` validation called by both backends before any duration arithmetic.
3. Reject out-of-range windows in the manifest validator at load time, so invalid policies never ship.

## Type of change

- Security hardening
- Bug fix

## How it was tested

```
make test    # Added regression tests for both backends and manifest validation
```

New tests:
- `TestInMemory_RejectsOutOfRangeWindow`: verifies in-memory backend rejects overflowing and non-positive windows on both `IncrementAndGet` and `Peek`.
- `TestRedis_RejectsOutOfRangeWindow`: verifies Redis backend rejects overflowing and non-positive windows on both methods.
- `manifest_validation_test.go`: verifies manifest validator rejects `windowSeconds` past `MaxWindowSeconds` at load time.

All existing tests pass; no behavior change for valid windows.

## Security considerations

**Fail-closed preserved:** On an out-of-range `windowSeconds`, the call-counter now returns an error instead of silently resetting the quota. The manifest validator rejects invalid policies at load, preventing them from ever reaching the runtime engine. This closes a quota-bypass vulnerability.

**Audit tape:** No impact — audit records are unaffected.

**Secrets:** No impact.

**Authority:** This change narrows what a manifest can express (rejects previously-accepted but broken values), closing a fail-open gap. No new authority is granted.

## Checklist

- [x] `make lint` passes — `go vet` + golangci-lint
- [x] Docs updated (`docs/capability-manifest-guide.md`) explaining the upper bound and why it exists
- [x] PR description explains *why* (quota bypass via overflow) and *what* (validation at load and runtime)

https://claude.ai/code/session_01Fs3z1CmEAbXnev884eiKQA


eunolabs:

pkg/callcounter/callcounter.go
```diff
@@ -4,11 +4,45 @@
...
+// turn windowSec into a time.Duration, and the longest product either computes
+// is the 2x cleanup/TTL margin — time.Duration(windowSec)*2*time.Second
+// (memory.go Cleanup, redis.go Expire). time.Duration is int64 nanoseconds, so a
+// windowSec above this bound overflows that multiplication, wraps to a negative
+// or tiny duration, and makes the sliding-window cutoff (or the Redis key TTL)
+// land at or after "now" — every call then reads as the first in its window and
+// the maxCalls quota silently fails open. ~4.6e9 seconds (~146 years) is far
+// beyond any real quota window, so rejecting larger values costs nothing while
+// closing the overflow.
+const MaxWindowSeconds = math.MaxInt64 / (2 * int64(time.Second))
```
eunolabs: The bound is exactly right today, but its correctness is pinned to the literal `2` here matching the `* 2` margin in `InMemory.Cleanup` (memory.go) and the `Expire(...)` TTL (redis.go) — three independent literals that have to stay in lockstep. If that cleanup/TTL margin is ever widened (say to `3` for more clock-skew slack), this divisor stays `2`, `MaxWindowSeconds` silently becomes too large, and the exact overflow this PR closes re-opens — with no test catching it, because the boundary tests only exercise `MaxWindowSeconds` itself, which would still be "accepted".

Consider extracting the margin as a single named constant shared by all three sites:

```go
const cleanupMarginFactor = 2 // also used by InMemory.Cleanup and Redis Expire
const MaxWindowSeconds = math.MaxInt64 / (cleanupMarginFactor * int64(time.Second))
```

and/or a guard test that pins the boundary arithmetic, e.g. assert `time.Duration(MaxWindowSeconds) * cleanupMarginFactor * time.Second > 0`. That keeps the security bound provably in sync with the arithmetic it guards rather than relying on a prose cross-reference.

---

pkg/callcounter/redis_test.go
```diff
@@ -33,6 +33,30 @@ func TestRedis_IncrementAndGet_Basic(t *testing.T) {
...
+	ctx := context.Background()
+
+	huge := int(callcounter.MaxWindowSeconds) + 1
+	_, err := counter.IncrementAndGet(ctx, "huge-ttl", huge)
+	require.Error(t, err, "overflowing windowSec must fail closed")
+	_, err = counter.Peek(ctx, "huge-ttl", huge)
+	require.Error(t, err, "overflowing windowSec must fail closed on Peek too")
+
+	_, err = counter.IncrementAndGet(ctx, "zero", 0)
+	require.Error(t, err)
```
eunolabs, line 57: Minor (test symmetry): `TestInMemory_RejectsOutOfRangeWindow` also asserts that the boundary value `MaxWindowSeconds` is *accepted* and counts normally — that "largest in-range window" case is the most valuable assertion for catching an off-by-one in the bound. This Redis test only checks the rejection side. Consider adding the symmetric accept case here so both backends pin the boundary, not just the over-limit branch:

```go
n, err := counter.IncrementAndGet(ctx, "ok", int(callcounter.MaxWindowSeconds))
require.NoError(t, err)
assert.Equal(t, int64(1), n)
```

---

eunolabs (commented):
## Review: `sec: reject overflowing windowSeconds`

Solid, well-scoped security fix. It closes #176 correctly and aligns with the repo's fail-closed invariant. Two minor, non-blocking suggestions inline; nothing here needs to change before merge.

### What it does
Adds an upper bound (`callcounter.MaxWindowSeconds`) on `maxCalls.windowSeconds` and enforces it in two places: the manifest validator at load time, and both call-counter backends (`IncrementAndGet` + `Peek`, in-memory and Redis) at runtime. An out-of-range window now fails closed with an error instead of overflowing the duration math and silently resetting the quota.

### Correctness — verified
- **The bound is exact.** `MaxWindowSeconds = MaxInt64 / (2 * int64(time.Second))` makes the first rejected value (`Max+1`) the first that overflows `time.Duration(windowSec) * 2 * time.Second`: `Max * 2e9 = 9223372036000000000 <= MaxInt64`, and `(Max+1) * 2e9` overflows. No off-by-one, and the largest accepted window leaves ~8.5e8 ns of headroom.
- **The 2x divisor is the true worst case.** Confirmed by reading both backends: the only duration multipliers are the 1x cutoffs (`IncrementAndGet`/`Peek`) and the 2x margin in `InMemory.Cleanup` and the Redis `Expire` TTL. The PR correctly uses the stricter 2x bound for *both* backends, matching the issue's note that the Redis TTL path needs `2 * windowSec * time.Second`.
- **`Cleanup` can't be poisoned.** `InMemory.Cleanup` does the 2x math without its own guard, but it only ever reads `e.windowSec`, which is written in `IncrementAndGet` — and the new guard runs *before* that assignment, so an overflowing value can never be stored. The invariant holds.
- **Single source of truth.** The manifest validator and the runtime backends both bound against the same exported `callcounter.MaxWindowSeconds`, so the load-time and runtime limits can't drift apart.
- **Defense in depth.** Rejecting both at load (invalid policies never ship) and at runtime (any path reaching the backend — e.g. `sequenceHistoryWindowSec`) keeps the guarantee even if a window arrives by a route other than `validateMaxCalls`.

### Tests & CI
- Regression tests mirror the two reproductions in #176 (in-memory `IncrementAndGet` overflow, Redis TTL overflow) and cover `Peek` plus the non-positive case. The in-memory test also asserts the boundary `MaxWindowSeconds` is accepted — good (see the inline note about mirroring that on the Redis side).
- All 42 CI checks green (Test, Lint, gofmt, License Headers, govulncheck, Trivy, CodeQL, enforcement compliance, demo stacks).
- Conventions look right: Apache-2.0 headers present, lowercase punctuation-free error strings, docs updated in the same PR, `sec:` commit type.

### Suggestions (both minor, inline)
1. **`callcounter.go`** — the `2` in the bound is duplicated as three independent literals (here + `Cleanup` + `Expire`); a shared `cleanupMarginFactor` const (and/or a boundary-overflow guard test) would keep the bound from silently drifting if that margin is ever widened.
2. **`redis_test.go`** — add the symmetric "largest in-range window accepted" assertion that the in-memory test already has.

One thing I'd flag but wouldn't act on: `windowSec int` and `int(MaxWindowSeconds)` assume a 64-bit `int` (on a 32-bit build, `MaxWindowSeconds` ~4.6e9 exceeds `MaxInt32`). Not a concern given the build matrix is entirely 64-bit, just noting the assumption.

---

eunolabs:

pkg/callcounter/callcounter.go
eunolabs: Done in 4c27d0a. Extracted the margin as a single `cleanupMarginFactor` constant used by all three sites (the `MaxWindowSeconds` divisor, `InMemory.Cleanup`, and the Redis `Expire` TTL), so the bound re-derives if the margin ever changes. Also added `TestMaxWindowSeconds_BoundaryDoesNotOverflow`, which pins the arithmetic: the bound's doubled-margin product stays positive and `(bound+1)` overflows — so a widened margin without a re-derived bound now fails a test rather than silently re-opening the overflow.

---

pkg/callcounter/redis_test.go
eunolabs: Done in 4c27d0a. Added the symmetric accept-the-boundary assertion to `TestRedis_RejectsOutOfRangeWindow`, matching the in-memory test so both backends now pin the off-by-one boundary, not just the over-limit branch.

---

pkg/callcounter/memory.go
```diff
@@ -138,10 +150,11 @@ func (m *InMemory) StartCleanup(ctx context.Context, interval time.Duration) {
...
-// An entry is dropped once its most-recent timestamp is older than twice the
-// largest window it was counted with — the same 2x safety margin the Redis
-// backend applies via Expire(key, windowSec*2) (redis.go), keeping the two
-// backends equivalent. Keying staleness to the entry's own window (rather than
+// An entry is dropped once its most-recent timestamp is older than
+// cleanupMarginFactor times the largest window it was counted with — the same
+// safety margin the Redis backend applies via Expire(key, windowSec*factor)
+// (redis.go), keeping the two backends equivalent. Keying staleness to the
+// entry's own window (rather than
 // a fixed 24h) is essential: a maxCalls quota with windowSeconds > 86400 (a
```
eunolabs: Nit (non-blocking, prose only): when `2x` became `cleanupMarginFactor times` here the paragraph wasn't re-wrapped, leaving `entry's own window (rather than` dangling on its own line with the parenthetical split awkwardly across the line break (`(rather than` / `a fixed 24h)`). Consider reflowing this comment block so the lines fill normally — purely cosmetic, no behavior impact.

---

eunolabs (commented):
## Independent re-review: `sec: reject overflowing windowSeconds`

I went through this fresh against the source (not just the diff) and ran the new tests locally — `pkg/callcounter` and `cmd/eunox-mcp` both pass with `-race`. This is a correct, well-scoped fail-closed fix; I'm comfortable with it. Nothing blocking. One cosmetic nit inline.

### Verified independently
- **Bound is exact and tight.** `MaxWindowSeconds = MaxInt64 / (cleanupMarginFactor * int64(time.Second))` makes `MaxWindowSeconds` the last accepted value and `MaxWindowSeconds+1` the first to overflow `time.Duration(windowSec) * cleanupMarginFactor * time.Second`. `TestMaxWindowSeconds_BoundaryDoesNotOverflow` pins both sides (positive at the bound, wraps non-positive at +1), and the int64-var trick correctly forces runtime evaluation so the overflow isn't a compile error.
- **2× is the true worst case.** Read every duration multiply in both backends: the only multipliers are the 1× cutoffs (`IncrementAndGet`/`Peek`) and the 2× margin in `InMemory.Cleanup` and the Redis `Expire` TTL. Bounding against the stricter 2× covers all of them, and the shared `cleanupMarginFactor` keeps the divisor, both backends, and the bound in lockstep (the prior round's suggestion — nicely done).
- **`Cleanup` can't be poisoned.** It reads `e.windowSec` without its own guard, but `checkWindowSec` runs *before* the assignment in `IncrementAndGet` (`memory.go:61` precedes `:81`), so an overflowing value can never be stored.
- **Fail-closed wiring confirmed at the engine.** The new backend errors surface as denies: `handleMaxCalls` → `ErrCodeConditionFailed` (`handlers.go:262`) and `handleSequenceBlock` `Peek` → `ErrCodeConditionFailed` (`handlers.go:726`). The one `IncrementAndGet` whose error is intentionally discarded (`engine.go:203`, session-history marker) only ever passes the constant `sequenceHistoryWindowSec = 86400`, which is far inside the bound — so no path silently swallows a relevant error.
- **Defense in depth holds.** Rejecting at load (`validateMaxCalls`) *and* at the backend means a `windowSeconds` arriving by any route other than the manifest validator (e.g. JWT-derived capabilities, or `sequenceHistoryWindowSec`) still fails closed against the same exported constant — no drift between load-time and runtime limits.

### Note (not raising as a change)
The earlier round already flagged the 64-bit `int` assumption (`int(MaxWindowSeconds)` ~4.6e9 > `MaxInt32`). Confirmed the build matrix is amd64/arm64 only, so this is a non-issue today — just keeping the assumption on record.

---

pkg/callcounter/memory.go
eunolabs: Done in a06e45c — reflowed the comment block so the lines fill normally and the `(rather than a fixed 24h)` parenthetical no longer splits across the break. Comment-only, no behavior change.

---

PR #185, opened 2026-06-13
fix: dedupe resource specificity/match scoring in pdp.go

### LINKED ISSUE #180: bug: `resSpecificity` in pdp.go duplicates `resourceSpecificity` in engine.go with a different implementation
## Summary

Pattern-specificity scoring is implemented twice with different code, creating a silent divergence risk. If one function is updated the other is not, the `ManifestPDP` and the raw `Engine` will rank overlapping glob patterns differently, producing different "most specific capability" selections for the same request.

## Locations

**`pkg/enforcement/engine.go:537–558`** — `resourceSpecificity`
```go
func resourceSpecificity(resource, toolName string) int {
    if resource == toolName {
        return 1000
    }
    if !strings.ContainsAny(resource, "*?[") {
        return 900
    }
    prefixLen := 0
    for _, r := range resource {
        if strings.ContainsRune("*?[", r) {
            break
        }
        prefixLen++
    }
    wildcardCount := 0
    for _, r := range resource {
        if strings.ContainsRune("*?[", r) {
            wildcardCount++
        }
    }
}
```

**`cmd/eunox-mcp/pdp.go:337–353`** — `resSpecificity`
```go
func resSpecificity(pattern, name string) int {
    bare := stripNamespacePrefix(pattern)
    if bare == name {
        return 1000
    }
    if !strings.ContainsAny(bare, "*?[") {
        return 900
    }
    prefix := 0
    for _, r := range bare {
        if strings.ContainsRune("*?[", r) {
            break
        }
        prefix++
    }
    wildcards := strings.Count(bare, "*") + strings.Count(bare, "?") + strings.Count(bare, "[")
    return prefix*10 - wildcards
}
```

The two implementations look equivalent today, but `strings.Count` vs rune-iteration diverge for multi-byte Unicode input — `[` in a Unicode-heavy pattern is counted differently by the two methods. More importantly, this is the kind of silent duplication that causes subtle bugs over time as callers evolve independently.

Note that `bareSpecificity` in `pdp.go` **correctly** delegates to `enforcement.ResourceSpecificity`:
```go
func bareSpecificity(pattern, name string) int {
    return enforcement.ResourceSpecificity(pattern, name)
}
```

`resSpecificity` is used only by `drift.go`, which also calls `matchResource` (a local reimplementation of the same glob match). Both `drift.go` helpers should delegate to the exported `enforcement.MatchesResource` and `enforcement.ResourceSpecificity` to keep the logic in one place.

## Fix

Remove `resSpecificity` from `pdp.go` and replace its call sites in `drift.go` with `enforcement.ResourceSpecificity` (already exported for exactly this purpose). Similarly replace the local `matchResource` call with `enforcement.MatchesResource`.

Add a test that asserts `resSpecificity` and `resourceSpecificity` return identical values for a representative set of patterns to catch any future divergence.

eunolabs:
## Summary

`resSpecificity` and `matchResource` in `cmd/eunox-mcp/pdp.go` were hand-rolled copies of the engine's `resourceSpecificity`/`matchesResource` in `pkg/enforcement`. Two implementations of the same glob-match-and-score logic created a silent-divergence risk: the `ManifestPDP`/drift path and the raw `Engine` could rank overlapping glob patterns differently for the same request as the copies evolved independently.

## What changed

- Both helpers are now thin wrappers that **strip the namespace prefix, then delegate** to the already-exported `enforcement.MatchesResource` / `enforcement.ResourceSpecificity` — mirroring how the sibling `matchBare`/`bareSpecificity` already delegate. The scoring/matching algorithm now lives in exactly one place.
- The prefix stripping is **retained** — it is load-bearing for drift detection (`drift.go` matches prefixed `tool:` targets against bare tool names) and for `pdp_jwt.go`'s allowed-values matching. Inlining the exported engine functions *without* the strip would make `"tool:foo"` never match bare `"foo"` and silently disable drift detection. So this is a pure dedup: **behavior is unchanged**.
- Dropped the now-unused `"path"` import.

## Test

Added `TestResSpecificity_MatchesEngine` asserting the proxy helpers return identical results to the engine functions across a representative pattern set (exact/non-match, prefix/suffix/multiple wildcards, `?`, `[abc]` class, bare `*`, and a Unicode-heavy pattern carrying `[` — the case flagged in the issue) for every namespace prefix. If anyone reintroduces a local reimplementation, this test fails.

## Verification

- `go build ./...`, `go vet ./cmd/eunox-mcp/`, and `gofmt` all clean
- `go test -race` on `cmd/eunox-mcp` and `pkg/enforcement` pass; full `go test ./...` green

Fixes #180

https://claude.ai/code/session_01UgD1Lxz6WyYyxqQKhtpJ1h

---
_Generated by [Claude Code](https://claude.ai/code/session_01UgD1Lxz6WyYyxqQKhtpJ1h)_


eunolabs:

cmd/eunox-mcp/coverage_gaps_test.go
```diff
@@ -145,6 +145,49 @@ func TestResSpecificity_NonWildcardNoMatch(t *testing.T) {
...
+		{"*_file", "read_file"},
+		{"a_*_b_*", "a_x_b_y"},
+		{"read_?ile", "read_file"},
+		{"read_[abc]", "read_a"},
+		{"*", "anything"},
+		{"naïve_[résumé]", "naïve_r"},
+	}
+	prefixes := []string{"tool", "resource", "prompt", "system"}
+	for _, tc := range cases {
+		wantScore := enforcement.ResourceSpecificity(tc.pattern, tc.name)
```
eunolabs, line 180: Useful regression guard, but worth being precise about what it can and can't catch.

Since `resSpecificity`/`matchResource` are now literally `enforcement.ResourceSpecificity(stripNamespacePrefix(…))`, this assertion reduces to *"`stripNamespacePrefix("tool:"+pattern) == pattern` and the call delegates."* It's structurally unable to detect a divergence between the engine's algorithm and the **old** proxy algorithm (the actual #180 concern), because that old algorithm no longer exists in-tree to diff against.

What it does well and is worth keeping: (a) it fails if anyone re-inlines a divergent local copy, and (b) it exercises `stripNamespacePrefix` for each prefix.

One concrete gap it could close: the wildcard **scoring magnitude** is pinned nowhere today — `TestResSpecificity_Wildcard` only asserts `> 0`, and this test only asserts wrapper == engine. So nothing fails if the formula changes (e.g. `read_*` should score `5*10 - 1 = 49`). Consider adding a hardcoded expected score for one wildcard case here to lock that contract. Non-blocking.

---

eunolabs (commented):
## Review: dedupe resource specificity/match scoring (#185)

**Overview** — Replaces the hand-rolled `resSpecificity`/`matchResource` in `cmd/eunox-mcp/pdp.go` with thin wrappers that strip the namespace prefix and delegate to the already-exported `enforcement.ResourceSpecificity`/`MatchesResource`, mirroring the existing `bareSpecificity`/`matchBare`. Drops the now-unused `path` import and adds a cross-check test.

### Correctness / behavior-equivalence — verified ✓
I diffed the old proxy bodies against the engine functions they now delegate to:
- **`matchResource`**: old `return err == nil && matched` is exactly equivalent to the engine's `if err != nil { return false }; return matched`. Identical.
- **`resSpecificity`**: the only textual difference is wildcard counting — old `strings.Count(bare,"*")+Count("?")+Count("[")` vs the engine's rune loop over `"*?["`. These are provably equal for **all** valid UTF-8: `*` `?` `[` are single-byte ASCII (`0x2A`/`0x3F`/`0x5B`) and can never appear as a UTF-8 continuation/lead byte, so the count is identical even for the `naïve_[résumé]` case. The prefix loop was already rune-based in both. So "behavior unchanged" holds, including the Unicode case #180 worried about. (Arguably the divergence #180 feared never actually existed for the count — but consolidating to one implementation is still the right call.)

### Prefix-stripping retention — correct
Keeping the strip is load-bearing: `drift.go` matches prefixed `tool:` targets against bare tool names, and `pdp_jwt.go`'s `matchesAllowedValues` relies on it. Inlining the engine funcs *without* the strip would silently disable drift detection, exactly as the PR description states. Good that this was preserved.

### Conventions
SPDX headers intact, Conventional Commit title, pure-dedup scope with an accompanying test, fail-closed semantics untouched. No docs change needed (no behavior/grammar/schema change).

### Verification I ran on the PR head (`287ab2f`)
- `go build ./...` — clean
- `go vet ./cmd/eunox-mcp/` — clean
- `gofmt -l` on the two changed files — clean
- `go test -race -count=1 ./cmd/eunox-mcp/ ./pkg/enforcement/` — **all green**

### Suggestions (non-blocking)
- See the inline note on the new test: it's a fine regression guard, but it's self-referential (asserts the wrapper equals the function it wraps) and doesn't pin the wildcard scoring *magnitude*, which no test currently does. One hardcoded expected score would close that gap.

No blocking issues — this is a clean, behavior-preserving dedup. Nice work.

---

eunolabs:

cmd/eunox-mcp/coverage_gaps_test.go
eunolabs: Good catch — addressed in afaa917. Rather than add the magnitude to the delegation test (which would mix two concerns), I pinned it in the dedicated `TestResSpecificity_Wildcard`, which previously only asserted `> 0`. It now locks `read_* -> 5*10 - 1 = 49`, so a change to the scoring formula fails a test instead of passing silently. (The engine side already pins this in `pkg/enforcement/coverage_test.go` via `4*10-1` for `file*`/`fileabc`; this mirrors it on the cmd side.) The delegation test stays as the re-inlining/strip guard you noted.

---

eunolabs (commented):
Independent re-review of the latest revision (afaa917).

Verified the dedup is genuinely behavior-preserving: the removed inlined `matchResource`/`resSpecificity` were byte-for-byte equivalent to the engine's `MatchesResource`/`ResourceSpecificity` they now delegate to —

- Same short-circuits (`bare == "*" || bare == name` for match; `== name -> 1000`, no-metachar `-> 900` for score).
- The old `strings.Count(bare,"*")+Count("?")+Count("[")` wildcard count equals the engine's rune-iteration count for **all** inputs, since only single-byte ASCII metacharacters are ever counted — so even the Unicode-`[` case the issue flagged cannot diverge.
- The namespace-prefix strip is correctly retained ahead of the delegation, which is load-bearing for drift.go / pdp_jwt.go.

The magnitude pin from the earlier thread (`read_* -> 49`) is in `TestResSpecificity_Wildcard`, and `TestResSpecificity_MatchesEngine` remains a good re-inlining guard. CI is green. No further comments from me.

---

PR #186, opened 2026-06-13
fix: surface directive serialization failures in BuildRegoInput instead of silently dropping

### LINKED ISSUE #182: bug: `directivesToRegoInput` silently discards directives that fail JSON marshaling
## Summary

`directivesToRegoInput` in `pkg/enforcement/handlers.go` silently drops any directive that fails `json.Marshal` or `json.Unmarshal`. A Rego policy condition that inspects `input.directives` to understand the obligations that will apply post-allow would see fewer directives than actually exist, potentially making decisions on incomplete information — with no log entry, no error, and no audit record.

## Location

`pkg/enforcement/handlers.go:77–91`

```go
func directivesToRegoInput(directives []capability.Directive) []interface{} {
    out := make([]interface{}, 0, len(directives))
    for _, d := range directives {
        b, err := json.Marshal(capability.DirectiveWrapper{Directive: d})
        if err != nil {
            continue   // <-- silently dropped
        }
        var m map[string]interface{}
        if err := json.Unmarshal(b, &m); err != nil {
            continue   // <-- silently dropped
        }
        out = append(out, m)
    }
    return out
}
```

## Why this matters

`input.directives` is advertised in `BuildRegoInput` as "always `[]` — never null" and the doc comment says "Expose matched directives on req so BuildRegoInput can include them in input.directives." A policy author who writes `count(input.directives) > 0` to gate a decision on whether redaction is in effect would get a wrong answer if a directive was silently dropped.

The manifests (trusted input) produced by `LoadManifest` only contain `RedactFieldsDirective`, which marshals cleanly today. But:
1. A future directive type that is not fully JSON-roundtrippable would be silently invisible to Rego.
2. A programmatically constructed `Constraint` carrying a misbehaving directive would produce the same silent gap.
3. The silent drop makes the code harder to reason about defensively.

## Fix

Return an error from `directivesToRegoInput` (or log the failure and include a sentinel in the output slice), so the caller can either abort or at least surface a diagnostic:

```go
func directivesToRegoInput(directives []capability.Directive) ([]interface{}, error) {
    out := make([]interface{}, 0, len(directives))
    for i, d := range directives {
        b, err := json.Marshal(capability.DirectiveWrapper{Directive: d})
        if err != nil {
            return nil, fmt.Errorf("directive[%d] marshal failed: %w", i, err)
        }
        var m map[string]interface{}
        if err := json.Unmarshal(b, &m); err != nil {
            return nil, fmt.Errorf("directive[%d] unmarshal failed: %w", i, err)
        }
        out = append(out, m)
    }
    return out, nil
}
```

The marshal/unmarshal round-trip could also be replaced by a more direct conversion to avoid the double allocation, but correctness is the priority here.

eunolabs:
## What

`directivesToRegoInput` (`pkg/enforcement/handlers.go`) used `continue` to skip any directive whose `json.Marshal`/`json.Unmarshal` round-trip failed, so `input.directives` could silently come up shorter than the obligations actually attached to a decision. A Rego policy that gates on those obligations — e.g. `count(input.directives) > 0` to confirm redaction is in effect — would then decide on incomplete information, with no log, error, or audit record. That is a fail-*open* gap, contrary to the repo's fail-closed invariant.

## Fix

- `directivesToRegoInput` now returns `([]interface{}, error)` and returns a wrapped error (with index and type) instead of dropping a directive.
- The public `BuildRegoInput` helper now returns `(map[string]interface{}, error)` and propagates the failure, so a `PolicyEvaluator` can deny rather than decide on a short slice. `BuildRegoInput` has no production callers — only external evaluators and tests — so the signature change is a clean pre-1.0 break.
- The marshal/unmarshal round-trip is preserved (it reuses `DirectiveWrapper`'s type-discriminator logic); correctness, not the double allocation, was the priority the issue called out.

**No behavior change for current manifests** — only `RedactFieldsDirective` exists today and it marshals cleanly. This closes the forward-looking gap: a future non-roundtrippable directive type, or a programmatically constructed `Constraint` carrying a misbehaving directive, would otherwise be invisible to Rego.

## Tests & docs

- Updated all `rego_input_test.go` call sites; the two spy evaluators now deny on error, modeling a real evaluator.
- Added `TestBuildRegoInput_UnserializableDirective_ReturnsError`, which feeds a directive whose payload type `marshalDirective` does not recognize and asserts a non-nil error + nil input (auto-included in the CI `TestBuildRegoInput_` run).
- Documented the error return and fail-closed contract in `docs/capability-manifest-guide.md`.

## Verification

`go build ./...`, `go vet ./...`, full `go test -race ./...`, `make check-fmt`, and `make check-license` all pass.

Fixes #182

https://claude.ai/code/session_01WCk9JWNhqiknG3MB9o4tRz

---
_Generated by [Claude Code](https://claude.ai/code/session_01WCk9JWNhqiknG3MB9o4tRz)_


eunolabs:

pkg/enforcement/rego_input_test.go
```diff
@@ -668,14 +705,49 @@ func TestBuildRegoInput_MultipleDirectives(t *testing.T) {
...
+
+// unserializableDirective is a Directive whose payload type is unknown to
+// capability.marshalDirective, so DirectiveWrapper.MarshalJSON returns an error.
+// It models a future or programmatically constructed directive that is not
+// JSON-roundtrippable — exactly the case that used to be silently dropped.
+type unserializableDirective struct{}
+
+func (unserializableDirective) DirectiveType() string { return "unserializable" }
+
+func TestBuildRegoInput_UnserializableDirective_ReturnsError(t *testing.T) {
```
eunolabs, line 729: Good direct coverage of the helper. One gap: the fail-closed *chain* through the engine isn't exercised. The new deny-on-error branches in `spyBuildRegoInputEvaluator` (L578-580) and `spyDirectiveEvaluator` (L938-940) are never hit — every test that wires those spies uses a valid `RedactFieldsDirective` or none, and this test calls `BuildRegoInput` directly rather than going through `EvaluateConditions`/`ValidateAction`.

Consider a sibling of `TestEvaluateConditions_PopulatesDirectivesOnReq` (L781) that puts `unserializableDirective{}` into `constraint.Directives`, runs `e.EvaluateConditions(...)`, and asserts the result is a denial. The harness is already there, so it's a few lines — and it would both cover those spy branches and prove the full engine → policy condition → evaluator → `BuildRegoInput` error → deny path actually fails closed, which is the behavior this PR exists to protect. Non-blocking.

---

pkg/enforcement/handlers.go
```diff
@@ -67,27 +79,32 @@ func BuildRegoInput(ctx context.Context, req *capability.EnforceRequest) map[str
...
+	for i, d := range directives {
 		b, err := json.Marshal(capability.DirectiveWrapper{Directive: d})
 		if err != nil {
-			continue
+			return nil, fmt.Errorf("directive[%d] (%T) marshal failed: %w", i, d, err)
 		}
 		var m map[string]interface{}
 		if err := json.Unmarshal(b, &m); err != nil {
-			continue
+			return nil, fmt.Errorf("directive[%d] (%T) unmarshal failed: %w", i, d, err)
```
eunolabs, line 103: Minor/defensive: with the only directive marshaler that exists today, a successful `json.Marshal` always yields a JSON object (and a nil `Directive` marshals to `null`, which `json.Unmarshal` accepts into a map without error), so this unmarshal-error branch can't actually fire with current types and is uncovered. Fine to keep as fail-closed symmetry with the marshal branch above — just flagging it's untested, in case a future directive's `MarshalJSON` ever emits non-object JSON and you want a test pinning this arm too. Non-blocking.

---

eunolabs (commented):
Reviewed the change (3 files, +124/-23). It converts `directivesToRegoInput` and `BuildRegoInput` to return an error instead of silently `continue`-dropping any directive that fails the JSON round-trip, and propagates the failure so a `PolicyEvaluator` can deny rather than decide on a short `input.directives`. This is the right call and matches the repo's fail-closed invariant.

**Verified locally on `a6c5cbc`:**
- `go build ./...`, `go vet ./pkg/enforcement/...`, `gofmt -l` (clean), and `go test -race ./pkg/enforcement/...` all pass.
- `TestBuildRegoInput_UnserializableDirective_ReturnsError` genuinely exercises the marshal-error path (the `default` arm of `capability.marshalDirective`), not a no-op.
- Confirmed the "no production callers" claim: `handlePolicy` passes the static `pc.Input` + `req` to the evaluator and never calls `BuildRegoInput` itself, so the only in-repo callers are tests. The signature change is a clean pre-1.0 break with no internal call site left unupdated.

**Design note (not a change request):** because `BuildRegoInput` is invoked *inside* each external evaluator, the engine can't enforce the new fail-closed contract on the evaluator's behalf — an adapter that ignores the returned error still fails open. The PR handles this the only way it can (return the error + document the contract in the manifest guide) and is honest that it's forward-looking, since `RedactFieldsDirective` marshals cleanly today. Worth remembering if an OPA/Cedar adapter ever lands in-repo.

Two non-blocking inline notes below (an end-to-end deny test that would also cover the now-unreachable spy branches, and an unmarshal branch that can't fire with current types). Nothing blocking — looks good to merge.

---

eunolabs:
Reviewed the current diff. No more comments from this review.

eunolabs (commented):
Independent re-review of the latest revision.

The fail-closed design is in the right place. I confirmed `BuildRegoInput` has no production callers (only `rego_input_test.go` plus comment references in engine.go), so the `(map[string]interface{}, error)` signature change is a clean pre-1.0 break. And critically, the engine's `handlePolicy` passes `req` to the evaluator and does **not** build the Rego input itself — the evaluator calls `BuildRegoInput` — so surfacing the error from the helper and requiring evaluators to deny on it is the only place this guard can live. There's no central engine hook to put it in without changing the `PolicyEvaluator` contract, so the approach is correct.

The chain test suggested in the earlier thread is now present (`TestEvaluateConditions_UnserializableDirective_FailsClosed`) and exercises the full engine -> policy condition -> evaluator -> `BuildRegoInput` error -> deny path, and asserts the evaluator captured nothing. The unmarshal-branch coverage note remains a fair, intentional fail-closed-symmetry call (it can't fire with today's only marshaler). CI is green. No further comments from me.

---

PR #187, opened 2026-06-13
fix: expose all JWT claims to Rego policy conditions

### LINKED ISSUE #181: usability: `jwtClaimsAsMap` exposes only 4 hardcoded JWT claims to Rego policy conditions
## Summary

`jwtClaimsAsMap` in `cmd/eunox-mcp/pdp.go` builds the `input.claims` object that Rego policy conditions receive, but it only copies four hardcoded fields (`sub`, `iss`, `task_id`, `agent_id`). Any other claim present in the JWT — `tenant_id`, `roles`, `region`, `org_id`, custom enterprise claims — is silently dropped and unavailable to Rego policies.

## Location

`cmd/eunox-mcp/pdp.go:667–686`

```go
func jwtClaimsAsMap(ctx context.Context) map[string]interface{} {
    c, ok := jwtClaimsFromContext(ctx)
    if !ok {
        return nil
    }
    m := make(map[string]interface{}, 4)
    if c.Subject != "" {
        m["sub"] = c.Subject
    }
    if c.Issuer != "" {
        m["iss"] = c.Issuer
    }
    if c.TaskID != "" {
        m["task_id"] = c.TaskID
    }
    if c.AgentID != "" {
        m["agent_id"] = c.AgentID
    }
    return m
}
```

## Impact

A Rego policy that references `input.claims.tenant_id` or any other IdP claim outside the four hardcoded fields will always see the field as `undefined`, which in OPA evaluates as falsy. This silently makes the policy less restrictive than intended — the policy author writes a claim-based restriction that never fires.

Since the `policy` condition is the primary mechanism for arbitrary claim-based access control, this limitation significantly constrains its utility without any warning to the operator.

## Reproduction

1. Write a Rego rule: `allow { input.claims.tenant_id == "acme" }`
2. Issue a JWT with `tenant_id: "acme"` and any other JWT
3. Both are indistinguishable from Rego's perspective: `input.claims.tenant_id` is `undefined` for both

## Root cause

`JWTClaims` (the struct stored in context) models only the four eunox-specific claims. Any additional claim from the IdP is not stored and therefore cannot reach Rego.

## Fix options

**Option A** — Add an `Extra map[string]interface{}` field to `JWTClaims` and populate it from the raw token claims during JWT verification. Merge it into `jwtClaimsAsMap` as the base map before adding the four named fields (so named fields always win on key collision).

**Option B** — Store the raw parsed JWT claims map alongside the structured `JWTClaims` and use it as the direct `input.claims` source.

Either option should document the claim-name precedence rules (what happens if a custom claim is named `"sub"`) and add a test that verifies non-standard claims reach Rego input.

eunolabs:
## Summary

`jwtClaimsAsMap` built the `input.claims` object for Rego policy conditions from only four hardcoded fields (`sub`, `iss`, `task_id`, `agent_id`). Any other claim an IdP issued — `tenant_id`, `roles`, `region`, `org_id`, custom enterprise claims — was silently dropped, so a policy like `allow { input.claims.tenant_id == "acme" }` always saw `undefined` (falsy in OPA) and was silently less restrictive than intended.

## Changes

- **`cmd/eunox-mcp/pdp_jwt.go`** — Add an `Extra map[string]interface{}` field to `JWTClaims` and capture the full raw top-level claim set during verification by adding a `map[string]interface{}` destination to the existing `tok.Claims(...)` call. Since `tok.Claims` returns before unmarshaling on a signature failure, `Extra` is only ever populated for a verified token (fail-closed preserved).
- **`cmd/eunox-mcp/pdp.go`** — Rewrite `jwtClaimsAsMap` to merge `Extra` as the base of `input.claims`, then overlay the canonical fields. The four canonical keys are *reserved* and sourced exclusively from the validated/verified values:
  - A token cannot override `sub`/`iss`/`task_id`/`agent_id` by planting a same-named top-level claim (canonical wins on collision).
  - When a canonical value is empty, the reserved key is **absent** rather than falling back to a raw claim — so `input.claims.task_id` is only ever the verified `mcp.task_id`.
  - All other claim names (and the nested `mcp` object) pass through verbatim.

## Tests & docs

- 3 unit tests (`pdp_target_claims_test.go`): custom-claim exposure, reserved-key precedence over spoofed top-level claims, empty-canonical fall-through guard.
- 1 end-to-end test (`pdp_jwt_test.go`): a signed token with `tenant_id`/`region`/`roles` is verified and those claims reach both `claims.Extra` and the `input.claims` map.
- Precedence rules documented in `docs/capability-manifest-guide.md` (policy-input section + table); curated **Fixed** entry added to `CHANGELOG.md`.

Existing `jwtClaimsAsMap` tests pass unchanged (the change is additive). Full `go test -race ./...`, `go vet`, and `gofmt` all pass.

Fixes #181

https://claude.ai/code/session_019ys3DZnknDEZYcSmFAv8rE

---
_Generated by [Claude Code](https://claude.ai/code/session_019ys3DZnknDEZYcSmFAv8rE)_


eunolabs:

cmd/eunox-mcp/pdp_jwt.go
```diff
@@ -92,6 +92,15 @@ type JWTClaims struct {
...
 	Issuer          string
+	// Extra holds every raw top-level claim from the verified token, keyed by
+	// claim name exactly as the IdP emitted it: the standard fields (sub, iss,
+	// exp, iat, aud, ...), the nested mcp object, and any custom claims such as
+	// tenant_id, roles, region, or org_id.  It is the source of input.claims for
+	// Rego policy conditions (see jwtClaimsAsMap), so a policy can reference an
+	// arbitrary IdP claim instead of only the four eunox-canonical fields above.
+	// Populated only after the token's signature is verified.  Nil when no token
+	// was parsed.
+	Extra map[string]interface{}
```
eunolabs, line 103: Worth a threat-model note. `Extra` captures the *entire* verified claim set, and through `jwtClaimsAsMap` that whole set now crosses the proxy→`PolicyEvaluator` boundary (OPA/Cedar/external PDP) on every evaluation — previously only four fields did. If an IdP puts PII in the token (`email`, `name`, group lists, …), that data now leaves the proxy where it didn't before.

The change is correct and desirable — policies need the claims, and the audit log is unaffected (`audit.go` still reads only `AgentID`/`TaskID`). But `docs/threat-model-mcp.md` carefully enumerates what reaches the audit log and flags it as sensitive, while saying nothing about the policy-evaluator boundary. Per CONTRIBUTING's "behavior changes need a docs update" / threat-model invariant, consider one sentence there noting the full claim set (including any IdP PII) is now serialized into `input.claims` for external evaluators. Non-blocking.

---

cmd/eunox-mcp/pdp.go
```diff
@@ -661,15 +661,52 @@ func agentIDFromContext(ctx context.Context) string {
...
 func jwtClaimsAsMap(ctx context.Context) map[string]interface{} {
 	c, ok := jwtClaimsFromContext(ctx)
 	if !ok {
 		return nil
 	}
-	m := make(map[string]interface{}, 4)
+	m := make(map[string]interface{}, len(c.Extra)+4)
+	// Base layer: every raw top-level claim, minus the reserved keys, which are
+	// set authoritatively from the canonical fields below.
+	for k, v := range c.Extra {
```
eunolabs, line 704: The reserved-key guard is correctly top-level only: the nested `mcp` object passes through verbatim, so `input.claims.mcp.task_id` / `input.claims.mcp.agent_id` are raw token values with no shadowing protection. Today that's harmless because the canonical `task_id`/`agent_id` are *sourced from* `mcp.*`, so the nested and top-level copies always agree.

Flagging only as a forward-looking watch-point: if the canonical sourcing ever changes (e.g. `task_id` starts coming from somewhere other than `mcp.task_id`), `input.claims.mcp.task_id` would silently diverge from the authoritative `input.claims.task_id` and could mislead a policy author who reaches into the nested object. A one-line doc note ("prefer the top-level reserved keys; the nested `mcp.*` copies are unguarded raw values") would future-proof it. Non-blocking.

---

cmd/eunox-mcp/pdp.go
```diff
@@ -661,15 +661,52 @@ func agentIDFromContext(ctx context.Context) string {
...
+// claim never overrides these four (the canonical value wins on collision), and
+// when a canonical value is empty the reserved key is omitted rather than
+// falling back to the raw claim.  All other claim names pass through verbatim.
 func jwtClaimsAsMap(ctx context.Context) map[string]interface{} {
 	c, ok := jwtClaimsFromContext(ctx)
 	if !ok {
 		return nil
 	}
-	m := make(map[string]interface{}, 4)
+	m := make(map[string]interface{}, len(c.Extra)+4)
```
eunolabs, line 701: Nit (non-blocking): the capacity hint over-allocates by the number of reserved keys actually present in `Extra` — commonly `sub`, since standard tokens carry it at the top level, so the real entry count is usually `len(c.Extra)+4` minus those. Negligible (a few buckets); noting only for completeness.

---

eunolabs (commented):
## Review: fix: expose all JWT claims to Rego policy conditions

Solid, security-conscious fix for #181. The core change is correct and the fail-closed posture is preserved. I checked the branch out, read the full functions (not just the diff hunks), traced the data flow, and ran the tests.

### What I verified
- **Fail-closed preserved.** The whole change rests on "`rawClaims` is only populated after signature verification." Confirmed against the dependency source: in `go-jose/v4@v4.1.4`, `JSONWebToken.Claims(key, dest...)` calls `t.payload(key)` (= `sig.Verify`) first and returns on error **before** the `json.Unmarshal` loop over the destinations. Adding `&rawClaims` as a 4th destination cannot populate it for an unverified token.
- **Spoofing prevented.** `reservedClaimKeys` are stripped from the `Extra` base layer and re-applied only from the canonical (validated/verified) values; canonical wins on collision; empty canonical → key absent (no fallback to a planted raw claim). The three unit tests cover exposure, precedence, and the empty-canonical guard exactly.
- **No audit-log leak.** `audit.go` reads only `c.AgentID` / `c.TaskID` from context, never `Extra`, so the wider claim set neither bloats nor leaks into the signed audit record. The full map flows only to the policy evaluator (the intended consumer).
- **Tests pass.** `go test -race -count=1 -run 'TestJWTClaimsAsMap|TestJWTPDP_ValidateToken' ./cmd/eunox-mcp/` → ok. `go vet` clean, `gofmt` clean on all four files, commit is DCO signed-off.
- Docs (guide + CHANGELOG) updated with the precedence table; the e2e test correctly exercises the JSON round-trip (`[]string` → `[]interface{}`).

### Comments left inline
- **(medium)** Threat-model coverage of the proxy→policy-evaluator boundary — `pdp_jwt.go`.
- **(low)** Nested `mcp` object is unguarded raw passthrough — forward-looking note, `pdp.go`.
- **(nit)** Map capacity-hint over-allocation — `pdp.go`.

None are blocking. The reserved-key design and the "absent rather than fall back to a raw claim" rule are the right calls.

---

eunolabs:
Reviewed the current diff. No more comments from this review.

eunolabs:
Thanks for the thorough review. Addressed the two doc comments in e162747:

- **Threat-model boundary (medium)** — Added §6.5 "Verified JWT claims at the policy-evaluator boundary" to `docs/threat-model-mcp.md`: `input.claims` now serializes the full verified claim set (incl. any IdP PII) to external `policy`-condition evaluators, noting it's verification-gated, never written to `audit.jsonl`, and stays in-process unless an operator's evaluator forwards it. Added change-log row 1.8.
- **Nested `mcp` passthrough (low)** — Added a note in the capability-manifest guide's precedence section: the reserved-key guard is top-level only, so prefer the top-level reserved keys over the unguarded nested `mcp.*` copies (they agree today only because the canonical values are sourced from `mcp.*`).

On the **capacity-hint nit**: leaving `len(c.Extra)+4` as-is. It's a deliberate upper bound — computing the exact post-strip count would need a pre-pass over `Extra`, trading a few unused map buckets for more code, which isn't a good trade for a size hint. Noted for the record.

---

docs/threat-model-mcp.md
```diff
@@ -599,6 +609,7 @@ We aim to acknowledge security reports within 48 hours and to provide an initial
...
 | 1.7     | 2026-06-12 | Eunolabs Platform Security | §7.3 added: Adversarial test coverage table mapping the published `cmd/eunox-mcp/adversarial_test.go` suite (Adv-1 prompt injection, Adv-2 tool poisoning, Adv-3 read-then-exfiltrate, Adv-4 overbroad JWT claim, Adv-5 novel-method fail-closed, Adv-6 rate-limit session-ID bypass) to the §3 attack classes and residual risks it exercises. Renumbered Reporting security issues to §7.4. |
+| 1.8     | 2026-06-13 | Eunolabs Platform Security | §6.5 added: `input.claims` now exposes the entire verified JWT claim set (not just `sub`/`iss`/`task_id`/`agent_id`) to external `policy`-condition evaluators, so any IdP PII in the token crosses the proxy→evaluator boundary. Documented that this is verification-gated, never written to the audit log, and stays in-process unless an operator-supplied evaluator forwards it. |
```
eunolabs: Merge conflict with `main` lands right here. Since this branch was opened, `main` gained its own `| 1.8 |` row (#171: JWKS-must-be-https, no-redirect upstream, right-most `X-Forwarded-For`, audit-drop markers), so this `| 1.8 |` row collides — the PR's `mergeable_state` is `dirty`, and `git merge-tree origin/main HEAD` reports a CONFLICT in this file.

To resolve: rebase on `main` and renumber this row to `| 1.9 |` (bump the version reference in the prose accordingly if you cite it anywhere). The `### 6.5` section itself is fine as-is — `main` only goes up to §6.4, so no section renumber is needed. Nothing in the code changed, just the changelog/threat-model table needs the rebase.

---

eunolabs (commented):
Independent re-review of the latest revision (e162747).

The code is correct and the three prior threads are addressed: the threat-model §6.5 boundary note, the nested-`mcp` "prefer top-level reserved keys" doc note, and the capacity-hint reply (keeping `len(c.Extra)+4` as a deliberate upper bound is fine). I also confirmed the mechanics:

- `tok.Claims(&matchingKeys[i], &stdClaims, &payload, &rawClaims)` populates `rawClaims` from the verified payload (go-jose unmarshals into each destination), and it's declared inside the per-key loop so a failed key can't leak stale claims into a later success. Verification-gated, as the doc says — Test CI is green.
- The reserved-key overlay (`Extra` base minus reserved keys, then canonical `sub`/`iss`/`task_id`/`agent_id` set only when non-empty) correctly prevents a token from shadowing identity, and the three unit tests pin precedence + the empty-canonical omission.
- Audit posture is unchanged — `audit.go` still reads only `AgentID`/`TaskID`; `input.claims` is never written to the log.

One blocking item before this can merge: **a merge conflict with `main` in `docs/threat-model-mcp.md`** (left inline). `main` already added a `| 1.8 |` row via #171, so this branch's `| 1.8 |` row needs to become `| 1.9 |` after a rebase. That's the only thing standing between this and green.

---

eunolabs:
Rebased onto `main` (030190c) and resolved the changelog conflict in `docs/threat-model-mcp.md`: kept `main`'s `| 1.8 |` row (#171) and renumbered this branch's row to `| 1.9 |`, bumping the document header **Version** to 1.9 to match. The `### 6.5` section is unchanged (no section renumber needed, as you noted). No code changes; `go build`, `go test -race ./...`, and `gofmt` are all clean against the rebased tree. Should be mergeable now.

---

eunolabs (commented):
Re-reviewed after the rebase (`7772f87`): the `docs/threat-model-mcp.md` conflict is resolved — the new entry is correctly `| 1.9 |` with main's `| 1.8 |` (#171) preserved, and §6.5 sits cleanly after §6.4 (no section renumber needed). I also confirmed no *new* conflict was introduced by the #185 and #186 merges that landed in the meantime — the `pdp.go` overlap with #185 and the `docs/capability-manifest-guide.md` overlap with #186 both merge cleanly (`git merge-tree origin/main HEAD` is clean). The rebase left the code unchanged from the version I already verified.

No further comments from me — CI is the only remaining gate. Resolved the outdated conflict thread.

---

PR #188, opened 2026-06-13
fix: validate ipRange CIDRs and timeWindow timestamps at manifest load

### LINKED ISSUE #179: usability: ipRange CIDRs and timeWindow timestamps are not validated at manifest load time
## Summary

The `ipRange` condition's CIDR list and the `timeWindow` condition's `notBefore`/`notAfter` timestamps are parsed per-request at runtime rather than once at manifest load time. A typo (e.g. `"10.0.1/24"` or `"2026-13-01T00:00:00Z"`) passes `LoadManifest` silently and only surfaces as a `CONDITION_FAILED` denial on the first live request, which is hard to diagnose.

## Locations

**Runtime CIDR parse** — `pkg/enforcement/handlers.go:197–204`
```go
for _, cidr := range ipr.CIDRs {
    _, network, err := net.ParseCIDR(cidr)
    if err != nil {
        return &ConditionError{
            ...
            Message: fmt.Sprintf("invalid CIDR in condition: %s", cidr),
        }
    }
    ...
}
```

**Runtime time parse** — `pkg/enforcement/handlers.go:123–130` and `145–152`
```go
notBefore, err := time.Parse(time.RFC3339, tw.NotBefore)
if err != nil {
    return &ConditionError{
        ...
        Message: fmt.Sprintf("invalid notBefore time: %s", tw.NotBefore),
    }
}
```

## Impact

A manifest with a malformed CIDR or timestamp:
1. Passes `eunox-mcp validate` (the subcommand calls `LoadManifest` which does not validate these fields)
2. Loads and starts the proxy without error
3. Denies the **first real request** that hits the affected capability with a `CONDITION_FAILED` code

The operator sees a policy denial — exactly what a bad request would produce — and may not immediately recognize it as a configuration error rather than a legitimate enforcement event. The audit log entry also looks like a normal denial rather than a load-time misconfiguration.

## Expected behavior

`LoadManifest` (or the `validate` subcommand) should parse CIDRs and RFC3339 timestamps in the loaded constraints and return a descriptive error at startup/validate time.

## Suggested fix

In `cmd/eunox-mcp/manifest.go`, inside the `validateLocalManifest` condition loop, add type-specific checks:

```go
case *capability.IPRangeCondition:
    for k, cidr := range v.CIDRs {
        if _, _, err := net.ParseCIDR(cidr); err != nil {
            return fmt.Errorf("capability %d, condition %d: invalid CIDR %q: %w", i, j, cidr, err)
        }
    }

case *capability.TimeWindowCondition:
    if v.NotBefore != "" {
        if _, err := time.Parse(time.RFC3339, v.NotBefore); err != nil {
            return fmt.Errorf("capability %d, condition %d: invalid notBefore %q: %w", i, j, v.NotBefore, err)
        }
    }
    if v.NotAfter != "" {
        if _, err := time.Parse(time.RFC3339, v.NotAfter); err != nil {
            return fmt.Errorf("capability %d, condition %d: invalid notAfter %q: %w", i, j, v.NotAfter, err)
        }
    }
```

As an additional performance benefit, the parsed `*net.IPNet` objects could be cached on the condition struct (or in a separate lookup table) to avoid repeated allocation on the hot enforcement path.

eunolabs:
fixes #179

## Problem

`ipRange` CIDRs and `timeWindow` `notBefore`/`notAfter` timestamps were parsed per-request at runtime, not at manifest load time. A typo (e.g. `"10.0.1/24"` or `"2026-13-01T00:00:00Z"`) passed `LoadManifest` — and therefore `eunox-mcp validate` and proxy startup — silently, and only surfaced as a `CONDITION_FAILED` denial on the first live request. That reads like a legitimate enforcement event, so the operator may not recognize it as a configuration error.

## Fix

Parse both at load time in `validateLocalManifest` via two new helpers that follow the existing `validateMaxCalls` / `validateAllowedValues` pattern (handling both value and pointer condition forms):

- **`validateIPRange`** — rejects an empty `cidrs` list and any entry that fails `net.ParseCIDR`.
- **`validateTimeWindow`** — rejects a window with neither bound, a non-RFC3339 `notBefore`/`notAfter`, and an inverted/empty window (`notBefore` at or after `notAfter`).

Errors are path-qualified (`capability at index N, condition M: …`) and wrap the underlying parse error. Alongside syntax, the degenerate shapes that deny *every* call are rejected too, matching the existing philosophy for `maxCalls` and `allowedValues`.

Because every manifest-loading path — single proxy, gateway routes, and the `validate` subcommand — funnels through `LoadManifest`, the validation applies uniformly.

The runtime parse paths in `pkg/enforcement/handlers.go` are **kept** as a fail-closed safety net for conditions that do not pass through manifest validation (JWT capability claims, library-constructed conditions).

## Tests

- New `TestLoadManifest_IPRangeAndTimeWindowValidation` — 12 table-driven cases (allow + deny + malformed, IPv4/IPv6, index-naming).
- `go test -race -count=1 ./cmd/eunox-mcp/ ./pkg/...` passes; `gofmt` and `go vet` clean.
- End-to-end with the built binary: `validate` rejects a bad CIDR and a bad timestamp with descriptive errors (exit 1) and accepts a well-formed manifest (exit 0).

## Docs

`docs/capability-manifest-guide.md` documents the `timeWindow` and `ipRange` conditions and their load-time validation.

https://claude.ai/code/session_01JfD2qC2CmJpQbHJxXC7bHA

---
_Generated by [Claude Code](https://claude.ai/code/session_01JfD2qC2CmJpQbHJxXC7bHA)_


eunolabs:

cmd/eunox-mcp/manifest.go
```diff
@@ -339,6 +357,56 @@ func validateMaxCalls(i, j int, v *capability.MaxCallsCondition) error {
...
+// legitimate enforcement event. An empty cidrs list matches no source IP and
+// denies every call, so it is almost certainly an authoring mistake (cf.
+// validateAllowedValues' empty-values check). Caught here, the operator learns
+// at validate-time rather than on the first denied call.
+func validateIPRange(i, j int, v *capability.IPRangeCondition) error {
+	if len(v.CIDRs) == 0 {
+		return fmt.Errorf("capability at index %d, condition %d: ipRange requires a non-empty 'cidrs' list; an empty list matches no source IP and denies every call", i, j)
+	}
+	for _, cidr := range v.CIDRs {
+		if _, _, err := net.ParseCIDR(cidr); err != nil {
```
eunolabs: Optional (non-blocking): `net.ParseCIDR` accepts a CIDR with host bits set — e.g. `10.0.0.5/8` parses fine and returns the masked network `10.0.0.0/8`, which `handleIPRange` then uses via `network.Contains`. So `cidrs: ["10.0.0.5/8"]` validates and silently behaves as `10.0.0.0/8`. That's consistent with the runtime, so nothing is broken.

But since this validator exists precisely to catch likely-typo'd CIDRs before they hit a live request, you could optionally also flag host bits (compare `cidr` against `network.String()`), turning a probable `/32`-vs-network mistake into a load-time error. Happy to leave as-is given the runtime parity — just flagging the choice.

---

cmd/eunox-mcp/manifest_validation_test.go
```diff
@@ -976,6 +976,194 @@ capabilities:
...
+version: "0.1.0"
+capabilities:
+  - target: "tool:get_invoice"
+    actions: [call]
+    conditions:
+      - type: timeWindow
+        notBefore: "2026-12-31T23:59:59Z"
+        notAfter: "2026-04-01T00:00:00Z"
+`,
+			wantErr: "the window is empty and denies every call",
```
eunolabs, line 1150: Nice, thorough table. One boundary it doesn't hit: `notBefore == notAfter` (same instant). The check in `validateTimeWindow` is `!notBefore.Before(notAfter)`, which rejects *equal* bounds as well as strictly-inverted ones — but every case here is strictly inverted (Dec → Apr) or single-bound.

Consider adding an equal-bounds case (identical timestamp for both) that expects rejection. It pins the `!Before` boundary, so a future refactor to e.g. `notAfter.Before(notBefore)` — which would silently start *accepting* the single-instant window — gets caught by a failing test.

---

eunolabs (commented):
## Review: load-time validation for `ipRange` / `timeWindow` (#188)

Solid, focused change. It does exactly what #179 asks — moves CIDR and RFC3339 parsing to manifest load so a typo (`10.0.1/24`, `2026-13-01T00:00:00Z`) surfaces at `validate`/startup instead of masquerading as a `CONDITION_FAILED` denial on the first live request — and it follows the established `validateMaxCalls` / `validateAllowedValues` pattern closely: both value- and pointer-form switch cases, path-qualified errors, `%w`-wrapped parse errors, and rejection of the degenerate "deny-everything" shapes.

**Verified while reviewing:**
- **Single choke point holds.** All four `LoadManifest` call sites (single-proxy `main.go`, gateway `main.go`, `route.go`, `doctor.go`) funnel through `validateLocalManifest`, so the new checks apply uniformly to single-proxy, gateway routes, and the `validate` subcommand, as the PR claims.
- **Runtime parity.** Validation uses the same `net.ParseCIDR` / `time.Parse(time.RFC3339, …)` as `handleIPRange` / `handleTimeWindow`, so nothing that loads can later fail to parse at runtime. Keeping the runtime parse paths as a fail-closed net for JWT/library-constructed conditions is the right call.
- **Fail-closed philosophy.** Rejecting empty `cidrs`, a bound-less `timeWindow`, and an inverted/empty window matches the existing "deny-everything shapes are almost certainly authoring mistakes" stance (cf. `validateAllowedValues`).
- No new `.go` files (SPDX N/A), docs updated in the same PR, and the only file under `schemas/` is the gateway *config* schema — the capability manifest itself has no JSON Schema to keep in sync here.

**Two non-blocking notes left inline** — an optional host-bits-in-CIDR thought on `validateIPRange`, and an equal-bounds (`notBefore == notAfter`) test-coverage suggestion.

**One question:** CONTRIBUTING/CLAUDE.md asks manifest-grammar changes to also land in the spec repo (`eunolabs/mcp-capability-manifest`). These read as validation-tightening rather than new grammar, so it may not apply — but "≥1 bound required / non-empty `cidrs` / non-inverted window" are now author-visible rules. Worth a companion note in the spec, or are these intentionally implementation-level only?

No blocking issues from me — nice work.

---

eunolabs:

cmd/eunox-mcp/manifest.go
eunolabs: Done in 17f155d. `validateIPRange` now rejects CIDRs with host bits set — comparing the parsed host IP to the masked network IP (`!ip.Equal(network.IP)`) rather than string-matching `network.String()`, so it doesn't false-positive on case/format. The error points at the canonical network address (or suggests `/32` for a single host). Docs updated.

---

cmd/eunox-mcp/manifest_validation_test.go
eunolabs: Added in 17f155d — an equal-bounds case (identical `notBefore`/`notAfter`) that expects rejection, pinning the `!notBefore.Before(notAfter)` boundary.

---

eunolabs (commented):
Independent re-review of the latest revision (17f155d).

Verified the validation reaches every load path, which is the load-bearing claim: single proxy (`main.go`), **gateway routes** (`route.go:227`), `doctor`, and the `validate` subcommand all funnel through `LoadManifest` -> `validateLocalManifest`, so no manifest-loading entry point skips the new checks.

- `validateIPRange`: the host-bit rejection via `!ip.Equal(network.IP)` is correct for IPv4 and IPv6 (an IP-value comparison, so it won't false-positive on formatting/case), and empty-`cidrs` is rejected like the sibling `allowedValues` empty check.
- `validateTimeWindow`: `!notBefore.Before(notAfter)` correctly rejects both equal and inverted bounds; the equal-bounds case from the earlier thread pins that boundary.
- Keeping the runtime `net.ParseCIDR` / `time.Parse` paths in `handlers.go` is the right call — they remain the fail-closed net for conditions that bypass manifest validation (JWT claims, library-constructed conditions).

Both prior threads are addressed and CI is green. No further comments from me.

---

PR #189, opened 2026-06-13
fix: enforce integer-ness for type: integer in argumentSchema

### LINKED ISSUE #178: bug: ArgumentSchema `integer` type constraint accepts any float64, including non-integer values
## Summary

The JSON Schema validator in `pkg/enforcement/schema.go` treats `integer` as an alias for `number`, meaning a value like `3.14` passes validation when the schema declares `type: integer`. This renders the `integer` type constraint a no-op.

## Location

`pkg/enforcement/schema.go:99–101` — `schemaTypeCompatible`

```go
func schemaTypeCompatible(gotType, schemaType string) bool {
    return gotType == schemaType || (schemaType == "integer" && gotType == "number")
}
```

JSON has a single numeric type. Both `42` and `3.14` decode as `float64` in Go, and `schemaJSONTypeOf` returns `"number"` for all of them. So `schemaTypeCompatible("number", "integer")` is unconditionally `true` — any float64 value satisfies an `integer` schema.

## Reproduction

```go
schema := &capability.ArgumentSchema{
    Type: capability.SchemaType{Single: "integer"},
}
err := enforcement.ValidateArgumentSchema(
    map[string]interface{}{"count": 3.14},
    schema,
)
// err == nil — should be a type error
```

## Impact

An operator who writes an `argumentSchema` with `type: integer` to constrain a parameter to whole numbers gets no protection. A call with `count: 3.14` passes the structural check and proceeds to condition evaluation.

This is a security-relevant gap when the condition logic or upstream tool interprets the argument differently depending on whether it is an integer (e.g. SQL `LIMIT 3.14` vs `LIMIT 3`).

## Fix

After confirming the value is a `float64`, check that it has no fractional part:

```go
func schemaValidateNumber(p string, v float64, s *capability.ArgumentSchema) error {
    if s.Type.Single == "integer" || slices.Contains(s.Type.Multiple, "integer") {
        if v != math.Trunc(v) {
            return fmt.Errorf("%s: expected integer, got non-integer number %g", p, v)
        }
    }
    if s.Minimum != nil && v < *s.Minimum {
        return fmt.Errorf(...)
    }
    ...
}
```

Alternatively, add the fractional check inside `schemaTypeCompatible` itself. Either way, a new test case in `pkg/enforcement/` should verify that `3.14`, `0.1`, and `-2.9` fail an `integer` schema, while `42`, `0`, and `-3` pass.

eunolabs:
Fixes #178

## Root cause

`schemaTypeCompatible` in `pkg/enforcement/schema.go` treated `integer` as an unconditional alias for `number`:

```go
return gotType == schemaType || (schemaType == "integer" && gotType == "number")
```

JSON has a single numeric type, so `42` and `3.14` both decode to `float64` and `schemaJSONTypeOf` reports both as `"number"`. The clause was therefore always `true` for any float — a `type: integer` schema was a no-op. The security angle: a value like `LIMIT 3.14` passed structural validation as if it were a whole number, then proceeded to condition evaluation / the upstream.

## The fix

Integer-ness is now part of type compatibility: a number satisfies `integer` only when it has no fractional part (`math.Trunc(v) == v`).

Handling this in `schemaTypeCompatible`/`schemaCheckType` rather than `schemaValidateNumber` (the issue's primary suggestion) keeps the multi-type case correct: each listed type is evaluated independently, so `3.14` still satisfies `[integer, number]` (via `number`) but is rejected by `[integer, string]`. The `schemaValidateNumber` approach with `slices.Contains(Multiple, "integer")` would have wrongly rejected `3.14` under `[integer, number]`.

The single-type integer mismatch yields a specific message (`expected integer, got non-integer number 3.14`) since the generic `expected integer, got number` is misleading — `42` is also `"number"` yet passes.

## Tests & docs

- `TestValidateArgumentSchema_IntegerType` — `3.14`/`0.1`/`-2.9` deny, `42`/`0`/`-3` pass (the cases requested in the issue).
- `TestValidateArgumentSchema_IntegerType_MultiType` — `[integer, number]` vs `[integer, string]`.
- `TestArgumentSchema_IntegerType_DeniedThroughEngine` — engine-level, asserts `INVALID_PARAMS` before any condition runs, plus the whole-number allow path.
- Updated `internal_coverage_test.go` for the new 3-arg signature (added a fractional-number deny case).
- Documented the `integer` vs `number` distinction in `docs/capability-manifest-guide.md`.

## Verification

- `go build ./...`, `go test -count=1 ./...` (full suite), `go vet ./pkg/enforcement/`, `golangci-lint run ./pkg/enforcement/` (0 issues) — all green.

https://claude.ai/code/session_01RFxKgxJdwoaaQQNimbEXpH

---
_Generated by [Claude Code](https://claude.ai/code/session_01RFxKgxJdwoaaQQNimbEXpH)_


eunolabs:

eunolabs (commented):
## Review — enforce integer-ness for `type: integer`

The fix is correct and the approach is the right one. What I verified:

- **Root-cause fix is sound.** `schemaTypeCompatible` now requires `math.Trunc(f) == f` for a number to satisfy `integer`, removing the previously-unconditional `integer ⇐ number` alias. `42`/`0`/`-3` pass; `3.14`/`0.1`/`-2.9` are denied with `INVALID_PARAMS` before any condition runs — closing the `LIMIT 3.14` gap from #178. `math.Trunc` is the right comparison here (avoids the `int64(f)` overflow trap).
- **Good call handling this in `schemaTypeCompatible` rather than `schemaValidateNumber`.** The issue's suggested `slices.Contains(Multiple, "integer")` check would wrongly reject `3.14` under `[integer, number]`. Evaluating each listed type independently keeps `[integer, number]` accepting `3.14` (via `number`) while `[integer, string]` rejects it — `TestValidateArgumentSchema_IntegerType_MultiType` locks this in.
- **The `val.(float64)` assertion in the tailored-error branch is safe** — it's guarded by `got == "number"`, which `schemaJSONTypeOf` only returns for `float64`. `Inf`/`NaN` can't reach it since `encoding/json` rejects them when decoding into `interface{}`.
- Tests cover allow/deny/multi-type/engine-level paths, the `internal_coverage_test.go` signature is updated with a fractional deny case, docs explain the `integer` vs `number` distinction, and CI is green across all 42 checks. Fail-closed behavior is preserved.

One optional, non-blocking nit: the tailored message (`expected integer, got non-integer number 3.14`) only fires on the single-type path. For a multi-type schema like `[integer, string]`, `3.14` falls through to the generic `value type "number" does not match any of [integer string]`, which reads slightly oddly since the value *is* a number. Reasonable to leave as-is given multi-type-with-`integer` is an edge case — just flagging it.

LGTM.

---

eunolabs:

PR #190, opened 2026-06-13
sec: fail closed on empty actions list in Engine.ValidateAction

### LINKED ISSUE #177: bug: Engine.ValidateAction fails open when constraint actions list is empty
## Summary

`Engine.ValidateAction` in `pkg/enforcement/engine.go` treats an empty `actions` list as "permit any action" (fail-open), violating the project's load-bearing fail-closed invariant.

## Location

`pkg/enforcement/engine.go:525–535` — `actionMatchScore`

```go
func actionMatchScore(actions []string, _ *capability.EnforceRequest) bool {
    if len(actions) == 0 {
        return true   // <-- fails open
    }
    for _, a := range actions {
        if a == "call" || a == "*" {
            return true
        }
    }
    return false
}
```

## Reproduction

Call `engine.ValidateAction` with a `Constraint` whose `Actions` field is nil or empty:

```go
engine := enforcement.New()
resp, _ := engine.ValidateAction(ctx, req, []capability.Constraint{
    {Target: "tool:read_file", Actions: nil, Conditions: nil},
})
// resp.Decision == "allow" — should be "deny"
```

The manifest loader (`validateLocalManifest`) rejects empty `actions` at load time, so this path is not reachable through `LoadManifest`. It is reachable when:
- Constraints are constructed programmatically (tests, embedded usage, gateway routes)
- A manifest is deserialized through a path that bypasses `LoadManifest`

## Expected vs Actual

| | Expected | Actual |
|---|---|---|
| `Actions: nil` | deny (AUTHORIZATION_FAILED) | **allow** |
| `Actions: []string{}` | deny | **allow** |

## Impact

Direct violation of the fail-closed invariant stated in `CONTRIBUTING.md`:
> On any ambiguity — missing manifest, unmapped method, malformed JWT, unset condition argument — deny.

An empty `actions` list is an ambiguous/incomplete constraint. The engine should deny, not allow.

Note that `ManifestPDP.Decide` is not affected because it performs its own action check via `containsAction` after `findConstraint` returns — but any caller that uses `Engine.ValidateAction` directly (e.g. external integrations, the hosted gateway's token-capability path) hits this fail-open.

## Fix

```go
func actionMatchScore(actions []string, _ *capability.EnforceRequest) bool {
    for _, a := range actions {
        if a == "call" || a == "*" {
            return true
        }
    }
    return false
}
```

Removing the empty-list early-return makes the function fail closed for any actions list that does not explicitly include `"call"` or `"*"`, consistent with how the manifest validator treats empty actions.

A test case for `enforcement_gaps_test.go` should cover `Actions: nil` and `Actions: []string{}` both returning `DecisionDeny`.

eunolabs:
## What was wrong

`actionMatchScore` in `pkg/enforcement/engine.go` had an early return treating an empty `actions` list as "permit any action":

```go
if len(actions) == 0 {
    return true   // fails open
}
```

This made `Engine.ValidateAction` **allow** a call against a constraint with `Actions: nil` or `Actions: []string{}` — a direct violation of the project's fail-closed invariant. The manifest loader rejects empty actions at load time, but the engine is reachable directly (external integrations, the gateway's token-capability path, or any deserialization that bypasses `LoadManifest`), so the engine itself must fail closed.

## The fix

Removed the early return so the function matches only when the list explicitly contains `"call"` or `"*"` — aligning the engine with how the manifest validator already treats empty actions. Updated the doc comment accordingly (and fixed its stale `Returns (0, true)` line, since the function returns a single `bool`).

## Tests

- **Updated** two tests that codified the old fail-open behavior:
  - `TestEngine_EmptyActions_MatchesAny` → `TestEngine_EmptyActions_FailsClosed` (now table-driven over `nil` and `[]string{}`, asserting deny + `AUTHORIZATION_FAILED`).
  - The `coverage_test.go` "empty actions matches" subtest → "empty actions fails closed".
- **Added** to `cmd/eunox-mcp/enforcement_gaps_test.go`:
  - `TestEmptyActions_ValidateAction_FailsClosed` — exercises the affected `Engine.ValidateAction` path directly for both `nil` and `[]string{}`.
  - `TestEmptyActions_ManifestPDP_StillDenies` — a regression guard for the issue's note that `ManifestPDP.Decide` is unaffected (its own `containsAction` check keeps denying).

## Verification

- Full repo `go test -race -count=1 ./...` — all green.
- `make check-fmt` clean, `go vet` clean.

No docs grammar update was needed — this is a behavior bug fix that brings the engine in line with the already-documented fail-closed invariant.

Fixes #177

https://claude.ai/code/session_01LRkm7bXB7iuQfUH56hxvTM

---
_Generated by [Claude Code](https://claude.ai/code/session_01LRkm7bXB7iuQfUH56hxvTM)_


eunolabs:

pkg/enforcement/coverage_test.go
```diff
@@ -500,13 +500,14 @@ func TestEngine_MatchingEdgeCases(t *testing.T) {
...
-	t.Run("empty actions matches", func(t *testing.T) {
+	t.Run("empty actions fails closed", func(t *testing.T) {
 		t.Parallel()
 		constraints := []capability.Constraint{{Target: "tool", Actions: nil}}
 		req := &capability.EnforceRequest{ToolName: "tool"}
 		resp, err := e.ValidateAction(context.Background(), req, constraints)
 		require.NoError(t, err)
-		assert.Equal(t, capability.DecisionAllow, resp.Decision)
+		assert.Equal(t, capability.DecisionDeny, resp.Decision)
+		assert.Equal(t, capability.ErrCodeAuthorizationFailed, resp.Denial.Code)
```
eunolabs, line 511: Minor (non-blocking): this dereferences `resp.Denial.Code` without first asserting it's non-nil, so a regression back to fail-open (Decision == Allow, Denial == nil) would `panic` here rather than fail cleanly. The sibling tests you added/updated guard this with `require.NotNil(t, resp.Denial)` (see `enforcement_test.go` and `enforcement_gaps_test.go`). Worth adding the same `require.NotNil(t, resp.Denial)` line here for consistency and a clean failure mode.

---

eunolabs (commented):
## Review: `sec: fail closed on empty actions list in Engine.ValidateAction`

**Verdict: looks good — correct, minimal, and well-tested.** One minor (non-blocking) test-robustness nit inline.

### What it does
Removes the `if len(actions) == 0 { return true }` early return in `actionMatchScore`, so a constraint with `Actions: nil`/`[]string{}` no longer matches "any action". An empty list is now a non-match, the constraint is skipped in `findMatchingCapability`, and `ValidateAction` falls through to deny with `AUTHORIZATION_FAILED`. This brings the engine in line with the fail-closed invariant.

### Correctness — verified against the surrounding code
- **Blast radius is exactly the right surface.** `ValidateAction`/`FindMatchingCapability` have no internal production callers; they're the public engine API used by embedded/external integrations and deserialization paths that bypass `LoadManifest` — precisely where an incomplete constraint could previously fail open.
- **Loader claim holds.** Empty actions are already rejected at load time (`cmd/eunox-mcp/manifest.go:126` — `'actions' must not be empty`), so no valid on-disk manifest is affected; this only hardens the programmatic path. No user-facing grammar change, so the "docs in same PR" convention doesn't apply here.
- **`ManifestPDP` is genuinely unaffected.** It uses its own `findConstraint` + `containsAction` check (`cmd/eunox-mcp/pdp.go:206`), independent of `actionMatchScore`. The `TestEmptyActions_ManifestPDP_StillDenies` regression guard is a good call to lock that in.
- Doc comment is corrected, including the stale `Returns (0, true)` line (the function returns a single `bool`).

### Tests
Thorough: table-driven over both `nil` and `[]string{}`, asserting deny + `AUTHORIZATION_FAILED`, and covering both the engine path and the `ManifestPDP` path. I ran the affected tests locally (`go test` for `pkg/enforcement` and `cmd/eunox-mcp`, filtered to the empty-actions cases) — all green.

### Nit
- `pkg/enforcement/coverage_test.go` dereferences `resp.Denial.Code` without a `require.NotNil` guard, unlike the sibling tests — see inline comment.

Nice, tightly-scoped security fix.

---

eunolabs:

PR #198, opened 2026-06-13
fix: reject malformed allowedValues glob patterns at manifest load

### LINKED ISSUE #196: bug: allowedValues glob patterns not validated at manifest load; malformed patterns silently dropped at runtime
## Summary

The `allowedValues` condition supports `path.Match` glob patterns in its `values` list. A malformed pattern (e.g., an unclosed character class `[invalid`) causes `path.Match` to return `path.ErrBadPattern`; the handler silently skips the pattern and moves on. The manifest validator does not check string values for valid glob syntax, so the operator receives no feedback at `eunox-mcp validate` time and no structured error at runtime.

### Runtime handler silently swallows `path.ErrBadPattern`

`pkg/enforcement/handlers.go`, lines 642–648:

```go
        if matched, err := path.Match(pattern, str); err == nil && matched {
            return nil
        }
    }
}
```

When `path.Match` returns `path.ErrBadPattern`, the `err == nil` guard is false and the pattern is treated as non-matching. No log, no structured error, no audit hint. The remaining values in the list are tried, and the request is denied if none match — so the behaviour is fail-closed. But the operator has no way to know a pattern is broken short of carefully auditing denied requests.

### Manifest validator does not check glob syntax

`cmd/eunox-mcp/manifest.go`, `validateAllowedValues` (lines 318–326) only checks for an empty `Argument` field and an empty `Values` list:

```go
func validateAllowedValues(i, j int, v *capability.AllowedValuesCondition) error {
    if v.Argument == "" { ... }
    if len(v.Values) == 0 { ... }
    return nil
}
```

A value like `[invalid` passes without error. The operator only discovers the problem when a call is unexpectedly denied with no `conditionType` or `details` pointing to the broken pattern.

## Impact

Operators writing complex `allowedValues` conditions with glob patterns (e.g. `/reports/[0-9]{4}/*`) who accidentally mistype the pattern will have a silently broken policy. The fail-closed behaviour prevents a bypass, but the operator is left with a policy that is more restrictive than intended and no clear error message.

## Fix

In `validateAllowedValues` (or a new `validateAllowedValuesPatterns` helper), iterate over the `Values` list and for each string value call `path.Match(v, "")`. If it returns `path.ErrBadPattern`, return a validation error:

```go
for k, val := range v.Values {
    if s, ok := val.(string); ok {
        if _, err := path.Match(s, ""); err != nil {
            return fmt.Errorf("capability at index %d, condition %d, value %d: invalid glob pattern %q: %w", i, j, k, s, err)
        }
    }
}
```

This catches bad patterns at `eunox-mcp validate` time and on proxy startup, consistent with how `ValidateResourcePattern` guards against bad target globs.

eunolabs:
## Summary

Fixes #196.

A string value in an `allowedValues` condition doubles as a `path.Match` glob pattern at runtime. A malformed pattern (e.g. an unclosed character class `[invalid`) makes `path.Match` return `path.ErrBadPattern`, which `handleAllowedValues` treats as a non-match and silently skips. The behaviour stays fail-closed, but the operator is left with a policy that is quietly **more restrictive** than intended and no error pointing at the broken pattern — neither at `eunox-mcp validate` time nor on proxy startup.

## Change

`validateAllowedValues` (`cmd/eunox-mcp/manifest.go`) now walks the `values` list and runs each **string** value through `path.Match(s, "")`, returning a structured load error on `path.ErrBadPattern`:

```
capability at index 0, condition 0, value 0: allowedValues contains invalid glob pattern "/reports/[invalid": syntax error in pattern
```

This catches the mistake at load — the same way `enforcement.ValidateResourcePattern` already guards target globs. Non-string values (numbers, bools) are matched by exact equality at runtime, never fed to `path.Match`, so they are skipped by the validation.

The runtime handler is unchanged: its fail-closed fall-through is still correct, and with load-time validation a malformed pattern can no longer reach it via `LoadManifest`.

## Tests

Added table cases to `TestLoadManifest_AllowedValuesAndMaxCallsValidation`:
- a malformed glob (`/reports/[invalid`) is rejected with `invalid glob pattern`
- a manifest mixing a non-string value (`10`) with a valid glob still loads

## Docs

Updated `docs/capability-manifest-guide.md`: the note previously said a malformed pattern "does not match and falls through to deny" — it now states such a pattern is rejected by `eunox-mcp validate` and on startup, consistent with invalid target globs.

`make test`, `gofmt`, and `go vet` are clean for the affected package.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DHfs3ByAPoLknpR752t6ZW)_


cmd/eunox-mcp/manifest.go
```diff
@@ -341,6 +342,23 @@ func validateAllowedValues(i, j int, v *capability.AllowedValuesCondition) error
...
+	// is quietly more restrictive than intended, with no error to tell the operator
+	// the pattern is broken. Reject bad globs here so the mistake surfaces at
+	// validate-time and on proxy startup, the same way ValidateResourcePattern
+	// guards target globs.
+	for k, val := range v.Values {
+		s, ok := val.(string)
+		if !ok {
+			continue
+		}
+		if _, err := stdpath.Match(s, ""); err != nil {
```
eunolabs, line 371: Worth confirming this trade-off is intended (non-blocking).

The runtime handler tries `reflect.DeepEqual` (exact match) **before** `path.Match` (`handlers.go:649-665`). So a string value with an unbalanced metacharacter — e.g. `/reports/[invalid` — was previously usable as a legitimate **exact-match literal**: the manifest loaded, and an argument whose value was exactly that string was allowed via the `DeepEqual` branch (the `path.Match` branch silently skipped it). This validation can't distinguish "author meant a literal" from "author meant a glob", so it now rejects such a manifest at load.

That's a defensible call, and the guide already steers literals-containing-metacharacters toward `argumentSchema` `enum`/`pattern`. The only nit: the doc note frames the rejection purely as a "malformed glob pattern", so an operator who was leaning on the exact-match path for an unbalanced-bracket literal won't immediately see why their previously-valid manifest now fails to load. A half-sentence in the guide noting that the exact-match escape hatch is closed for such strings too would remove the surprise.

---

eunolabs (commented):
## Review: reject malformed `allowedValues` glob patterns at load

**Verdict: looks good.** Focused, well-motivated fix for a real fail-closed-but-silent gap, and it follows the codebase's established pattern closely.

### What's solid
- **Mirrors the existing precedent exactly.** `path.Match(s, "")` to surface `ErrBadPattern` without a real candidate string is precisely what `enforcement.ValidateResourcePattern` (`engine.go:510`) already does for target globs — same technique, same load-time-rejection rationale.
- **Correctly scoped to strings.** The runtime handler matches numbers/bools/nil by `reflect.DeepEqual` and never feeds them to `path.Match`, so validating only the `string` values is right. The `[10, "/data/*"]` case pins this.
- **Structured, indexed error** (`capability … condition … value …`) consistent with the sibling validators (`validateMaxCalls`, `validateIPRange`, …), and it wraps the underlying `path.ErrBadPattern`.
- **`stdpath` import alias is the correct move** — `LoadManifest(path string)` already binds `path` as a parameter name, so an unaliased `path` import would shadow it.
- **Docs updated in the same PR**, and the runtime handler is left untouched — its fail-closed fall-through stays correct, and `TestEngine_AllowedValues_MalformedGlobFallsThrough` still exercises the handler directly (constructs the engine without `LoadManifest`), so it remains valid rather than contradictory.

### One thing to confirm (non-blocking)
See the inline note on `manifest.go`: because the runtime tries exact match before glob match, this also closes the (exotic) exact-match-literal path for unbalanced-bracket strings. Defensible, but a half-sentence in the guide would avoid a surprising load failure for anyone relying on it.

### Scope note (not a defect)
`pdp_jwt.go`'s `matchesAllowedValues` runs the same `path.Match` on glob values arriving from IdP JWT capability claims, which aren't validated at load (they're dynamic, and JWT can only restrict). A malformed glob there silently skips → strictly more restrictive → still fail-closed, so leaving it unguarded is fine. Noting the asymmetry only for the record.

### Verified locally (on `claude/beautiful-hawking-smufyo`)
- `gofmt -l` on both changed `.go` files: clean
- `go vet ./cmd/eunox-mcp/`: clean
- `go test -run TestLoadManifest_AllowedValuesAndMaxCallsValidation ./cmd/eunox-mcp/`: pass

---

cmd/eunox-mcp/manifest.go
eunolabs: Confirmed — intended. The loader can't distinguish "author meant a literal" from "author meant a glob", so failing closed at load (rather than silently honoring the exact-match path for one case while breaking it for globs) is the consistent choice. Added the half-sentence you suggested to the guide in 11396f5: it now notes that the exact-`DeepEqual` path is closed for unbalanced-bracket strings too, with `/reports/[invalid` as the worked example. Literals that must contain a metacharacter should go through `argumentSchema` `enum`/`pattern` instead.

---

eunolabs: Reads well — the worked example and the "DeepEqual runs before the glob match" framing make the load-time rejection unsurprising. That was my only note, and it's a non-blocking one, so nothing else from me. Resolving this thread.

---

eunolabs:
No more comments from this review. The current head includes the doc clarification for the exact-match-literal tradeoff, and I did not find additional issues.

---

PR #199, opened 2026-06-13
fix: deny non-string array items in allowedExtensions and recipientDomain

### LINKED ISSUE #197: bug: non-string array items silently dropped in allowedExtensions and recipientDomain handlers
## Summary

The `allowedExtensions` and `recipientDomain` condition handlers accept both a single-string argument and an array-of-strings argument for multi-file / multi-recipient calls. When the argument is a `[]interface{}` array, non-string items (e.g. numbers, booleans, null) are silently ignored via a failed type assertion. This means a structurally invalid MCP request slips through without a type error, and the condition evaluates only the subset of items that happen to be strings.

## Location

`pkg/enforcement/handlers.go` — `handleAllowedExtensions` (the `recipientDomain` handler has the same pattern):

```go
case []interface{}:
    for _, item := range t {
        if s, ok := item.(string); ok {   // non-string items: ok == false, silently skipped
            if s = strings.TrimSpace(s); s != "" {
                paths = append(paths, s)
            }
        }
    }
```

### Concrete example

An MCP tool call with argument:

```json
{ "paths": ["/data/report.pdf", 42, null] }
```

…is treated identically to:

```json
{ "paths": ["/data/report.pdf"] }
```

The request proceeds to the extension allowlist check as if `42` and `null` were never present. No denial, no audit detail, no error to the caller.

## Impact

1. **Correctness**: A request that is structurally malformed (wrong element type in array) is not rejected at the condition layer. Downstream tools may handle the mixed-type array differently (e.g., error on the number element), producing a confusing upstream failure with no corresponding proxy audit record.
2. **Consistency**: The `allowedValues` condition uses `reflect.DeepEqual` for non-string types, which would catch them. The `allowedExtensions` and `recipientDomain` conditions silently swallow them. The behaviour is inconsistent across condition types.

## Fix

Change the inner loop to return `ErrCodeConditionFailed` when a non-string item is encountered:

```go
case []interface{}:
    for _, item := range t {
        s, ok := item.(string)
        if !ok {
            return &ConditionError{
                ConditionType: capability.ConditionTypeAllowedExtensions,
                Message:       fmt.Sprintf("argument %q array contains a non-string item: %T", ae.Argument, item),
                Details:       map[string]interface{}{"argument": ae.Argument},
            }
        }
        if s = strings.TrimSpace(s); s != "" {
            paths = append(paths, s)
        }
    }
```

Apply the same fix to `handleRecipientDomain`. Failing closed on a type mismatch is consistent with the proxy's fail-closed invariant.

eunolabs:
## Summary

Fixes #197.

The `allowedExtensions` and `recipientDomain` condition handlers accept both a single-string argument and a `[]interface{}` array (for multi-file reads and multi-recipient sends). When the array carried a non-string element — a number, boolean, or null — the inner `item.(string)` type assertion failed silently and the item was skipped. A structurally malformed array such as:

```json
{ "paths": ["/data/report.pdf", 42, null] }
```

was therefore evaluated exactly as if it had been the valid subset `["/data/report.pdf"]` — no denial, no audit detail, no error to the caller. This violated the proxy's fail-closed invariant and was inconsistent with `allowedValues`, which catches non-string types via `reflect.DeepEqual`.

## Change

Both `handleAllowedExtensions` and `handleRecipientDomain` now return a `CONDITION_FAILED` `ConditionError` the moment a non-string item appears in the array, identifying the offending argument and the item's Go type. The single-string and all-strings-array paths are unchanged.

## Tests

- Updated `TestAllowedExtensions_ArrayArgument`: the prior `"array of non-strings is missing context"` case asserted the old silent-skip behavior (it ended up as `MissingContext` because every item was dropped). It is replaced by two cases that assert `Deny` / `CONDITION_FAILED` — one for a mixed array (`["/data/a.csv", 42, nil]`) and one entirely of non-strings (`[1, 2]`).
- Added a non-string fail-closed case to `TestRecipientDomain_ErrorBranches`.

## Docs

- `docs/capability-manifest-guide.md`: documented the fail-closed rule on non-string array items for `allowedExtensions` (noting `recipientDomain` shares it).
- `CHANGELOG.md`: added a Fixed entry under Unreleased.

`make check-fmt`, `go vet`, and the full `go test -race ./...` suite pass.

https://claude.ai/code/session_01LFKoaXnXVi2xFrK1vLRX2X

---
_Generated by [Claude Code](https://claude.ai/code/session_01LFKoaXnXVi2xFrK1vLRX2X)_


pkg/enforcement/handlers.go
```diff
@@ -388,11 +388,18 @@ func (e *Engine) handleAllowedExtensions(_ context.Context, cond capability.Cond
...
 			for _, item := range t {
-				if s, ok := item.(string); ok {
-					if s = strings.TrimSpace(s); s != "" {
-						paths = append(paths, s)
+				s, ok := item.(string)
+				if !ok {
+					return &ConditionError{
+						Code:          capability.ErrCodeConditionFailed,
+						ConditionType: capability.ConditionTypeAllowedExtensions,
+						Message:       fmt.Sprintf("argument %q array contains a non-string item: %T", ae.Argument, item),
```
eunolabs, line 396: **Consistency gap: the third array-accepting handler, `allowedTables`, still silently drops non-string items.**

This fix is correct, and the rationale here ("matching the array handling in `allowedTables` and `recipientDomain`" / "consistent with the type-strictness of `allowedValues`") is the right one. But after this PR, `allowedTables` is the odd one out — `parseTableArgument` (`handlers.go:838-865`) has the exact silent-skip pattern this PR is removing:

```go
case []interface{}:
    var out []capability.TableAccess
    for _, item := range t {
        out = append(out, parseTableArgument(item)...) // non-string item -> returns nil -> dropped
    }
    return out
```

A non-string element (number, bool, null) hits none of the `string` / `[]interface{}` / `map` cases, so the recursive call returns `nil` and the item is silently swallowed — `{"tables": ["users", 42, null]}` is evaluated exactly as `["users"]`. The same is true for non-string **columns** at `handlers.go:856-861` (`if col, ok := c.(string); ok && col != ""` skips non-strings). This is the identical correctness/consistency problem #197 describes, just in the SQL-table handler instead of the path/recipient handlers.

Issue #197 only named `allowedExtensions` and `recipientDomain`, so leaving `allowedTables` out is a defensible scope decision — but since the PR's whole justification is fail-closed consistency across the array-accepting conditions, I'd either extend the same `ConditionFailed`-on-non-string guard to `parseTableArgument` here, or file a follow-up issue so it's tracked rather than quietly left inconsistent.

---

eunolabs (commented):
## Review — fix: deny non-string array items in `allowedExtensions` and `recipientDomain`

The core fix is correct and resolves #197. Both `handleAllowedExtensions` and `handleRecipientDomain` now return a structured `CONDITION_FAILED` the instant a non-string array element appears, instead of letting the `item.(string)` assertion silently shrink a malformed array to its valid subset. This is the right fail-closed behavior and matches the proxy invariant. `nil`/`null` is correctly covered (it fails the assertion → deny), the single-string and all-strings paths are untouched, `fmt` is already imported, and the `ConditionType` on each error is correct.

**Tests / docs / CI:** Test coverage is solid — mixed array and all-non-strings for `allowedExtensions`, mixed array for `recipientDomain`, all asserting `Deny` + `CONDITION_FAILED` + the `"non-string item"` message. The misleading `"array of non-strings is missing context"` case is correctly replaced. CHANGELOG and the manifest guide are both updated. All CI checks are green (one `Types Sync` job was still queued at review time; nothing failing).

**One thing worth resolving before merge (non-blocking):** the third array-accepting handler — `allowedTables`, via `parseTableArgument` — still has the identical silent-skip pattern this PR removes (non-string table entries and non-string columns are dropped silently). Details in the inline comment. Since the PR's stated motivation is fail-closed *consistency* across these conditions, I'd either extend the guard there too or open a follow-up so it's tracked. Left as-is, the codebase fixes two of the three array handlers and leaves the third inconsistent.

**Minor / optional:** within an array, whitespace-only string items are still trimmed to empty and dropped (e.g. `["/data/a.csv", "  "]` → `["/data/a.csv"]`) rather than denied. That's defensible normalization and pre-existing, so not a blocker — just flagging it as the same "malformed array silently becomes a smaller valid one" shape for completeness.

Nice, tightly-scoped fix overall.

---

eunolabs:
Addressed the consistency gap in 36e8675: extended the same fail-closed guard to `allowedTables`.

`parseTableArgument` now returns an error on a non-string array item or a non-string `columns` entry, and `handleAllowedTables` maps it to `CONDITION_FAILED` with the argument name. So `{"tables": ["users", 42, null]}` now denies instead of being evaluated as `["users"]`, and a non-string column denies likewise. Added deny tests for both, and updated the manifest guide and CHANGELOG so all three array-accepting conditions (`allowedExtensions`, `recipientDomain`, `allowedTables`) are now consistent.

On the minor point — whitespace-only string items are still trimmed and dropped rather than denied. I left that pre-existing normalization as-is since it's defensible and out of scope for this fix; happy to tighten it in a follow-up if you'd prefer it also fail closed.

---

pkg/enforcement/handlers.go
eunolabs, line 396: The fix here is correct and faithfully implements the issue's prescription. One consistency gap remains, though: `allowedTables` still has the *same* silent-drop pattern this PR is closing for its two siblings — and after this PR it becomes the lone array-accepting condition that still treats a structurally-malformed array as a smaller valid one (the exact vuln class issue #197 describes).

It lives in `parseTableArgument` further down this file:

```go
case []interface{}:
    var out []capability.TableAccess
    for _, item := range t {
        out = append(out, parseTableArgument(item)...) // non-string item -> recursion returns nil -> silently dropped
    }
    return out
...
// and the columns list:
if col, ok := c.(string); ok && col != "" { // non-string column silently dropped
```

So `{"table": ["users", 42]}` is evaluated as `["users"]`, and `{"table":"users","columns":["id", 7]}` as `columns:["id"]` — no denial, no audit detail.

This also makes two claims in the PR slightly overstated: the comment a few lines up ("matching the array handling in allowedTables and recipientDomain") and the CHANGELOG's "matching ... the type-strictness of allowedValues". Worth either extending this PR to cover `parseTableArgument` (note it returns `[]TableAccess`, not `*ConditionError`, so the fix shape differs — the type check would have to move into `handleAllowedTables` or `parseTableArgument` would need to grow an error return), or filing a follow-up so the fail-closed treatment is uniform across all three.

---

eunolabs (commented):
## Review — `fix: deny non-string array items in allowedExtensions and recipientDomain`

### Overview
Closes the silent-drop bug from #197: `handleAllowedExtensions` and `handleRecipientDomain` now return `CONDITION_FAILED` the instant a `[]interface{}` array carries a non-string element, instead of skipping it and evaluating the string subset. The implementation matches the fix prescribed in the issue verbatim, the single-string and all-strings-array paths are untouched, and it strengthens the fail-closed invariant. Tests, docs, and CHANGELOG are all updated in the same PR. Nicely scoped.

### What's good
- **Correct and fail-closed.** The non-string check runs during the parse phase, before the allowlist loop, so a non-string anywhere in the array denies the whole call regardless of the other entries — `["/data/a.csv", 42]` now denies even though `.csv` is allowed. That's exactly the regression the issue called out.
- **Good audit shape.** `CONDITION_FAILED` + `Details{"argument": ...}` + a `%T` in the message reuses existing patterns; no new top-level audit field, so no threat-model churn.
- **Tests match the convention.** Allow + deny + malformed cases; the old `"array of non-strings is missing context"` case that asserted the buggy behavior is correctly replaced rather than left to rot.

### Findings
1. **(Main — consistency) `allowedTables` still has the same silent-drop gap.** See the inline comment on `handlers.go`. `parseTableArgument` drops non-string table-name items and non-string `columns` entries the same way the old code did. After this PR it's the only array-accepting condition that still treats a malformed array as a smaller valid one, which also makes the in-code comment ("matching the array handling in allowedTables") and the CHANGELOG line ("matching ... the type-strictness of allowedValues") read as overstated. Extend here or file a follow-up — your call, but it should be acknowledged.

2. **(Minor) Empty/whitespace-only string items are still silently dropped.** `["/data/a.csv", "  ", ""]` is evaluated as `["/data/a.csv"]` because the trim drops the blanks. The PR's stated principle — "a structurally malformed argument is never treated as a smaller, valid one" — arguably covers an empty-string element too. This is pre-existing and probably intentional cleanup, but worth a conscious decision rather than an accident.

3. **(Nit) Audit detail could name the offending element.** `Details` carries only the argument name; the item's *index* (and/or value/type) isn't captured in a structured field — only `%T` in the free-text message. For a multi-element array, an `"index"` key would let audit consumers pinpoint the bad element without re-parsing. Optional.

4. **(Nit) Test symmetry.** `allowedExtensions` got both a mixed-array and an all-non-string case; `recipientDomain` got only the mixed case. A one-line all-non-string case there would mirror it.

### Verdict
The core fix is correct, well-tested, and well-documented — no blocking issues in the changed lines. Finding #1 is the one I'd want a decision on (fix-now vs. follow-up) before this merges, since the PR's own framing leans on a cross-condition consistency that `allowedTables` doesn't yet satisfy. The rest are optional polish.

---

eunolabs:
No more comments from this review. The current head extends the fail-closed non-string handling to `allowedTables` as well, and I did not find additional issues.

---

eunolabs (commented):
## Re-review of the new head (`36e8675`)

The second commit extends the fail-closed treatment to `allowedTables`, which resolves my main finding. Verified:

- **`parseTableArgument` now returns an error instead of silently dropping.** The `[]interface{}` case explicitly rejects any item that isn't a `string` or `map` (number/bool/null → error rather than a recursive `nil`), and the `columns` loop errors on a non-string entry. `handleAllowedTables` propagates it as `CONDITION_FAILED` with the argument name in `Details`. Fail-closed and consistent with the two sibling handlers.
- **Tests** cover both new paths — the non-string table item and the non-string column.
- **Docs** (`CHANGELOG.md`, `capability-manifest-guide.md`) now accurately describe all three array-accepting conditions plus the `columns` list; the two claims I'd flagged as overstated are now correct.
- CI is green on this head (42/42 checks) and the PR is mergeable-clean.

No remaining blocking issues. The only leftover items are the optional nits from my first pass (empty/whitespace-string array items are still trimmed away across all three handlers; audit `Details` doesn't carry the offending item's index) — both pre-existing, uniform, and fail-closed, so fine to leave or defer.

LGTM. Resolving my earlier `allowedTables` thread as addressed.

---

eunolabs:
Quick follow-up on the remaining review nits:

- **Test symmetry** — added the all-non-string `recipientDomain` case in 0053d15 to mirror `allowedExtensions`, so all three array-accepting conditions now have symmetric mixed + all-non-string deny coverage.
- **Audit index of the offending element** — leaving as-is for now. It's a reasonable enhancement but would apply uniformly across all three handlers and touches the audit `Details` shape, so I'd rather do it as a focused follow-up than widen this fix.
- **Whitespace-only string items** — unchanged, as noted above: pre-existing normalization, deliberate.

The two `allowedTables` consistency threads are resolved (fixed in 36e8675).

---

PR #200, opened 2026-06-13
fix: re-open circuit breaker on a failed half-open probe

### LINKED ISSUE #191: Circuit breaker can stay closed after a failed half-open probe
## Summary

`pkg/circuitbreaker` closes the breaker on the first successful half-open probe. If `HalfOpenMaxProbes > 1`, another probe from the same half-open window can fail afterward, but `RecordFailure` then sees `StateClosed` and only increments the closed-state failure counter. With a normal threshold greater than 1, the breaker remains closed even though a half-open probe failed.

That violates the package contract in `breaker.go`: half-open allows limited probes, success closes, and failure re-opens. More importantly, it can return traffic to an upstream that is still failing under concurrent half-open probes.

## Reproducer

```go
func TestBreaker_HalfOpenProbeFailureAfterSuccessReopens(t *testing.T) {
    var mu sync.Mutex
    current := time.Date(2026, 1, 1, 0, 0, 0, 0, time.UTC)
    clock := func() time.Time {
        mu.Lock()
        defer mu.Unlock()
        return current
    }
    advanceClock := func(d time.Duration) {
        mu.Lock()
        defer mu.Unlock()
        current = current.Add(d)
    }

    cfg := circuitbreaker.Config{
        FailureThreshold:  3,
        CooldownDuration:  5 * time.Second,
        HalfOpenMaxProbes: 2,
    }
    b := circuitbreaker.New(cfg, circuitbreaker.WithClock(clock))

    for i := 0; i < cfg.FailureThreshold; i++ {
        b.RecordFailure()
    }
    advanceClock(6 * time.Second)

    require.True(t, b.Allow())
    require.True(t, b.Allow())

    b.RecordSuccess()
    b.RecordFailure()

    require.Equal(t, circuitbreaker.StateOpen, b.State())
}
```

Actual result on current `main`:

```text
expected a failed half-open probe to re-open the breaker, got "closed"
```

## Likely cause

`RecordSuccess` unconditionally transitions from `StateHalfOpen` to `StateClosed` on the first successful probe. A later `RecordFailure` from another admitted half-open probe no longer has any way to know it belonged to that half-open generation.

## Suggested fix

Track half-open probe outcomes by generation/window. Do not close until the breaker has seen the required successful probe result(s), and ensure any failure from an admitted half-open probe re-opens the breaker even if another probe completed successfully first.

eunolabs:
Fixes #191

## Problem

`RecordSuccess` unconditionally transitioned the breaker `StateHalfOpen → StateClosed` on the **first** successful half-open probe. With `HalfOpenMaxProbes > 1`, a second probe admitted in the same half-open window could fail afterward, but by then `RecordFailure` saw `StateClosed` and only incremented the closed-state consecutive-failure counter. With `FailureThreshold > 1`, the breaker stayed closed and routed traffic back to an upstream that was still failing under concurrent half-open probes — violating the package contract (`half-open allows limited probes; success closes, failure re-opens`).

## Fix

Track half-open probe successes per window (`halfOpenSuccess`), reset on every transition into half-open and on `Reset`:

- The breaker closes **only once every admitted probe** in the window has succeeded (`HalfOpenMaxProbes` successes), so it stays half-open while probes are still outstanding.
- **Any** failure from an admitted half-open probe re-opens the breaker immediately, even if an earlier probe in the same window already succeeded.
- Default `HalfOpenMaxProbes=1` behavior is unchanged: one success still closes, one failure still re-opens.

## Tests

- `TestBreaker_HalfOpenProbeFailureAfterSuccessReopens` — the reproducer from the issue (success then failure across two probes must re-open).
- `TestBreaker_HalfOpenClosesOnlyAfterAllProbesSucceed` — closes only after `HalfOpenMaxProbes` successes, not before.

`make test`, `make lint`, `make check-fmt`, and `make check-license` all pass.

https://claude.ai/code/session_01AXZUiEC3dobvymiujKGznW

---
_Generated by [Claude Code](https://claude.ai/code/session_01AXZUiEC3dobvymiujKGznW)_


pkg/circuitbreaker/breaker.go
```diff
@@ -129,9 +135,24 @@ func (b *Breaker) RecordSuccess() {
...
+			b.halfOpenSuccess = 0
+			b.lastTransitionAt = b.now()
+		}
+	}
+}
+
+// requiredHalfOpenSuccesses is the number of successful probes needed to close
+// the breaker from half-open. It mirrors HalfOpenMaxProbes (the probes admitted
+// per window) and is clamped to at least 1 so a degenerate config still closes.
+func (b *Breaker) requiredHalfOpenSuccesses() int {
```
eunolabs, line 162: `requiredHalfOpenSuccesses()` ties "successes needed to close" to `HalfOpenMaxProbes`, which overloads that config field. Before this PR `HalfOpenMaxProbes` meant only "max probes admitted per half-open window"; now it also means "successes required to close." Two follow-ups:

1. **Stale doc.** The `Config.HalfOpenMaxProbes` field comment still reads *"the number of probe requests allowed in half-open state"* — it no longer tells the whole story, since raising it also raises the bar to close. CONTRIBUTING requires a docs update in the same PR for behavior changes; please refresh that field comment (and the `Allow` doc, which is fine, but the Config one is the gap).

2. **Coupling.** This collapses two independent knobs into one. You can no longer express "admit N concurrent probes, close if any one succeeds" (the gobreaker `MaxRequests` model). The all-admitted-probes-must-succeed semantics are a defensible, conservative choice — just flagging it's a real semantic change, not only a bugfix. If you later want them decoupled, a separate `HalfOpenRequiredSuccesses` field would do it cleanly.

---

pkg/circuitbreaker/breaker.go
```diff
@@ -129,9 +135,24 @@ func (b *Breaker) RecordSuccess() {
 	b.consecutiveFails = 0
 
 	if b.state == StateHalfOpen {
-		b.state = StateClosed
-		b.lastTransitionAt = b.now()
+		b.halfOpenSuccess++
+		if b.halfOpenSuccess >= b.requiredHalfOpenSuccesses() {
```
eunolabs, line 150: Liveness edge case introduced for `HalfOpenMaxProbes > 1`: closing now requires `HalfOpenMaxProbes` distinct success records. If an admitted probe never reports an outcome, the breaker wedges half-open — `halfOpenProbes == HalfOpenMaxProbes` blocks new admissions in `Allow()`, `halfOpenSuccess < required` blocks closing here, and there is no failure to re-open it. It then rejects all traffic until `Reset()`.

Concrete in-tree trigger: `Do` (do.go) has no `recover`, so if `fn` panics *after* `Allow()` has consumed a probe slot, that slot is spent with no `RecordSuccess`/`RecordFailure`. Pre-PR a single success always closed the breaker, so one completed probe recovered it; with this change a single panicking probe can wedge a `>1` breaker indefinitely.

Worth either (a) documenting the "every admitted probe must report exactly one outcome" contract on the breaker, or (b) adding a `recover` in `Do` that records a failure before re-panicking. Default `HalfOpenMaxProbes=1` is unaffected.

---

pkg/circuitbreaker/breaker_test.go
```diff
@@ -204,6 +204,105 @@ func TestBreaker_HalfOpenFailureReopens(t *testing.T) {
 	}
 }
 
+// TestBreaker_HalfOpenProbeFailureAfterSuccessReopens guards the contract that a
+// failed half-open probe re-opens the breaker even when an earlier probe in the
+// same window already succeeded. With HalfOpenMaxProbes > 1 the first success
+// must not prematurely close the breaker (see issue #191).
+func TestBreaker_HalfOpenProbeFailureAfterSuccessReopens(t *testing.T) {
```
eunolabs, line 211: Both new tests admit two probes and then record outcomes sequentially. Issue #191 and the PR description frame the bug as a concurrency problem ("routed traffic back to an upstream that was still failing under concurrent half-open probes"). A `-race` test with two goroutines — both admitted, then one calling `RecordSuccess` and one `RecordFailure` — would reproduce the reported scenario more faithfully and guard the mutex interaction directly.

Not blocking, since all state transitions are fully serialized under `b.mu` and the sequential tests already pin the logic. Just noting the tests don't exercise the concurrency the PR motivates itself with.

---

eunolabs (commented):
Reviewed the half-open probe fix.

**The core change is correct.** Tracking `halfOpenSuccess` per window and re-opening on *any* admitted probe failure does fix the reported bug: a first probe success no longer prematurely closes the breaker while later probes in the same window are still outstanding, so `RecordFailure` correctly sees `StateHalfOpen` and re-opens. The counter is reset on every transition into half-open (`Allow`), on close (`RecordSuccess`), on re-open (`RecordFailure`), and in `Reset()` — all paths covered. The `HalfOpenMaxProbes=1` default behavior is unchanged. Clear commit message and a faithful reproducer test.

Left three inline notes — none block the fix, but the first two are worth addressing:

1. **Stale `Config.HalfOpenMaxProbes` doc + semantic overload.** That field now also dictates "successes required to close," and its doc comment doesn't say so (CONTRIBUTING wants the docs updated in-PR).
2. **Liveness edge for `HalfOpenMaxProbes > 1`.** A probe that never reports an outcome (e.g. a panic in `Do`, which has no `recover`) can wedge the breaker half-open until `Reset()`. Pre-PR a single success always recovered it.
3. **Test fidelity.** The new tests are sequential; the issue is framed around *concurrent* probes, so a `-race` goroutine test would match the motivation better.

Submitting as a comment rather than requesting changes — (1) and (2) are the only ones I'd push on, and neither breaks the fix.

---

eunolabs:
No additional comments from this review. The existing notes on `HalfOpenMaxProbes` docs/semantics and the half-open no-outcome liveness edge are still the current items to address.

---

pkg/circuitbreaker/breaker.go
eunolabs: Fixed the stale doc: `Config.HalfOpenMaxProbes` now documents that it also gates the number of successful probes required to close. On the coupling — keeping the single knob with the all-admitted-probes-must-succeed semantics for now, since (as you noted) it's a defensible conservative choice; a separate `HalfOpenRequiredSuccesses` field remains the clean way to decouple if we want that later. Pushed in a1876e3.

---

pkg/circuitbreaker/breaker.go
eunolabs: Went with remedy (b): `Do` now records a failure if the guarded `fn` panics, before the panic unwinds, so an admitted probe always reports exactly one outcome — a panicking probe re-opens the breaker instead of wedging it half-open. The panic is not recovered; it still propagates to the caller. Also documented the contract on `Breaker`. Covered by `TestDo_PanicRecordsFailureBeforePropagating` and `TestDo_PanicReopensHalfOpenBreaker`. Pushed in a1876e3.

---

pkg/circuitbreaker/breaker_test.go
eunolabs: Added `TestBreaker_HalfOpenConcurrentProbes_FailureWins`: it admits two probes, then records one success and one failure from separate goroutines and asserts the breaker ends open regardless of interleaving (runs under `-race`). Pushed in a1876e3.

---

eunolabs (commented):
Re-reviewed `a1876e3` — all three notes are correctly addressed, nothing further from me.

1. **Docs** — `Config.HalfOpenMaxProbes` now documents that it also gates successes-to-close, and the "each admitted probe must report exactly one outcome" contract is documented on `Breaker`. Keeping the single knob is a fine call.
2. **Liveness** — the `Do` panic path is correct: `completed` is set only after `fn` returns, so the deferred `RecordFailure` fires exactly once on a panic and is skipped on the normal success/error paths (no double-record), while the panic still propagates unrecovered. A panicking probe now re-opens the breaker instead of wedging it half-open.
3. **Concurrency** — `TestBreaker_HalfOpenConcurrentProbes_FailureWins` admits two probes and records one success + one failure from separate goroutines, asserting `Open` for either interleaving under `-race`. A lone success can't reach `halfOpenSuccess >= 2`, so the assertion holds.

Code review LGTM. CI for the new commit is still queued; I'll flag if anything goes red.

---

PR #201, opened 2026-06-13
fix: replace error-code string literals in pdp.go with constants (#193)

### LINKED ISSUE #193: bug: pdp.go uses string literals "CAPABILITY_DENIED" and "ENFORCEMENT_ERROR" instead of defined constants
## Summary

`cmd/eunox-mcp/pdp.go` uses bare string literals for error codes that are either already defined as constants in `pkg/capability/errors.go` or are entirely undocumented.

### `"CAPABILITY_DENIED"` — constant exists, literal used

`ErrCodeCapabilityDenied = "CAPABILITY_DENIED"` is defined in `pkg/capability/errors.go:9`, yet three call sites in `pdp.go` use the raw string:

```go
// pdp.go:210 (Decide — actions check)
Code: "CAPABILITY_DENIED",

// pdp.go:409 (DecideResourceRead — actions check)
Code: "CAPABILITY_DENIED",

// pdp.go:496 (DecidePromptGet — actions check)
Code: "CAPABILITY_DENIED",
```

If the constant is ever renamed (rename refactoring or a future breaking change), the three literals will silently diverge, producing an undocumented error code on the wire without any compile-time failure.

### `"ENFORCEMENT_ERROR"` — no constant defined at all

Three further call sites use `"ENFORCEMENT_ERROR"`:

```go
// pdp.go:256
Denial: &capability.DenialInfo{Code: "ENFORCEMENT_ERROR", Message: err.Error()},

// pdp.go:431
Denial: &capability.DenialInfo{Code: "ENFORCEMENT_ERROR", Message: err.Error()},

// pdp.go:518
Denial: &capability.DenialInfo{Code: "ENFORCEMENT_ERROR", Message: err.Error()},
```

This code is not listed in `pkg/capability/errors.go`, is not in the `JSONRPCCode*` mapping in `cmd/eunox-mcp`, is not mentioned in `docs/threat-model-mcp.md`, and has no corresponding test. Callers that need to distinguish an internal enforcement failure from a policy denial have no stable constant to match against.

## Fix

1. Add `ErrCodeCapabilityDenied` usages to the three `"CAPABILITY_DENIED"` literals (already imported as `capability.ErrCodeCapabilityDenied`).
2. Define `ErrCodeEnforcementError = "ENFORCEMENT_ERROR"` in `pkg/capability/errors.go`, add it to the JSONRPC code mapping, update `docs/threat-model-mcp.md` denial_code table, and switch all three literals to the constant.

eunolabs:
Fixes #193.

## Problem

`cmd/eunox-mcp/pdp.go` used bare string literals for two error codes:

- **`"CAPABILITY_DENIED"`** (3 sites) duplicated the existing `capability.ErrCodeCapabilityDenied` constant — a rename would silently diverge the literals from the constant with no compile-time failure.
- **`"ENFORCEMENT_ERROR"`** (3 sites) had no constant at all: it was absent from `pkg/capability/errors.go`, missing from the `denialToJSONRPCCode` mapping (so it fell through to the `-32002` `CAPABILITY_DENIED` wire code), undocumented in the threat model, and untested. Callers had no stable constant to distinguish an internal enforcement failure from a policy denial.

## Changes

- **`pkg/capability/errors.go`**: add `ErrCodeEnforcementError = "ENFORCEMENT_ERROR"` and the wire-code constant `JSONRPCCodeEnforcementError = -32603`. An enforcement-engine failure during condition evaluation is a server-side internal error, not a policy decision, so it uses the standard JSON-RPC internal-error code — the same one the transport layer already uses for upstream/infrastructure failures — distinguishing it from `-32001`/`-32002`/`-32003` on the wire.
- **`cmd/eunox-mcp/jsonrpc.go`**: add an explicit `ErrCodeEnforcementError` case to `denialToJSONRPCCode` instead of relying on the `-32002` default.
- **`cmd/eunox-mcp/pdp.go`**: switch the three `"CAPABILITY_DENIED"` and three `"ENFORCEMENT_ERROR"` literals to the constants.
- **`docs/threat-model-mcp.md`**: document `ENFORCEMENT_ERROR` in the §2.2 enforcement flow and add a change-log entry (v1.10).
- **`cmd/eunox-mcp/jsonrpc_error_codes_test.go`**: add a unit test for the new mapping, the constant value, and an entry in the table-driven `TestJSONRPCCode_AllCodes_ExactIntegers`.

## Verification

`make check-fmt`, `make build`, `make lint`, and `go test -race -count=1 ./...` all pass.

https://claude.ai/code/session_01MmrrWtzd8BcysdBkXkiunm

---
_Generated by [Claude Code](https://claude.ai/code/session_01MmrrWtzd8BcysdBkXkiunm)_


docs/threat-model-mcp.md
```diff
@@ -105,7 +105,7 @@ Three policy sources can be combined:
...
-5. All attached `conditions` pass: `TimeWindow`, `IPRange`, `AllowedOperations`, `AllowedExtensions`, `AllowedTables`, `MaxCalls`, `RecipientDomain`, `AllowedValues`, `Policy`, `Custom`. `conditions` are strictly boolean predicates.
+5. All attached `conditions` pass: `TimeWindow`, `IPRange`, `AllowedOperations`, `AllowedExtensions`, `AllowedTables`, `MaxCalls`, `RecipientDomain`, `AllowedValues`, `Policy`, `Custom`. `conditions` are strictly boolean predicates. If the engine itself errors while evaluating a condition (a handler returns an error rather than a clean pass/fail — e.g. a counter backend is unreachable), the call is denied fail-closed with `ENFORCEMENT_ERROR` (JSON-RPC `-32603`, the internal-error code), which is distinct from a clean condition rejection (`CONDITION_FAILED`) and from a capability miss (`CAPABILITY_DENIED`) so callers can tell an internal failure apart from a policy decision.
```
eunolabs: This example is incorrect and now contradicts both the code and §5.3 below.

Condition handlers return `*capability.ConditionError`, and `Engine.EvaluateConditions` always converts that into a **clean deny** carrying the handler's own code (`pkg/enforcement/engine.go:440-452`) — it never returns a Go `error`. The "counter backend is unreachable" case specifically returns `CONDITION_FAILED`, not `ENFORCEMENT_ERROR`:

- `TestEngine_MaxCalls_CounterError` asserts `ErrCodeConditionFailed` (`pkg/enforcement/enforcement_test.go:2046`) — I ran it on this branch and it passes.
- §5.3 just below (line 419) still says a `MaxCalls` counter error "fails closed and the call is denied with `CONDITION_FAILED`".

So §2.2 and §5.3 now disagree about the same scenario. In fact `EvaluateConditions` has **no** `return …, err` path today (both returns pass `nil`), so the `ENFORCEMENT_ERROR` branch in `pdp.go` is currently unreachable.

Suggest either describing this as a defensive guard for a *hypothetical* internal engine failure and dropping the counter-backend example, or reconciling §5.3 — but the docs shouldn't name a concrete trigger that the engine actually routes to a different code.

---

cmd/eunox-mcp/pdp.go
```diff
@@ -252,7 +252,7 @@ func (p *ManifestPDP) Decide(ctx context.Context, sessionID string, target Enfor
 	if err != nil {
 		return capability.EnforceResponse{
 			Decision: capability.DecisionDeny,
-			Denial:   &capability.DenialInfo{Code: "ENFORCEMENT_ERROR", Message: err.Error()},
+			Denial:   &capability.DenialInfo{Code: capability.ErrCodeEnforcementError, Message: err.Error()},
```
eunolabs, line 260: The literal→constant swap itself is correct. Flagging for completeness (not introduced by this PR): `EvaluateConditions` never returns a non-nil error (`pkg/enforcement/engine.go:410-476` — every path returns `nil`), so this branch and the two identical ones (`pdp.go:416`, `pdp.go:503`) are currently dead code.

This PR does change the wire code this path *would* emit from `-32002` (the old unmapped fallthrough) to `-32603`, but the new test only checks the static `denialToJSONRPCCode` mapping — nothing drives an actual `ENFORCEMENT_ERROR` denial end-to-end. If you want to lock the behavior in (and demonstrate the path is reachable), a stubbed engine/PDP returning an error here, asserting the resulting `denial_code` + `-32603` wire code, would do it. Otherwise it'd be more honest to document the constant as reserved.

---

eunolabs (commented):
## Review: error-code literals → constants (#193)

The mechanical part of this change is correct and worth doing: the three `"CAPABILITY_DENIED"` literals now reference `capability.ErrCodeCapabilityDenied`, so a future rename can't silently diverge the literal from the constant. The new `denialToJSONRPCCode` case and its tests are right, and `-32603` is a sound wire code — the transport layer (`stdio.go`/`http.go`/`stdio_http_upstream.go`) already uses it for internal/upstream failures.

One substantive issue, raised inline on two threads, that I'd resolve before merge:

**The `ENFORCEMENT_ERROR` path is currently unreachable, and the new docs example mis-describes it.**

- `Engine.EvaluateConditions` (`pkg/enforcement/engine.go:410-476`) returns `nil` as the error on *every* return path; condition handlers return `*ConditionError`, never a Go `error`. So the `if err != nil { … ENFORCEMENT_ERROR … }` branches at `pdp.go:252`, `:413`, and `:500` cannot execute today. Renaming the literal to a constant is fine, but no test drives an actual `ENFORCEMENT_ERROR` denial — the new test only checks the static `denialToJSONRPCCode` mapping.
- The new §2.2 text names "a counter backend is unreachable" as the trigger. That scenario actually returns `CONDITION_FAILED` — confirmed by `TestEngine_MaxCalls_CounterError` (`pkg/enforcement/enforcement_test.go`, asserts `ErrCodeConditionFailed`) and by §5.3 line 419, which still says a `MaxCalls` counter error denies with `CONDITION_FAILED`. As written, §2.2 and §5.3 now disagree about the same case.

Recommendation — pick one:
1. **Make the path reachable + lock it in:** drive an `ENFORCEMENT_ERROR` denial end-to-end via a stubbed engine/PDP that returns an error from `EvaluateConditions`, asserting both `denial_code == ENFORCEMENT_ERROR` and the `-32603` wire code. Then drop the inaccurate counter-backend example.
2. **Document it as reserved:** keep the constant as a fail-closed guard for a hypothetical future internal failure, and reword §2.2 to describe it that way (no concrete trigger), leaving §5.3 authoritative for the counter case.

Either way the §2.2/§5.3 contradiction should be reconciled. Minor, optional: the new `JSONRPCCodeEnforcementError` constant coexists with ~16 bare `-32603` literals in the transport layer that mean the same thing — out of scope here, but a candidate follow-up if the goal is to eliminate the magic number everywhere.

See the two inline threads for specifics.

---

eunolabs:
No additional comments from this review. The existing notes about `ENFORCEMENT_ERROR` reachability and the §2.2/§5.3 docs mismatch are still the current items to address.

---

docs/threat-model-mcp.md
eunolabs: Fixed in 7502e7d. You're right — confirmed against `engine.go:440-452` and `TestEngine_MaxCalls_CounterError` (asserts `CONDITION_FAILED`). §2.2 now states condition rejections (including a `MaxCalls` counter-backend error) fail closed with `CONDITION_FAILED` per §5.3, and frames `ENFORCEMENT_ERROR` as a reserved fail-closed guard not emitted on any reachable path today. Dropped the counter-backend example.

---

cmd/eunox-mcp/pdp.go
eunolabs: Agreed the branch is currently unreachable. I went with your second option — documenting the constant as reserved — rather than the stubbed-engine test, because making the path genuinely reachable would require changing `EvaluateConditions` to surface a Go error, which is out of scope for a literal→constant cleanup and a real behavior change to the engine. I kept the defensive guard (per the fail-closed invariant) and added comments at all three sites noting it's unreachable today. The static mapping/`denialResult` round-trip test locks in the `-32603` wire code the guard would emit. 7502e7d.

---

eunolabs (commented):
## Re-review of 7502e7d — both threads resolved

Verified the follow-up commit against the actual patch; both items from my first pass are addressed and I've resolved the threads:

- **§2.2 / §5.3 contradiction (docs):** fixed. §2.2 now attributes a condition rejection — including an unreachable `MaxCalls` counter backend — to `CONDITION_FAILED` per §5.3, and frames `ENFORCEMENT_ERROR` as a reserved guard "not emitted on any reachable path today." Consistent with the code and `TestEngine_MaxCalls_CounterError`.
- **`ENFORCEMENT_ERROR` dead code (pdp.go):** addressed via option 2. The three guard sites, the `errors.go` constant doc, and the test comment now all describe it as a reserved, defensive fail-closed guard that's unreachable today. Reasonable call to leave the path-reachability change (an engine behavior change) out of a literal→constant cleanup, and keeping the fail-closed guard is consistent with the project invariant.

No new issues — `7502e7d` is comment/prose-only with no logic change. The `CAPABILITY_DENIED` → constant swap (the original point of the PR) stands. CI is green on the completed checks and re-running on the new head.

Nothing further from me. (Can't formally APPROVE since this PR is authored by the same account.)

---

PR #202, opened 2026-06-13
fix: guard shipped JWT PDP JWKS fetches with pkg/circuitbreaker (#192)

### LINKED ISSUE #192: Shipped JWT PDP does not use pkg/circuitbreaker for JWKS fetches
## Summary

`docs/architecture.md` says `pkg/circuitbreaker` guards JWKS fetches so a flapping IdP endpoint does not hammer the network and JWT verification fails closed while the breaker is open. The shipped `cmd/eunox-mcp` JWT path does not appear to wire that breaker in.

The proxy path in `cmd/eunox-mcp/main.go` builds a `JWTPDP` via `wrapRoutesWithJWT(...)`. That code uses the private `jwksCache` in `cmd/eunox-mcp/pdp_jwt.go`, whose `fetch` path calls the HTTP client directly. There is no `pkg/circuitbreaker` import or `circuitbreaker.Do` call under `cmd/eunox-mcp`.

The only runtime use of `pkg/circuitbreaker` I found is the optional `Breaker` field in `pkg/capability.JWKSClientConfig`, but that client is not the shipped `eunox-mcp proxy` JWT PDP path.

## Impact

When the IdP/JWKS endpoint is down or flapping and cached keys are unavailable or stale, the shipped proxy can issue a fresh JWKS HTTP request for each sequential verification attempt. Singleflight deduplicates only concurrent refreshes; it does not provide a cooldown window across sequential failures. This contradicts the documented operating behavior and can amplify IdP outages.

## Evidence

Relevant references:

- `docs/architecture.md`: says `pkg/circuitbreaker` guards JWKS fetches.
- `cmd/eunox-mcp/main.go`: `wrapRoutesWithJWT(routes, JWTPDPOptions{...})` constructs the shipped JWT PDP.
- `cmd/eunox-mcp/pdp_jwt.go`: `jwksCache.fetch` directly calls `c.client.Do(req)`.
- `rg "pkg/circuitbreaker|circuitbreaker\." cmd pkg docs README.md CLAUDE.md` finds no `cmd/eunox-mcp` circuitbreaker usage.

## Suggested fix

Either wire `pkg/circuitbreaker` into `cmd/eunox-mcp`'s `jwksCache` refresh path, or update the architecture docs to state that circuit breaker protection exists only for the library `pkg/capability.JWKSClient` when callers explicitly provide a breaker.

eunolabs:
## Summary

Fixes #192.

`docs/architecture.md` states that `pkg/circuitbreaker` guards JWKS fetches so a flapping IdP endpoint does not hammer the network, and that JWT verification fails closed while the breaker is open. The shipped `cmd/eunox-mcp` JWT path (`jwksCache` in `pdp_jwt.go`) did not actually wire the breaker in — its `fetch` path called `c.client.Do(req)` directly. The only runtime breaker usage was the optional `Breaker` field on the *library* `pkg/capability.JWKSClient`, which is not the path the `eunox-mcp proxy` JWT PDP uses.

Because singleflight only deduplicates **concurrent** refreshes (not sequential ones), each verification against a down/stale endpoint with no usable cached keys would issue its own HTTP request — contradicting the documented behavior and amplifying IdP outages.

## Changes

- **`cmd/eunox-mcp/pdp_jwt.go`**
  - Add a `breaker *circuitbreaker.Breaker` to `jwksCache` and route both `doRefresh` fetch sites through a new `guardedFetch`, which runs the HTTP fetch via `circuitbreaker.Do`. When the breaker is open, the fetch short-circuits with `circuitbreaker.ErrOpen` and no HTTP request is made — verification fails closed (propagates to HTTP 401).
  - `newJWKSCache` installs a `circuitbreaker.DefaultConfig()` breaker when none is supplied, so the shipped proxy always has the protection. The gateway path (`wrapRoutesWithJWT`) shares one validating cache, so its breaker is active.
  - `JWTPDPOptions.Breaker` lets callers/tests override the breaker config or clock.
- **`cmd/eunox-mcp/pdp_jwt_test.go`** — add `TestJWKSCache_CircuitBreaker_OpensAndStopsFetching` (after `FailureThreshold` consecutive failures the endpoint is no longer hit and the error wraps `ErrOpen`) and `TestJWKSCache_CircuitBreaker_RecoversAfterCooldown` (a half-open probe closes the breaker once the IdP recovers, using an injected clock).
- **`docs/architecture.md`** — clarify that both the shipped `jwksCache` and the library `JWKSClient` route refreshes through the breaker, and that singleflight covers only concurrent refreshes while the breaker adds the cooldown across sequential failures.

## Testing

- `go test -race -count=1 ./cmd/eunox-mcp/ ./pkg/...` — all pass
- `make lint` / `make check-fmt` — clean

This preserves the fail-closed invariant: a breaker-open JWKS fetch returns a structured error rather than forwarding without verification.

https://claude.ai/code/session_016stwz6Ypv9JQqbPbkw6C2a

---
_Generated by [Claude Code](https://claude.ai/code/session_016stwz6Ypv9JQqbPbkw6C2a)_


cmd/eunox-mcp/pdp_jwt.go
```diff
@@ -283,6 +300,25 @@ func (c *jwksCache) doRefresh(ctx context.Context, force bool) (*jose.JSONWebKey
...
+// guardedFetch performs the JWKS HTTP fetch through the circuit breaker. When
+// the breaker is open it returns an error wrapping circuitbreaker.ErrOpen
+// without issuing an HTTP request, so a flapping IdP endpoint is not hammered
+// across sequential verification attempts. A nil breaker (never the case for
+// caches built by newJWKSCache) falls back to a direct fetch.
+func (c *jwksCache) guardedFetch(ctx context.Context) (*jose.JSONWebKeySet, error) {
+	if c.breaker == nil {
+		return c.fetch(ctx)
+	}
+	jwks, err := circuitbreaker.Do(ctx, c.breaker, c.fetch)
```
eunolabs: **Client-induced cancellations can trip the breaker (discussion point, non-blocking).**

`circuitbreaker.Do` only short-circuits on `ctx.Err()` *before* calling `fn`; any error returned *by* `fn` is recorded as a failure — including the case where `c.fetch` fails because the request-scoped `ctx` was canceled or hit its deadline (host disconnect, per-request timeout), which surfaces here as `fetching JWKS: context canceled`.

Because this cache is a per-validator singleton shared across every session, a burst of ~`FailureThreshold` (5 by default) consecutive client cancellations landing on the leader refresh could open the breaker and then fail *all* JWT verification for the 30s cooldown — turning a client-side condition into a server-side auth outage that's unrelated to IdP health.

Worth considering whether caller-induced cancellations (`errors.Is(err, context.Canceled)` / `context.DeadlineExceeded`) should be excluded from breaker accounting. Since `Do` is shared with the upstream-call path, the right fix may live in `pkg/circuitbreaker` rather than here — flagging as a discussion point, not a blocker.

---

eunolabs (commented):
## Review — guard shipped JWT PDP JWKS fetches with `pkg/circuitbreaker`

Reviewed end-to-end and verified locally against the PR head (`fc5d4fe`).

**Verification (all clean):**
- `go build ./...`
- `go test -race -count=1 ./cmd/eunox-mcp/ ./pkg/circuitbreaker/` — pass, including the two new tests
- `go vet ./cmd/eunox-mcp/` and `gofmt -l` on the changed files — clean

**What's solid**
- Core fix is correct: `doRefresh` routes both fetch sites through `guardedFetch` → `circuitbreaker.Do`, and `newJWKSCache` installs a `DefaultConfig` breaker so the shipped proxy is always protected. Fail-closed is preserved — a breaker-open fetch returns a wrapped `ErrOpen` and never forwards an unverified token.
- Gateway path checks out: `wrapRoutesWithJWT` builds the validator via `NewJWTPDP` (so its cache gets a breaker) and the per-route wrappers share `validator.cache`, so the breaker is active there too — matches the PR description.
- The two new tests are meaningful — exact fetch-count assertion for the open transition, and an injected-clock half-open recovery rather than a real sleep.

**Comments**
1. *(inline, `pdp_jwt.go`)* Context-cancellation accounting — a client-induced cancel/timeout during a refresh is recorded as a breaker failure; raised as a discussion point.
2. The existing thread on `docs/architecture.md:306` is **valid** — I confirmed `pkg/capability.JWKSClient` only routes through the breaker when a caller supplies one (`NewJWKSClient` sets no default; `fetch` guards on `if c.breaker != nil`). The "Both … route refreshes through the breaker" phrasing should distinguish always-on (shipped proxy) from opt-in (library). Worth resolving alongside this PR.

**Minor (non-blocking)**
Once the breaker opens, the still-held last-known-good keys in `c.jwks` aren't served past TTL — every verification fails for the full cooldown rather than serving slightly-stale keys. That's the correct fail-closed posture per the repo invariants; just noting the availability tradeoff so it reads as intentional.

Focused, well-tested fix overall.

---

cmd/eunox-mcp/pdp_jwt.go
eunolabs: Good catch — fixed in c9b389d. `guardedFetch` no longer goes through `circuitbreaker.Do`; it does the Allow/RecordSuccess/RecordFailure accounting itself and **skips `RecordFailure` when `ctx.Err() != nil`**, so a caller-induced cancel/deadline (host disconnect, per-request timeout) on the shared per-validator cache can't open the breaker.

I used `ctx.Err()` rather than `errors.Is(err, context.DeadlineExceeded)` deliberately: the HTTP client's own timeout against a slow/unreachable JWKS endpoint also surfaces as `DeadlineExceeded`, but leaves the request context's `Err()` nil — that case *should* still count as an IdP-health failure, and it does. Only cancellation of the request-scoped context we passed in is excluded.

Kept the fix local rather than in `pkg/circuitbreaker.Do` so the upstream-call path keeps counting deadline failures (a hung upstream arguably should trip its breaker). Added `TestJWKSCache_CircuitBreaker_IgnoresCallerCancellation` covering it.

---

eunolabs (commented):
Found one new issue in the latest head. The cancellation-accounting fix avoids false breaker failures in closed state, but it can wedge a half-open breaker because a consumed probe is left without an outcome.

cmd/eunox-mcp/pdp_jwt.go
```diff
@@ -283,6 +299,47 @@ func (c *jwksCache) doRefresh(ctx context.Context, force bool) (*jose.JSONWebKey
...
+		return nil, err
+	}
+	if !c.breaker.Allow() {
+		return nil, fmt.Errorf("fetching JWKS: %w", circuitbreaker.ErrOpen)
+	}
+	jwks, err := c.fetch(ctx)
+	if err != nil {
+		if ctx.Err() != nil {
+			// Caller-induced cancel/timeout, not an IdP failure — leave the breaker alone.
+			return nil, err
```
eunolabs, line 350: This returns after `Allow()` has already consumed a breaker slot, but it records neither success nor failure. That is fine while the breaker is closed, but it wedges recovery when the breaker is half-open: `Allow()` transitions open -> half-open and sets `halfOpenProbes = 1`; if the request context is then canceled during `fetch`, this branch returns with the breaker still half-open and no outcome recorded. With the default `HalfOpenMaxProbes=1`, every later JWKS refresh is denied by the half-open probe limit (`halfOpenProbes < HalfOpenMaxProbes` is false) until process reset, even if the IdP has recovered.

The fix needs to preserve the "client cancellations do not count as IdP failures" behavior without leaving an admitted half-open probe unreleased. One option is to record a failure only when the cancellation happened after consuming a half-open probe; another is to add explicit breaker support for abandoning/releasing a probe.

cmd/eunox-mcp/pdp_jwt.go
eunolabs: Re-reviewed `c9b389d`. The closed-state case is handled correctly, but skipping `RecordFailure` on `ctx.Err()` opens a new failure mode in the **half-open** path that's strictly worse than the original 30s outage: a single caller cancellation landing on the once-per-cooldown probe **permanently wedges** the breaker.

Mechanism:
1. Breaker is open; cooldown elapses.
2. `guardedFetch` passes the top `ctx.Err()` check, then `c.breaker.Allow()` transitions Open→HalfOpen and **consumes the one probe slot** (`halfOpenProbes=1`), returning true.
3. The probe `fetch` is canceled by the caller → `ctx.Err() != nil` → we `return nil, err` **without** `RecordFailure`.
4. Breaker is now stranded in half-open with no probe budget. The only exits from half-open are `RecordSuccess`/`RecordFailure`, but `Allow()` now returns false for every future call (`halfOpenProbes(1) < max(1)` is false), so `fetch` never runs and neither is ever called.

Result: every subsequent JWKS fetch returns `ErrOpen` forever — a permanent fail-closed auth outage until process restart, triggered by one client disconnect during the probe. `TestJWKSCache_CircuitBreaker_IgnoresCallerCancellation` doesn't catch it because the breaker is closed throughout that test, so the probe slot is never involved.

Reproduction (drop into `cmd/eunox-mcp`, fails on the current branch — reuses `newTestKey`):

```go
func TestJWKSCache_CircuitBreaker_HalfOpenProbeCancellation_Recovers(t *testing.T) {
	var mode atomic.Int32 // 0=500, 1=block-until-cancel, 2=healthy
	key := newTestKey(t, "k1")
	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		switch mode.Load() {
		case 0:
			w.WriteHeader(http.StatusInternalServerError)
		case 1:
			<-r.Context().Done() // block until the client gives up
		default:
			jwks := jose.JSONWebKeySet{Keys: []jose.JSONWebKey{{Key: key.priv.Public(), KeyID: key.kid, Use: "sig"}}}
			w.Header().Set("Content-Type", "application/json")
			_ = json.NewEncoder(w).Encode(jwks)
		}
	}))
	defer srv.Close()

	var clkMu sync.Mutex
	clk := time.Now()
	nowFn := func() time.Time { clkMu.Lock(); defer clkMu.Unlock(); return clk }
	advance := func(d time.Duration) { clkMu.Lock(); clk = clk.Add(d); clkMu.Unlock() }

	const cooldown = 30 * time.Second
	br := circuitbreaker.New(circuitbreaker.Config{FailureThreshold: 1, CooldownDuration: cooldown, HalfOpenMaxProbes: 1}, circuitbreaker.WithClock(nowFn))
	cache := newJWKSCache(srv.URL, time.Millisecond, nil, br)

	mode.Store(0) // trip it open
	if _, err := cache.getKeys(context.Background()); err == nil {
		t.Fatal("expected failure from 500")
	}

	advance(cooldown + time.Second) // half-open probe, canceled mid-fetch
	mode.Store(1)
	ctx, cancel := context.WithTimeout(context.Background(), 40*time.Millisecond)
	if _, err := cache.getKeys(ctx); err == nil {
		t.Fatal("expected caller-cancel error on the probe")
	}
	cancel()

	mode.Store(2) // IdP healthy, well past cooldown — must recover
	for i := 0; i < 5; i++ {
		advance(cooldown + time.Second)
		if _, err := cache.getKeys(context.Background()); err == nil {
			return
		}
	}
	t.Fatalf("WEDGED: breaker never recovered despite healthy IdP; state=%q", br.State())
}
```

Output on the current branch: `WEDGED: breaker never recovered despite healthy IdP; state="half-open"`.

Suggested direction: the "skip accounting" shortcut is only safe while the breaker is **closed**. When the canceled fetch consumed a half-open probe, the breaker must not be left stranded — e.g. gate the skip on `c.breaker.State() == circuitbreaker.StateClosed`, or on a canceled probe call `RecordFailure()` so the breaker re-opens and re-arms the cooldown (self-healing after the next window). Either still prevents the original "burst of client disconnects opens a *closed* breaker" scenario, since that's the Closed→Open path, which stays guarded. Happy to push a patch if useful.

Unresolving so this stays visible.

---

eunolabs:
No further comments beyond the inline half-open cancellation liveness issue from this review.

---

cmd/eunox-mcp/pdp_jwt.go
eunolabs: Confirmed and fixed in 66e54f5 — this was a real liveness regression, and worse than the original. Thank you for the precise mechanism and repro.

`guardedFetch` now captures whether `Allow()` admitted it as a half-open probe (`c.breaker.State() != StateClosed`, read immediately after `Allow()` and before the fetch). The "skip accounting on caller-cancel" shortcut now only applies when **no probe was consumed** (closed breaker). When a cancellation lands on a consumed half-open probe it falls through to `RecordFailure()`, which re-opens the breaker and re-arms the cooldown so the next window gets a fresh probe — self-healing rather than wedged. The closed-state guard (a burst of client disconnects must not open a healthy breaker) is unchanged, since that's the Closed→Open path.

I added your reproduction as `TestJWKSCache_CircuitBreaker_HalfOpenProbeCancellation_Recovers` and verified it: it fails on the prior commit with `state="half-open"` (never recovers) and passes now. Kept the fix local to `guardedFetch` rather than adding an abandon/release API to `pkg/circuitbreaker`, since recording a failure for an abandoned probe is the correct fail-closed posture here.

---

eunolabs (commented):
## Re-review — all findings resolved, verified locally at `66e54f5`

Independently verified the fix chain end-to-end:
- `go test -race -count=1 ./cmd/eunox-mcp/ ./pkg/circuitbreaker/` — green, including the new `TestJWKSCache_CircuitBreaker_HalfOpenProbeCancellation_Recovers` (my repro, now a permanent regression guard): it fails on `c9b389d` with `state="half-open"` and passes here.
- `go vet` + `gofmt` clean; CI on the head is green (Test / Lint / gofmt / all platform builds / demos / OPA comparisons).

Resolution trail:
- **Docs** overstatement (library breaker is opt-in, not always-on) → `7bb480e`. ✓
- **Caller-cancel tripping a *closed* breaker** → `c9b389d`. ✓
- **Half-open probe wedge** (liveness regression, worse than the original) → `66e54f5`: `guardedFetch` captures whether `Allow()` consumed a probe (`State() != StateClosed`, read right after `Allow()`) and only skips accounting when no probe was consumed; a cancel on a consumed probe falls through to `RecordFailure()` → re-opens + re-arms the cooldown → self-heals. Fail-closed posture preserved throughout. ✓

The core change does what `docs/architecture.md` promises — the shipped JWT PDP path is now actually breaker-guarded, and the gateway shares that guarded cache. Solid, well-tested fix.

**Non-blocking nit (no change needed):** `consumedProbe` is inferred from a `State()` read just after `Allow()`. Under a concurrent forced+non-forced refresh (the `doRefresh` force-doesn't-join-non-forced path), a state transition could land in that window and mis-tag a closed-admit as a consumed probe or vice-versa. I worked through both directions — neither can re-introduce the wedge, and the worst case is one spurious failure tally on an already-opening breaker, so it's benign. Flagging only for awareness; an `Allow()` that returns whether it admitted a probe would make this exact, if you ever revisit `pkg/circuitbreaker` for other reasons.

---

PR #203, opened 2026-06-13
perf: cache compiled argumentSchema patterns; compile at manifest load

### LINKED ISSUE #195: perf: schemaValidateString recompiles regex pattern on every request
## Summary

`pkg/enforcement/schema.go` compiles the regex from an `argumentSchema.pattern` constraint on every single invocation of `schemaValidateString`. For a proxy under sustained load with a `pattern`-constrained tool, this is unnecessary repeated CPU work.

## Location

`pkg/enforcement/schema.go`, lines 103–111:

```go
func schemaValidateString(p, v string, s *capability.ArgumentSchema) error {
    if s.Pattern != "" {
        re, err := regexp.Compile(s.Pattern)   // compiled on every call
        if err != nil {
            return fmt.Errorf("%s: invalid pattern %q: %w", p, s.Pattern, err)
        }
        if !re.MatchString(v) {
            return fmt.Errorf("%s: value does not match pattern %q", p, s.Pattern)
        }
    }
    ...
```

The pattern is static — it comes from the manifest and never changes at runtime. Each request that hits this code path compiles the same pattern from scratch.

## Impact

Go's `regexp` package does have an internal LRU cache (`regexp.Compile` checks it before re-parsing), but that cache is limited in size and shared across the whole process. For a single manifest with several distinct `pattern` constraints, the patterns will repeatedly evict each other. Measured overhead is tens of microseconds per compile; under high-QPS production load this adds up.

Additionally, the compile error on line 106–108 can never be reached in a normally loaded manifest because `validateArgumentSchema` in `validateLocalManifest` calls `regexp.Compile` during manifest load and would have rejected a bad pattern already. The per-call defensive re-compilation is wasted work.

## Fix

Compile the pattern once at manifest load time (in `validateLocalManifest` or a new `ArgumentSchema.Compile()` step) and store the compiled `*regexp.Regexp` on the schema node. Alternatively, use a package-level `sync.Map` keyed by the pattern string. Either approach eliminates the repeated compilation on the hot enforcement path.

eunolabs:
Fixes #195.

## Problem

`schemaValidateString` in `pkg/enforcement/schema.go` recompiled the
`argumentSchema.pattern` regex on **every** `tools/call` against a
pattern-constrained tool. The pattern is static — it comes from the capability
manifest and never changes at runtime — so the per-request `regexp.Compile` was
wasted CPU on the enforcement hot path.

While investigating I found the issue's premise was slightly off: `validateLocalManifest`
only validated argumentSchema *keywords* (`checkArgumentSchemaKeywords`); it never
actually compiled the `pattern`. So a malformed regex loaded silently and only
failed per-request — the opposite of this codebase's fail-closed posture.

## Fix

- **Hot-path cache.** A package-level compiled-regexp cache in `pkg/enforcement`
  keyed by the pattern string. The per-request path now does a map lookup instead
  of a compile. `argumentSchema` patterns are manifest-only (JWT capability claims
  never carry an `argumentSchema` — they build `AllowedValues`/`AllowedOperations`
  conditions programmatically), so the key space is finite and bounded.
- **Compile once at load.** `CompileSchemaPatterns` walks the schema tree and is
  called from `validateLocalManifest`. A malformed regex is now rejected up front
  with a path-qualified error (fail closed), matching the other load-time
  condition validators, and each valid pattern is primed into the cache so the
  first live request is already a hit.

This is the issue's explicitly-sanctioned alternative ("a package-level `sync.Map`
keyed by the pattern string") — chosen over storing on the schema node because it
requires no changes to `capability.ArgumentSchema` (no DeepEqual/JSON/copylock
risk) and covers the public `ValidateArgumentSchema` regardless of caller.

## Tests

- White-box (`pkg/enforcement`): cache reuses the same `*regexp.Regexp` instance;
  a non-compiling pattern errors and is not cached; `CompileSchemaPatterns` recurses
  through `properties`/`items` and reports the first bad pattern with its JSON path.
- `cmd/eunox-mcp` (through `LoadManifest`): a valid `pattern` loads; malformed
  patterns at top level, in a property, and nested in `items` are rejected at load.
- Existing direct-call behavior preserved: the `[unclosed` invalid-pattern test
  still gets an "invalid pattern" error.

New functions are at 100%/93.8% coverage. `make test`, `make lint`,
`make check-fmt`, and `make check-license` all pass.

## Docs

`docs/capability-manifest-guide.md` updated: a malformed `argumentSchema`
`pattern` is now reported at load by `validate` rather than surfacing as a
per-request denial.

https://claude.ai/code/session_0117Rjv4Nwi4KYvCmSYAYj5M

---
_Generated by [Claude Code](https://claude.ai/code/session_0117Rjv4Nwi4KYvCmSYAYj5M)_


cmd/eunox-mcp/manifest.go
```diff
@@ -175,6 +175,16 @@ func validateLocalManifest(m *LocalManifest) error {
...
 			}
 		}
+		// Compile every argumentSchema `pattern` once, at load: a malformed regex
+		// is rejected here (fail closed) instead of loading silently and denying
+		// the first live request with a CONDITION_FAILED that reads like a real
+		// enforcement event, and each valid pattern is primed into the enforcement
+		// regexp cache so the per-request hot path never recompiles it.
+		if c.ArgumentSchema != nil {
+			if err := enforcement.CompileSchemaPatterns(c.ArgumentSchema); err != nil {
+				return fmt.Errorf("capability at index %d: argumentSchema %w", i, err)
```
eunolabs: Minor / non-blocking: the path dialect in this load-time error diverges from the sibling load-time validator.

`CompileSchemaPatterns` reports JSON-Pointer-style paths (`$`, `$.path`, `$.tags[*]`), so this produces messages like:

- `capability at index 0: argumentSchema $.path: invalid pattern "[unclosed"`
- `capability at index 0: argumentSchema $: invalid pattern "*nostart"` (top-level — the bare `argumentSchema $:` reads a bit cryptically)

But `checkArgumentSchemaKeywords` — the other thing that rejects a bad `argumentSchema` at load — reports the structural manifest path instead:

- `capabilities[0].argumentSchema.properties.path: unknown keyword "format"`

So an operator debugging the same `argumentSchema` node can get two different path notations (and two different capability prefixes, `capability at index N` vs `capabilities[N]`) depending on whether the keyword check or the pattern compile fires first.

I see the rationale — `$.`-style matches the *runtime* validator's path (`schemaValidateValue`), so the same path shows up at load and at request time, which is a defensible choice. Just flagging the divergence: it'd be worth either aligning the two load-time validators on one dialect, or adding a one-line comment noting the intent so the mismatch doesn't read as accidental. Not blocking.

---

eunolabs (commented):
## Review — looks solid

Reviewed the implementation, ran the tests (`pkg/enforcement` + `cmd/eunox-mcp` both pass with `-race`), `go vet`, and `gofmt`. This is a clean, well-reasoned fix for #195.

**What's right:**
- **Correct & complete tree walk.** `compileSchemaPatterns` recurses `Properties` and `Items`, which are the only two nestable-schema fields on `ArgumentSchema` (`AdditionalProperties` is `*bool`, `Enum` is values) — so every `pattern` the runtime validator could reach via `schemaValidateValue` is primed at load. No node is missed.
- **Fail-closed at load.** Compiling in `validateLocalManifest` turns a malformed regex into an up-front load rejection instead of a first-request `CONDITION_FAILED` that masquerades as a real enforcement event. This also covers gateway mode, since each route loads through `LoadManifest` → `validateLocalManifest` (`route.go:227`).
- **Hot-path cache is sound.** `sync.Map` + `LoadOrStore` collapses the concurrent first-compile race to one shared `*regexp.Regexp`; a failed compile is deliberately not cached. The package-global cache shared across gateway routes is fine (and dedupes identical patterns) because the key is the pattern string and RE2 compilation is deterministic.
- **Deterministic errors** via sorted property names, and tests that actually assert the JSON-path-qualified message at top level, in a property, and nested in `items`.
- No ReDoS exposure from caching — Go's `regexp` is RE2 (linear time) — and the keyspace is operator-controlled and bounded as the comments note.

**One minor, non-blocking note** (inline on `manifest.go`): the load-time pattern error uses `$.`-style paths while the sibling load-time keyword validator uses `capabilities[N].argumentSchema.properties.x` — two path dialects for the same node. Worth aligning or documenting the intent, but not a blocker.

Nice work, especially catching that the issue's premise was slightly off (`validateLocalManifest` never actually compiled the pattern before) and fixing the silent-load gap alongside the perf win.

---

cmd/eunox-mcp/manifest.go
eunolabs: Good catch — aligned the two load-time validators on one dialect in 5e052e5. `CompileSchemaPatterns` now takes a `schemaPath` root and extends it with `.properties.<name>` / `.items`, and `validateLocalManifest` passes `capabilities[N].argumentSchema`, so a bad pattern now reports e.g. `capabilities[0].argumentSchema.properties.path: invalid pattern "[unclosed"` — same notation as `checkArgumentSchemaKeywords`, and the cryptic top-level `argumentSchema $:` becomes `capabilities[0].argumentSchema:`. Since this is a schema walk (not a data instance), schema-location paths are also the more accurate choice. Tests updated.

---

eunolabs: Verified 5e052e5 locally — both load-time validators now speak one dialect (`.properties.<name>` / `.items` rooted at `capabilities[N].argumentSchema`), the top-level case reads cleanly, and `go build`/`go vet`/`gofmt` plus the updated tests pass with `-race`. Agreed that schema-location paths are the more accurate choice for a schema walk. Resolving.

---

PR #209, opened 2026-06-13
fix: namespace maxCalls counter key by target type (#205)

### LINKED ISSUE #205: bug: maxCalls counter key not namespaced by MCP target type — tool and prompt with the same bare name share one quota bucket
## Summary

`handleMaxCalls` in `pkg/enforcement/handlers.go` builds the counter key as:

```go
```

`req.ToolName` is the **bare name** of the target, set to:
- the tool name for `tools/call` (via `ManifestPDP.Decide` → `target.Name`)
- the URI for `resources/read` (via `ManifestPDP.DecideResourceRead` → `req.ToolName = uri`)
- the prompt name for `prompts/get` (via `ManifestPDP.DecidePromptGet` → `req.ToolName = promptName`)

A **tool named "X"** and a **prompt named "X"** will produce the identical key `maxcalls:<session>:X`. If both manifest entries carry a `maxCalls` condition, they share the same quota counter: calls to the tool consume budget from the prompt's limit, and vice versa.

## Location

`pkg/enforcement/handlers.go`:
```go
// line 277
```

`pkg/enforcement/engine.go` for comparison — `sequenceHistoryKey` is also not namespaced by type, but collisions there are less likely because the key includes the literal prefix `"seq:"`:
```go
}
```

## Concrete scenario

```yaml
capabilities:
  - target: tool:export
    actions: [call]
    conditions:
      - type: maxCalls
        count: 10
        windowSeconds: 3600

  - target: prompt:export
    actions: [get]
    conditions:
      - type: maxCalls
        count: 100
        windowSeconds: 3600
```

Both generate the key `maxcalls:<session>:export`. After 10 tool calls, the prompt `export` is also rate-limited (its counter reads 10, exceeding the tool's limit of 10 — or the other way around depending on which condition runs first).

Resource URIs are typically immune because they include a scheme (`file:///…`, `https://…`) that no bare tool name would match, but the tool/prompt collision is realistic whenever shared vocabulary is used for both.

## Fix

Include the target type in the counter key:

```go
targetType := ""
if req.Target != nil {
    targetType = req.Target.Type
}
key := fmt.Sprintf("maxcalls:%s:%s:%s", req.SessionID, targetType, req.ToolName)
```

`req.Target` is populated by `ManifestPDP.Decide`, `DecideResourceRead`, and `DecidePromptGet`. For callers that go through `engine.ValidateAction` directly and leave `req.Target` nil, fall back to the current schema (no type prefix) to preserve backwards compatibility, or require callers to populate `req.Target`.

This change is a breaking change to the counter key schema (existing running counters for tool keys would reset). Given the pre-1.0 no-backwards-compat-shim policy this is acceptable; a note in the CHANGELOG is sufficient.

eunolabs:
## Summary

Fixes #205.

`handleMaxCalls` in `pkg/enforcement/handlers.go` built the per-session counter key from the **bare target name** only:

```go
```

`req.ToolName` is the tool name for `tools/call`, the prompt name for `prompts/get`, and the URI for `resources/read`. So a **tool named `export`** and a **prompt named `export`** both produced the key `maxcalls:<session>:export` and shared a single quota bucket — calls to the tool consumed the prompt's budget and vice versa, and whichever `maxCalls` limit was lower silently throttled both.

## Fix

Include the target type in the key:

```go
targetType := ""
if req.Target != nil {
    targetType = req.Target.Type
}
key := fmt.Sprintf("maxcalls:%s:%s:%s", req.SessionID, targetType, req.ToolName)
```

`req.Target` is populated by every `ManifestPDP` entry point (`Decide` / `DecideResourceRead` / `DecidePromptGet`) with `"tool"`, `"resource"`, or `"prompt"`. A direct `ValidateAction` caller that leaves `req.Target` nil gets an empty type segment — no worse than the pre-fix schema for that caller. The fixed-vocabulary type segment is ordered before the caller-influenced name, so a name containing `:` can't shift an earlier segment's meaning.

Resource URIs were already effectively immune because their scheme prefix (`file://`, `https://`) never matches a bare tool or prompt name; the realistic collision is tool-vs-prompt, which this closes.

## Compatibility

This is a breaking change to the **runtime** counter-key schema: in-flight `maxCalls` windows reset on upgrade. Counters are in-memory or short-TTL Redis entries, so the practical impact is one window's worth of recounting. This is acceptable under the pre-1.0 no-back-compat-shim policy; a CHANGELOG note is included.

## Out of scope

`sequenceHistoryKey` (`seq:<session>:<name>`) has the same structural property but is intentionally left unchanged to keep this PR scoped to `maxCalls`, as #205 intended. Tracked as a follow-up in **#221**.

## Tests

- `TestEngine_MaxCalls_TargetTypeNamespacesQuota` — exhausts a tool's quota and asserts the same-named **prompt and resource** each keep their full, independent quota (all three namespaces mutually independent), then are independently rate-limited on their own third call. Genuinely fails on the pre-fix key.
- `TestEngine_MaxCalls_NilTargetUsesEmptyTypeSegment` — pins the documented empty-segment fallback (`maxcalls:<session>::<name>`): a nil-`Target` request and a `"tool"`-typed request with the same session and bare name land in different buckets.
- `go test -race -count=1 ./...` — all packages pass. `go vet ./...` clean; changed files gofmt-clean.

https://claude.ai/code/session_01Fa3eCscuRnL6rNmopopSkH


pkg/enforcement/handlers.go
```diff
@@ -270,11 +270,25 @@ func (e *Engine) handleMaxCalls(ctx context.Context, cond capability.Condition,
...
+	//
+	// Route scoping is not needed because session IDs are server-assigned UUIDs
+	// (one per connection) and the HTTP transport rejects requests whose session
+	// belongs to a different route, making cross-route key collision
+	// architecturally impossible.
+	targetType := ""
+	if req.Target != nil {
+		targetType = req.Target.Type
+	}
+	key := fmt.Sprintf("maxcalls:%s:%s:%s", req.SessionID, targetType, req.ToolName)
```
eunolabs, line 322: The identical root cause still lives in `sequenceHistoryKey` (`pkg/enforcement/engine.go:79`): `"seq:" + sessionID + ":" + tool` is also built from the bare name, so a tool and a prompt that share a name record into the same `seq:` bucket. A `sequenceBlock` keyed on prior tool `export` would then be tripped by a `prompts/get` for a prompt also named `export`. The PR explicitly scopes this out and #205 calls it out too, which is reasonable — but since it's the same bug shape, it'd be good to file a tracking issue so the second half doesn't get lost. Non-blocking.

---

pkg/enforcement/enforcement_test.go
```diff
@@ -368,6 +368,70 @@ func TestEngine_MaxCalls_EmptySessionIDDenies(t *testing.T) {
 	assert.Equal(t, capability.ConditionTypeMaxCalls, resp.Denial.ConditionType)
 }
 
+func TestEngine_MaxCalls_TargetTypeNamespacesQuota(t *testing.T) {
```
eunolabs, line 378: Good regression test — it genuinely fails on the pre-fix key (the first prompt call would land on the shared, already-exhausted bucket). Two optional additions to lock the contract down further: (1) a `resource`-type request, to demonstrate all three namespaces are mutually independent rather than just tool-vs-prompt; and (2) a nil-`Target` request, to pin the documented empty-segment fallback (`maxcalls:<session>::<name>`) so a future refactor can't silently change it. Nit, non-blocking.

---

eunolabs (commented):
## Review: namespace `maxCalls` counter key by target type

### Overview
Fixes #205. `handleMaxCalls` built the per-session counter key from `req.SessionID` + `req.ToolName`, where `ToolName` is the *bare* target name (tool name for `tools/call`, prompt name for `prompts/get`, URI for `resources/read`). A tool and a prompt sharing a name (e.g. `export`) collided on `maxcalls:<session>:export` and drained one shared quota bucket. The fix inserts the target type — `maxcalls:<session>:<type>:<name>` — giving each namespace its own bucket.

### Correctness — looks right
- `req.Target` is in fact populated with a distinct `Type` by all three `ManifestPDP` entry points: `Decide` (`pdp.go:242` → `"tool"`), `DecideResourceRead` (`pdp.go:405` → `"resource"`), `DecidePromptGet` (`pdp.go:492` → `"prompt"`), so the realistic tool-vs-prompt collision is closed.
- Ordering the fixed-vocabulary `type` segment *before* the variable, caller-influenced `name` segment is the right choice — a name containing `:` can't shift the meaning of an earlier segment.
- The nil-`Target` fallback (empty type segment) is internal-only; every production path sets `Target`, and the implementation matches the snippet in #205. (Minor: #205's prose said "fall back to the current no-prefix schema", but its *code* — which the PR followed — keeps the 3-segment shape with an empty middle. The PR's code comment describes the real behavior accurately, so this is just an issue-text quibble.)

### Tests
- `TestEngine_MaxCalls_TargetTypeNamespacesQuota` is a genuine regression test: it would fail pre-fix, since the first prompt call lands on the already-exhausted shared bucket. 
- Optional hardening suggested inline (resource-type + nil-`Target` cases).

### Risks / blockers
- **Merge conflict (blocks merge).** GitHub currently reports the PR as not mergeable (`mergeable_state: dirty`). The almost-certain culprit is the `### Fixed` section of `CHANGELOG.md`, which sibling fix PRs also append to — `main` has since gained other `Fixed` entries. A rebase on `main` is needed before this can merge; the code itself is unaffected.
- **Same bug shape remains in `sequenceHistoryKey`** (flagged inline). Explicitly scoped out, which is defensible, but worth a tracking issue.
- The breaking runtime counter-key change is correctly documented in the CHANGELOG and is acceptable under the pre-1.0 no-shim policy; impact is one window's worth of recounting.

### Verdict
Tight, well-scoped, well-documented fix; all 42 checks are green. No code changes requested. Please rebase to clear the `CHANGELOG.md` conflict before merging, and consider a follow-up issue for `sequenceHistoryKey`.

---

pkg/enforcement/enforcement_test.go
eunolabs: Done in 68ec4fd. Added both:
- A `resource`-type leg, so the test now asserts tool / prompt / resource are mutually independent (each keeps its own quota and is rate-limited on its own third call).
- A new `TestEngine_MaxCalls_NilTargetUsesEmptyTypeSegment` that pins the documented empty-segment fallback (`maxcalls:<session>::<name>`): a nil-`Target` request and a `"tool"`-typed request with the same session and bare name land in different buckets, and the nil-`Target` request still enforces its own quota.

---

pkg/enforcement/handlers.go
eunolabs: Good catch — filed #221 to track the same fix for `sequenceHistoryKey`. Kept it out of this PR to keep the change scoped to `maxCalls`, as #205 intended; #221 captures the `seq:` key collision and a fix sketch (the lookup side needs the antecedent's type from each `afterTools` entry's prefix).

---

PR #210, opened 2026-06-13
fix: strip namespace prefix in handleMaxCalls counter key (#206)

### LINKED ISSUE #206: bug: handleMaxCalls does not strip namespace prefix from ToolName, inconsistent with handleSequenceBlock and recordSessionCall
## Summary

Three sites in `pkg/enforcement/` that build per-key strings from `req.ToolName` handle the optional namespace prefix (`tool:`, `resource:`, etc.) inconsistently:

| Site | Strips prefix? |
|------|---------------|
| `handleSequenceBlock` (handlers.go:736) | **Yes** — `blockedTool := stripEnginePrefix(req.ToolName)` |
| `recordSessionCall` (engine.go:199) | **Yes** — `tool := stripEnginePrefix(req.ToolName)` |
| `handleMaxCalls` (handlers.go:277) | **No** — uses `req.ToolName` raw |

## Location

```go
// pkg/enforcement/handlers.go — handleMaxCalls, line 277

// contrast — handleSequenceBlock, line 736
// …
count, err := history.Peek(ctx, sequenceHistoryKey(req.SessionID, priorTool), sequenceHistoryWindowSec)

// contrast — recordSessionCall, engine.go:199
```

## Impact

`ManifestPDP.Decide` always sets `req.ToolName = target.Name` (already bare), so the inconsistency is **invisible on the production path**. However:

1. Direct callers of `engine.ValidateAction` or `engine.EvaluateConditions` that pass a prefixed `ToolName` (e.g. `"tool:read_file"`) will get a different `maxcalls` bucket than the `sequenceBlock`/history buckets for the same logical tool. A `maxCalls` limit of N on tool `read_file` passed as `"tool:read_file"` would track under `maxcalls:<s>:tool:read_file`, while `sequenceBlock` tracks history under `seq:<s>:read_file`. These diverge.
2. The asymmetry is a latent correctness hazard for future transport or embedding work that may pipe prefixed names into the engine.
3. It is inconsistent with the stated invariant (CLAUDE.md): _"Conditions match a specific argument name; never silently match alternatives."_ — by analogy, condition key construction should also be stable and normalised.

## Fix

Apply the same `stripEnginePrefix` call in `handleMaxCalls`:

```go
// pkg/enforcement/handlers.go, handleMaxCalls
key := fmt.Sprintf("maxcalls:%s:%s", req.SessionID, toolName)
```

This is a one-line change with a one-line test addition (assert that a prefixed and bare ToolName produce the same counter key). No key schema migration is needed for users of the production ManifestPDP since that path always provides bare names today.

eunolabs:
## Summary

Fixes #206.

`handleMaxCalls` (`pkg/enforcement/handlers.go`) built its quota bucket key from `req.ToolName` **raw**, while the two other per-key sites in the package normalise the optional engine namespace prefix (`tool:`, `resource:`, …) via `stripEnginePrefix`:

| Site | Strips prefix? |
|------|----------------|
| `handleSequenceBlock` | Yes |
| `recordSessionCall` | Yes |
| `handleMaxCalls` | **No** → fixed here |

A direct caller of `engine.ValidateAction` / `engine.EvaluateConditions` that passes a prefixed name (e.g. `"tool:read_file"`) therefore tracked `maxCalls` under `maxcalls:<sess>:tool:read_file`, while the sequence-history bucket for the same logical tool used `seq:<sess>:read_file`. The two diverge, so a `maxCalls` limit was not enforced across the two spellings.

## Change

```go
// pkg/enforcement/handlers.go — handleMaxCalls
key := fmt.Sprintf("maxcalls:%s:%s", req.SessionID, toolName)
```

## Why it's safe

The production `ManifestPDP.Decide` path always sets `req.ToolName = target.Name` (already bare), so this is invisible on the production path and needs **no key-schema migration**. The fix removes a latent correctness hazard for direct embedders and future transport/embedding work that may pipe prefixed names into the engine, and aligns with the stated invariant that per-key construction should be stable and normalised.

## Tests

Added `TestEngine_MaxCalls_StripsNamespacePrefix`: a bare call (`read_file`) consumes a `Count: 1` quota, then a prefixed call (`tool:read_file`) must hit the **same** bucket and be denied with `ErrCodeRateLimited`. Verified the test fails against the pre-fix code (returns `allow`) and passes with the fix.

- `make check-fmt`, `make check-license`: clean
- `go test -race -count=1 ./...`: all packages pass

https://claude.ai/code/session_01E71kDyqMHMk5kLa71QNTua

---
_Generated by [Claude Code](https://claude.ai/code/session_01E71kDyqMHMk5kLa71QNTua)_


pkg/enforcement/handlers.go
```diff
@@ -274,7 +274,15 @@ func (e *Engine) handleMaxCalls(ctx context.Context, cond capability.Condition,
...
 	// route, making cross-route key collision architecturally impossible.
-	key := fmt.Sprintf("maxcalls:%s:%s", req.SessionID, req.ToolName)
+	//
+	// Strip the engine namespace prefix ("tool:", "resource:", ...) so a
+	// prefixed and a bare ToolName for the same logical tool share one quota
+	// bucket. This matches handleSequenceBlock and recordSessionCall, which
+	// already normalise via stripEnginePrefix; without it a direct caller of
+	// ValidateAction/EvaluateConditions passing "tool:read_file" would track
+	// under a different bucket than the sequence-history key for "read_file".
+	toolName := stripEnginePrefix(req.ToolName)
```
eunolabs: Fix is correct and the right call — aligning with `handleSequenceBlock`/`recordSessionCall`/`FindMatchingCapability`, all of which `stripEnginePrefix`. Two things worth recording, neither blocking:

1. **Cross-type collision after stripping.** Once the prefix is dropped, `tool:read_file` and `resource:read_file` map to the same `maxcalls:<sess>:read_file` bucket and so share one quota. That's a behavior change only for direct embedders that pass *prefixed* names to distinguish types; it's already how the rest of the engine treats names (sequence history + matching both strip), and the production `ManifestPDP.Decide` path passes bare `target.Name`, so it's invisible there. Calling it out so it's a deliberate property rather than a surprise.

2. **No empty-name guard here, and that's correct.** `recordSessionCall` skips when `stripEnginePrefix(...) == ""`, but doing the same in `handleMaxCalls` would fail *open* (skip the limit) — so the asymmetry is the right choice for fail-closed. A degenerate `"tool:"` just buckets under the empty bare name, still scoped by session. No change requested; noting it in case the divergence from `recordSessionCall` looks accidental later.

---

eunolabs (commented):
## Review — `fix: strip namespace prefix in handleMaxCalls counter key (#206)`

**Verdict: looks good, no blocking issues.** Clean, well-scoped one-line correctness fix with a genuine regression test.

### What it does
`handleMaxCalls` built its quota key from raw `req.ToolName`, while the package's other two per-key sites (`handleSequenceBlock`, `recordSessionCall`) normalise via `stripEnginePrefix`. A caller passing a prefixed name (`tool:read_file`) therefore tracked `maxCalls` under `maxcalls::tool:read_file` while the sequence-history bucket used `seq::read_file` — the two diverge and a `maxCalls` limit wasn't enforced across spellings. This aligns `handleMaxCalls` with the rest of the engine.

### Correctness — verified locally (PR head `e7a2378`)
- `go test -race -run TestEngine_MaxCalls ./pkg/enforcement/` → **pass**; `go vet` clean; `gofmt -l` clean.
- Reverted *only* the source change (kept the test): the new test fails with `allow` instead of `deny` → confirms it's a real regression guard, not a tautology.
- Fix is consistent with `stripEnginePrefix` usage at `engine.go:199` (record), `engine.go:353` (match), and `handlers.go:736-738` (sequence). The grep for counter-key construction in the package shows only these three sites, so the inconsistency is now fully closed.

### Conventions
- Conventional Commit `fix:` ✓, scoped to `pkg/enforcement` ✓, no new files / license headers needed ✓.
- No manifest-grammar, audit-shape, or config-key change, and the production path (`ManifestPDP.Decide` → bare `target.Name`) is byte-identical before/after, so no key-schema migration and no docs/threat-model update required. PR body already reasons through this correctly.
- Fail-closed preserved (and see the inline note on why *not* guarding the empty-name case is the right call).

### Test coverage
Solid: allow-then-deny across bare→prefixed spellings, asserting `Decision`, `Denial.Code` (`ErrCodeRateLimited`), and `ConditionType`. Optional nice-to-have (not requested): the reverse order (prefixed→bare) for symmetry, but it exercises the same key path so it adds little.

One non-blocking awareness note left inline re: cross-type bucket collision after stripping. Nothing here needs to change to merge.

---

eunolabs:
Closing as **superseded by #246** (`93e5beb fix: normalize tool name in maxCalls counter key (#240) (#246)`), which merged to `main` while this PR was open.

#246 already implements the exact fix this PR set out to make — and a bit more:

- **Same key shape:** `maxcalls:<sessionID>:<targetType>:<stripEnginePrefix(ToolName)>`. The bare name is normalized with `stripEnginePrefix`, so a prefixed (`tool:read_file`) and a bare (`read_file`) call share one quota bucket, consistent with `handleSequenceBlock` and `recordSessionCall`.
- It also keeps the `targetType` segment from #205/#209, so distinct capability types (a tool vs. a prompt both named `export`) stay in separate buckets — the constraint this PR's original approach would have violated.
- Extras this PR lacked: a fail-closed guard that denies an empty/unidentifiable bare name (`MISSING_CONTEXT`), plus regression tests `TestEngine_MaxCalls_PrefixNormalizedKey` (equivalent to this PR's `StripsNamespacePrefix`) and `TestEngine_MaxCalls_EmptyToolNameDenies`.

Net result: `main` is a strict superset of this PR, so there's nothing left to merge here. The underlying issue (#206) is resolved on `main`. No code changes needed.

---

PR #211, opened 2026-06-13
perf: release InMemory call-counter backing array after a burst drains

### LINKED ISSUE #207: perf: InMemory call counter never releases backing-array capacity after a call burst, leaking heap proportional to peak concurrency
## Summary

`InMemory.IncrementAndGet` in `pkg/callcounter/memory.go` prunes expired timestamps by reusing the existing slice's backing array:

```go
valid := e.timestamps[:0]          // length 0, capacity unchanged
    if ts.After(cutoff) {
        valid = append(valid, ts)
    }
}
valid = append(valid, now)
e.timestamps = valid
```

After a burst of N concurrent calls within one window, `e.timestamps` reaches capacity N. When the window expires and the next call prunes all N expired timestamps and inserts one new one, `e.timestamps` ends up with length 1 but **still holds a backing array of capacity N**. That capacity is never released.

## Location

`pkg/callcounter/memory.go`, lines 86–95.

## Impact

For a tool with `maxCalls` that sees an occasional traffic burst (e.g. 10 000 concurrent calls during an incident, then steady-state 1 req/s), the `InMemory` store retains a 10 000-element array per key indefinitely. In a gateway with many such tools this can cause a significant and permanent heap bump that `Cleanup` does not address (it deletes entire stale entries, not oversized live ones).

`Cleanup` itself does not shrink the capacity of retained entries — it only deletes entries whose **most-recent** timestamp is older than `cleanupMarginFactor × windowSec`. A live entry (receiving steady traffic) retains its peak capacity forever.

## Reproducer sketch

```go
// 10 000 concurrent increments with a 60 s window
for i := 0; i < 10000; i++ {
    go counter.IncrementAndGet(ctx, "k", 60)
}
// … wait for window to expire …
// Next call prunes all 10 000 timestamps, inserts one, but cap remains 10 000
counter.IncrementAndGet(ctx, "k", 60)
// cap(entry.timestamps) == 10 000
```

## Fix

After filtering, check whether the backing array is significantly oversized and copy to a fresh slice:

```go
valid := e.timestamps[:0]
    if ts.After(cutoff) {
        valid = append(valid, ts)
    }
}
valid = append(valid, now)

// Release excess capacity when utilisation drops below 25%.
if cap(valid) > 64 && len(valid)*4 < cap(valid) {
    compact := make([]time.Time, len(valid))
    copy(compact, valid)
    valid = compact
}
e.timestamps = valid
```

The threshold (64 / 4×) is tunable; the key invariant is that the backing array shrinks proportionally when the active window is much smaller than the burst peak. Alternatively, use `slices.Clip(valid)` (Go 1.21+) to trim to exact length after each prune.

eunolabs:
## Summary

Fixes #207 — `InMemory.IncrementAndGet` (`pkg/callcounter/memory.go`) leaked heap proportional to peak burst concurrency.

`IncrementAndGet` prunes expired timestamps by reslicing the entry's backing array with `[:0]`. That keeps appends amortised O(1) but **never shrinks capacity**: after a burst of N concurrent calls within one window drains to a steady-state handful, the entry pins an N-element array indefinitely. `Cleanup` doesn't help — it deletes whole *stale* entries, never shrinks a *live* one — so a key receiving steady traffic (e.g. 10 000 calls during an incident, then 1 req/s) kept its peak capacity forever. In a gateway with many such tools this was a permanent, unbounded heap bump.

## Fix

After pruning, reclaim the slice once its backing array is **both large (capacity > 64) and mostly empty (live length below a quarter of capacity)** by copying the live timestamps into a right-sized slice, letting the oversized one be garbage-collected.

The 25% threshold deliberately keeps the copy **off the steady-state hot path**: ordinary sliding-window churn stays near capacity and is left untouched, so only a genuine burst-then-drain pays for a single copy. Call **counts are unchanged**.

I chose threshold-gated compaction over the `slices.Clip` alternative mentioned in the issue: `Clip` only reslices the capacity, so it doesn't release the array until the next append and would force a reallocation on *every* steady-state append — defeating the amortised-O(1) property the prune-in-place was designed for.

## Tests

In-package (capacity isn't observable from the external test package):

- `TestInMemory_IncrementAndGet_ReleasesBurstCapacity` — regression test: after a 1000-call burst expires, the next call reclaims the oversized array (cap drops from burst peak to ≤ 64).
- `TestInMemory_IncrementAndGet_SteadyStateKeepsCapacity` — hot-path guard: a near-full slice receiving in-window traffic is **not** reallocated.
- `TestCompactTimestamps` — table-driven coverage of the thresholds (small / at-threshold / just-above / mostly-empty / near-capacity / exactly-25% / empty), confirming live contents always survive.

## Verification

- `make test` (`go test -race -count=1 ./...`) — all packages pass
- `make lint` (`go vet` + golangci-lint) — 0 issues
- `make check-fmt`, `make check-license` — clean
- `make build` — binary builds
- `pkg/callcounter` coverage: 98.1% (above the 80% gate)

CHANGELOG "Fixed" (Unreleased) updated. No manifest grammar / audit shape / schema / config change — purely an internal memory optimization, so no threat-model or guide update is required.

Closes #207

https://claude.ai/code/session_018yj7wRSNMeyZy7vFZnHHCS

---
_Generated by [Claude Code](https://claude.ai/code/session_018yj7wRSNMeyZy7vFZnHHCS)_


pkg/callcounter/memory.go
```diff
@@ -92,11 +92,41 @@ func (m *InMemory) IncrementAndGet(_ context.Context, key string, windowSec int)
...
+// IncrementAndGet prunes expired timestamps by reslicing the backing array with
+// [:0], which keeps appends amortised O(1) but never shrinks capacity: after a
+// burst of N concurrent calls within one window drains to a steady-state
+// handful, e.timestamps would otherwise pin an N-element array for as long as
+// the key stays live. Cleanup does not help — it deletes whole stale entries,
+// never shrinks a live one (issue #207). The 25% threshold keeps this copy off
+// the steady-state hot path: ordinary sliding-window churn stays near capacity
+// and is left untouched, so only a genuine burst-then-drain pays for one copy.
+func compactTimestamps(ts []time.Time) []time.Time {
+	if cap(ts) > compactMinCap && len(ts)*4 < cap(ts) {
```
eunolabs, line 131: Minor, non-blocking — comment wording. The doc says *"only a genuine burst-then-drain pays for one copy."* That's exactly right for a **one-shot** burst. A key that recurringly spikes to N and drains below N/4 each window will recompact once per cycle (then regrow on the next spike). That's cheap — the copy is of the small live set and is dwarfed by the regrowth it would do anyway — and arguably the behaviour you want, so no code change is needed. Just flagging that "one copy" is per burst-cycle, not per lifetime, in case you want to make the comment precise.

Separately, this condition is the one place worth being explicit that the steady-state safety rests on `append`'s growth factor: a sliding window that only grows sits at ~50% utilization right after each grow (≈80% above 1024 elems) and climbs toward 100%, so it never reaches the 25% trigger. Picking 25% rather than 50% is what keeps that margin — nice. No action needed; recording the reasoning for the next reader.

---

eunolabs (commented):
**Review — looks correct and well-targeted. No blocking issues.**

I walked through the logic and the threshold math:

- **Correctness:** `compactTimestamps` only ever returns the same slice or an order-preserving full copy of the live timestamps, under the existing `m.mu` lock. Counts (`len(valid)`) and contents are unchanged, and the fresh allocation can't alias the old array. `Peek` and `Cleanup` are untouched, which is right — a pure read must not mutate, and `Cleanup`'s `len == 0` delete path still holds since `IncrementAndGet` always appends `now` first.

- **The hot-path claim is genuinely sound.** `append`'s growth (2× below 1024 elems, ~1.25× above) means that right after a grow the slice sits at ~50% utilization (≈80% for large slices) and climbs toward 100% as the window fills. Pure sliding-window churn therefore never dips to the 25% trigger — only an actual *drop* in live count (a drained burst) does. So 25% is comfortably below the natural floor; choosing it over 50% is what preserves the margin. The "off the steady-state hot path" claim checks out.

- **The `slices.Clip` rejection is the right call** — clip-then-append would reallocate on every steady-state append, defeating the amortised-O(1) prune-in-place.

- **Tests** cover the boundary precisely: `len*4 < cap` vs the exactly-25% case, just-above / at `compactMinCap`, and empty; plus the end-to-end burst-drain regression and the steady-state no-shrink guard. 98.1% coverage is fine, and the in-package placement is justified (cap isn't observable externally).

- **CHANGELOG** entry is accurate and scoped; agree no threat-model/guide/schema change is needed since manifest grammar, audit shape, and config are all untouched.

One minor, non-blocking wording note left inline on `compactTimestamps`. Otherwise this is good to merge from my read.

---

pkg/callcounter/memory.go
eunolabs: Good call — made the comment precise on both points in 1ce9e11: the copy is now described as paid once per burst-then-drain *cycle* (not per lifetime), and the steady-state margin is spelled out as resting on `append`'s growth factor (~50% utilisation right after a grow, climbing to 100%, so churn never reaches the 25% trigger; 25% over 50% is what preserves the margin). Comment-only, no behaviour change.

---

PR #217, opened 2026-06-13
sec: stop maxCalls from counting denied calls (#204)

### LINKED ISSUE #204: bug: handleMaxCalls increments the counter for denied requests, enabling unbounded counter growth and self-extending lockout
## Summary

`pkg/enforcement/handlers.go` `handleMaxCalls` calls `e.counter.IncrementAndGet` **before** comparing the returned count to the limit. Every over-limit request therefore writes a new timestamp to the counter store, even though the request is denied immediately afterward. This has two distinct consequences:

1. **Unbounded counter state growth**: A rate-limited client that keeps retrying causes the sorted set (Redis) or timestamps slice (InMemory) to grow without bound — O(retries) storage per key.
2. **Self-extending lockout**: The denied calls' timestamps occupy the sliding window. If a client retries at a rate ≥ `count/windowSeconds`, the counter never drops below the limit and the lockout persists indefinitely, even past the original window expiry. A legitimate client that accidentally exceeded the quota and then retries to check whether the window has cleared will hold themselves in the locked-out state as long as they keep retrying.

## Location

`pkg/enforcement/handlers.go`, lines 279–299:

```go
count, err := e.counter.IncrementAndGet(ctx, key, mc.WindowSeconds)  // always writes
if err != nil {
    return &ConditionError{...}
}

if count > int64(mc.Count) {
    return &ConditionError{...}  // denial, but timestamp already committed
}
return nil
```

## Concrete scenario

Limit = 5, window = 60 s, client hammers at 1 req/s after being rate-limited:

| Time | Event | Counter state |
|------|-------|---------------|
| T=0–4 | 5 allowed calls | 5 entries |
| T=5 | 6th call: `IncrementAndGet` → 6, denied | 6 entries |
| T=6 | 7th: `IncrementAndGet` → 7, denied | 7 entries |
| … | … | … |
| T=60 | The 5 original timestamps expire (T=0–4 fall out) | But T=5–60 denied-call timestamps remain: 55 entries |
| T=61 | Counter = 56, still well above limit of 5 | Lockout continues |

The lockout clears only when the client **stops retrying for a full window**. This is not the standard sliding-window semantic (where waiting one window period is sufficient to recover from any burst).

## Impact

* **Resource exhaustion**: A single rate-limited session can drive the Redis sorted set or InMemory slice to grow without limit until the key TTL fires (2× window seconds), occupying O(retries × member size) storage.
* **Unexpected operator behaviour**: Operators configuring `maxCalls` to limit tool use to N per window expect that the window clears after `windowSeconds`. The self-extending property makes recovery depend on the client stopping rather than on time passing.

## Fix directions

Two approaches:

**Option A – peek, then conditionally increment (TOCTOU risk under concurrent load):**
```go
current, _ := counter.Peek(ctx, key, mc.WindowSeconds)
if current >= int64(mc.Count) {
    return &ConditionError{...}
}
e.counter.IncrementAndGet(ctx, key, mc.WindowSeconds)
return nil
```
This avoids writing for already-over-limit calls but introduces a TOCTOU window under high concurrency (two goroutines both peek at N-1 and both increment to N, allowing N+1 through).

**Option B – increment, check, and decrement on deny (not supported by the sliding-window backend):**
The sorted-set / timestamp approach does not support a cheap "undo last increment."

**Option C – only record when the call is allowed (change semantics):**
Move `IncrementAndGet` to after the limit check using `Peek` first (conservatively compare against `mc.Count - 1` to close the TOCTOU gap for single-replica deployments), or document that denied calls are counted against the window and expose a `Retry-After` hint in the denial details so callers know when to retry.

At minimum the `Details` map in the denial should include a `retryAfter` field so callers can back off correctly.

eunolabs:
## Summary

Fixes #204.

`handleMaxCalls` (`pkg/enforcement/handlers.go`) called `IncrementAndGet` **before** comparing the returned count to the limit, so every over-limit retry wrote a new timestamp even though the call was denied. Two consequences:

1. **Unbounded counter growth** — a rate-limited client that keeps retrying grows the Redis sorted set / in-memory slice by O(retries) per key until the key TTL fires.
2. **Self-extending lockout** — denied calls occupy the sliding window, so a client retrying at ≥ `count/windowSeconds` holds itself locked out indefinitely; the window only clears after the client *stops* retrying for a full window, not after `windowSeconds` elapses.

## Fix

Replace increment-then-compare with an **atomic check-and-record**:

- **New optional `RateLimiter` interface** (`IncrementIfBelow`) in `pkg/enforcement`. `maxCalls` requires it and **fails closed** on a counter that does not implement it — exactly mirroring how `sequenceBlock` requires `CallHistory`.
- **InMemory** does the prune + count + record under a single mutex; **Redis** does it in a single **Lua script** (`EVAL`). This closes the TOCTOU gap a separate `Peek`-then-`IncrementAndGet` would leave open under concurrent load, and adds **no new runtime dependency** (the Lua runs through the existing go-redis client).
- **Denied calls record nothing** and prune expired entries, so a rate-limited key *shrinks* rather than grows, and the window clears `windowSeconds` after the last **admitted** call regardless of retry rate.
- `RATE_LIMITED` denials now carry a **`retryAfter`** hint (whole seconds until a slot frees up) in their details so callers can back off.

`IncrementAndGet` is retained unchanged for the `sequenceBlock` history path.

## Tests

- `pkg/callcounter`: `IncrementIfBelow` admit-up-to-limit, out-of-range window fail-closed, and a **#204 regression** for both InMemory and Redis (retry flood does not grow state or extend the lockout; recovery happens on time via the injected clock).
- `pkg/enforcement`: engine-level #204 regression asserting the `retryAfter` detail and on-time recovery, plus a fail-closed test for a counter without `RateLimiter`. Compile-time conformance asserts for both backends.
- `go test -race ./...`, `go vet ./...`, and `golangci-lint run ./...` all clean.

## Docs

- `docs/capability-manifest-guide.md` — documents that `maxCalls` counts only admitted calls and that denials carry `retryAfter`.
- `docs/threat-model-mcp.md` §3.6 — updated for the atomic, admitted-only counting and the closed self-extending-lockout / unbounded-growth bug.

Closes #204

---
_Generated by [Claude Code](https://claude.ai/code/session_01E7sPKpbbRmpb6okjycKAgq)_


pkg/callcounter/redis.go
```diff
@@ -114,6 +114,83 @@ func (r *Redis) IncrementAndGet(ctx context.Context, key string, windowSec int)
...
+if limit < 1 or count >= limit then
+  local retry = 0
+  local oldest = redis.call('ZRANGE', KEYS[1], 0, 0, 'WITHSCORES')
+  if limit >= 1 and oldest[2] then
+    retry = (tonumber(oldest[2]) + tonumber(ARGV[6])) - tonumber(ARGV[2])
+    if retry < 0 then retry = 0 end
+  end
+  return {0, count, retry}
+end
+redis.call('ZADD', KEYS[1], ARGV[2], ARGV[3])
```
eunolabs, line 146: Nit (non-blocking, consistency): this path stores scores as **integer** microseconds (`ARGV[2] = now.UnixMicro()`), whereas `IncrementAndGet` above stores them as **float**-formatted micros (`float64(now.UnixMicro())` / `fmt.Sprintf("%f", cutoff)`). This is harmless today — `maxCalls` keys (`maxcalls:…`) and `sequenceBlock` keys (`seq:…`) never land in the same sorted set, and `float64(UnixMicro())` is exact for present-day timestamps — so there's no correctness issue. But the two methods now write the same logical value in two different formats into the same key *namespace*; worth a one-line note (or aligning on int micros) so a future reader who ever points both at one key isn't surprised by mixed score encodings.

---

pkg/enforcement/handlers.go
```diff
@@ -284,15 +299,27 @@ func (e *Engine) handleMaxCalls(ctx context.Context, cond capability.Condition,
...
 		return &ConditionError{
 			Code:          capability.ErrCodeRateLimited,
 			ConditionType: capability.ConditionTypeMaxCalls,
 			Message:       "call limit exceeded",
 			Details: map[string]interface{}{
-				"limit":   mc.Count,
-				"current": count,
-				"window":  mc.WindowSeconds,
+				"limit":      mc.Count,
+				"current":    count,
```
eunolabs, line 320: Heads-up (non-blocking): the semantics of the `current` denial detail change with this fix. Pre-fix, `current` was the post-increment count, so it climbed past the limit on every denied retry (`limit + N`) and acted as a rough flood-magnitude signal. Now `current` is the capped in-window admitted count, so it sits at ~`limit` no matter how hard a client hammers. That's the *correct* behavior given denied calls are no longer recorded — just flagging it because anything dashboarding on `current` to gauge abuse volume should switch to counting `RATE_LIMITED` audit records (and/or the new `retryAfter`) instead. No code change required.

---

eunolabs (commented):
## Review — `sec: stop maxCalls from counting denied calls (#204)`

Reviewed the diff and the surrounding code; checked out the branch and ran `go test -race -count=1 ./pkg/callcounter/... ./pkg/enforcement/...` (all green) plus `go vet`. This is a solid, correct security fix. Summary below; two non-blocking inline notes.

### What's right
- **Core fix is correct.** Replacing increment-then-compare with an atomic check-and-record is the right shape for the bug. A denied (over-limit) call now records nothing, so the sliding window holds only admitted calls and clears `windowSeconds` after the oldest admitted call — directly closing both the self-extending lockout and the unbounded counter growth in #204.
- **Atomicity is real on both backends.** InMemory does prune+count+record under the single existing mutex; Redis does it in one `EVAL` Lua script, closing the TOCTOU gap a separate `ZCOUNT`-then-`ZADD` would leave under concurrent/multi-replica load. No new runtime dependency (Lua runs through the existing go-redis client) — consistent with the one-static-binary invariant.
- **Fail-closed posture preserved and extended.** `maxCalls` now requires the new `RateLimiter` capability and denies (`CONDITION_FAILED`) on a counter that doesn't implement it, exactly mirroring how `sequenceBlock` requires `CallHistory`. Out-of-range windows still fail closed via `checkWindowSec`, ahead of any TTL arithmetic. `Count >= 1` / window bounds remain enforced at manifest load (`validateMaxCalls`), so the `limit < 1` branch is purely defensive.
- **`retryAfter` math checks out** in both backends (verified the `valid[cur-limit]` derivation and the Lua `oldest_score + window - now`), and the engine's round-up + full-window fallback is sound. Returning `retryAfter` only on admitted-slot-expiry (not refreshing the Redis TTL on the deny path) is the correct call — denied calls extend nothing.
- **Test coverage is on point.** Per-backend admit-up-to-limit, out-of-range fail-closed, and the #204 retry-flood regression (state doesn't grow, lockout lifts on time via injected clock), plus an engine-level regression asserting the `retryAfter` detail and a counter-without-`RateLimiter` fail-closed test, plus compile-time conformance asserts. `errorCounter` correctly implements the new path so the counter-error branch is still exercised.
- Docs updated in the same PR (`capability-manifest-guide.md` + threat-model §3.6), per the behavior-change convention.

### Non-blocking notes (inline)
1. `redis.go` — `IncrementIfBelow` stores integer-micro scores while `IncrementAndGet` stores float-formatted micros; harmless (disjoint keys, exact values) but worth aligning or noting.
2. `handlers.go` — the `current` denial detail is now capped at the limit rather than climbing with denied retries; fine and intended, just a heads-up for anything dashboarding on it.

No blocking issues. Nice work — the fix, the atomicity argument, and the regression tests all line up.

---

pkg/callcounter/redis.go
eunolabs: Good call — added a one-line note in `IncrementIfBelow` (08a9616) explaining the integer-micros encoding and that the `maxcalls:*` and `seq:*` keys never share a sorted set, so the two encodings can't meet. Kept `IncrementAndGet` as-is to avoid churning the existing tested path.

---

pkg/enforcement/handlers.go
eunolabs: Agreed — that semantic shift is intended: with denied calls no longer recorded, `current` correctly caps at ~`limit`. Abuse-volume dashboards should count `RATE_LIMITED` audit records (and/or use the new `retryAfter`) rather than `current`. No code change.

---

PR #218, opened 2026-06-13
fix: reject malformed recipient emails in recipientDomain

### LINKED ISSUE #216: bug: recipientDomain condition accepts "user@" (empty domain) and produces misleading error
## Summary

`handleRecipientDomain` (`pkg/enforcement/handlers.go:592-613`) splits a recipient on `@` with `strings.SplitN(recipient, "@", 2)` and checks `len(parts) != 2` to reject strings that contain no `@`. However, it does not reject the case where the domain part (after `@`) is an empty string.

## Reproduction

Recipient string: `"user@"` (technically malformed — nothing after the `@`).

```go
parts := strings.SplitN("user@", "@", 2)
// parts = ["user", ""]
// len(parts) == 2, so the len check passes
// domain = ""
```

The handler then looks up `domainSet[""]`. Unless the manifest has an (unusual) allowed domain of `""`, this correctly denies the request — but with the message:

```
recipient domain "" is not allowed
```

rather than the appropriate:

```
invalid recipient email: user@
```

An operator reading the audit log sees an `allowedDomains` field and a `domain` field of `""`, which is confusing. The denial looks like a legitimate domain-restriction event rather than a malformed input.

## Why it matters

The `recipientDomain` condition is the only line of defence against misrouted emails in MCP tool calls. A malformed recipient (`user@`, `@@example.com`) should be caught with an explicit "invalid email" error distinct from a policy denial, because:

1. It makes audit records actionable — operators can distinguish "agent tried to send to an unapproved domain" from "agent's tool received a malformed recipient".
2. An empty domain could coincide with a manifest typo (`domains: [""]`) and silently allow the call.

## Suggested fix

After splitting, add an explicit check for an empty domain before the allowlist lookup:

```go
if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
    return &ConditionError{
        ConditionType: capability.ConditionTypeRecipientDomain,
        Message:       fmt.Sprintf("invalid recipient email: %s", recipient),
    }
}
```

Checking `parts[0] == ""` also catches `@example.com` (empty local part), which is similarly malformed and currently produces domain `"example.com"` — an unintended match if `"example.com"` is an allowed domain.

A test case should be added for `"user@"`, `"@example.com"`, and `"@@example.com"`, asserting that all three produce an `invalid recipient email` error, not a domain-restriction error.

## Location

`pkg/enforcement/handlers.go:592-600` — `handleRecipientDomain`

eunolabs:
Fixes #216.

## Problem

`handleRecipientDomain` (`pkg/enforcement/handlers.go`) split a recipient on `@` with `strings.SplitN(recipient, "@", 2)` and only rejected strings containing no `@` (`len(parts) != 2`). Two malformed cases slipped through:

- **Empty domain** (`user@`) → `parts = ["user", ""]`, then `domainSet[""]` is looked up, producing a misleading `recipient domain "" is not allowed` **policy denial** rather than an "invalid email" error. An operator reading the audit log sees an `allowedDomains`/`domain` field and a `domain` of `""`, which looks like a legitimate domain-restriction event.
- **Empty local part** (`@example.com`, `@@example.com`) → matches on the domain alone, an unintended allow if that domain is permitted.

The `recipientDomain` condition is the only line of defence against misrouted emails, so a malformed recipient must be caught with an explicit "invalid email" error distinct from a policy denial.

## Fix

After splitting, reject an empty local part or empty domain before the allowlist lookup:

```go
if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
    return &ConditionError{
        ConditionType: capability.ConditionTypeRecipientDomain,
        Message:       fmt.Sprintf("invalid recipient email: %s", recipient),
    }
}
```

Malformed recipients now fail closed with `invalid recipient email` (code `CONDITION_FAILED`), distinct from the `recipient domain "…" is not allowed` denial. This also prevents a `domains: [""]` manifest typo from silently allowing an empty-domain recipient.

## Tests

Added table-driven cases to `TestRecipientDomain_ErrorBranches` for `user@`, `@corp.example`, and `@@corp.example` (against an allowlist of `corp.example`), asserting each is denied with an `invalid recipient email` message and **not** a `not allowed` domain-restriction message.

## Docs

Documented the malformed-email handling in `docs/capability-manifest-guide.md` alongside the `recipientDomain` example.

This is a runtime input-validation fix; the manifest grammar (`type`/`argument`/`domains`) is unchanged, so no spec-repo (`mcp-capability-manifest`) update is required.

### Verification

- `go build ./...` — clean
- `go vet ./pkg/enforcement/` — clean
- `gofmt -l` — clean
- `go test -race -count=1 ./pkg/enforcement/ ./pkg/capability/` — pass

https://claude.ai/code/session_013zGjUEBrRmwWzHhDTN6of3

---
_Generated by [Claude Code](https://claude.ai/code/session_013zGjUEBrRmwWzHhDTN6of3)_


pkg/enforcement/handlers.go
```diff
@@ -614,7 +614,12 @@ func (e *Engine) handleRecipientDomain(_ context.Context, cond capability.Condit
 
 	for _, recipient := range recipients {
 		parts := strings.SplitN(recipient, "@", 2)
-		if len(parts) != 2 {
+		// Reject anything without exactly one usable local part and domain:
+		// no "@" (len != 2), empty domain ("user@"), or empty local part
+		// ("@example.com", "@@example.com"). An empty domain would otherwise
+		// look up domainSet[""] and surface a misleading domain-restriction
+		// denial; an empty local part would match on the domain alone.
+		if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
```
eunolabs: The condition is correct and the `SplitN(..., 2)` semantics line up with each case in the comment: `@@example.com` splits to `["", "@example.com"]`, so `parts[0] == ""` catches it, and `user@` splits to `["user", ""]`, caught by `parts[1] == ""`. Failing closed with `CONDITION_FAILED` before the allowlist lookup is the right call, and the negative `NotContains(..., "not allowed")` assertion in the test is a good regression guard.

Optional, non-blocking nit (same "malformed email surfaces as a domain-restriction denial" class as #216, but lower severity since it still fails closed): a recipient with internal whitespace around the `@`, e.g. `"user@ example.com"`, isn't caught here — `TrimSpace` only runs on the whole string, so `parts[1]` becomes `" example.com"`, which misses the allowlist and produces a `recipient domain " example.com" is not allowed` message rather than `invalid recipient email`. Could trim `parts[0]`/`parts[1]` (or reject parts containing whitespace) if you want full message-clarity coverage, but I'd be fine deferring this — it's out of the issue's stated scope.

---

eunolabs (commented):
## Review — fix: reject malformed recipient emails in `recipientDomain`

Clean, well-targeted fix. It implements the suggested fix from #216 exactly, fails closed with a distinct `CONDITION_FAILED` / `invalid recipient email` error before the allowlist lookup, and the table-driven cases (`user@`, `@corp.example`, `@@corp.example`) assert both the new error *and* the absence of the misleading `not allowed` message. Docs are updated in the same PR and the "no spec-repo update needed" reasoning is correct (manifest grammar `type`/`argument`/`domains` is unchanged). I confirmed the three malformed inputs are caught by the new condition and that no legitimate address can be falsely rejected, since an empty local part or empty domain is never valid.

One inline note on `handlers.go` (optional, non-blocking).

### Follow-up suggestion (separate PR, not blocking)
The PR description and docs lean on the `domains: [""]` typo scenario as motivation — and the runtime guard does neutralize it now, since no valid recipient can reach `domainSet[""]`. But the manifest loader still accepts `domains: [""]` (and `domains: []`) silently: there's no validation of `RecipientDomainCondition.Domains` in `pkg/capability/validate.go` (the only non-test references to `Domains` are the struct declaration and this handler). Rejecting empty/whitespace-only domain entries at manifest load would surface the typo as a config error rather than a silent no-op entry, which fits the "fail closed + actionable" posture. Worth a follow-up issue; it's orthogonal to this fix.

LGTM to merge as-is.

---

pkg/enforcement/handlers.go
eunolabs: Addressed in 0f0b710. Rather than trimming (which would normalize a malformed address into an accepted one), the check now rejects any internal whitespace in either part via `strings.ContainsFunc(part, unicode.IsSpace)`, so `user@ example.com`, `user @corp.example`, and `us er@corp.example` all fail closed with `invalid recipient email`. Added those to the table-driven coverage.

---

eunolabs:
Also handled your `domains: [""]` follow-up in this PR (0f0b710) rather than deferring it. `validateRecipientDomain` in `cmd/eunox-mcp/manifest.go` now rejects, at manifest load, an empty `domains` list and any empty/whitespace-only entry — mirroring the existing `ipRange` empty-cidrs and `allowedValues` empty-values checks. So the typo surfaces as a config error at `validate` time instead of lingering as a silent dead allowlist entry. Added manifest-validation cases for `domains: []`, `domains: [""]`, and a whitespace-only entry, plus a docs note in the manifest guide.

This is a load-time validation tightening, not a grammar change, so no `mcp-capability-manifest` spec update is needed.

---

pkg/enforcement/handlers.go
eunolabs: Verified `0f0b710` — looks good, no further concerns from me. Rejecting internal whitespace via `strings.ContainsFunc(part, unicode.IsSpace)` is the right call over trimming (which would have normalized a malformed address into an accepted one), and it correctly catches tabs/newlines too. Nice bonus that you also landed the manifest-load validation (`validateRecipientDomain` rejecting empty/whitespace-only `domains` entries) in the same change — that closes the `domains: [""]` typo at config time, not just at runtime. Tests and docs cover all the new cases. CI is green on the new head.

---

PR #219, opened 2026-06-13
sec: validate sequenceBlock afterTools at manifest load

### LINKED ISSUE #212: bug: sequenceBlock condition has no manifest-time validation
## Summary

`sequenceBlock` is the only condition type that is not validated in the `validateLocalManifest` switch statement (`cmd/eunox-mcp/manifest.go:202-268`). This means two classes of authoring mistakes pass `eunox-mcp validate` silently and only surface at runtime — in some cases by failing **open** rather than failing closed.

### 1. Empty `afterTools` not caught at validate time

Every other condition type that requires a non-empty list (e.g. `allowedValues`, `ipRange`) has a corresponding load-time check that tells the operator immediately:

```
capability at index 0, condition 1: ipRange requires a non-empty 'cidrs' list
```

`sequenceBlock` is documented identically in the guide ("AfterTools is required and must be non-empty; an empty list **fails closed**"), but the runtime handler is the only guard. An operator running `eunox-mcp validate` on a manifest with `afterTools: []` sees no error; they discover it only on the first live call, and they see a `CONDITION_FAILED` denial rather than a configuration error.

### 2. Unrecognized namespace prefix in `afterTools` entries silently fails open

`stripEnginePrefix` (`pkg/enforcement/engine.go:379`) recognises `tool:`, `resource:`, `prompt:`, and `system:` prefixes. An entry with any other prefix (e.g. `mcp:read_file`, or a case mismatch like `Tool:read_file`) is returned **unchanged** (`"mcp:read_file"`).

- History is **recorded** under the bare name: `"seq:<session>:read_file"` (because `req.ToolName` comes through the transport stripped).
- `sequenceBlock` **looks up** key `"seq:<session>:mcp:read_file"`, which is never populated.
- Result: the sequenceBlock condition always passes (`count == 0`), silently failing open.

There is no validate-time check that catches this. An operator writing `afterTools: [mcp:read_file]` instead of `afterTools: [read_file]` or `afterTools: [tool:read_file]` gets no warning from `eunox-mcp validate`.

## Expected behaviour

`eunox-mcp validate` should reject a manifest with:
- a `sequenceBlock` condition whose `afterTools` is empty
- a `sequenceBlock` condition whose `afterTools` contains an entry that, after stripping, produces the same string (i.e. the prefix is not a recognised namespace)

Consistent with the pattern for all other condition types:
- `allowedValues` — rejects empty `values`
- `allowedOperations` — rejects empty `argument`
- `ipRange` — rejects empty `cidrs` and malformed CIDRs
- `timeWindow` — rejects both-empty bounds and empty windows

## Location

- `cmd/eunox-mcp/manifest.go` — `validateLocalManifest`, lines 202-268 (switch missing a `SequenceBlockCondition` case)
- `pkg/enforcement/engine.go` — `stripEnginePrefix`, lines 379-387
- `pkg/enforcement/handlers.go` — `handleSequenceBlock`, lines 737-741 (runtime-only guard)

## Suggested fix

Add a `validateSequenceBlock` function in `manifest.go` and call it from the condition switch, mirroring `validateIPRange`. Validation should:

1. Reject `len(afterTools) == 0`.
2. For each entry, call `stripEnginePrefix(entry)`. If `entry` contains a colon and `stripEnginePrefix` returns the entry unchanged (i.e. the prefix is not a recognised namespace), reject the manifest with a clear message pointing at the recognised prefixes.

A test case should be added to `pkg/capability/capability_test.go` (malformed-input cases) and to `cmd/eunox-mcp/manifest_validation_test.go`.

eunolabs:
Fixes #212.

## Problem

`sequenceBlock` was the only condition type with no case in `validateLocalManifest` (`cmd/eunox-mcp/manifest.go`). Two authoring mistakes therefore passed `eunox-mcp validate` silently:

1. **Empty `afterTools`** — only the runtime handler guarded it, so the operator learned of the mistake on the first live call, via a `CONDITION_FAILED` denial indistinguishable from a real enforcement event.
2. **Unrecognized namespace prefix** (e.g. `mcp:read_file`, or a case mismatch like `Tool:read_file`) — `stripEnginePrefix` recognizes only `tool:`/`resource:`/`prompt:`/`system:` and returns any other prefix **unchanged**. Session history is recorded under the bare name (`read_file`) but looked up under the prefixed string (`mcp:read_file`), so the count is always zero and the condition **silently fails open** — a bypass of the sequencing rule.

## Fix

- Add `validateSequenceBlock` in `manifest.go` and wire it into the condition switch (value + pointer cases), mirroring `validateIPRange`. It rejects:
  - an empty `afterTools` list;
  - an entry that strips to the empty string (`""` or a bare `tool:`), which names no tool and is silently skipped at runtime;
  - an entry whose colon-prefix is not a recognized namespace.
- Export `StripEnginePrefix` from `pkg/enforcement` (mirroring the existing `MatchesResource`/`ResourceSpecificity` wrappers) so validation and the engine share **one** source of truth for the recognized-prefix set — add a prefix to the engine and validation stays in sync.

This preserves the fail-closed invariant: a misconfigured manifest is now rejected at load (and at proxy startup) with a clear, indexed error rather than loading and failing open at runtime.

## Tests / docs

- `cmd/eunox-mcp/manifest_validation_test.go` — `TestLoadManifest_SequenceBlockValidation` (deny: empty/missing list, unrecognized prefix, case mismatch, one-bad-among-several, empty-string entry, prefix-only entry; allow: bare names and all four recognized prefixes; plus an index-naming assertion).
- `pkg/capability/capability_test.go` — adds the missing `sequenceBlock` case to `TestConditionRoundTripByType` (it was the only condition type absent).
- `pkg/enforcement/arch_review_fixes_test.go` — `TestStripEnginePrefix` pins the recognized-namespace set behind the new export.
- `docs/capability-manifest-guide.md` — documents the new validate-time checks under `sequenceBlock`.

All of `go test -race ./...`, `go vet ./...`, `gofmt`, and `golangci-lint` pass; `pkg/` coverage stays well above the 80% gate.

https://claude.ai/code/session_016zmfqnHM4JWpM12GMHVDC5

---
_Generated by [Claude Code](https://claude.ai/code/session_016zmfqnHM4JWpM12GMHVDC5)_


cmd/eunox-mcp/manifest.go
```diff
@@ -456,6 +464,48 @@ func validateTimeWindow(i, j int, v *capability.TimeWindowCondition) error {
...
+	for k, entry := range v.AfterTools {
+		stripped := enforcement.StripEnginePrefix(entry)
+		if stripped == "" {
+			return fmt.Errorf("capability at index %d, condition %d, afterTools entry %d: sequenceBlock entry %q names no tool; use a bare tool name (e.g. read_file) or a tool: prefix (e.g. tool:read_file)", i, j, k, entry)
+		}
+		// A colon-bearing entry that stripEnginePrefix returns unchanged carries a
+		// prefix that is not a recognized namespace. The runtime would look up the
+		// prefixed string against history keyed by the bare name, never match, and
+		// silently pass (fail open). Reject it and point at the recognized prefixes.
+		if strings.Contains(entry, ":") && stripped == entry {
```
eunolabs, line 518: Non-blocking, but worth a thought: this check also rejects a **bare resource URI** like `file:///secrets`, even though that entry would actually match a prior resource read at runtime.

A resource read records session history under the bare URI — `pdp.go` sets `ToolName: uri` (no prefix), and `stripEnginePrefix("file:///secrets")` returns it unchanged because `file` isn't a recognized namespace. So `afterTools: ["file:///secrets"]` strips to `file:///secrets` and *would* fire. But here `strings.Contains(entry, ":") && stripped == entry` is true, so it's rejected as having "an unrecognized namespace prefix" — where the `:` is really the URI scheme, not a namespace prefix.

There's no functional loss: the `resource:`-prefixed form (`resource:file:///secrets`) strips to the same key and is accepted, and forcing the explicit prefix in a security policy is defensible. But the rejection is asymmetric (prefixed resource URI accepted, bare one rejected) and the message is a little misleading for a legitimate URI. If rejecting the bare form is intended, consider rewording to something like "ambiguous colon-bearing entry; prefix with one of tool:/resource:/prompt:/system:" rather than asserting it's an unrecognized *namespace prefix*.

---

eunolabs (commented):
## Review — `sec: validate sequenceBlock afterTools at manifest load`

Solid, well-scoped fix that correctly closes #212. The change moves the only un-validated condition type into the `validateLocalManifest` switch and preserves the fail-closed invariant: a misconfigured `sequenceBlock` is now rejected at load instead of failing **open** at runtime.

**What I checked**
- **Correctness vs. runtime.** Verified `validateSequenceBlock` mirrors `handleSequenceBlock`'s real behavior. The two fail-open classes from the issue are caught: an empty `afterTools` (which the runtime denies, but only on the first live call) and an entry with an unrecognized prefix (recorded under the bare name, looked up under the prefixed string → count always 0 → silent pass). Empty-string / prefix-only entries (`""`, `tool:`) are also rejected.
- **Single source of truth.** Exporting `StripEnginePrefix` so validation and the engine share one recognized-prefix set is the right call — add a namespace to the engine and validation stays in sync automatically. Pinned by `TestStripEnginePrefix`.
- **Recognized prefixes are meaningful.** Confirmed `resource:`/`prompt:`/`system:` antecedents aren't dead config: resource reads and prompt gets both flow through `recordSessionCall`, and the bare URI/name they record under matches the stripped prefixed form (e.g. `resource:file:///x` → `file:///x`).
- **Tests / hygiene.** Table-driven allow + deny + malformed cases, plus the index-naming assertion, are thorough. Built the branch and ran `go build ./...`, the touched tests with `-race`, `go vet`, and `gofmt -l` — all green. Adding the previously-missing `sequenceBlock` case to `TestConditionRoundTripByType` is a nice incidental cleanup.

**One non-blocking note** (inline on `manifest.go`): the unrecognized-prefix check also rejects a *bare* colon-bearing entry such as a resource URI `file:///secrets`, which would in fact match a prior resource read at runtime. No functional loss (the `resource:`-prefixed form works), but the rejection is asymmetric and the message is slightly misleading for a URI scheme. Worth a glance to decide whether that's intended.

No blockers from me — the fix is correct and the invariant is preserved.

---

cmd/eunox-mcp/manifest.go
eunolabs: Good catch, and confirmed — `pdp.go` sets `ToolName: uri` for resource reads, so `file:///secrets` records under the bare URI and a bare `afterTools: ["file:///secrets"]` entry does fire at runtime.

I kept the rejection rather than special-casing URIs: at load time a bare URI is indistinguishable from a prefix typo (`mcp:read_file`), and accepting bare colon-bearing entries would re-admit exactly the fail-open #212 closes — so requiring the explicit prefix keeps the policy fail-closed. But you're right the message was wrong. Reworded it (cec5166) to call it an *ambiguous colon-bearing entry* and point at the recognized prefixes / the working `resource:file:///secrets` form, updated the function + inline comments to note the bare-URI nuance, added a load-time test case locking in the asymmetry, and updated the guide.

---

eunolabs: Re-reviewed `cec5166` — this resolves it. Agreed the load-time rejection is the right call: a bare URI is indistinguishable from a prefix typo there, and admitting bare colon-bearing entries would re-open the exact fail-open #212 closes. The "ambiguous colon-bearing entry" wording + the `resource:file:///secrets` pointer is clear, and the new bare-URI test case nicely pins the deliberate asymmetry. Build, the validation test, and gofmt all pass locally. Resolving the thread.

---

PR #220, opened 2026-06-13
perf: pre-compile ipRange CIDRs at manifest load (#215)

### LINKED ISSUE #215: perf: ipRange CIDRs re-parsed on every request despite validated at load time
## Summary

`validateIPRange` (`cmd/eunox-mcp/manifest.go:389-408`) already validates and parses every CIDR in an `ipRange` condition at manifest load time to catch host-bits errors and malformed notation. Despite this, `handleIPRange` (`pkg/enforcement/handlers.go:213-224`) re-parses each CIDR string with `net.ParseCIDR` on **every** incoming request:

```go
for _, cidr := range ipr.CIDRs {
    _, network, err := net.ParseCIDR(cidr)   // <-- called on every request
    if err != nil {
        return &ConditionError{...}
    }
    if network.Contains(ip) {
        return nil
    }
}
```

`net.ParseCIDR` allocates and parses on each call. For deployments that:
- have `ipRange` conditions with large CIDR allowlists (e.g. RFC-1918 blocks split into many /24s)
- handle high request volumes (the HTTP transport's hot enforcement path)

this is unnecessary work that could be eliminated entirely.

## How other conditions handle this

- `validateTimeWindow` parses RFC3339 timestamps at load time and stores them; the runtime handler re-parses them too (same issue, though time parsing is cheaper than CIDR parsing).
- `validateIPRange` calls `net.ParseCIDR` during validation but the result is discarded — the parsed `*net.IPNet` is never stored.

## Suggested fix

Store pre-parsed `*net.IPNet` values alongside (or instead of) the string slice. Options:

**Option A — parallel slice on the struct (source-compatible):**
Add a `parsed []net.IPNet` field to `IPRangeCondition` (or a separate `IPRangeConditionCompiled` type) populated by the manifest loader. The handler uses the compiled slice when non-nil and falls back to parsing if nil (for programmatically-constructed constraints).

**Option B — pre-compile in the manifest loader:**
After validating, replace the string CIDRs in the condition with a new type that holds `[]*net.IPNet`. Since this is pre-1.0 and backwards-compatibility shims are explicitly out of scope per CONTRIBUTING.md, a clean type change is acceptable.

Either way, `handleIPRange` would then be a simple loop over `[]*net.IPNet`:

```go
for _, network := range ipr.Parsed {
    if network.Contains(ip) {
        return nil
    }
}
```

The manifest loader already validates every CIDR, so the handler's `err != nil` branch (line 215) could also be removed as unreachable from validated manifests — though it may be kept for defensiveness when conditions are constructed programmatically.

## Location

- `pkg/enforcement/handlers.go:186-236` — `handleIPRange`
- `cmd/eunox-mcp/manifest.go:389-408` — `validateIPRange` (where pre-parsing could live)

eunolabs:
Fixes #215.

## Problem

`validateIPRange` (`cmd/eunox-mcp/manifest.go`) already parses every CIDR in an `ipRange` condition at manifest load time to catch host-bits errors and malformed notation — but it discarded the parsed `*net.IPNet`. As a result `handleIPRange` (`pkg/enforcement/handlers.go`) called `net.ParseCIDR` for **each** CIDR on **every** incoming request. That is wasted allocation/parsing on the HTTP transport's hot enforcement path, growing with the allowlist size (e.g. RFC-1918 blocks split into many /24s) and request volume.

## Fix

Implements **Option A** from the issue — store the pre-parsed networks with the condition:

- **`pkg/capability/condition.go`** — `IPRangeCondition` gains an unexported `parsed []*net.IPNet` field plus `Compile()` (parse-and-cache, idempotent, commits only on full success) and `Networks() ([]*net.IPNet, bool)`. The field is unexported, so it never serializes and never round-trips through a manifest.
- **`cmd/eunox-mcp/manifest.go`** — after `validateIPRange` confirms the CIDRs parse, the loader calls `Compile()` on the condition the engine evaluates at runtime (the unmarshaled `*IPRangeCondition` pointer). This covers stdio, HTTP single-upstream, and gateway mode, all of which load via `LoadManifest`.
- **`pkg/enforcement/handlers.go`** — `handleIPRange` matches against `Networks()` when compiled, with a fallback that parses per request for programmatically-constructed conditions. The malformed-CIDR branch is **kept** in the fallback so hand-built conditions still fail closed (`CONDITION_FAILED`).
- **`jsonFieldKeys` fix** — taught to skip unexported, non-anonymous fields so the runtime-only `parsed` field is not mistaken for a permitted manifest key (anonymous embeds still promote their exported subfields, matching `encoding/json`).

## Behavior

No observable behavior change — same allow/deny outcomes, same error codes, same audit `allowedCIDRs` details, identical manifest JSON shape (no grammar/schema change, so no spec-repo update). Purely a hot-path allocation/parse elimination.

## Tests

- `pkg/capability`: table-driven `Compile`/`Networks` coverage — valid compile + membership, uncompiled reports no networks, malformed CIDR errors without clobbering a prior result, idempotent recompile, empty-CIDR set.
- `pkg/enforcement`: `TestIPRange_CompiledPath` proves the compiled hot path allows in-range / denies out-of-range (existing `TestIPRange_ErrorBranches` continues to exercise the uncompiled fallback, including malformed CIDRs).
- `cmd/eunox-mcp`: `TestLoadManifest_PreCompilesIPRangeCIDRs` proves the loader compiles the exact condition the engine evaluates.

`make test` (race), `go vet`, `golangci-lint`, `make check-fmt`, and `make check-license` all pass. CHANGELOG updated under Unreleased → Changed.

https://claude.ai/code/session_01UorVotqumsXxnfDRKwBfcK

---
_Generated by [Claude Code](https://claude.ai/code/session_01UorVotqumsXxnfDRKwBfcK)_


pkg/enforcement/handlers.go
```diff
@@ -210,15 +210,29 @@ func (e *Engine) handleIPRange(_ context.Context, cond capability.Condition, req
...
+			if err != nil {
+				return &ConditionError{
+					Code:          capability.ErrCodeConditionFailed,
+					ConditionType: capability.ConditionTypeIPRange,
+					Message:       fmt.Sprintf("invalid CIDR in condition: %s", cidr),
+				}
 			}
+			networks = append(networks, network)
 		}
+	}
```
eunolabs: Subtle behavior change in the fallback (uncompiled) path. The old handler parsed each CIDR and checked `Contains` in the **same** loop, returning `allow` on the first match — so a malformed CIDR positioned *after* a matching one was never reached. This version parses every CIDR up front, then matches in a separate loop, so a malformed CIDR anywhere now returns `CONDITION_FAILED` (deny) even when an earlier CIDR already matched the source IP.

For manifest-loaded conditions this is unreachable (the loader rejects malformed CIDRs, and they take the compiled path), so the PR's "no observable behavior change" claim holds there. But the fallback exists precisely for programmatically-constructed conditions, and for those this is a real allow→deny flip on e.g. `["10.0.0.0/8", "bad-cidr"]` with an in-range IP.

It's arguably *more* fail-closed, so I'd be fine keeping it — but then it's worth a CHANGELOG line + a test to lock it in, rather than being folded under "behavior unchanged." Alternatively, preserve the old interleaved form so the fallback stays byte-for-byte identical (this also drops the extra slice allocation on that path):

```go
networks, ok := ipr.Networks()
if !ok {
    for _, cidr := range ipr.CIDRs {
        _, network, err := net.ParseCIDR(cidr)
        if err != nil {
            return &ConditionError{
                ConditionType: capability.ConditionTypeIPRange,
                Message:       fmt.Sprintf("invalid CIDR in condition: %s", cidr),
            }
        }
        if network.Contains(ip) {
            return nil
        }
    }
    return notInRangeDeny() // the existing trailing deny
}
for _, network := range networks {
    if network.Contains(ip) {
        return nil
    }
}
```

Not blocking — just flag which way you intend it.

---

cmd/eunox-mcp/manifest.go
```diff
@@ -258,6 +258,15 @@ func validateLocalManifest(m *LocalManifest) error {
 				if err := validateIPRange(i, j, v); err != nil {
 					return err
 				}
+				// Pre-compile the CIDRs into *net.IPNet once, here at load, so the
+				// per-request enforcement hot path never re-parses them (issue #215).
+				// validateIPRange has already confirmed they parse, so this cannot
+				// error for a valid manifest; fail closed if it somehow does. Only the
+				// pointer case is the condition the engine evaluates at runtime — the
+				// value case above is a copy, so compiling it would not persist.
+				if err := v.Compile(); err != nil {
```
eunolabs, line 267: Minor / optional: `validateIPRange` immediately above already calls `net.ParseCIDR` on every CIDR (it needs the parsed `network` for the host-bits check), and `Compile()` then parses them all a second time. It's load-time only so the cost is negligible, and keeping `Compile` self-contained + idempotent for direct callers is a reasonable reason to leave it as-is — just noting the redundant parse in case you'd prefer `validateIPRange` to hand the already-parsed networks straight to the condition and skip the re-parse.

---

eunolabs (commented):
Reviewed the pre-compilation change end to end and it's solid, well-tested work.

**Traced the data path:** `unmarshalCondition` returns `*IPRangeCondition` → loader's pointer-case calls `Compile()` → engine ranges over `matched.Conditions` directly → `asIPRange` → `Networks()`. So the compiled pointer is genuinely the one the hot path evaluates (the value-case in the switch is correctly skipped since compiling a copy wouldn't persist). `validateIPRange` rejects empty / parse-error / host-bits CIDRs *before* `Compile()` runs, so the loader's fail-closed branch is unreachable for a valid manifest, exactly as the comment claims. The `jsonFieldKeys` unexported-field skip is correct per `encoding/json`, and anonymous-embed promotion is preserved.

**Verified locally:** `go build ./...` clean; `go test -race -count=1` passes for `pkg/capability`, `pkg/enforcement`, and `cmd/eunox-mcp`.

**Nice touches:** compiling at load (single-threaded) and only ever *reading* `parsed` at request time sidesteps the data race a lazy in-handler cache would have introduced; `Compile` committing `c.parsed` only on full success keeps a condition from ending up half-compiled on a mid-list error.

Two notes inline — one substantive (a narrow allow→deny flip in the *uncompiled fallback* path that the "no observable behavior change" framing doesn't quite cover) and one minor/optional (a redundant load-time parse). Neither is blocking; the substantive one is mostly "confirm which behavior you intend."

---

pkg/enforcement/handlers.go
eunolabs: Good catch. Went with option (b): restored the interleaved parse-and-match in the fallback so it's byte-for-byte identical to the pre-#215 handler (first containing CIDR wins; a malformed CIDR after a match is never reached), which also drops the extra slice allocation you noted. For a perf-only change I'd rather keep behavior unchanged than fold an allow→deny flip under "more fail-closed." Added a regression test (`TestIPRange_FallbackMatchPrecedesMalformedCIDR`) pinning it, so the "no observable behavior change" claim now holds on the fallback path too. Fixed in e2ca11c.

---

cmd/eunox-mcp/manifest.go
eunolabs: Leaving as-is for the reason you noted: keeping `Compile` self-contained and idempotent for direct callers (and for the programmatic-construction fallback) is worth the negligible, load-time-only double parse. Threading the already-parsed networks out of `validateIPRange` would couple the two and only save work that happens once at startup.

---

PR #245, opened 2026-06-13
fix: scan for newest timestamp in InMemory.Cleanup (#242)

### LINKED ISSUE #242: bug: InMemory.Cleanup staleness check assumes timestamps are chronologically ordered — breaks under non-monotonic WithTimeFunc clock
## Summary

`Cleanup` decides whether an entry is stale by inspecting only its **last** timestamp:

```go
// memory.go:218
if e.timestamps[len(e.timestamps)-1].Before(now.Add(-staleAfter)) {
    delete(m.entries, key)
}
```

The implicit assumption is that `timestamps[last]` is the most recent timestamp in the slice. `IncrementAndGet` appends `now` at the end after pruning, so under a monotonically-increasing clock (production `time.Now()`) this invariant holds: each new timestamp is at least as recent as every previous one.

Under a non-monotonic `WithTimeFunc` clock — routinely used in tests — the invariant can be violated: if the clock jumps backward between two calls, the appended timestamp is older than some earlier entries in the slice, and `timestamps[last]` is no longer the most recent.

## Concrete failure

```go
clk := time.Unix(1_000_000, 0)
counter := callcounter.NewInMemory(callcounter.WithTimeFunc(func() time.Time { return clk }))

// Call 1 at t=1000000
counter.IncrementAndGet(ctx, "k", 3600)

// Jump clock forward, then call again — most recent entry is second
clk = clk.Add(2 * time.Hour)
counter.IncrementAndGet(ctx, "k", 3600)

// Jump clock backward to somewhere between the two calls
clk = clk.Add(-1 * time.Hour)

// Cleanup sees timestamps[last] = t=1000000+2h; "now" = t=1000000+1h
// staleAfter = 2*3600s = 7200s = 2h
// timestamps[last].Before(now - 2h) = (t+2h).Before(t+1h-2h) = (t+2h).Before(t-1h) = false
// Entry is NOT cleaned up — correct in this case.

// But with a different clock pattern:
clk = time.Unix(1_000_000, 0).Add(3 * time.Hour)  // far future first call
counter.IncrementAndGet(ctx, "k2", 60)  // timestamps[0] = t+3h

clk = time.Unix(1_000_000, 0)           // reset to past
counter.IncrementAndGet(ctx, "k2", 60)  // timestamps[1] = t (OLDER than timestamps[0])

// Now timestamps[last] = t (the second call, which is earlier)
// Cleanup uses timestamps[last] as "most recent" → incorrect staleness decision
```

In the second pattern, `Cleanup` uses `timestamps[last] = t` (which looks old) as the most-recent timestamp. If `now - staleAfter < t`, the entry is correctly kept. But if the first call's timestamp (`t+3h`) is what should be driving staleness and it is much newer, `Cleanup` may incorrectly delete the entry because it only looked at the last (older) element.

## Impact

**Production**: zero impact. `time.Now()` in Go uses a monotonic clock; wall-clock jumps do not affect monotonic reads, and `IncrementAndGet` always appends a timestamp ≥ all previous ones.

**Tests**: tests using `WithTimeFunc` with a clock that goes backward (e.g. to simulate window expiry by jumping back, or to re-run a scenario) can hit this. `TestInMemory_Cleanup_PreservesLongWindowQuota` and similar tests jump the clock forward and then call `Cleanup` — these are safe because the jump is always forward. But a test that jumps backward could get a spurious cleanup deletion or spurious retention.

**Test reliability risk**: a future test that sets the clock backward to test "what happens when the last call was in the future and now we're in the past" would get incorrect Cleanup behaviour without an obvious reason.

## Suggested fix

**Option A — sort before taking the last element:**
```go
// In Cleanup, find the actual most-recent timestamp, not just the last:
latest := e.timestamps[0]
for _, ts := range e.timestamps[1:] {
    if ts.After(latest) {
        latest = ts
    }
}
if latest.Before(now.Add(-staleAfter)) {
    delete(m.entries, key)
}
```
This is O(n) per entry but `Cleanup` is already O(n) so the asymptotic cost is unchanged.

**Option B — document the invariant:**
Add a comment to `Cleanup` noting that `timestamps[last]` is the most-recent element under the assumption of a monotonically non-decreasing clock (production `time.Now()`), and that `WithTimeFunc` clocks must be monotonically non-decreasing for correct cleanup behaviour.

Option B is the minimal fix for the test-reliability concern; Option A is more robust.

## Affected code

`pkg/callcounter/memory.go` — `Cleanup` (lines 200–222).

eunolabs:
## Summary

Fixes #242.

`InMemory.Cleanup` decided staleness by inspecting only the **last** timestamp of an entry:

```go
if e.timestamps[len(e.timestamps)-1].Before(now.Add(-staleAfter)) {
    delete(m.entries, key)
}
```

This assumes `timestamps[last]` is the most recent timestamp. `IncrementAndGet` appends in call order, so the invariant holds under a monotonically non-decreasing clock — production `time.Now()`, whose monotonic reading is immune to wall-clock jumps. But a non-monotonic `WithTimeFunc` clock (used in tests, and able to jump backward) can leave an *older* timestamp at the tail, so a last-element check reads a stale tail and can wrongly drop a still-live entry (or retain a dead one).

## Fix

Adopt **Option A** from the issue: scan the slice for the genuine maximum so staleness keys off the truly newest call regardless of clock monotonicity.

- The scan is `O(n)` per entry over the already-`O(n)` total work and runs only on the periodic cleanup tick.
- **Production behaviour is unchanged** — under a monotonic clock the last element already *is* the maximum.
- The empty-slice case is still handled by the existing guard above, so `timestamps[0]` is safe.

The `Cleanup` doc comment now explains why `timestamps[last]` cannot be trusted under a non-monotonic clock.

## Test

Added `TestInMemory_Cleanup_NonMonotonicClock_KeepsLiveEntry` (in `cleanup_internal_test.go`), which reproduces the issue's pattern: a first call lands at `t+3h`, the clock resets and a second call appends `t` (older, at the tail), and `Cleanup` runs at `t+4h`. The newest call (`t+3h`, 1h old) is well inside the 2h margin, so the entry must survive.

Verified the test **fails on the old last-element logic** and **passes with the scan**. Full `pkg/...` suite passes; `gofmt` and `go vet` are clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UcV31Y9PSgsg4hbeNzoowS)_


pkg/callcounter/memory.go
```diff
@@ -289,8 +300,16 @@ func (m *InMemory) Cleanup() {
...
+		// Find the genuinely most-recent timestamp; see the doc comment for why
+		// timestamps[last] cannot be trusted under a non-monotonic clock.
+		latest := e.timestamps[0]
+		for _, ts := range e.timestamps[1:] {
+			if ts.After(latest) {
+				latest = ts
+			}
+		}
+		// If even the most recent timestamp is old enough, the whole entry is stale.
+		if latest.Before(now.Add(-staleAfter)) {
```
eunolabs, line 353: Verified correct. The empty-slice case is handled by the `len(e.timestamps) == 0` branch above (it `continue`s), so `e.timestamps[0]` is always safe here, and the loop over `[1:]` correctly finds the true maximum. This resolves the Cleanup staleness site from #242.

One related note, out of scope for this PR: Cleanup was *one* order-dependent read of the timestamp slice, but not the only one. `IncrementIfBelow`'s `retryAfter` estimate also assumes oldest-first ordering — it takes `valid[cur-limit]` as the oldest in-window call. Under the same backward-jumping `WithTimeFunc` clock this PR guards against, that index is no longer guaranteed to be the oldest, so the returned `retryAfter` could be off. It's only a hint (no correctness or security impact, and production's monotonic clock is unaffected), so it needn't block here — just flagging it as the same class of latent assumption worth tracking.

---

eunolabs (commented):
## Review — fix #242 (`InMemory.Cleanup` under a non-monotonic clock)

Reviewed against the linked issue and the surrounding `callcounter` code. This is a clean, correctly-scoped fix.

**Correctness**
- The scan for the true maximum timestamp is right, and `e.timestamps[0]` is safe because the `len(e.timestamps) == 0` branch above already `continue`s on empty slices.
- The regression test is valid — I traced it both ways. Under the old `timestamps[last]` logic the slice is `[t+3h, t]`, so the tail `t` is judged 4h old at `now=t+4h` and the live entry is wrongly deleted (test fails); the scan finds `t+3h` (1h old, inside the 2h margin) and keeps it (test passes).
- Production behaviour is unchanged, as the PR states: `time.Now()` carries a monotonic reading, so each appended timestamp is `>=` all prior ones and `timestamps[last]` already is the max.

**Non-blocking notes**
1. `IncrementIfBelow`'s `retryAfter` computation shares the same oldest-first ordering assumption (`valid[cur-limit]`) that this PR removes from Cleanup. It's only a retry hint and production is monotonic, so no action is needed here — flagged inline as the same class of latent assumption.
2. Option A (scan) vs Option B (document the invariant) is a judgment call. Given the issue is test-only, A is the more robust choice and the added cost is negligible (O(n) over already-O(n) work, on the periodic cleanup tick). Reasonable.

**CI**: `Test`, `Lint`, `Formatting (gofmt)`, and `License Headers` are green on the head commit; the remaining build-matrix jobs were still in progress at review time with nothing failing.

LGTM. Leaving this as a comment rather than a formal approval.

---

pkg/callcounter/memory.go
eunolabs: Good catch on the sibling site. Addressed in 0a36a18 by documenting the shared invariant rather than fixing with a sort.

Reasoning: `IncrementIfBelow`'s `retryAfter` runs on the rate-limited **deny hot path** (per denied call), whereas `Cleanup` runs on a cold periodic tick where the O(n) scan is free. `retryAfter` is purely advisory — the caller clamps it and falls back to the full window — and carries no correctness or security weight, so selecting the true `(cur-limit)`-th oldest would tax production for a clock artifact that production's monotonic `time.Now()` never produces. The comment now spells out that the oldest-first ordering holds only under a monotonically non-decreasing clock, why a backward-jumping `WithTimeFunc` clock makes the hint imperfect, and why the scan that #242 added to `Cleanup` is deliberately *not* mirrored here.

Net: #242's actual staleness bug is fixed with the scan; this sibling is now a documented, intentional assumption instead of a silent one.

---

eunolabs: Confirmed — re-reviewed `0a36a18`. It's docs-only (no behavioral change), and the added comment accurately captures the invariant: `valid` stays sorted ascending only under a monotonically non-decreasing clock, production's `time.Now()` never regresses, and the imperfect-hint case is test-only. Agree with documenting the intentional assumption rather than sorting the deny hot path. Resolution looks good.

---

PR #246, opened 2026-06-13
fix: normalize tool name in maxCalls counter key (#240)

### LINKED ISSUE #240: correctness: handleMaxCalls key uses raw req.ToolName without prefix normalization, inconsistent with recordSessionCall
## Summary

`handleMaxCalls` builds its counter key using the raw `req.ToolName`, while `recordSessionCall` strips the engine prefix before building the session-history key. If `req.ToolName` ever carries a `"tool:"` prefix (which `recordSessionCall`'s defensive use of `stripEnginePrefix` implies is possible), the `maxCalls` quota bucket and the `sequenceBlock` history bucket track different identifiers for the same tool.

## Code

```go
// pkg/enforcement/handlers.go:277 — maxCalls key, NO stripping:

// pkg/enforcement/engine.go:199 — sequenceBlock history key, STRIPS prefix:
key := sequenceHistoryKey(req.SessionID, tool)  // "seq:<sessionID>:<bare_tool>"
```

## Concrete inconsistency

If `req.ToolName = "tool:read_file"`:

| Use | Key |
|-----|-----|
| maxCalls counter | `maxcalls:<SESSION>:tool:read_file` |
| sequenceBlock history | `seq:<SESSION>:read_file` |

After 5 allowed calls with `req.ToolName = "tool:read_file"`, switching to bare `"read_file"` would address a **different** `maxCalls` bucket (`maxcalls:<SESSION>:read_file`) — effectively resetting the quota.

Whether this is currently exploitable depends on whether the transport layer always normalizes `req.ToolName` to a bare name. The fact that `recordSessionCall` defensively strips the prefix suggests the engine does not rely on the transport enforcing this. `handleMaxCalls` should be consistent.

## Fix

Apply `stripEnginePrefix` to `req.ToolName` before building the `maxCalls` key:

```go
key := fmt.Sprintf("maxcalls:%s:%s", req.SessionID, toolName)
```

This matches the normalization in `recordSessionCall` and makes both quota and history keys consistent.

## Affected code

- `pkg/enforcement/handlers.go:277` (`handleMaxCalls`)
- `pkg/enforcement/engine.go:199` (`recordSessionCall`)

eunolabs:
## Summary

Fixes #240.

`handleMaxCalls` built its rate-limit counter key from the **raw** `req.ToolName`, while `recordSessionCall` strips the engine prefix (`stripEnginePrefix`) before building the `sequenceBlock` history key. Because the constraint matcher (`findMatchingCapability`) strips the `tool:` prefix from **both** sides, the same constraint matches whether a request arrives as `tool:read_file` or the bare `read_file` — so the two forms addressed **different** quota buckets:

| Use | Key (before fix) |
|-----|------------------|
| maxCalls counter (`tool:read_file`) | `maxcalls:<sess>:tool:read_file` |
| maxCalls counter (`read_file`)      | `maxcalls:<sess>:read_file` |
| sequenceBlock history               | `seq:<sess>:read_file` |

A caller could therefore **reset its own `maxCalls` quota** simply by toggling the namespace prefix on the tool name, and the quota bucket disagreed with the `sequenceBlock` history bucket for the same tool.

## Fix

Apply `stripEnginePrefix` to `req.ToolName` before building the `maxCalls` key (`pkg/enforcement/handlers.go`), so the quota bucket and the `sequenceBlock` history key both key on the same bare identifier — matching the normalization already done in `recordSessionCall`.

## Tests

- Added `TestEngine_MaxCalls_PrefixNormalizedKey`: fills a `maxCalls: 2` window using the prefixed form (`tool:read_file`), then asserts the bare form (`read_file`) hits the **same** exhausted bucket and is denied with `RATE_LIMITED`. Verified this test **fails** before the fix (the bare form was wrongly admitted) and **passes** after.
- Full suite green: `go test -race -count=1 ./...`, plus `make check-fmt`, `make check-license`, and `go vet` all clean.

## Docs

Documented the tool-name normalization in `docs/threat-model-mcp.md` alongside the existing `maxcalls:<sessionID>:<toolName>` key description, noting that prefixed and bare forms now share one quota bucket.

https://claude.ai/code/session_01FcmRnGJphT45AmMKyhCQQE

---
_Generated by [Claude Code](https://claude.ai/code/session_01FcmRnGJphT45AmMKyhCQQE)_


pkg/enforcement/handlers.go
```diff
@@ -305,7 +305,17 @@ func (e *Engine) handleMaxCalls(ctx context.Context, cond capability.Condition,
...
 	// the HTTP transport rejects requests whose session belongs to a different
 	// route, making cross-route key collision architecturally impossible.
-	key := fmt.Sprintf("maxcalls:%s:%s", req.SessionID, req.ToolName)
+	//
+	// Normalize the tool name with stripEnginePrefix so the quota bucket keys on
+	// the same bare identifier that recordSessionCall's sequenceBlock history uses
+	// (engine.go). The constraint matcher already strips the "tool:" prefix from
+	// both sides, so the same constraint matches whether req.ToolName arrives as
+	// "tool:read_file" or "read_file"; without normalizing here those two forms
+	// would address different buckets ("maxcalls::tool:read_file" vs
```
eunolabs: Correct fix. This aligns the `maxCalls` bucket with the prefix-stripped key that `recordSessionCall` already uses for `seq:` history (`engine.go:223`) and with `findMatchingCapability` stripping the prefix on both sides (`engine.go:377`). I confirmed locally that the regression test fails without this line (bare form wrongly `allow`ed) and passes with it.

Minor consistency nit (optional, non-blocking): the two other call sites that strip the prefix off `req.ToolName` both guard against an empty result — `recordSessionCall` (`engine.go:224`, `if tool == "" { return }`) and `handleSequenceBlock` (`handlers.go:805`, `if priorTool == "" { continue }`). Here an input like `tool:` strips to `""`, producing the key `maxcalls:<session>:`. It's benign (it fails toward a single, more-restrictive shared bucket, not a bypass), but adding the same empty-name guard would keep the three sites consistent.

---

docs/threat-model-mcp.md
```diff
@@ -273,7 +273,7 @@ The audit key is generated once at first startup using `crypto/rand` (Go's CSPRN
...
+Counter keys include the session ID (`maxcalls:<sessionID>:<toolName>`). The `<toolName>` component is normalized with the engine's prefix stripper, so a tool addressed as `tool:read_file` and one addressed as the bare `read_file` share a single quota bucket — a client cannot reset its own quota by toggling the namespace prefix, and the quota bucket stays consistent with the `sequenceBlock` history key, which strips the same prefix. This means Redis addresses the *cross-instance* bypass (the same session's quota is visible to all proxy replicas) but does **not** prevent *session-rotation* bypass: a client that opens a fresh MCP session gets a new `Mcp-Session-Id` and therefore a fresh counter, on both backends equally. The session-rotation bypass is a documented residual risk (see `adversarial_test.go` Adv-6).
```
eunolabs: Doc precision (low priority). The note frames this as the prefixed vs. bare forms of the *same* tool sharing a bucket. Since the key is now keyed solely on the prefix-stripped name (`maxcalls:<sessionID>:<bareName>`), the bucket is shared across *any* capabilities that strip to the same bare identifier — e.g. a `tool:foo` constraint and a `prompt:foo` constraint would draw from one quota bucket.

This isn't new behavior introduced by the fix: in the real proxy path, `pdp.go` already sets `ToolName` to the bare `target.Name`/`promptName`/uri for all three method types, so same-named capabilities already collided pre-fix. But since the threat-model wording is being (re)written here, it'd be worth stating the keying precisely — the bucket is per-`(sessionID, bareName)`, not per-tool.

---

eunolabs (commented):
## Review: fix #246 — normalize tool name in `maxCalls` counter key

**Verdict: LGTM.** The diagnosis in the description is accurate and the fix is minimal and well-targeted.

What I checked:

- **Root cause is real.** `handleMaxCalls` built the rate-limit key from raw `req.ToolName`, while `findMatchingCapability` (`engine.go:377`) and `recordSessionCall`'s `seq:` history (`engine.go:223`) both strip the engine prefix. So a single constraint matched both `tool:read_file` and `read_file`, yet the two forms consumed different `maxcalls:` buckets — a caller could refill its own quota by toggling the prefix. Routing the key through the same `stripEnginePrefix` restores consistency between the quota bucket, the matcher, and the `sequenceBlock` history.
- **Regression test is genuine.** I ran `TestEngine_MaxCalls_PrefixNormalizedKey` on the branch (passes) and with the one-line fix reverted (fails — bare form wrongly `allow`ed, exactly the described bug). Good guard against regression.
- **Fix is complete.** `maxcalls:` was the only counter key still built from the raw tool name; the `seq:` history already normalized. No other call sites need the same change.
- **Invariants honored.** Fail-closed paths untouched; no new audit field; threat-model doc updated in the same PR per CONTRIBUTING.
- **CI:** all 42 checks green.

Two non-blocking nits left inline (an empty-name guard for consistency with the other two strip sites, and a doc-precision note on what the bucket is keyed on). Neither blocks merge.

---

pkg/enforcement/handlers.go
eunolabs: Added the guard in f69381d. One deliberate difference from the two sibling sites: `recordSessionCall` and `handleSequenceBlock` *skip* on an empty name (a side-effect and a loop-`continue` respectively), but `handleMaxCalls` is a gate — returning early without counting would be an allow, i.e. a quota bypass. So it instead **denies** with `MISSING_CONTEXT`, mirroring the empty-`sessionID` guard right above it (fail-closed, structured, logged). Covered by `TestEngine_MaxCalls_EmptyToolNameDenies`.

---

PR #248, opened 2026-06-13
fix: make InMemory.StartCleanup idempotent (#235)

### LINKED ISSUE #235: bug: InMemory.StartCleanup is not idempotent — multiple calls start redundant goroutines that cannot be stopped independently
## Summary

`InMemory.StartCleanup` unconditionally starts a new background goroutine on each call. There is no guard against multiple invocations:

```go
// memory.go:171-187
func (m *InMemory) StartCleanup(ctx context.Context, interval time.Duration) {
    // No guard — every call starts another goroutine.
    go func() {
        t := time.NewTicker(interval)
        defer t.Stop()
        for {
            select {
            case <-t.C:
                m.Cleanup()
            case <-ctx.Done():
                return
            }
        }
    }()
}
```

### Double-start in the same component

If `StartCleanup` is called twice with the same context (e.g. a constructor that also calls `StartCleanup`, combined with a caller that does so too), two goroutines both call `Cleanup()` on every tick. The mutex inside `Cleanup` prevents data races, but the CPU cost is doubled and the goroutine count leaks upward with every call.

### Stopped-then-restarted pattern

If `ctx` is cancelled (intent: "stop cleanup") and `StartCleanup` is called again on a new context to restart it, the original goroutine has already exited, which is correct. But if the caller cancels `ctx` and then starts a second goroutine **before** the first goroutine observes the cancellation (a scheduling window), two goroutines are live simultaneously for a brief period. In the degenerate case where `StartCleanup` is called in a retry loop, goroutines accumulate.

### No way to detect misuse

Because `StartCleanup` returns nothing (no `io.Closer`, no cancel function, no goroutine handle), callers have no way to detect that they have called it twice.

## Suggested fix

Guard the startup with a `sync.Once` or an `atomic.Bool`:

```go
type InMemory struct {
    mu         sync.Mutex
    entries    map[string]*entry
    now        func() time.Time
    cleanupOnce sync.Once
}

func (m *InMemory) StartCleanup(ctx context.Context, interval time.Duration) {
    m.cleanupOnce.Do(func() {
        if interval <= 0 {
            interval = DefaultCleanupInterval
        }
        go func() {
            t := time.NewTicker(interval)
            defer t.Stop()
            for {
                select {
                case <-t.C:
                    m.Cleanup()
                case <-ctx.Done():
                    return
                }
            }
        }()
    })
}
```

This makes `StartCleanup` safe to call multiple times from different code paths (e.g. both a library constructor and a caller's init) without leaking goroutines. The trade-off is that the interval of the first call wins; if callers need to reconfigure, the pattern needs to change (e.g. return a `context.CancelFunc` to stop the existing goroutine before starting a new one).

Alternatively, return an error or `bool` indicating whether cleanup was already started, letting callers detect misuse.

## Affected code

`pkg/callcounter/memory.go` — `StartCleanup` (lines 170–187).

eunolabs:
## Summary

Fixes #235. `InMemory.StartCleanup` previously started a new background goroutine on **every** call, with no guard and no handle to stop an extra one. Two code paths both starting cleanup — or a cancel-then-restart in a retry loop — left redundant goroutines, each calling `Cleanup` on every tick: doubled per-tick CPU cost and a goroutine count that leaked upward with every call.

## Fix

- Add a `sync.Once` (`cleanupOnce`) to `InMemory` and launch the goroutine inside `cleanupOnce.Do(...)`, so at most one cleanup goroutine runs per counter no matter how many code paths call it. The first call wins — its `ctx` and `interval` govern the goroutine for the counter's lifetime.
- `StartCleanup` now returns a `bool`: `true` when this call started the goroutine, `false` when cleanup was already running — so callers can detect a double-start (the issue's third failure mode, "no way to detect misuse").
- Expanded the doc comment to spell out the idempotency guarantee and the "first interval wins" trade-off (a counter whose cleanup must outlive a canceled context needs a fresh `InMemory`).

This follows the issue's suggested fix (`sync.Once`) and folds in its alternative (return a value indicating whether cleanup was already started).

## Tests

Added to `pkg/callcounter/callcounter_test.go`:
- `TestInMemory_StartCleanup_Idempotent` — first call returns `true`, the next five return `false`. Because the goroutine is launched only inside the `sync.Once` block, a `false` return proves no second goroutine was started.
- `TestInMemory_StartCleanup_FirstIntervalWins` — a later call with an hour-long interval does not override the first call's fast interval (stale entry is still evicted on a fake clock).

The existing `StartCleanup` tests construct a fresh counter each time, so they are unaffected and still pass.

## Notes

- The only production caller (`cmd/eunox-mcp/main.go`) starts cleanup exactly once and ignores the new return — no change needed there; ignoring a `bool` return as a statement compiles cleanly.
- `StartCleanup` is not part of the `Store` interface (it is `InMemory`-specific, reached via type assertion), so the signature change breaks no interface.
- `CHANGELOG.md` updated under Unreleased → Fixed.

## Verification

- `go build ./...`
- `make check-fmt`, `make check-license`, `make lint` — clean

https://claude.ai/code/session_01A8sLLrWQzkqrDck6Pufjib

---
_Generated by [Claude Code](https://claude.ai/code/session_01A8sLLrWQzkqrDck6Pufjib)_


pkg/callcounter/callcounter_test.go
```diff
@@ -266,6 +266,72 @@ func TestInMemory_StartCleanup_StopsOnContextCancel(t *testing.T) {
...
+
+	// Age the entry past the eviction threshold and let the fast (first) interval
+	// fire. If the second call's hour-long interval had taken effect, no tick
+	// would have run and the stale entry would survive.
+	advanceClock(25 * time.Hour)
+	time.Sleep(100 * time.Millisecond)
+
+	count, err = counter.IncrementAndGet(context.Background(), "k", 60)
+	require.NoError(t, err)
+	assert.Equal(t, int64(1), count, "the first call's fast interval must still govern cleanup")
```
eunolabs: This test doesn't actually exercise "first interval wins" — the final assertion passes whether or not cleanup ever runs, so it gives false confidence about the contract it names.

`IncrementAndGet` prunes a key's *own* expired timestamps before counting (`memory.go`: `valid := e.timestamps[:0]` then the `ts.After(cutoff)` loop). So after `advanceClock(25 * time.Hour)`, the second `IncrementAndGet("k", 60)` returns `1` **regardless** of whether `Cleanup` evicted the entry: the stale 12:00 timestamp is outside the 60s window and gets dropped by the increment itself, then the new `now` is appended → count 1.

Concretely, this test would still pass if the hour-long second interval had won, or even if no cleanup goroutine ran at all. The comment "If the second call's hour-long interval had taken effect ... the stale entry would survive" is true of the internal map, but a surviving entry is invisible at the `count` level because `IncrementAndGet` prunes it anyway.

The only observable effect unique to `Cleanup` is the entry being deleted from the internal `entries` map, and that's not reachable from the external `callcounter_test` package. To make this load-bearing, put it in the internal `_test.go` (package `callcounter`, like `cleanup_internal_test.go`) and assert on map state after the sleep — e.g. `len(counter.entries) == 0` (the fast first interval fired) vs `== 1` (the hour interval suppressed it). That assertion distinguishes the two intervals; the current one cannot.

(Note: `TestInMemory_StartCleanup_RemovesStaleEntries` above has the same blind spot, so it isn't introduced here. But that test only claims "cleanup runs at all," whereas this one's whole purpose is to tell the two intervals apart — which is exactly what the post-prune count can't see.)

---

eunolabs (commented):
Reviewed against issue #235. The core fix is correct and follows the issue's suggested approach.

**Solid:**
- The `sync.Once` guard launches exactly one cleanup goroutine per counter. `started` is set only inside `Do`, so the first caller gets `true` and all others `false` — correct even under concurrent first-calls, since `Once.Do` serializes them.
- `StartCleanup` is not part of the `Store` interface (`callcounter.go`) or the `CallCounter` interface (`pkg/capability/enforce.go`), so widening the return to `bool` breaks no interface. The sole production caller (`cmd/eunox-mcp/main.go:303`) and the two existing `StartCleanup` tests use it as a bare statement and still compile.
- `TestInMemory_StartCleanup_Idempotent` is a genuine regression test: because the goroutine is launched only inside `Once.Do`, the `false` returns really do prove no second goroutine started.
- Doc comment and CHANGELOG are accurate, and the "first call wins / needs a fresh `InMemory` to restart after cancel" trade-off is exactly the one issue #235 accepted.

**One thing to address (non-blocking for the production code):**
- `TestInMemory_StartCleanup_FirstIntervalWins` doesn't actually test what its name claims — its final assertion holds whether or not cleanup runs, because `IncrementAndGet` self-prunes expired timestamps. Details + a suggested fix in the inline comment.

CI is green on the merge-relevant jobs (Test, Lint, gofmt, license headers, builds); the remaining checks are still in progress. Nice, thorough issue write-up and PR description.

---

pkg/callcounter/callcounter_test.go
eunolabs: Good catch — fixed in 10f6a4f. Moved `TestInMemory_StartCleanup_FirstIntervalWins` into the internal package (`cleanup_internal_test.go`) and replaced the post-prune count assertion with `len(m.entries)` after the sleep: 0 means the fast first interval fired and evicted the aged entry, 1 means the hour-long second interval wrongly suppressed cleanup. That distinguishes the two intervals, which the external count could not. Left a note in `callcounter_test.go` explaining why the contract can't be tested from the external package.

---

eunolabs: Verified — this resolves the finding. The internal test now asserts `len(m.entries)` under `m.mu`, which does distinguish the two intervals: at `windowSec=60` the 25h advance puts the entry past the `2x` staleness threshold, so the 10ms ticker evicts it within the 100ms window (→ 0), whereas the hour-long interval would leave it in the map (→ 1). Production code is unchanged. LGTM.

---

PR #249, opened 2026-06-13
fix: strip namespace prefix from handleMaxCalls counter key (#223)

### LINKED ISSUE #223: bug: handleMaxCalls counter key does not strip namespace prefix from req.ToolName — inconsistent with sequenceBlock and recordSessionCall
## Summary

`handleMaxCalls` constructs its counter key directly from `req.ToolName` without stripping the namespace prefix. Every other site that builds a session-scoped key (`recordSessionCall`, `handleSequenceBlock`) consistently calls `stripEnginePrefix(req.ToolName)` first. The inconsistency means that a caller who passes a prefixed tool name (e.g. `"tool:read_file"`) gets a **different counter bucket** than the sequence-block history, and two callers that use the same logical tool with different `ToolName` forms (bare vs. prefixed) get **different quota buckets**, allowing the per-session call limit to be silently exceeded.

## Code evidence

```go
// handlers.go:277 — maxCalls key uses raw ToolName (NO prefix stripping)

// engine.go:199 — recordSessionCall strips the prefix
_, _ = e.counter.IncrementAndGet(ctx, sequenceHistoryKey(req.SessionID, tool), ...)

// handlers.go:759 — handleSequenceBlock also strips
```

If `req.ToolName = "tool:read_file"`:
- `maxCalls` key: `"maxcalls:<sess>:tool:read_file"`
- `sequenceBlock` history key: `"seq:<sess>:read_file"`

They address different keys in the counter backend, so a `sequenceBlock` condition on `"read_file"` and a `maxCalls` condition on the same tool operate on inconsistent state.

## Production impact

`ManifestPDP.Decide` currently passes `target.Name` (already bare, from `ParseTarget`) as `req.ToolName`, so the production code path is unaffected today. However:

- Any caller that embeds `enforcement.Engine` directly and passes a prefixed `ToolName` via `ValidateAction` or `EvaluateConditions` silently gets split quota buckets.
- The inconsistency is a latent bug waiting for a future refactor that changes how `ToolName` is populated.

## Suggested fix

Apply `stripEnginePrefix` in `handleMaxCalls` when building the counter key:

```go
// handlers.go — handleMaxCalls
toolKey := stripEnginePrefix(req.ToolName)
key := fmt.Sprintf("maxcalls:%s:%s", req.SessionID, toolKey)
```

This mirrors the existing pattern in `recordSessionCall` and `handleSequenceBlock` and makes all session-keyed counter operations use the same canonical (bare) tool name.

eunolabs:
## Summary

Fixes #223.

`handleMaxCalls` built its counter key directly from `req.ToolName` without stripping the engine namespace prefix, unlike every other site that builds a session-scoped key — `recordSessionCall` (`engine.go`) and `handleSequenceBlock` (`handlers.go`) both canonicalize via `stripEnginePrefix` first.

The inconsistency meant a caller passing a prefixed tool name (e.g. `"tool:read_file"`) landed in a **different quota bucket** (`maxcalls:<sess>:tool:read_file`) than its bare form (`maxcalls:<sess>:read_file`), silently splitting the per-session call limit, and diverging from the session history (`seq:<sess>:read_file`) that a `sequenceBlock` condition inspects for the same logical tool.

## Change

```go
// handlers.go — handleMaxCalls
toolKey := stripEnginePrefix(req.ToolName)
key := fmt.Sprintf("maxcalls:%s:%s", req.SessionID, toolKey)
```

This mirrors the existing pattern in `recordSessionCall` and `handleSequenceBlock` so every session-keyed counter operation uses the same canonical (bare) tool name.

## Production impact

The shipped `ManifestPDP.Decide` path passes `target.Name` (already bare, from `ParseTarget`) as `req.ToolName`, so production behavior is unchanged. This closes the latent split-bucket bug for callers that embed `enforcement.Engine` directly and pass a prefixed `ToolName` via `ValidateAction`/`EvaluateConditions`.

## Testing

- Added `TestEngine_MaxCalls_PrefixedAndBareShareBucket`: fills the window with two bare-name calls, then asserts a third call using the prefixed name hits the **same** bucket and is denied (`RATE_LIMITED`). Verified the test fails on the pre-fix code (the prefixed call was allowed) and passes with the fix.
- `go test -race -count=1 ./pkg/enforcement/` passes; `gofmt` clean; `go vet` clean; `go build ./...` clean.

https://claude.ai/code/session_01Uz8WJQYNprG4KvT6UCirUq

---
_Generated by [Claude Code](https://claude.ai/code/session_01Uz8WJQYNprG4KvT6UCirUq)_


pkg/enforcement/handlers.go
```diff
@@ -301,11 +301,18 @@ func (e *Engine) handleMaxCalls(ctx context.Context, cond capability.Condition,
...
+	// quota bucket is keyed on the canonical (bare) tool name, matching
+	// recordSessionCall and handleSequenceBlock. Without this, a prefixed
+	// ToolName ("tool:read_file") and its bare form ("read_file") would land in
+	// different buckets, silently splitting the per-session call limit and
+	// addressing different state than a sequenceBlock condition on the same tool.
+	// Route scoping is not needed because session IDs are server-assigned UUIDs
+	// (one per connection) and the HTTP transport rejects requests whose session
+	// belongs to a different route, making cross-route key collision
+	// architecturally impossible.
+	toolKey := stripEnginePrefix(req.ToolName)
```
eunolabs, line 330: Optional nit (non-blocking): the two sibling sites you're aligning with both guard the empty result after stripping — `recordSessionCall` returns when `tool == ""` (engine.go:224) and `handleSequenceBlock` skips `priorTool == ""` (handlers.go:805). Here there's no such guard, so a prefix-only `ToolName` like `"tool:"` now strips to `""` and collapses into the same `maxcalls:<sess>:` bucket as an empty tool name.

The blast radius is essentially nil — a matched constraint always carries a real target name, and merging buckets only ever *tightens* quota (so it fails closed, not open) — so this is fine to leave as-is. Just flagging the asymmetry with the call sites the comment cites, in case you'd prefer exact parity.

---

eunolabs (commented):
## Review: `fix: strip namespace prefix from handleMaxCalls counter key (#223)`

### Overview
Fixes the latent split-quota bug from #223. `handleMaxCalls` was building its rate-limit key straight off `req.ToolName`, so a prefixed name (`tool:read_file`) and its bare form (`read_file`) landed in different `maxcalls:<sess>:<tool>` buckets — silently doubling the effective per-session limit and diverging from the `seq:` history a `sequenceBlock` condition inspects for the same logical tool. The fix routes the name through `stripEnginePrefix` before building the key.

### Correctness — looks right ✅
- `stripEnginePrefix` (engine.go:403) only strips the known `tool:`/`resource:`/`prompt:`/`system:` prefixes, so arbitrary tool names containing a `:` are left untouched — no risk of over-stripping.
- The change makes `handleMaxCalls` consistent with the other two session-keyed sites: `recordSessionCall` (engine.go:223) and `handleSequenceBlock` (handlers.go:802) already canonicalize via `stripEnginePrefix`. The `maxcalls:` and `seq:` keyspaces now agree on the bare name for the same logical tool.
- The "production unchanged" claim holds: the shipped `ManifestPDP.Decide` path passes the already-bare `target.Name`, so stripping is a no-op there; this only fixes direct `enforcement.Engine` embedders that pass a prefixed `ToolName`.

### Conventions ✅
- Conventional Commit `fix:` with issue reference, DCO, no emoji.
- Preserves the fail-closed posture (sessionID and atomic-limiter guards untouched).
- Comment is updated to explain the stripping and retains the original route-scoping rationale.

### Tests ✅
- `TestEngine_MaxCalls_PrefixedAndBareShareBucket` is a precise regression test: it fills the window with two bare calls, then asserts the prefixed call hits the *same* bucket and is denied with `RATE_LIMITED` / `ConditionTypeMaxCalls`. The PR notes it was confirmed to fail on the pre-fix code, which is exactly the right validation.

### Security
- No new attack surface. Merging the two namespaces for a shared bare name only ever tightens a quota, never relaxes it. This namespace-merge is already the established engine behavior for `seq:` history, so the fix brings `maxcalls:` in line rather than introducing a new semantic.

### CI
- Green so far — Test, Lint, gofmt, and License Headers all passed; remaining build/demo jobs queued or running with no failures at review time.

### Verdict
Clean, well-scoped, well-tested bug fix that improves cross-handler consistency. One optional non-blocking nit inline re: the empty-after-strip guard the sibling sites have. Nothing blocking from my side.

---

pkg/enforcement/handlers.go
eunolabs: Good catch — addressed in 2fc7141. `handleMaxCalls` now guards the empty-after-strip case. Since this is an enforcement gate (unlike the side-effect-only `recordSessionCall`/`handleSequenceBlock`), it fails **closed** — denies with `ErrCodeMissingContext` — rather than skipping, so the degenerate `"tool:"` input can never share a bucket. A matched constraint always carries a real target, so production behavior is unchanged. Added `TestEngine_MaxCalls_EmptyToolNameDenies` to lock it in.

---

eunolabs: Confirmed — re-reviewed 2fc7141 and this fully resolves the nit. Failing closed with `ErrCodeMissingContext` (matching the empty-`sessionID` guard) is the right call here, and distinguishing the enforcement gate from the side-effect-only sibling sites is a nice touch. `TestEngine_MaxCalls_EmptyToolNameDenies` uses a `*` target so the constraint actually reaches the handler — correct. No further comments from me; LGTM.

---

eunolabs:
Closing as superseded by #246 (`93e5beb`, "fix: normalize tool name in maxCalls counter key (#240)").

While this PR was open, #246 landed on `main` and combined both maxCalls key fixes — the target-type namespacing from #209/#205 and the `stripEnginePrefix` name canonicalization that this PR (for #223) introduced — into:

```go
key := fmt.Sprintf("maxcalls:%s:%s:%s", req.SessionID, targetType, stripEnginePrefix(req.ToolName))
```

plus the same fail-closed empty-name guard (`ErrCodeMissingContext`). The result on `main` is behaviorally identical to this PR. Test coverage is also already present on `main`: `TestEngine_MaxCalls_PrefixNormalizedKey` covers the prefixed-vs-bare shared-bucket case (equivalent to this PR's `TestEngine_MaxCalls_PrefixedAndBareShareBucket`), alongside `EmptyToolNameDenies`, `TargetTypeNamespacesQuota`, and `NilTargetUsesEmptyTypeSegment`.

Merging this PR now would produce no behavioral change — only comment rewording, a variable rename (`toolKey`/`toolName`), and a duplicate test — so closing rather than merging. The underlying issue #223 is resolved on `main`.

---

PR #250, opened 2026-06-13
fix: guard auditSink.Close against concurrent Record (send on closed channel)

### LINKED ISSUE #231: bug: `auditSink.Close()` panics if called concurrently with an in-flight `Record()` — send on closed channel
## Summary

`Close()` closes the drainer channel (`close(s.records)`) via `closeOnce`, and `record()` sends to it with a non-blocking select:

```go
// audit.go:344-351 — record()
select {
case s.records <- rec:    // panics if s.records is already closed
default:
    s.dropped.Add(1)
}
```

In Go, **sending to a closed channel always panics**, even when the send is the selected arm of a `select` statement. If a goroutine is executing the `case s.records <- rec` arm at the same moment that another goroutine executes `close(s.records)` in `Close()`, the runtime panics with "send on closed channel".

## When this can happen

The audit sink's lifecycle relies on an **implicit, external ordering guarantee**: all request-handling goroutines must finish calling `Record()` before `Close()` is called. This is not enforced by the `auditSink` itself — there is no shutdown semaphore, `sync.WaitGroup` on the producer side, or other coordination primitive.

In the HTTP transport, `Close()` is called after the HTTP server's `Shutdown()` returns. Go's `http.Server.Shutdown` waits for active connections to drain, but a goroutine that has already completed its HTTP handler and returned the response can still be executing `Record()` after `Shutdown()` reports done — the audit call happens before the handler returns, but the scheduling window between the channel send and `Close()` closing the channel is narrow and real.

In the stdio transport, the same race applies during graceful shutdown if the proxy receives a SIGTERM while a request is being processed.

## Severity

A panic in the drainer or a send-to-closed-channel terminates the process immediately (no graceful shutdown, no partial flush, chain head not advanced). Because the panic happens in a goroutine that is not the drainer, the drainer's `wg.Wait()` never completes, and `Close()`'s `sync.Once` body is stuck — on a multi-goroutine panic the behaviour is undefined.

## Suggested fix

Guard the channel send against a closed channel. The idiomatic approach is a dedicated shutdown flag:

```go
type auditSink struct {
    ...
    closed atomic.Bool
}

func (s *auditSink) record(...) {
    if s.closed.Load() {
        s.dropped.Add(1)
        return
    }
    select {
    case s.records <- rec:
    default:
        s.dropped.Add(1)
    }
}

func (s *auditSink) Close() error {
    var closeErr error
    s.closeOnce.Do(func() {
        s.closed.Store(true)
        // Small window between closed.Store and close(s.records) is safe:
        // any record that slips through after Store(true) is either
        // accepted (channel not yet closed) or hits the closed-channel
        // panic. Use a recover in drain() to handle the latter, or
        // use a mutex around the send + close.
        close(s.records)
        ...
    })
    return closeErr
}
```

A more robust fix pairs the `closed` flag with a `sync.Mutex` around the send and the close so the two cannot interleave:

```go
// producer side
s.mu.Lock()
if !s.closed.Load() {
    select {
    case s.records <- rec:
    default:
        s.dropped.Add(1)
    }
}
s.mu.Unlock()

// Close()
s.mu.Lock()
s.closed.Store(true)
close(s.records)
s.mu.Unlock()
```

The mutex is only held for the duration of the non-blocking send (no I/O, no allocation after the struct is built), so contention is negligible.

## Affected code

`cmd/eunox-mcp/audit.go`, `record()` (lines 344–351) and `Close()` (lines 511–528).

eunolabs:
## Summary

Fixes #231 — `auditSink.Close()` could panic the process with **"send on closed channel"** when it ran concurrently with an in-flight `Record()`.

`Close()` closes the drainer's record channel (`close(s.records)`) to flush and stop the drainer, while `record()` enqueues with a non-blocking `select` send. In Go a send on a closed channel **always panics, even from the selected arm of a `select`** — so a goroutine executing the `case s.records <- rec` arm at the instant `Close()` closed the channel crashed the process. Because the panic fires on a request-handling goroutine rather than the drainer, `Close()`'s `wg.Wait()` never completes and shutdown wedges.

The window is reachable on graceful shutdown:
- **HTTP transport** — a handler goroutine can still be inside `Record()` after `http.Server.Shutdown` returns (the audit call happens before the handler returns, but the scheduling gap to `Close()` is real).
- **stdio transport** — a SIGTERM can land mid-request during graceful shutdown.

## Fix

Make the send and the close mutually exclusive with an `RWMutex` plus a `closed` flag:

- `record()` takes the **read** lock for the span of the non-blocking send. Concurrent producers still enqueue in parallel (channel sends are themselves concurrency-safe); the lock only serializes them against shutdown.
- `Close()` takes the **write** lock while it sets `closed = true` and closes the channel. The write lock cannot be acquired until every in-flight send has released its read lock, so a send can never be mid-flight when `close(s.records)` runs. The lock is released before `wg.Wait()` so post-close `Record()` calls aren't blocked behind the drainer's final flush.
- Any `Record()` arriving after `closed` is set is counted as a drop (surfaced by `DroppedRecords()` and `Close()`'s shutdown warning) instead of sending on a closed channel.

Idempotency of `Close()` (`closeOnce`) is preserved.

## Tests

`cmd/eunox-mcp/audit_close_race_test.go`:
- `TestAuditSink_CloseRace_NoSendOnClosedChannel` — 64 producers hammer `Record()` (500 iterations each), released together, while `Close()` runs concurrently. Must complete without panicking; also verifies a second `Close()` stays a safe no-op.
- `TestAuditSink_RecordAfterClose_DropsWithoutPanic` — deterministic post-close path: `Record()` after `Close()` counts a drop and returns instead of panicking.

Both run green under `go test -race` (CI default), and the full `go test -race -count=1 ./...` suite passes.

Closes #231.

https://claude.ai/code/session_01Q3bxXV2NhYFnM8J1KRpwbF

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q3bxXV2NhYFnM8J1KRpwbF)_


cmd/eunox-mcp/audit.go
```diff
@@ -341,7 +354,21 @@ func (s *auditSink) record(ctx context.Context, upstream, policyVersion, policyS
...
+	// send is skipped — the record is counted as dropped rather than sent on a
+	// closed channel (which would panic).
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	if s.closed {
+		// Sink is shutting down; the drainer has been signalled to exit and may
+		// already be gone, so there is nothing to enqueue onto. Count the loss
+		// (surfaced by Close()'s shutdown warning and DroppedRecords) without the
+		// queue-full message, which would misattribute the cause.
+		s.dropped.Add(1)
```
eunolabs, line 369: Non-blocking note on audit completeness during shutdown (not a defect — strictly better than the prior panic):

Drops counted on this post-`closed` path are **not** recorded as a signed `AUDIT_RECORDS_DROPPED` chain marker. The drainer writes its final `flushDropMarker()` and exits inside `wg.Wait()`, so any `record()` that observes `closed == true` *after* the drainer is gone increments `dropped` but never makes it into the tamper-evident chain. It's only observable via a later `DroppedRecords()` read.

That also means the `dropped` snapshot the warning in `Close()` prints (`if n := s.dropped.Load(); n > 0`) right after `wg.Wait()` can under-count: producers still running after `Close()` returns (exactly what the race test does) keep incrementing the counter past that read.

Given this repo's emphasis on audit completeness as a core property, it'd be worth a one-line note in `docs/threat-model-mcp.md` (or the CHANGELOG entry) making explicit that decisions made during the shutdown window are counted-dropped but not chained — i.e. the shutdown drop window is a known, bounded gap rather than a guarantee. No code change required.

---

eunolabs (commented):
## Review — looks correct, well-reasoned, well-tested

I traced the fix against `record()`, `Close()`, and the drainer, and the race in #231 is genuinely closed.

**Correctness**
- `record()` holds `mu.RLock()` across *both* the `closed` check and the non-blocking send, and `Close()` sets `closed = true` + `close(s.records)` under `mu.Lock()`. The write lock can't be acquired until every in-flight read-locked send has released, and any send arriving after the close observes `closed` and drops — so a send can never be on the `case s.records <- rec` arm at the instant the channel closes. The "send on closed channel" panic is eliminated.
- No deadlock: the send is `select { … default }`, so `record()` never blocks while holding `RLock`, and `Close()` always makes progress to the write lock.
- `closed` is only ever read/written under `mu`, so a plain `bool` (no atomic) is race-free — confirmed it's clean under `-race`.
- Releasing the lock *before* `wg.Wait()` is the right call (post-close `record()` calls drop immediately instead of blocking behind the drainer's final flush), and the comment explains it.
- Ordering of `closed = true` before `close()` under the same write lock means a reader sees both or neither — no torn observation.

**Design** — `RWMutex` over a plain `Mutex` is a good choice here: producers stay parallel on the hot path (channel sends are themselves concurrency-safe) and only serialize against the single `Close()`. Matches the "robust" approach suggested in #231.

**Tests** — `TestAuditSink_CloseRace_NoSendOnClosedChannel` (64×500 producers released together, concurrent `Close()`, + idempotent second `Close()`) plus the deterministic `TestAuditSink_RecordAfterClose_DropsWithoutPanic` cover both the probabilistic race and the exact post-close drop path. Good coverage.

**Housekeeping** — CHANGELOG entry is accurate, license header present on the new file, Conventional Commit title. No new manifest grammar / audit field, so no other docs are strictly required.

**One non-blocking note** (inline on `audit.go`): drops on the post-`closed` path aren't written as a signed `AUDIT_RECORDS_DROPPED` chain marker and the `Close()` warning's `dropped` snapshot can under-count late producers. Strictly better than the previous panic, but given how central audit completeness is here, a one-line mention in the threat model / CHANGELOG that the shutdown window is a known counted-but-unchained gap would be worth considering.

CI was still in flight at review time (the `Test` job — which runs these race tests — was queued); nothing had failed. I'll watch the PR and re-review on new pushes.

---

cmd/eunox-mcp/audit.go
eunolabs: Good catch — documented in c52ca9f. Added a "Shutdown drop window (bounded, counted-but-unchained)" paragraph to threat model §3.6 making explicit that post-`closed` drops are counted via `DroppedRecords()` but not folded into a signed `AUDIT_RECORDS_DROPPED` marker (drainer has already flushed and exited), and that `Close()`'s printed `dropped` snapshot can under-count late producers. Also tightened the L-3 risk-register row and the CHANGELOG entry. No code change, as you noted.

---

PR #251, opened 2026-06-13
fix: deny allowedTables map argument lacking 'table' key with accurate error

### LINKED ISSUE #225: bug: parseTableArgument returns nil (not an error) for map with missing "table" key, yielding misleading "argument missing or empty" denial
## Summary

When the argument for an `allowedTables` condition is a map that lacks the `"table"` key (e.g. `{"columns": ["id", "name"]}`), `parseTableArgument` returns `(nil, nil)` — no error, empty slice. The calling code in `handleAllowedTables` then reaches the `len(tables) == 0` guard and returns a `ErrCodeMissingContext` denial saying `"argument %q is missing or empty"`. The argument is neither missing nor empty: it is present and structurally populated, just with an absent `"table"` key inside.

## Code path

```go
// handlers.go — parseTableArgument
case map[string]interface{}:
    tableName, _ := t["table"].(string)
    if tableName = strings.TrimSpace(tableName); tableName == "" {
        return nil, nil   // ← silent nil on missing/empty table key
    }
```

```go
// handlers.go — handleAllowedTables
tables, perr := parseTableArgument(v)
// perr is nil, tables is nil
...
if len(tables) == 0 {
    return &ConditionError{
        Code:    capability.ErrCodeMissingContext,
        Message: fmt.Sprintf("argument %q is missing or empty", at.Argument),
    }
}
```

**Scenario**: operator's tool sends `{"table": {"columns": ["id", "name"]}}` with no `"table"` sub-key.
- The argument IS present (`req.Arguments["table"]` is a map).
- `parseTableArgument` returns `nil, nil` because the map's `"table"` key is absent.
- `handleAllowedTables` says "argument 'table' is missing or empty" — misleading.

The error code `ErrCodeMissingContext` is also wrong: the argument is present in the request.

## Suggested fix

Return an explicit error from `parseTableArgument` when the map has a non-empty `"columns"` list but no `"table"` key:

```go
case map[string]interface{}:
    tableName, _ := t["table"].(string)
    tableName = strings.TrimSpace(tableName)
    if tableName == "" {
        // Distinguish: is the whole map empty, or just missing the "table" sub-key?
        if _, hasColumns := t["columns"]; hasColumns {
            return nil, fmt.Errorf("map argument is missing required 'table' key")
        }
        return nil, nil
    }
```

Or, more consistently, always return an error when the map is present but yields no table name:

```go
if tableName == "" {
    return nil, fmt.Errorf("table specification map is missing a non-empty 'table' key")
}
```

Either way, the error surfaces as `ErrCodeConditionFailed` with a message that accurately describes the structural problem, rather than `ErrCodeMissingContext` implying the argument was absent.

eunolabs:
Fixes #225

## Problem

When the argument for an `allowedTables` condition is a map that is present and populated but lacks the `"table"` key (e.g. `{"columns": ["id", "name"]}`), `parseTableArgument` returned `(nil, nil)` — no error, empty slice. `handleAllowedTables` then reached the `len(tables) == 0` guard and denied with `ErrCodeMissingContext` and the message `argument "table" is missing or empty`.

That denial is misleading on two counts:
- The argument is **not** missing or empty — it is present in the request and structurally populated, it just lacks the required `"table"` sub-key.
- `ErrCodeMissingContext` (MISSING_CONTEXT) is the wrong code; the argument is present.

## Fix

`parseTableArgument` now returns a non-nil error when a map argument carries no non-empty `"table"` key. The caller already routes a non-nil parse error through the `ErrCodeConditionFailed` (CONDITION_FAILED) branch with a message describing the real problem:

```
argument "table" is a map without the required 'table' key
```

This is the same fail-closed treatment already applied to non-string array items and non-string column entries (per `parseTableArgument`'s doc comment), so the change is consistent with existing behavior — a structurally malformed argument is denied with an accurate structural error rather than silently evaluating a valid subset or emitting a misleading "missing or empty" denial. The decision is a deny either way (fail-closed preserved); only the error code and message change to be accurate.

Genuinely-missing (`req.Arguments[arg]` absent), empty-string, and empty-array arguments continue to deny with `ErrCodeMissingContext` "missing or empty" as before — those are accurately described as missing/empty.

## Tests

Added regression subtests to `TestAllowedTables_ColumnsAndShapes` (the test dedicated to `parseTableArgument` map handling):
- map present with `columns` but no `table` key → deny, `CONDITION_FAILED`, message describes the structural problem and is **not** "missing or empty"
- empty map `{}` → deny, `CONDITION_FAILED`
- array containing a map item without a `table` key → deny, `CONDITION_FAILED` (fail-closed through the array branch)

`go build ./...`, `go test -race ./pkg/enforcement/`, `gofmt -l`, and `go vet` all pass.

https://claude.ai/code/session_017UTVqqUt3vabBDWvbCdPTy

---
_Generated by [Claude Code](https://claude.ai/code/session_017UTVqqUt3vabBDWvbCdPTy)_


pkg/enforcement/handlers.go
```diff
@@ -916,7 +917,14 @@ func parseTableArgument(v interface{}) ([]capability.TableAccess, error) {
...
 		if tableName = strings.TrimSpace(tableName); tableName == "" {
-			return nil, nil
+			// The map form is an explicit table specification. A map that
+			// carries no non-empty "table" key (e.g. {"columns": ["id"]}) is
+			// structurally malformed, not a missing or empty argument: the
+			// value is present and populated, it just lacks the required key.
+			// Return an error so the caller denies fail-closed with
+			// CONDITION_FAILED and an accurate message, rather than the
+			// misleading MISSING_CONTEXT "argument is missing or empty" denial.
+			return nil, fmt.Errorf("is a map without the required 'table' key")
```
eunolabs: Two small things on this error fragment (neither blocks the fix, which is correct):

1. **Double-meaning of "table" in the common case.** The caller formats this as `argument %q %s`, so when the argument is itself named `table` (the canonical example used throughout the guide and every test here), the full message reads `argument "table" is a map without the required 'table' key`. "table" then appears in two different senses — the argument name and the required sub-key — which is a touch confusing. Consider disambiguating the sub-key, e.g. `is a map with no non-empty 'table' entry`.

2. **Array-element context is lost.** When this map sits inside an array argument (the third new test), the error propagates verbatim through the `[]interface{}` branch, so the message doesn't indicate it was one element of an array — unlike the sibling `array contains a non-string item: %T` error, which does. Minor, but if you reword for (1) you could fold in the array case too. Purely cosmetic; the deny + `CONDITION_FAILED` code is right.

---

eunolabs (commented):
Reviewed the fix for #225. **The change is correct and well-scoped**, and I verified locally on the PR head (`94e08aa`): `go vet`, `gofmt -l`, and `go test -race ./pkg/enforcement/` all pass.

### What's good
- Addresses #225 exactly: a map argument that is present and populated but lacks a non-empty `table` key now denies with `CONDITION_FAILED` and an accurate structural message, instead of the misleading `MISSING_CONTEXT` "argument is missing or empty". Fail-closed is preserved (deny either way).
- Consistent with the existing treatment of non-string array items and non-string column entries — a structurally malformed argument is never silently narrowed to a valid subset.
- The array regression test is the important one: a map-without-`table` *inside an array* now fails the whole call closed rather than silently evaluating the valid element. Good catch to cover that path.
- The doc comment on `parseTableArgument` is updated to describe the new case.

### One item to address before merge
**Docs update missing.** Per `CLAUDE.md`/`CONTRIBUTING.md`, behavior changes need a docs update in the same PR. `docs/capability-manifest-guide.md` (the "non-string item … fail-closed" paragraph, ~lines 865–871) explicitly enumerates the `allowedTables` structurally-malformed cases (non-string array item, non-string column entry) and states "a structurally malformed argument is never treated as a smaller, valid one." This PR adds a new such case — a map argument with no non-empty `table` key, including within an array — so that paragraph should be extended to mention it.

### Behavior nuance worth confirming (looks intentional)
The empty map `{}` now denies with `CONDITION_FAILED` ("…without the required 'table' key") rather than the previous `MISSING_CONTEXT` "missing or empty". This is the unified "any present map yielding no table name is malformed" treatment (the issue's second suggested approach) and you call it out + test it in the PR description, so flagging only for visibility — no change needed.

Plus one cosmetic nit on the error wording left inline. Nothing here is a correctness blocker.

---

eunolabs:
Addressed in a8bebc0:

- **Docs (before-merge item):** extended the "structurally malformed argument is never treated as a smaller, valid one" paragraph in `docs/capability-manifest-guide.md` to cover the new `allowedTables` object-form case (`{table, columns}` with no non-empty `table` entry, including inside an array) → denied fail-closed with `CONDITION_FAILED`, not reported as missing/empty.
- **Inline nit (1):** reworded to `is a map with no non-empty 'table' entry` so it no longer reads as a duplicate of the argument name when the argument is itself named `table`.
- **Inline nit (2):** the `[]interface{}` branch now wraps element errors with an `array element` prefix, mirroring the sibling `array contains a non-string item` error, so array context is preserved.

No behavior change — deny + `CONDITION_FAILED` is unchanged. Regression assertions updated to the new wording and to check the array-element context. `go build`, `go test -race ./pkg/enforcement/`, `gofmt -l`, and `go vet` pass locally.

The empty-map `{}` → `CONDITION_FAILED` behavior is intentional (unified "any present object yielding no table name is malformed"), as noted.

---

PR #252, opened 2026-06-13
fix: preserve large-integer details in audit HMAC verification (#230)

### LINKED ISSUE #230: bug: `VerifyRecord` JSON round-trip converts `int64` values in `Details` to `float64`, causing HMAC mismatch for records with large integer counts
## Summary

`writeRecord` signs the marshaled JSON of an `auditRecord` whose `Details` field contains Go-typed values:

```go
// audit.go:464-471 — writeRecord
body, err := json.Marshal(rec)   // Details values retain their original Go types (e.g. int64)
mac := hmac.New(sha256.New, s.key)
mac.Write(body)
rec.HMAC = "sha256:" + hex.EncodeToString(mac.Sum(nil))
```

`VerifyRecord` reconstructs the body via a JSON round-trip:

```go
// audit.go:536-547 — VerifyRecord
var rec auditRecord
json.Unmarshal(line, &rec)       // Details["dropped"] arrives as float64 (JSON default)
storedHMAC := rec.HMAC
rec.HMAC = ""
body, err := json.Marshal(rec)   // re-marshals float64 — may produce different bytes
```

Because `Details` is `map[string]interface{}`, `encoding/json` decodes all JSON numbers into `float64`. For values that cannot be represented exactly as `float64` (integers larger than 2⁵³), the re-marshaled JSON differs from the original, and the recomputed HMAC does not match the stored one: **`VerifyRecord` returns `false` for a record that was correctly signed**.

## Concrete case

`flushDropMarker` writes `AUDIT_RECORDS_DROPPED` records whose `details` carry the drop counts as `int64`:

```go
// audit.go:435-438
    "dropped":       newDrops,        // int64
    "dropped_total": cur,             // int64
},
```

If either value exceeds 2⁵³ (≈ 9 × 10¹⁵) — admittedly extreme but possible on a long-running, saturated instance — `float64` cannot represent it exactly. The re-marshal produces a different decimal string, the HMAC diverges, and `audit-verify` reports the `AUDIT_RECORDS_DROPPED` marker as `INVALID`.

The same issue affects any custom `details` map containing large integers passed by the enforcement engine.

## Why this matters

`AUDIT_RECORDS_DROPPED` is the only in-chain signal that records were lost. If it fails HMAC verification, `audit-verify` treats it as tampered — exactly the kind of false alarm that erodes operator confidence in the audit trail.

More broadly, the correctness of HMAC verification should not depend on whether integer values in `Details` happen to be small enough for float64 precision.

## Suggested fix

Use `json.Number` decoding to preserve numeric fidelity during the round-trip, or switch `Details` to `json.RawMessage` and preserve the original JSON bytes without re-encoding:

**Option A — `json.Decoder` with `UseNumber()`:**

```go
func (s *auditSink) VerifyRecord(line []byte) (bool, error) {
    var rec auditRecord
    d := json.NewDecoder(bytes.NewReader(line))
    d.UseNumber()
    if err := d.Decode(&rec); err != nil {
        return false, err
    }
    storedHMAC := rec.HMAC
    rec.HMAC = ""
    body, err := json.Marshal(rec)  // json.Number marshals as its original string
    ...
```

**Option B — change `Details` to `json.RawMessage`** in `auditRecord`, store the raw JSON bytes at write time, and skip re-encoding entirely during verification.

Option A is less invasive. Option B gives exact round-trip fidelity at the cost of changing the `Details` field type.

## Affected code

`cmd/eunox-mcp/audit.go`, `VerifyRecord` (lines 536–560) and `flushDropMarker` (lines 421–445).

eunolabs:
Fixes #230.

## Problem

`VerifyRecord` (`cmd/eunox-mcp/audit.go`) recomputes a record's HMAC by JSON round-tripping it: `json.Unmarshal(line, &rec)` then `json.Marshal(rec)`. Because `auditRecord.Details` is `map[string]interface{}`, `encoding/json` decodes **all** JSON numbers in `details` to `float64`. Integers larger than 2⁵³ (≈ 9 × 10¹⁵) have no exact `float64` representation, so the re-marshaled bytes diverge from the originally-signed bytes and the recomputed HMAC does not match — a correctly-signed record is reported `INVALID`.

The concrete trigger is an `AUDIT_RECORDS_DROPPED` marker on a long-running, saturated instance whose `dropped` / `dropped_total` counts exceed 2⁵³ (written as `int64` by `flushDropMarker`). Any custom `details` map carrying large integers from the enforcement engine was affected too. This is exactly the kind of false alarm that erodes operator confidence in the audit trail.

## Fix

Decode in `VerifyRecord` with `json.Decoder` + `UseNumber()` (the issue's suggested Option A — least invasive). Each number is preserved as `json.Number`, i.e. its exact textual form, which `encoding/json` re-marshals **byte-for-byte**. The verification round-trip is therefore exact at any magnitude.

- **Signing path unchanged** — `writeRecord` still signs the marshaled `int64`/native Go values.
- **On-disk record format unchanged** — no new fields, no audit-shape change, no threat-model impact.
- **Backward compatible** — previously written tapes verify correctly without rewriting.

## Tests

`cmd/eunox-mcp/audit_largeint_test.go`:

- `TestAuditVerify_LargeIntDetailsRoundTrip` — a record whose `details` carries `2^53 + 1` round-trips through `VerifyRecord` and verifies; asserts the integer is on disk verbatim.
- `TestAuditVerify_LargeDropCountMarker` — reproduces the concrete issue: a saturated `AUDIT_RECORDS_DROPPED` marker with counts above 2⁵³ verifies per-record and across the tamper-evident chain.

Both tests were confirmed to **fail** against the pre-fix code (record reported `INVALID`) and pass with the fix. Full `cmd/eunox-mcp` suite, `make check-fmt`, and `make check-license` all pass.

`CHANGELOG.md` updated under Unreleased → Fixed.

---
_Generated by [Claude Code](https://claude.ai/code/session_011rRku68J9GVLVXUxSFYUE3)_


cmd/eunox-mcp/audit_largeint_test.go
```diff
@@ -0,0 +1,116 @@
...
+			t.Fatalf("VerifyRecord: %v", err)
+		}
+		if !ok {
+			t.Errorf("record failed HMAC verification: %s", line)
+		}
+		if strings.Contains(string(line), "AUDIT_RECORDS_DROPPED") {
+			sawMarker = true
+			if !strings.Contains(string(line), strconv.FormatInt(bigInt, 10)) {
+				t.Errorf("drop marker lost precision on large count: %s", line)
+			}
```
eunolabs, line 104: Non-blocking suggestion: consider adding a *negative* case to lock in that the fix doesn't weaken tamper-evidence at large magnitudes. Both new tests assert the positive direction (a correctly-signed large-int record verifies). What's not yet covered is that a **tampered** large-int detail is still reported `INVALID` after the `UseNumber()` change — e.g. take the on-disk `AUDIT_RECORDS_DROPPED` line, edit `dropped_total` to a different >2^53 value, and assert `VerifyRecord` returns `false`.

This matters specifically here because `AUDIT_RECORDS_DROPPED` is a security-relevant signal: an attacker who can edit the tape might want to *shrink* the drop count to hide the scale of a loss window. The existing tamper test (`audit_chain_test.go:151`) exercises the chain fields (`seq`/`prev_hmac`) but not a large-integer `details` value going through the new decode path. `json.Number` preserves the exact textual bytes, so editing the count will change the re-marshaled body and break the HMAC — this would just prove it and prevent a future regression from silently turning into a false-negative.

---

eunolabs (commented):
## Review: large-integer fidelity in audit HMAC verification

Reviewed against the codebase (checked out `claude/eager-bell-prwqd9`). This is a clean, correctly-scoped fix for #230 — the issue's recommended Option A (`json.Decoder` + `UseNumber()`), applied at the one place it's needed.

### What I verified

- **Root cause & fix are right.** `Details` is `map[string]interface{}`, so a plain `json.Unmarshal` coerces every JSON number to `float64`; integers > 2^53 lose precision and the re-marshaled body diverges from the signed bytes. `UseNumber()` keeps each number as `json.Number` (its exact on-disk text), which re-marshals byte-for-byte. Because the on-disk `details` bytes *are* the signed bytes, the round-trip is now exact at any magnitude — actually more robust than relying on `int64`, since it also covers floats/exponential forms.

- **The fix is complete — `VerifyRecord` is the single HMAC-recompute chokepoint.** I traced every JSON round-trip in `cmd/eunox-mcp` that touches audit records:
  - `audit.go:233` (chain resume) and `audit.go:683` (`verifyAuditLog`) both recompute HMAC *through* `VerifyRecord`, so they inherit the fix.
  - `verifyAuditLog`'s own `json.Unmarshal` (`audit.go:648`) only reads top-level typed fields (`Seq uint64`, `PrevHMAC`/`HMAC` strings) for chain linkage — none are subject to the `float64` bug.
  - `main.go:1524` (stats), `suggest.go:117`, and `doctor.go:486` (redact bundle) decode into narrow structs / redact details and never recompute HMAC.
  - `UseNumber()` has no effect on the concrete typed top-level fields — they still parse normally.

- **Tests are genuine regression tests.** I reverted just the `VerifyRecord` change and re-ran: both new tests **fail** (`record ... failed HMAC verification`, chain `invalid:1`); with the fix they **pass**. Broader `Audit|Verify|Chain|Drop` suite passes; `gofmt` clean; license header present; `CHANGELOG` updated under Unreleased → Fixed.

- **No audit-shape / threat-model impact.** No new fields, on-disk format unchanged, previously written tapes still verify. Consistent with the "audit records are append-only and structured" invariant.

### Suggestion (non-blocking)

One inline comment on `audit_largeint_test.go`: add a negative case asserting that a *tampered* large-int `details` value (e.g. an edited `dropped_total`) is still reported `INVALID`, to prove the `UseNumber()` change preserves tamper-evidence at >2^53 magnitudes and not just that valid records pass.

LGTM overall — the suggestion is optional hardening, not a blocker. (Heads-up: `mergeable_state` is currently `unstable` only because CI was still running at review time.)

---

cmd/eunox-mcp/audit_largeint_test.go
eunolabs: Done in cb00e39. Added `TestAuditVerify_LargeDropCountTamperDetected`: it writes a saturated `AUDIT_RECORDS_DROPPED` marker with a `dropped_total` of `9999999999999999` (>2^53, float64-inexact), then shrinks it on disk to `9007199254740993` (still >2^53, still inexact) — the exact "hide the loss window" edit you described — and asserts `VerifyRecord` reports the marker `INVALID`. Confirms the `UseNumber()` change preserves tamper-evidence at large magnitudes rather than turning into a false-negative.

---

eunolabs: Verified — checked out `cb00e39`, built and ran the new test. `TestAuditVerify_LargeDropCountTamperDetected` passes and correctly reports the shrunk `dropped_total` marker as `INVALID`; the `sawMarker` guard ensures it's actually exercising the marker, and the `"dropped_total":`-prefixed replace cleanly avoids the equal-valued `dropped` field. Together with the positive round-trip test (which I confirmed fails pre-fix), the pair now covers both directions. Thanks for the quick turnaround — no further comments from me.

---

PR #254, opened 2026-06-13
sec: prevent same-second audit rotations from overwriting prior rotated file

### LINKED ISSUE #227: bug: audit log rotation silently overwrites prior rotated file when two rotations occur within the same second
## Summary

`rotate()` in `cmd/eunox-mcp/audit.go` builds the rotated filename using a **second-precision** UTC timestamp:

```go
// audit.go:700
rotated := s.logPath + "." + time.Now().UTC().Format("20060102T150405Z")
if err := os.Rename(s.logPath, rotated); err != nil {
```

On POSIX systems (Linux, macOS), `os.Rename` maps to `rename(2)`, which **atomically replaces the target if it already exists** — there is no error returned, and no data is preserved. If two rotations happen within the same wall-clock second, the second `rename` silently overwrites the first rotated file, permanently destroying its records.

## Reproduction

Set `--audit-rotate-size 1` (or `rotateSizeBytes: 1` in config) and write two audit records in quick succession. The first rotated file (`audit.jsonl.20260613T120000Z`) is silently replaced by the second rotation. Because both renames succeed with no error, the drainer goroutine proceeds normally — there is no stderr warning and no chain marker for the lost file.

In tests this is easy to trigger: `TestAuditRotate_TriggeredByMaxBytes` uses `maxBytes=1` and writes one record, but a test writing two records under the same sink would expose the collision.

## Impact

Audit records in the first rotated file are permanently lost with no indication in the active log, the chain, or stderr. This undermines the tamper-evident audit guarantee for the affected window.

## Root cause

`time.Now().UTC().Format("20060102T150405Z")` has one-second resolution. Combined with POSIX rename semantics, identical timestamps overwrite the target.

## Suggested fix

Use nanosecond or at minimum sub-second resolution:

```go
rotated := s.logPath + "." + time.Now().UTC().Format("20060102T150405.000000000Z")
```

Or add a monotonic counter suffix that increments when the target path already exists:

```go
base := s.logPath + "." + time.Now().UTC().Format("20060102T150405Z")
rotated := base
for i := 1; ; i++ {
    if _, err := os.Lstat(rotated); os.IsNotExist(err) {
        break
    }
    rotated = fmt.Sprintf("%s.%d", base, i)
}
```

The nanosecond approach is simpler and avoids the stat loop. Either fix should also add a test that writes enough records to force two consecutive rotations and asserts both rotated files survive.

## Affected code

`cmd/eunox-mcp/audit.go`, line 700 (`rotate()`).

eunolabs:
Fixes #227

## Problem

`rotate()` in `cmd/eunox-mcp/audit.go` built the rotated filename from a **second-precision** UTC timestamp:

```go
rotated := s.logPath + "." + time.Now().UTC().Format("20060102T150405Z")
if err := os.Rename(s.logPath, rotated); err != nil {
```

On POSIX systems `os.Rename` maps to `rename(2)`, which **atomically replaces an existing target** with no error and no trace. When two rotations occur within the same wall-clock second (easy to hit with a small `--audit-rotate-size`), the second rename silently overwrites the first rotated file — permanently destroying its records with no stderr warning and no chain marker. This undermines the tamper-evident audit guarantee for the affected window.

## Fix

- New `rotatedPath()` helper builds the rotated-file suffix at **nanosecond resolution** (`audit.jsonl.20060102T150405.000000000Z`), making same-second collisions vanishingly unlikely.
- As a deterministic backstop it probes (`os.Lstat`) for a target name that does **not** yet exist, appending a numeric suffix if needed, so an existing rotated file can never be replaced regardless of clock resolution. Each iteration probes a distinct name, so the loop always terminates; an undeterminable stat error falls through to `rotate()`'s existing rename-error branch (which keeps the open fd).
- `rotate()` keeps its fail-closed behavior unchanged otherwise.

This preserves the load-bearing invariants: audit records are never silently lost, and the change adds no new runtime dependencies.

## Test

`TestAuditRotate_TwoRotationsSameSecond` forces two consecutive rotations (both landing in the same wall-clock second) and asserts both rotated files survive with their distinct records. On the old code the first rotated file's record is missing; with the fix both are present.

## Docs

`docs/architecture.md` now describes the rotated-file naming and its collision-safety guarantee.

## Verification

- `go build ./...`
- `gofmt -l` clean, `go vet ./cmd/eunox-mcp/` clean

https://claude.ai/code/session_01XKKsfS7WNcMAG4jbyN8mMy

---
_Generated by [Claude Code](https://claude.ai/code/session_01XKKsfS7WNcMAG4jbyN8mMy)_


cmd/eunox-mcp/audit.go
```diff
@@ -715,6 +719,32 @@ func (s *auditSink) rotate() {
...
+func (s *auditSink) rotatedPath() string {
+	base := s.logPath + "." + time.Now().UTC().Format("20060102T150405.000000000Z")
+	candidate := base
+	for i := 1; ; i++ {
+		// os.Lstat returns nil only when the path already exists; any error
+		// (ErrNotExist, or an undeterminable stat failure) means it is safe to
+		// use this candidate — a failed rename then falls through to rotate()'s
+		// existing rename-error branch, which keeps the open fd. Each iteration
+		// probes a distinct name, so the loop always terminates.
+		if _, err := os.Lstat(candidate); err != nil {
```
eunolabs, line 758: The logic here is correct, and two details are worth calling out as deliberately good: choosing `Lstat` over `Stat` means a dangling-symlink name still counts as "occupied" (so a rename won't clobber through it), and returning the candidate on *any* stat error is the right fail-safe — a genuinely undeterminable name just falls through to `rotate()`'s rename-error branch, which keeps the fd. Loop termination is also fine since `base` is fixed before the loop.

One non-blocking note: the uniqueness guarantee is time-of-check (Lstat) to time-of-use (the `os.Rename` in `rotate()`), so it relies on no other writer creating that exact target in between. That holds today because `rotate()` is only ever reached from `writeRecord` on the single `drain()` goroutine — but nothing in this function pins that assumption. A one-line comment noting "safe because rotation is single-goroutine (drain)" would protect a future caller from invoking `rotate()` concurrently and quietly reintroducing the overwrite this PR closes.

---

cmd/eunox-mcp/audit_rotate_keys_test.go
```diff
@@ -100,6 +100,71 @@ func TestAuditRotate_NilFileGuard(t *testing.T) {
...
+	sink.written = 1
+	sink.rotate()
+
+	// Second rotation, immediately after (same wall-clock second): a different
+	// marker into the freshly reopened log, rotate it out too.
+	if _, err := sink.f.WriteString(`{"marker":"second"}` + "\n"); err != nil {
+		t.Fatalf("seed second: %v", err)
+	}
+	sink.written = 1
+	sink.rotate()
```
eunolabs, line 134: Solid regression guard — on the old second-precision code this reliably fails (two `rotate()` calls land microseconds apart, same second, second rename clobbers the first).

Optional: which path the *fixed* code takes here depends on host clock granularity. On a fine-grained clock the two rotations get distinct nanosecond suffixes and you get 2 files without the `Lstat` backstop loop ever running; the same-second collision path (the `base.%d` fallback) is only exercised when the clock is coarse enough to return the same nanosecond twice. So the assertion is robust either way, but the backstop itself isn't deterministically covered. If you want guaranteed coverage of the fallback, you could call `sink.rotatedPath()` directly, pre-create that exact file, then assert the next `rotatedPath()` returns a `.1`-suffixed name. Not blocking.

---

eunolabs (commented):
Reviewed the fix for #227 (same-second audit rotation overwrite). This is a clean, well-scoped, well-documented security fix and I think it's correct.

**What's right**
- The root cause is accurately diagnosed: POSIX `os.Rename` → `rename(2)` atomically replaces an existing target, so a second-precision suffix silently destroys the earlier rotated file (and it's not POSIX-only — Go's `os.Rename` on Windows uses `MOVEFILE_REPLACE_EXISTING` too, so the `Lstat` backstop hardens every platform).
- Defense in depth is the right shape: nanosecond suffix makes collisions vanishingly unlikely, and the `Lstat` uniqueness probe is a deterministic backstop even on a coarse clock.
- Fail-closed behavior is preserved — an undeterminable target name falls through to the existing rename-error branch, which keeps the open fd, so records are never lost.
- No new runtime dependencies; one static binary intact.
- I checked the consumption side: nothing parses the rotated-filename format programmatically (`audit-verify` opens a single file by explicit `--audit-log` path, and rotated files aren't auto-discovered/sorted), so the format change is safe and breaks no readers. The new two-dot suffix still sorts chronologically.

**Non-blocking notes** (see inline):
1. The Lstat→Rename uniqueness is time-of-check/time-of-use; it's safe only because `rotate()` runs solely on the single `drain()` goroutine. Worth a one-line code comment pinning that invariant so a future caller doesn't reintroduce the race.
2. The regression test passes regardless of clock granularity, but the `base.%d` backstop loop is only exercised on a coarse clock — consider a direct `rotatedPath()` test to deterministically cover the fallback.
3. Optional: since this strengthens an audit-integrity guarantee, a one-liner in `docs/threat-model-mcp.md` §3.4 (Audit Log Tampering) would keep the threat model in step with `docs/architecture.md`, which you already updated.

None of these block merge. CI was still queued at review time. Nice work.

---

eunolabs:
Thanks for the thorough review. Addressed all three non-blocking notes in `cadf136`:

1. **Single-goroutine invariant** — `rotatedPath`'s doc comment now spells out that the Lstat→Rename uniqueness is time-of-check/time-of-use and safe only because `rotate()` runs solely on the `drain()` goroutine, so a future caller can't quietly reintroduce the overwrite. Also noted Windows `os.Rename` replaces via `MOVEFILE_REPLACE_EXISTING`, per your point that the backstop hardens every platform.
2. **Deterministic backstop coverage** — split the Lstat loop into `uniqueRotatedPath(base)` and added `TestUniqueRotatedPath_BackstopSuffix`, which drives the `.N` fallback directly (free base → base; occupied base + base.1 → base.2) regardless of clock granularity, plus a dangling-symlink case confirming `Lstat` (not `Stat`) treats it as occupied.
3. **Threat model** — added a "Rotation integrity" note to §3.4 so it stays in step with the architecture doc.

---

PR #255, opened 2026-06-13
fix: declare Peek/IncrementIfBelow store contracts in pkg/callcounter (#236)

### LINKED ISSUE #236: interface: callcounter.Store doesn't declare Peek — sequenceBlock requirement is invisible to implementors
## Summary

`pkg/callcounter/callcounter.go` defines the `Store` interface with only `IncrementAndGet`. Both `InMemory` and `Redis` implement `Peek`, but `Peek` is not part of any interface in the `callcounter` package. The enforcement engine defines a separate `CallHistory` interface in `pkg/enforcement/engine.go` and detects `Peek` support via a runtime type assertion.

## The trap

A developer writing a third-party (or new built-in) `Store` implementation satisfies the public `Store` interface but has no signal that `Peek` is also required. The code compiles and runs. At runtime, any manifest that uses `sequenceBlock` produces:

```
"configured call counter does not support the session-history lookups required by sequenceBlock"
```

This is fail-closed (the condition denies), but the error message is opaque and there is no compile-time or test-time signal that the implementation is incomplete.

## Current state

```go
// pkg/callcounter/callcounter.go
type Store interface {
    // Peek is NOT here
}

// pkg/enforcement/engine.go
type CallHistory interface {
}

// detection at runtime:
history, ok := e.counter.(CallHistory)
if !ok {
    return &ConditionError{...} // fail closed but silent
}
```

## Proposed fix

**Option A** — Define a `PeekableStore` (or `FullStore`) interface in `pkg/callcounter` that embeds `Store` and adds `Peek`. Move or alias `enforcement.CallHistory` to refer to it. Both `InMemory` and `Redis` already implement it; new implementations get a compile error if they omit `Peek`.

**Option B** — Add `Peek` directly to `Store`. This is a breaking change (pre-1.0 so acceptable per CLAUDE.md) that makes the requirement unambiguous.

**Option C** — Move `CallHistory` to `pkg/callcounter` so the requirement is discoverable in the same package as `Store`.

Any of these closes the documentation and discoverability gap.

## Affected code

- `pkg/callcounter/callcounter.go:15–21` (`Store` interface)
- `pkg/enforcement/engine.go:63–68` (`CallHistory` interface)
- `pkg/enforcement/engine.go:750–751` (runtime type assertion)

eunolabs:
## Summary

Fixes #236.

`callcounter.Store` declared only `IncrementAndGet`, but the enforcement engine reaches for two more methods at runtime via type assertion:

- `Peek` — session-history lookup for the `sequenceBlock` condition (`enforcement.CallHistory`)
- `IncrementIfBelow` — atomic rate check for the `maxCalls` condition (`enforcement.RateLimiter`)

A backend (third-party or a new built-in) that satisfied `Store` compiled and ran, but had **no signal** that those two methods were also required. The result was fail-closed but opaque: every `sequenceBlock`/`maxCalls` condition denied with `configured call counter does not support ...`, with no compile-time or test-time warning that the implementation was incomplete.

## Approach

This implements **Option A** from the issue, placing the discoverable interfaces in `pkg/callcounter` where backends are actually written:

- **`HistoryStore`** — embeds `Store`, adds `Peek`. Required by `sequenceBlock`.
- **`LimiterStore`** — embeds `Store`, adds `IncrementIfBelow`. Required by `maxCalls`.

Each interface's doc names the condition it unlocks, and the `Store` doc now describes the full taxonomy. Both built-in backends (`InMemory`, `Redis`) are pinned to both interfaces with compile-time assertions, so a new built-in that forgets a method is a **build error**, not a silent runtime deny. A third-party backend can pin itself the same way (`var _ callcounter.HistoryStore = (*MyStore)(nil)`).

The issue scoped the gap to `Peek`/`sequenceBlock`; `IncrementIfBelow`/`maxCalls` is the identical sibling trap, so both are documented together rather than fixing one and leaving the other half-documented.

### On the `enforcement.CallHistory` alias

The issue suggested aliasing `enforcement.CallHistory` to the new interface. I deliberately **did not** alias, to preserve the existing layering: `capability.CallCounter` already duplicates `callcounter.Store` rather than importing it, explicitly "so that consumers can import the interface without importing the implementation package." Aliasing would add a new production `enforcement → callcounter` import, contradicting that decoupling. Instead, `CallHistory`/`RateLimiter` stay as the consumer-side mirrors and now cross-reference their producer-side counterparts in their doc comments. The structural identity means the runtime type assertions and the built-in backends are unaffected.

**No behavior change** for the shipped backends — this is purely a discoverability/ergonomics fix.

## Changes

- `pkg/callcounter/callcounter.go` — add `HistoryStore` and `LimiterStore`; enrich the `Store` doc with the extended-contract taxonomy.
- `pkg/callcounter/callcounter_test.go` — compile-time assertions pinning `InMemory`/`Redis` to both interfaces (matches the existing convention in `enforcement_test.go`).
- `pkg/enforcement/engine.go` — cross-reference the producer-side interfaces from the `CallHistory`/`RateLimiter` docs.
- `docs/architecture.md` — extension-points row and call-counter bullet now spell out the `HistoryStore`/`LimiterStore` requirements.
- `CHANGELOG.md` — `Fixed` entry under Unreleased.

## Verification

- `go build ./...` — clean
- `go vet ./pkg/callcounter/... ./pkg/enforcement/...` — clean
- `go test -race -count=1 ./pkg/callcounter/... ./pkg/enforcement/...` — pass
- `gofmt -l` on changed files — clean

https://claude.ai/code/session_0115D2Mcgy952LJsKHwT1uWC

---
_Generated by [Claude Code](https://claude.ai/code/session_0115D2Mcgy952LJsKHwT1uWC)_


CHANGELOG.md
```diff
@@ -82,7 +82,21 @@ Section conventions:
...
+  `pkg/enforcement`. A backend that satisfied `Store` compiled and ran but, with
+  no signal that those two methods were also required, made every `sequenceBlock`
+  or `maxCalls` condition fail closed with an opaque "configured call counter does
+  not support …" deny. `pkg/callcounter` now declares the two extended contracts
+  next to `Store` — `HistoryStore` (`Store` + `Peek`) and `LimiterStore` (`Store`
+  + `IncrementIfBelow`) — with the conditions each one unlocks documented on the
+  interface, and both built-in backends are pinned to them by compile-time
+  assertion. A new backend that targets the extended interface gets a build error
+  for an omitted method instead of a silent runtime deny. No behavior change for
+  the shipped backends. (#236)
```
eunolabs, line 148: **This change clobbered the previous `Fixed` entry.** The new bullet was inserted by *replacing the heading line* of the existing entry rather than prepending a new bullet. The old entry's lead sentence — `- **The array-accepting conditions no longer silently drop non-string items.**` — is gone, and its body (now lines 100-109: `` `allowedExtensions`, `recipientDomain`, and `allowedTables` all accept an array argument ... `` through `...type-strictness of `allowedValues`.`) is now dangling under *this* `#236` bullet.

Net effect:
- The new `#236` bullet ends with `...No behavior change for the shipped backends. (#236)` and then immediately runs into unrelated array-condition prose — one malformed bullet.
- The "array-accepting conditions" fix has effectively lost its changelog entry.

Fix: keep the new bullet self-contained and restore the array-conditions entry as its own bullet, e.g.

```
- **A custom `callcounter` backend now learns at build time ... (#236)

- **The array-accepting conditions no longer silently drop non-string items.**
  `allowedExtensions`, `recipientDomain`, and `allowedTables` all accept an array
  ...
```

(Worth a quick `git diff` of this hunk — the deletion of the old heading line is the giveaway.)

---

pkg/enforcement/engine.go
```diff
@@ -63,6 +63,11 @@ type Clock interface {
 // session?" without itself counting as a call. Both built-in counter backends
 // (callcounter.InMemory, callcounter.Redis) implement it; the engine detects it
 // by type assertion and fails closed on a counter that does not.
+//
+// It is the consumer-side mirror of callcounter.HistoryStore (kept separate so
+// the engine need not import the producer package, the same decoupling
+// capability.CallCounter applies to callcounter.Store); a backend declares the
+// capability by implementing callcounter.HistoryStore.
```
eunolabs, line 70: The decoupling rationale (no production `enforcement → callcounter` import) is sound and well-argued. One gap it leaves, though: there are now **two independent declarations of the same contract** (`callcounter.HistoryStore` and this `CallHistory`, likewise `LimiterStore`/`RateLimiter`), and nothing pins them together. If one signature is ever edited and the other isn't, a backend pinned to `callcounter.HistoryStore` would silently stop satisfying `enforcement.CallHistory` and fall back to the exact opaque runtime deny this PR set out to prevent — just relocated.

A cheap, import-only-in-test guard closes that loop. In `enforcement_test.go` (which already imports `callcounter`):

```go
// The producer-side extended contracts must remain supersets of the
// consumer-side capability interfaces; if they drift, a backend pinned to
// callcounter.HistoryStore/LimiterStore would silently stop satisfying these.
var (
	_ enforcement.CallHistory = (callcounter.HistoryStore)(nil)
	_ enforcement.RateLimiter = (callcounter.LimiterStore)(nil)
)
```

Optional, but it makes the "structural identity" the doc relies on a build-enforced fact rather than a convention.

Minor nit while here: line 68 reads `...need not import the producer package, the same decoupling capability.CallCounter applies to callcounter.Store)` — the clause is missing a word, e.g. `...the same decoupling **that** capability.CallCounter applies to callcounter.Store)`.

---

eunolabs (commented):
## Review

The core change is correct and well-motivated. `HistoryStore`/`LimiterStore` mirror the real backend signatures exactly (`Peek`, `IncrementIfBelow`, `IncrementAndGet` all match `memory.go`/`redis.go`), the compile-time assertions in `callcounter_test.go` pin both built-ins to both interfaces, and the deliberate choice *not* to alias `enforcement.CallHistory` (to avoid a production `enforcement → callcounter` import, consistent with the existing `capability.CallCounter` ↔ `callcounter.Store` decoupling) is the right call and is clearly explained. Documenting the `maxCalls`/`IncrementIfBelow` sibling alongside the issue's `sequenceBlock`/`Peek` gap is a good completeness call. Verified locally: `go build ./...`, `go vet`, and `go test -race ./pkg/callcounter/... ./pkg/enforcement/...` all pass, gofmt clean.

Findings:

1. **`CHANGELOG.md` — must fix (defect).** The new `Fixed` entry was inserted by *replacing the heading line* of the existing "array-accepting conditions" entry, not by prepending a new bullet. The old entry lost its lead sentence and its body is now dangling under the `#236` bullet, producing one malformed bullet and erasing the array-conditions entry. Inline detail + suggested fix below.

2. **`engine.go` — suggestion.** The PR now keeps two independent declarations of the same contract (`callcounter.HistoryStore`/`LimiterStore` vs `enforcement.CallHistory`/`RateLimiter`) with nothing pinning them together; an edit to one could silently reintroduce the opaque runtime deny this PR fixes. A test-only compile-time assertion would make the structural identity build-enforced. Plus a one-word grammar nit. Inline below.

Once the CHANGELOG entry is restored this is good to go; (2) is optional hardening.

---

CHANGELOG.md
eunolabs: Fixed in 81ac187. Restored `- **The array-accepting conditions no longer silently drop non-string items.**` as its own bullet; the `#236` bullet is now self-contained. Verified with `git diff CHANGELOG.md`.

---

pkg/enforcement/engine.go
eunolabs: Done in 81ac187. Added the build-enforced drift guard in `enforcement_test.go` (`_ enforcement.CallHistory = (callcounter.HistoryStore)(nil)` and the `RateLimiter`/`LimiterStore` pair), and fixed the missing-word nit ("the same decoupling **that** capability.CallCounter applies to ..."). `go test -race ./pkg/enforcement/...` passes.

---

eunolabs:
Re-reviewed at `81ac187` — both review threads are correctly resolved and I have no further comments.

- **CHANGELOG.md**: the `#236` bullet is now self-contained and the `**The array-accepting conditions...**` entry is restored as its own bullet with its body intact — two well-formed entries.
- **enforcement_test.go**: the build-enforced drift guard (`_ enforcement.CallHistory = (callcounter.HistoryStore)(nil)` + the `RateLimiter`/`LimiterStore` pair) is in place, so producer/consumer interface drift now fails the build instead of silently reintroducing the opaque runtime deny. The `engine.go` grammar nit is fixed too.

Verified locally on this commit: `go build ./...`, `go test -race ./pkg/enforcement/...`, and gofmt all clean. No new issues — LGTM (can't formally Approve since this is the same account that opened the PR).

---

pkg/enforcement/engine.go
eunolabs: Heads-up: the drift-guard snippet I suggested trips `gocritic`'s `typeUnparen` and is failing the **Lint** check on `81ac187`:

```
pkg/enforcement/enforcement_test.go:35:30: typeUnparen: could simplify (callcounter.HistoryStore) to callcounter.HistoryStore (gocritic)
pkg/enforcement/enforcement_test.go:36:30: typeUnparen: could simplify (callcounter.LimiterStore) to callcounter.LimiterStore (gocritic)
```

That's on me — my suggested form had redundant parens. An interface-type conversion doesn't need them (unlike the `(*callcounter.InMemory)(nil)` pointer conversions just above, where the parens are required). Drop them on just those two lines:

```go
var (
	_ enforcement.CallHistory = callcounter.HistoryStore(nil)
	_ enforcement.RateLimiter = callcounter.LimiterStore(nil)
)
```

Same compile-time guarantee, and `golangci-lint run` goes clean. Leave the existing `(*callcounter.InMemory)(nil)` / `(*callcounter.Redis)(nil)` lines as-is.

---

PR #256, opened 2026-06-13
perf: split InMemory.Cleanup into scan + batched-delete phases

### LINKED ISSUE #237: perf: InMemory.Cleanup() holds exclusive mutex for entire map scan, blocking all callers
## Summary

`InMemory.Cleanup()` acquires `m.mu.Lock()` (exclusive) at the top and holds it for the entire iteration over `m.entries`. Every concurrent `IncrementAndGet` and `Peek` call blocks for the full duration of the cleanup pass.

```go
// pkg/callcounter/memory.go:200-222
func (m *InMemory) Cleanup() {
    m.mu.Lock()
    defer m.mu.Unlock()

    now := m.now()
    for key, e := range m.entries {  // entire scan under lock
        if len(e.timestamps) == 0 {
            delete(m.entries, key)
            continue
        }
        ...
        if e.timestamps[len(e.timestamps)-1].Before(now.Add(-staleAfter)) {
            delete(m.entries, key)
        }
    }
}
```

## Impact

In a deployment with many unique session×tool pairs (e.g., a gateway with many concurrent clients each calling many tools), `m.entries` can hold tens of thousands of keys. The default cleanup interval is 5 minutes (`DefaultCleanupInterval`). Each cleanup pass:

1. Takes O(n) time proportional to the number of live keys
2. Blocks every `IncrementAndGet` (which is on the hot enforcement path) for that duration

Under a sustained load of 10k req/s with 50k live keys, even a 1ms cleanup pass produces a visible latency spike on the enforcement hot path.

## Suggested fix

**Option A — Collect keys under lock, delete after releasing:**

```go
func (m *InMemory) Cleanup() {
    m.mu.Lock()
    now := m.now()
    var toDelete []string
    for key, e := range m.entries {
        if shouldDelete(e, now) {
        }
    }
    m.mu.Unlock()

    if len(toDelete) > 0 {
        m.mu.Lock()
        for _, key := range toDelete {
            delete(m.entries, key)
        }
        m.mu.Unlock()
    }
}
```

Note: re-checking staleness before the second lock avoids deleting entries that became live again between the two lock acquisitions (correctness). The first pass is still O(n) but the lock is held only for the scan (reads + append), not for the deletes.

**Option B — Batch cleanup:**  
Process entries in fixed-size batches, releasing and re-acquiring the lock between batches to allow callers through.

**Option C — Use `sync.Map`:**  
`sync.Map` is optimized for high read-to-write ratios and its `Range` does not prevent concurrent loads/stores. Suitable if the entry type can be made atomic.

## Affected code

- `pkg/callcounter/memory.go:200–222` (`Cleanup`)
- `pkg/callcounter/memory.go:171–187` (`StartCleanup`)

eunolabs:
Fixes #237

## Problem

`InMemory.Cleanup()` acquired the exclusive `m.mu` lock and held it for the **entire** `O(n)` iteration over `m.entries`, deleting stale entries inline. Every concurrent `IncrementAndGet` / `IncrementIfBelow` / `Peek` — all on the hot enforcement path — blocked for the full pass. With tens of thousands of live session×tool keys, the periodic (5-minute) cleanup produced a visible latency spike.

## Fix

Split `Cleanup()` into two phases (issue's Option A + Option B):

1. **Scan** — under the lock, walk `m.entries` once and collect the keys of stale entries into a slice. The lock is held only for the scan (map reads + slice appends), never for deletes.
2. **Batched delete** — delete the collected keys in bounded batches of `cleanupDeleteBatch` (256), re-acquiring the lock per batch so waiting callers get through *between* batches instead of blocking for the whole delete pass.

The staleness rule is extracted into a `shouldDeleteEntry` helper. **Dropped-entry behavior is unchanged**: an entry is reclaimed when it has no timestamps or its newest timestamp has aged past `windowSec * cleanupMarginFactor` (24h floor for entries with no recorded window) — still matching the Redis backend's `Expire` TTL.

### Correctness (fail-closed preserved)

Between the two phases — and between batches — a concurrent `IncrementAndGet` may refresh or recreate a key that phase 1 saw as stale. So phase 2 **re-checks staleness against a fresh `now` under the lock before each delete**. Without this, Cleanup could delete a counter that just went live again, silently resetting a `maxCalls` quota — a fail-open rate-limit bypass. The re-check makes the delete safe.

Note: the phase-1 scan is still `O(n)` under the lock (inherent to a single mutex + map). This change removes the deletes from the critical section and bounds the maximum *contiguous* hold; fully eliminating the scan would require sharding or `sync.Map` (the issue's Option C), a larger change left as a potential follow-up.

## Tests

- `TestShouldDeleteEntry` — table-driven (empty / recent / past-margin / long-window-within-margin / 24h-floor / most-recent-wins) pinning the staleness rule independent of locking.
- `TestInMemory_Cleanup_ConcurrentIncrementNotDropped` — `-race` regression: seeds every key stale (so phase 1 collects it), then hammers all keys with concurrent increments that race the deletes; asserts no live counter is dropped (final in-window count == increments performed) and no data race.
- Existing `Cleanup` tests (empty, stale, long-window-preservation, double-window-drop) pass unchanged.

`make check-fmt`, `make check-license`, `make lint` (0 issues), and `go test -race ./...` all green. Coverage on the changed functions is 100%.

---
_Generated by [Claude Code](https://claude.ai/code/session_0156pErDwpANQCUngFvfJXjM)_


pkg/callcounter/memory.go
```diff
@@ -261,37 +261,80 @@ func (m *InMemory) StartCleanup(ctx context.Context, interval time.Duration) {
...
+// refreshed or recreated a key that phase 1 saw as stale, and deleting it then
+// would drop a live counter, itself a fail-open quota reset. Re-checking the
+// current entry against a fresh now makes the delete safe.
 func (m *InMemory) Cleanup() {
 	m.mu.Lock()
-	defer m.mu.Unlock()
-
 	now := m.now()
+	var toDelete []string
 	for key, e := range m.entries {
```
eunolabs, line 466: Scope note (not blocking): phase 1 still holds `m.mu` for the full `O(n)` range over `m.entries`. That's the right call for a single mutex and it's a strict improvement — deletes are out of the critical section now — but it's worth being explicit that the win is partial: a concurrent `IncrementAndGet`/`Peek` still blocks for the entire phase-1 scan; only the *delete* cost is batched/yielded. With 50k mostly-live keys, the contiguous lock hold is still dominated by that scan. The full fix (sharding / `sync.Map`, the issue's Option C) remains future work, as the PR body already calls out — just flagging so the perf expectation is calibrated.

Minor, optional: in the mass-eviction case this targets, `toDelete` grows by repeated append-realloc *under the lock*. Pre-sizing the slice (e.g. `make([]string, 0, len(m.entries))`, or a fraction thereof) would move those reallocations off the critical section. Negligible in the common case, so purely optional.

---

pkg/callcounter/memory.go
```diff
@@ -261,37 +261,80 @@ func (m *InMemory) StartCleanup(ctx context.Context, interval time.Duration) {
...
+		if end > len(toDelete) {
+			end = len(toDelete)
 		}
-		// If the most recent timestamp is old enough, the whole entry is stale.
-		if e.timestamps[len(e.timestamps)-1].Before(now.Add(-staleAfter)) {
-			delete(m.entries, key)
+		m.mu.Lock()
+		now := m.now()
+		for _, key := range toDelete[start:end] {
+			if e, ok := m.entries[key]; ok && shouldDeleteEntry(e, now) {
```
eunolabs, line 486: This re-check is the load-bearing, fail-closed part of the change and it's correct. Deferring deletes opens a window where a concurrent `IncrementAndGet` revives a key phase 1 flagged; deleting it then would silently reset a live `maxCalls` quota — a fail-open rate-limit bypass. Re-fetching `m.entries[key]` under the lock covers both refresh-in-place (`IncrementAndGet` mutates the existing `*entry`, so the re-read sees the fresh newest timestamp) and delete-then-recreate (`ok` is false → skipped), and the fresh `now` only ever makes an entry *more* stale, so the check can flip to "keep" only on a genuine refresh. 

Flagging mainly so the `ok &&` guard and the second `shouldDeleteEntry(e, now)` are not "simplified" away in a later cleanup — both are required to preserve the fail-closed invariant.

---

eunolabs (commented):
## Review — `perf: split InMemory.Cleanup into scan + batched-delete phases`

Reviewed the diff plus surrounding context (`memory.go`, `redis.go`, `callcounter.go`, issue #237). This is a clean, well-reasoned change. No correctness concerns; comments below are scope/calibration notes, not blockers.

### What it does
Splits `InMemory.Cleanup()` into (1) a locked scan that collects stale keys and (2) a batched delete (256/lock acquisition) that re-checks staleness against a fresh `now` before each delete. The staleness predicate is extracted into `shouldDeleteEntry`.

### Correctness — fail-closed preserved ✅
- **Faithful extraction.** `shouldDeleteEntry` reproduces the old inline logic exactly (empty-timestamps → reclaim; else `lastTimestamp.Before(now - windowSec*cleanupMarginFactor)`, 24h floor when `windowSec == 0`). Dropped-entry behavior is unchanged.
- **The phase-2 re-check is the crux and is sound.** Deferring deletes creates a collect→delete gap; without re-checking, a concurrent `IncrementAndGet` that revived a key would have its live counter dropped — a fail-open `maxCalls` reset. Re-fetching `m.entries[key]` under the lock with a fresh `now` correctly handles both refresh-in-place and delete-then-recreate. (Inline note on the exact line so it isn't simplified away later.)
- **Backend equivalence holds.** Confirmed `cleanupMarginFactor` is the same factor `redis.go` applies via `Expire(key, windowSec*cleanupMarginFactor)`, so the in-memory and Redis reclamation windows stay aligned.

### Performance
Strict improvement — deletes leave the critical section and the per-acquisition hold is bounded to `O(batch)`. One calibration point (inline): **phase 1 still holds the lock for the full `O(n)` scan**, so concurrent callers still block for that scan; only the delete cost is yielded. The complete fix (sharding / `sync.Map`) is correctly scoped out as follow-up.

### Tests ✅
- `TestShouldDeleteEntry` — table-driven, covers empty / within-margin / past-margin / long-window / 24h-floor / most-recent-wins. Good isolation of the rule from the locking.
- `TestInMemory_Cleanup_ConcurrentIncrementNotDropped` — solid `-race` regression for the collect/delete window; the frozen clock makes the "all 50 increments land in-window → count must be 50" assertion precise, so a dropped live counter reliably shows up as a short count. Traced it through and the seeded 3h-old timestamp is correctly pruned on the first increment.

### Conventions ✅
License headers intact, no emojis, Conventional `perf:` type. No manifest/audit/config/schema/grammar change and no behavior change, so no docs update is required by CONTRIBUTING.

Nice work — the inline notes are optional polish (pre-sizing `toDelete`) and a guardrail comment on the re-check.

---

pkg/callcounter/memory.go
eunolabs: On the scope note — agreed, and it's called out in both the PR body and the `Cleanup` doc comment. The phase-1 `O(n)` scan under the lock is inherent to a single mutex + map; sharding / `sync.Map` (Option C) remains the follow-up for closing that.

On pre-sizing `toDelete`: I evaluated it and am keeping the incremental append. Pre-sizing to `len(m.entries)` would move an `O(n)` allocation **and zeroing** (strings are pointer-containing, so the backing array must be zeroed for GC safety) *under the lock* on every pass — including the common case where few/no keys are stale and `toDelete` currently stays `nil` and allocates nothing. That trades a guaranteed per-pass `O(n)` alloc+zero for avoiding `log(d)` reallocations only in the rare mass-eviction case — a net regression on the dominant path. The append already allocates proportional to the actual stale count, so I think `nil`-start is the better default here.

---

pkg/callcounter/memory.go
eunolabs: Agreed — both guards are required to preserve the fail-closed invariant. Added an inline comment at the re-check (b4b55a8) flagging the `ok &&` and the repeated `shouldDeleteEntry` as load-bearing, so neither is simplified away in a later cleanup.

---

pkg/callcounter/memory.go
eunolabs: Concede — you're right, and my nit under-weighed the common path. `make([]string, 0, len(m.entries))` forces an `O(n)` alloc **and** a GC-safe zeroing of the pointer-containing backing array on *every* pass, whereas the `nil`-start allocates nothing when no keys are stale (the dominant case) and only as much as the actual stale count otherwise. Trading that guaranteed per-pass cost for avoiding `log(d)` reallocations in the rare mass-eviction case is a net regression. `nil`-start is the correct default. Thanks for the careful evaluation.

---

PR #257, opened 2026-06-13
fix: deep-copy audit details map and obligations slice before drainer handoff

### LINKED ISSUE #229: bug: `record()` sends `details` map and `obligs` slice to the drainer goroutine by reference — concurrent callers can trigger a data race
## Summary

`record()` stores the caller-supplied `details` (`map[string]interface{}`) and `obligs` (`[]string`) directly into the `auditRecord` struct and sends it to the drainer channel:

```go
// audit.go:320-342
rec := auditRecord{
    ...
    Details:   details,    // map — reference type, not deep-copied
    Obligations: obligs,   // slice — reference type, not deep-copied
    ...
}
select {
case s.records <- rec:
```

The channel send copies the `auditRecord` value, but **not** the underlying map or slice storage. Both the caller and the drainer goroutine now hold references to the same backing data.

## Where this is triggered

Two call sites pass live data structures that outlive the `Record()` call:

**`http.go:774-777` and `stdio.go:516-519` — allowed tools/call in audit mode:**

```go
details = params.Arguments  // map[string]interface{} from the parsed JSON-RPC request
rt.sink.Record(ctx, ..., details, oblNames, rt.audit)
```

`params.Arguments` is the raw argument map parsed from the incoming request. The drainer goroutine may still be marshaling this map while the request-handling goroutine has returned (the variable is on the stack frame of the handler, but the underlying map storage lives on the heap and is kept alive by the record in the channel). If anything modifies the map's backing store after `Record()` returns — e.g., a future refactor that reuses the parsed params, or Go's runtime zeroing the map during GC — the race fires.

**All `denial.Details` paths (`http.go:735`, `stdio.go:471`, etc.):**

```go
rt.sink.Record(ctx, ..., denial.Details, nil, ...)
```

`denial.Details` is produced by the enforcement engine. If the engine or any middleware retains a reference to the same `DecisionResult` and modifies `Details` after the call, the drainer races on it. Currently the engine does not do this, but the API does not enforce it.

## Impact

- **Data race**: the Go race detector (`-race`) will flag this if the map is touched after `Record()` returns. In a `-race`-enabled CI run that exercises the audit-mode allow path, this could surface as a `DATA RACE` failure.
- **Corrupt audit records**: if the race is lost by the drainer, the marshaled JSON may capture a partially-modified map, producing a record with scrambled or missing details — without any error, signature failure, or chain break to signal the corruption.

## Suggested fix

Deep-copy `details` and `obligs` inside `record()` before the channel send:

```go
func copyDetails(m map[string]interface{}) map[string]interface{} {
    if m == nil {
        return nil
    }
    out := make(map[string]interface{}, len(m))
    for k, v := range m {
        out[k] = v  // shallow copy of values is sufficient if values are not mutable
    }
    return out
}

rec := auditRecord{
    ...
    Details:     copyDetails(details),
    Obligations: append([]string(nil), obligs...),
    ...
}
```

A shallow copy of `Details` values is sufficient because the values stored in the map (`string`, `float64`, `bool`, `[]interface{}`, nested `map[string]interface{}`) are either immutable scalars or containers that the enforcement engine does not modify after the call.

## Affected code

`cmd/eunox-mcp/audit.go`, `record()` function (lines 305–352).

eunolabs:
Fixes #229

## Problem

`record()` (`cmd/eunox-mcp/audit.go`) stored the caller-supplied `details` (`map[string]interface{}`) and `obligs` (`[]string`) directly into the `auditRecord` and sent it to the drainer goroutine over a channel. The channel send copies the `auditRecord` value but **not** the underlying map/slice storage, so the calling goroutine and the drainer goroutine end up holding references to the same backing data.

Two call sites pass live structures that outlive the `Record()` call:

- **Audit-mode allow path** (`http.go`, `stdio.go`): `details = params.Arguments` — the raw argument map parsed from the incoming JSON-RPC request.
- **Denial paths** (`denial.Details` across the handlers): the enforcement engine's `DecisionResult.Details`.

If anything touches the structure after `Record()` returns, the Go race detector flags a `DATA RACE`, and — worse — if the drainer loses the race it can marshal a half-mutated map into a signed record, producing corrupt audit details with no error, signature failure, or chain break to signal it.

## Fix

Clone `details` and `obligs` inside `record()` before constructing the record, so each record owns its own top-level storage:

- `Details: copyDetails(details)` — a new `copyDetails` helper does a shallow clone (returns `nil` for a `nil` map so the `omitempty` field stays absent). A shallow clone of the values is sufficient: they are immutable scalars (`string`, `float64`, `bool`) or containers that no caller mutates after the handoff — only the map's own backing store was at risk of concurrent reuse.
- `Obligations: append([]string(nil), obligs...)` — clones the slice (`nil` stays `nil`).

Cloning happens before signing, so the HMAC covers the snapshot and the record stays self-contained. No audit-record shape change, no new field — the on-disk output is identical.

## Tests (`cmd/eunox-mcp/audit_copy_test.go`)

- `TestAuditSink_RecordClonesDetailsAndObligations` — **deterministic** proof of the snapshot semantics. It gates the drainer with the `writeLine` seam: a warm-up record parks the drainer mid-write, the record under test is enqueued behind it (still unmarshaled), the caller's map/slice are mutated, and only then is the drainer released to marshal. Asserts the written record holds the original values and still verifies end-to-end. Confirmed to **fail without the fix** (`details[path] = MUTATED`, deleted key gone, injected key present) and pass with it.
- `TestAuditSink_RecordNoRaceOnDetailsMutation` — concurrency guard: 32 goroutines hand live maps/slices to `Record` and keep mutating them while the drainer serializes in the background. Passes cleanly under `-race`.

## Verification

- `go test -race -count=1 ./cmd/eunox-mcp/` — pass
- `make check-fmt check-license` — pass
- `make build` — pass

---
_Generated by [Claude Code](https://claude.ai/code/session_01GAFrD67DjK1eHqJV7E4CSy)_


cmd/eunox-mcp/audit.go
```diff
@@ -351,6 +361,25 @@ func (s *auditSink) record(ctx context.Context, upstream, policyVersion, policyS
...
 }
 
+// copyDetails returns a shallow clone of m so the audit record handed to the
+// drainer goroutine owns its own top-level map storage, never one shared with
+// the caller. A shallow clone suffices: the values are immutable scalars
+// (string, float64, bool) or containers (nested map[string]interface{} /
+// []interface{}) that no caller mutates after passing them to Record — only the
+// map's own backing store is at risk of concurrent reuse, and a fresh map breaks
+// that sharing. Returns nil for a nil map so the omitempty `details` field stays
+// absent rather than serializing as `{}`.
```
eunolabs: The shallow clone is correct for the top-level map, and the deterministic test proves it (I confirmed it fails without the fix and passes with it). One thing to flag, though: the clone is **shallow**, so the comment's framing — defending against "a future refactor [that] touches the structure after `Record` returns" and making the record "self-contained" — only holds one level deep.

The values (`nested map[string]interface{}` / `[]interface{}`) are still shared between the caller's structure and the queued record. The audit-mode allow path logs `params.Arguments`, which is arbitrary tool-call JSON and therefore routinely contains nested objects/arrays. So the same class of future refactor this PR is meant to guard against — one that mutates a *nested* field in place after `Record()` — would silently reintroduce both the data race and the corrupt-record outcome, and neither new test would catch it (both mutate only top-level keys).

I verified this empirically against this branch: with the fix in place, mutating a nested map value after `Record()` returns still writes the mutated value into the signed record:

```
written nested secret = MUTATED
LEAK: nested mutation reached the signed record
```

This matches what issue #229 explicitly accepted ("shallow copy ... is sufficient because the enforcement engine does not modify [it] after the call"), so it's not a regression — but the in-code comment promises more than the code delivers. Two options:

1. **Match the comment to the code:** state plainly that nested containers remain aliased and that "self-contained" relies on the invariant that no caller mutates a value after handoff (and ideally point at where that invariant is owned).
2. **Match the code to the comment:** do a real deep copy (recursive, or a marshal round-trip) so the record is genuinely self-contained. Audit is off the policy hot path and this helper only runs for non-nil `details` (denials + audit-mode allow), so the cost is acceptable.

I'd lean toward (2) given the whole point of the change is "self-contained record," but (1) is fine if you'd rather keep it shallow — just don't let the comment overstate the guarantee.

---

cmd/eunox-mcp/audit.go
```diff
@@ -336,9 +336,19 @@ func (s *auditSink) record(ctx context.Context, upstream, policyVersion, policyS
...
+		// storage it points at, so without these clones the calling goroutine and
+		// the drainer goroutine would share the same map and slice. The audit-mode
+		// allow path passes the live params.Arguments map straight through, and
+		// denial paths pass the engine's DecisionResult.Details; either becomes a
+		// data race the instant a caller (now or after a future refactor) touches
+		// the structure after Record returns, and can corrupt a record if the
+		// drainer marshals a half-mutated map. Cloning here makes the record
+		// self-contained.
+		Details:     copyDetails(details),
+		Obligations: append([]string(nil), obligs...),
```
eunolabs: Minor / optional: `go.mod` is on Go 1.25 and this package already imports `slices` in tests, so the standard library covers both of these and would let you drop the hand-rolled `copyDetails`:

```go
Details:     maps.Clone(details),
Obligations: slices.Clone(obligs),
```

`maps.Clone(nil)` and `slices.Clone(nil)` both return `nil`, preserving the `omitempty`-stays-absent behavior you rely on. Purely a readability suggestion — note it's orthogonal to the shallow-vs-deep question above (`maps.Clone` is also a shallow copy), so if you go with a deep copy this point is moot.

---

eunolabs (commented):
## Review: deep-copy audit details/obligs before drainer handoff

The diagnosis in #229 is sound and the fix resolves it. Cloning the top-level `details` map and `obligs` slice before the channel send severs the sharing between the request goroutine and the drainer, which is the data race / corrupt-record window the issue describes. Nice touch keeping the clone before signing so the HMAC covers the snapshot.

**Verified locally (`go test -race`):**
- `TestAuditSink_RecordClonesDetailsAndObligations` is genuinely deterministic (the `writeLine` warm-up gate is a clean way to park the drainer) and it **fails without the fix** / passes with it — I reverted the two clone lines and confirmed all four assertions fire. Good regression guard.
- `TestAuditSink_RecordNoRaceOnDetailsMutation` passes clean under `-race`.
- Hot-path cost is negligible: on the common non-audit allow path `details`/`obligs` are nil, and both clones short-circuit to nil.

**One substantive point** (inline on `copyDetails`): the clone is shallow, so the "self-contained record" guarantee only holds one level deep — nested maps/slices inside `params.Arguments` / `denial.Details` stay aliased with the caller. I confirmed against this branch that a nested mutation after `Record()` still leaks into the signed record. It's consistent with what #229 accepted, so not a regression, but the in-code comment currently promises more than a shallow copy delivers. Suggest either softening the comment to name the relied-upon invariant, or doing a real deep copy (audit is off the hot path). Details and a repro inline.

**Minor:** `maps.Clone` / `slices.Clone` (Go 1.25) could replace the hand-rolled helper — inline, optional.

No audit-record shape change, fail-closed semantics preserved, license headers present. Once the nested-copy comment/behavior question is settled this looks good to merge. CI is still queued at time of review.

---

cmd/eunox-mcp/audit.go
eunolabs: Went with option 2. `copyDetails` is now a recursive deep copy: `copyDetailValue` recurses through nested `map[string]interface{}` / `[]interface{}`, so the queued record shares no storage with the caller at any level `params.Arguments` can nest. Scalars are copied by value, and the only non-scalars the engine puts in denial details are `[]string` slices borrowed from the already-loaded, immutable manifest (allowedCIDRs/allowedOperations/...), which are never mutated after a decision — the doc comment now names that invariant explicitly rather than claiming a blanket "self-contained".

Extended the deterministic test to mutate nested object **and** array values (including a doubly-nested array) after `Record()` returns. It reproduces your leak on a shallow copy and passes on the deep one — verified by reverting `copyDetailValue` to a passthrough:

```
details[nested][secret] = map[secret:MUTATED] ...; want keep
details[list][0] = [MUTATED [MUTATED]] ...; want a
```

Fixed in 876069d.

---

cmd/eunox-mcp/audit.go
eunolabs: Adopted `slices.Clone(obligs)` for the obligations clone. Kept a custom helper for `details` since it's now a deep copy — `maps.Clone` is shallow, as you noted, so it's moot there. In 876069d.

---

cmd/eunox-mcp/audit.go
eunolabs: Re-reviewed 876069d — this fully resolves the point. `copyDetailValue` recursing through `map[string]interface{}` / `[]interface{}` means the queued record shares no storage with the caller at any depth, and documenting the `[]string`-from-immutable-manifest invariant for the passthrough case is exactly the right call rather than over-copying.

I verified locally against the pushed commit: the suite passes under `-race`, and reverting `copyDetailValue` to a passthrough makes the extended deterministic test fail on precisely the nested object and nested array assertions (`details[list][0] = [MUTATED [MUTATED]]`) — so the deep recursion is doing real work and is properly guarded. `slices.Clone(obligs)` on the obligations side is good too. No further comments from me.

---

PR #258, opened 2026-06-13
test: lock in dotfile handling for allowedExtensions (#243)

### LINKED ISSUE #243: bug: allowedExtensions denies dotfiles (`.env`, `.gitignore`) as "file has no extension" — filepath.Ext returns "" for dotfiles
## Summary

`handleAllowedExtensions` uses `filepath.Ext(filePath)` to extract the extension:

```go
// handlers.go:421-432
ext := strings.ToLower(filepath.Ext(filePath))
if ext == "" {
    return &ConditionError{
        Message: "file has no extension",
            "filePath":          filePath,
            "allowedExtensions": ae.Extensions,
        },
    }
}
```

Go's `filepath.Ext` returns `""` for filenames that start with a dot but have no further dot (the "dotfile" convention on Unix):

```go
filepath.Ext(".env")        // ""  ← treated as "no extension"
filepath.Ext(".gitignore")  // ""  ← treated as "no extension"
filepath.Ext(".bashrc")     // ""  ← treated as "no extension"
filepath.Ext(".htaccess")   // ""  ← treated as "no extension"
```

As a result, a manifest that allows `.env` files:

```yaml
conditions:
  - type: allowedExtensions
    argument: path
    extensions: [".env"]
```

will **deny** all requests for `.env` files with the message `"file has no extension"`, because `filepath.Ext(".env") == ""`. The policy author's intent (allow `.env`) is not achievable with the current implementation.

## Reproduction

```go
    Arguments: map[string]interface{}{"path": "/app/.env"},
}
cond := &capability.AllowedExtensionsCondition{
    Argument:   "path",
    Extensions: []string{".env"},
}
// Result: CONDITION_FAILED "file has no extension"
// Expected: allow
```

## Impact

Operators who want to allow or restrict access to dotfiles (`.env`, `.gitignore`, `.htaccess`, `.npmrc`, etc.) cannot express that policy with `allowedExtensions`. Any `allowedExtensions` condition silently denies ALL dotfiles regardless of what is in the `extensions` list — a fail-closed outcome, but not the one the operator intended.

This is particularly relevant for tools that manage configuration files (`.env` secrets, `.gitignore` templates, etc.) because those files are often exactly what a policy author wants to restrict.

## Suggested fix

Extend the extension extraction to handle dotfiles by treating the entire filename as the "extension" for dotfile-style names (leading dot, no further dot):

```go
ext := strings.ToLower(filepath.Ext(filePath))
if ext == "" && strings.HasPrefix(base, ".") {
    // Dotfile: the whole name is the extension (e.g. ".env", ".gitignore")
    ext = strings.ToLower(base)
}
```

Then the allowlist check `allowed[ext[1:]]` becomes `allowed[ext]` for dotfiles (or adjust `allowed` normalization to handle both cases):

```go
// Normalize the allowlist: strip leading dot for regular extensions,
// keep full name for dotfiles.
for _, e := range ae.Extensions {
    normalized := strings.ToLower(strings.TrimPrefix(e, "."))
    allowed[normalized] = true
}
// Extension is the full filename for dotfiles, or the .ext portion stripped of ".".
extKey := strings.TrimPrefix(ext, ".")
if !allowed[extKey] {
    return &ConditionError{...}
}
```

This requires updating the manifest documentation and possibly `validateAllowedExtensions` to clarify that dotfile names (`.env`) are valid extension entries.

## Affected code

`pkg/enforcement/handlers.go` — `handleAllowedExtensions` (lines 421–448).

eunolabs:
Re: #243.

## TL;DR

The bug in #243 **does not reproduce** — its premise is incorrect for Go. This PR adds a regression test that proves and locks in the correct behavior, plus a docs clarification. No production code changes.

## Why the issue doesn't reproduce

#243 claims `filepath.Ext(".env") == ""`, causing `allowedExtensions` to deny dotfiles as `file has no extension`. That is **Node.js's `path.extname` behavior, not Go's**. Go's `filepath.Ext` returns the suffix beginning at the **final** dot in the last path element — and for a dotfile, the leading dot *is* the final dot:

```
filepath.Ext(".env")        = ".env"        // issue claims ""
filepath.Ext(".gitignore")  = ".gitignore"  // issue claims ""
filepath.Ext("/app/.env")   = ".env"
filepath.Ext("noext")       = ""            // genuinely no extension
```

So dotfiles flow through the normal allowlist check. Running the issue's exact reproduction through the real handler:

| path | extensions | result |
|------|-----------|--------|
| `/app/.env` | `[".env"]` | **allow** (issue expects allow; claims it denies) |
| `/app/.gitignore` | `[".gitignore"]` | **allow** |
| `/app/.bashrc` | `[".bashrc"]` | **allow** |
| `/app/.env` | `["env"]` (no dot) | **allow** |
| `/app/.env.local` | `[".local"]` | **allow** |
| `/app/.env` | `[".txt"]` | **deny** `file extension ".env" is not allowed` (correct mismatch — not "file has no extension") |

The suggested fix would rewrite currently-correct extraction logic and risk regressions, so it's intentionally not applied.

## What this PR does

- **`pkg/enforcement/coverage_test.go`** — adds `TestAllowedExtensions_Dotfiles`, a regression guard covering: dotted/bare allowlist entries (`.env` ≡ `env`), `.gitignore`/`.bashrc`, a dotfile among regular entries, multi-dot names matching their final segment (`.env.local` → `.local`), and a dotfile outside the allowlist denied as a normal mismatch (asserting the message is `not allowed`, **not** `no extension`). This locks the behavior so a future refactor of the extraction logic can't silently regress it.
- **`docs/capability-manifest-guide.md`** — documents that dotfile names are valid `allowedExtensions` entries and how the final-dot rule applies.

### Verification

- `go test -race -count=1 -run TestAllowedExtensions ./pkg/enforcement/` — pass
- `go vet ./pkg/enforcement/`, `go build ./...`, `gofmt -l` — clean

I'll close #243 as works-as-intended with a pointer to this PR.

https://claude.ai/code/session_013zGjUEBrRmwWzHhDTN6of3

---
_Generated by [Claude Code](https://claude.ai/code/session_013zGjUEBrRmwWzHhDTN6of3)_


docs/capability-manifest-guide.md
```diff
@@ -879,6 +879,14 @@ conditions apply the same rule: `recipientDomain` to its recipient list, and
       extensions: [".csv", ".json", ".txt"]
 ```
 
+Dotfile names work as ordinary entries: `extensions: [".env"]` matches `/app/.env`,
+and `[".gitignore"]` matches `.gitignore`. The extension is taken from the final
+dot in the file name, so a leading dot with no further dot (`.env`, `.bashrc`) is
+itself the extension, while a multi-dot name matches its last segment (`.env.local`
+matches `.local`). A file with no dot at all (`Makefile`, `id_rsa`) has no
+extension and is denied — it can never satisfy an extension allowlist. Entries may
+be written with or without the leading dot (`".env"` and `"env"` are equivalent).
```
eunolabs, line 888: Optional (non-blocking) — since `.env` ≡ `env` after normalization, the equivalence is bidirectional and worth one more sentence of warning here. Because the handler strips the leading dot when normalizing both the allowlist entry and the extracted extension (`allowed[strings.TrimPrefix(strings.ToLower(e), ".")]` vs `ext[1:]`), an entry can't distinguish a bare dotfile from a same-suffixed regular file:

- `extensions: [".env"]` (or `["env"]`) allows **both** the dotfile `/app/.env` **and** any regular `/app/config.env`.

That's a security-relevant subtlety given the issue framed dotfiles (`.env`, `.npmrc`) as exactly the sensitive files an author wants to gate: allow-listing `*.env` config files silently also permits the bare `.env` secret, and vice versa. A reader scanning these docs to lock down secrets would benefit from that being called out explicitly. The new `allow/dotfile matches bare entry` test case already encodes this behavior — just not the caveat.

---

eunolabs (commented):
## Review: solid — central claim verified

I checked the premise of this PR against the actual Go behavior and the handler source, and it holds up.

**The bug in #243 does not reproduce — confirmed.** The issue describes Node's `path.extname`, not Go's `filepath.Ext`. Ran it directly:

```
filepath.Ext(".env")          = ".env"
filepath.Ext(".gitignore")    = ".gitignore"
filepath.Ext("/app/.env")     = ".env"
filepath.Ext("/app/.env.local") = ".local"
filepath.Ext("noext")         = ""        // genuinely no extension
filepath.Ext("Makefile")      = ""
```

Go's `filepath.Ext` scans back to the final dot in the last path element, and for a dotfile the leading dot *is* that final dot — so dotfiles flow through the normal allowlist check in `handleAllowedExtensions` (`pkg/enforcement/handlers.go`) and are never hit by the `"file has no extension"` branch. Declining to apply #243's suggested "fix" is the right call; it would rewrite correct extraction logic.

**Test is a good regression guard.** `go test -race -count=1 -run TestAllowedExtensions_Dotfiles ./pkg/enforcement/` passes. Cases cover dotted/bare-entry equivalence, `.gitignore`/`.bashrc`, a dotfile among regular entries, the multi-dot final-segment rule, and a deny case that asserts the message is `not allowed` and **not** `no extension` — exactly the misclassification #243 feared. Nicely scoped, no production change.

**Note:** the genuine "no dot at all" denial that the docs reference (`Makefile`, `id_rsa`) isn't re-asserted by the new test, but it's already locked by `TestAllowedExtensions_ErrorBranches` ("no extension" → `/data/noext`), so there's no coverage gap.

One optional, non-blocking docs suggestion inline (the `.env` ≡ `env` collision cuts both ways and is worth calling out for secret-gating). Nothing here blocks merge once CI is green.

---

docs/capability-manifest-guide.md
eunolabs: Good catch — added in a7094fe. The docs now spell out that the leading dot is stripped on both sides, so an entry can't distinguish a bare dotfile from a same-suffixed regular file: `[".env"]` allows both `/app/.env` and `/app/config.env`, bidirectionally. Called out the secret-gating implication explicitly and pointed readers at constraining the path (e.g. an `allowedValues` glob on the same argument) when they need to admit one but not the other.

---

eunolabs: Verified the cross-reference holds: `handleAllowedValues` does glob via `path.Match` (string pattern + string arg), so `allowedValues: ["/app/config.env"]` on the path argument admits the regular file without the bare `.env` — the doc pointer is accurate, not hand-wavy. One caveat readers should keep in mind: `path.Match` treats `/` as a separator and `*` won't cross it, so a glob like `/app/*.env` matches `/app/config.env` but not `/app/sub/config.env` (use a stricter exact value or a per-segment pattern if that matters). Resolved on my end — re-review of a7094fe is clean, docs-only, no behavior change.

---

eunolabs: Folded your `path.Match` caveat into the docs in be52368: the pointer now gives a concrete example (`allowedValues: ["/app/config.env"]`) and spells out the boundary — `*` doesn't cross `/`, so `/app/*.env` matches `/app/config.env` but not `/app/sub/config.env`; use an exact value or a per-segment pattern when a single `*` is too narrow. Thanks for verifying the cross-reference.

---

PR #259, opened 2026-06-13
sec: fail closed when sequenceBlock afterTools entries strip to empty

### LINKED ISSUE #234: security: sequenceBlock silently ignores afterTools entries that reduce to empty string after prefix-stripping
## Summary

`handleSequenceBlock` silently skips `afterTools` entries whose value reduces to an empty string after `stripEnginePrefix`. If **all** entries in `afterTools` strip to empty, the block never fires and the condition passes — fail-open for a declared security rule.

## Code path

```go
// pkg/enforcement/handlers.go:760-764
for _, prior := range sb.AfterTools {
    priorTool := stripEnginePrefix(prior)
    if priorTool == "" {
        continue   // ← silently skipped
    }
    // ... check history ...
}
```

The up-front guard at line 716 only catches a completely empty `afterTools` list:

```go
if len(sb.AfterTools) == 0 {
    return &ConditionError{...}
}
```

It does not catch the case where all entries are non-empty strings that strip to `""` (e.g., `["tool:", "resource:"]`).

## Reproduction

Manifest:
```yaml
conditions:
  - type: sequenceBlock
    afterTools:
      - "tool:"       # strips to ""
      - "resource:"   # strips to ""
```

Expected: condition authoring error → deny at manifest load or fail-closed denial at runtime.  
Actual: both entries are skipped, loop exits without checking history, `nil` is returned → **allow**.

## Why this matters

A manifest author may intend "block this tool after any of these other tools have run" and accidentally write `tool:` (missing name) or similar. The current behavior silently treats the condition as satisfied instead of surfacing the authoring mistake.

From CLAUDE.md:
> **Fail closed.** On any ambiguity … deny.

An `afterTools` list whose every entry strips to empty is an unambiguous authoring error and should be treated as a condition failure.

## Suggested fix

**Option A — Validate at manifest load time:**  
Add a validation pass in `pkg/capability/validate.go` that rejects `sequenceBlock` conditions where any `afterTools` entry strips to empty.

**Option B — Treat as a condition failure at evaluation time:**  
Accumulate empty-after-strip entries and return a `CONDITION_FAILED` error if any are found, so the failure surfaces clearly in the audit log.

## Affected code

- `pkg/enforcement/handlers.go:716–785` (`handleSequenceBlock`)
- `pkg/capability/validate.go` (missing validation)

eunolabs:
## Summary

Fixes #234.

`handleSequenceBlock` stripped the namespace prefix from each `afterTools` entry and **silently skipped** any that reduced to an empty string. A list whose every entry strips to empty (e.g. `["tool:", "resource:"]`, or an accidental bare `tool:`) therefore checked nothing, the loop fell through, and the condition returned **allow** — fail-open for a declared security rule. The existing up-front guard only caught a *completely empty* `afterTools` list, not entries that are non-empty strings reducing to `""` after stripping.

Per the `Fail closed` invariant, an `afterTools` entry that names no tool is an unambiguous authoring error and must surface as a denial, not be silently honoured.

## Fix (both layers — defense in depth)

`pkg/enforcement` is an importable engine that can be used without the cmd-layer manifest validation, so the fix lands in both:

- **Runtime (`pkg/enforcement/handlers.go`)** — `handleSequenceBlock` now denies fail-closed with `CONDITION_FAILED` the moment any `afterTools` entry names no tool after prefix-stripping, reporting the offending entries in `Details.emptyAfterTools`. The check runs *before* the dry-run short-circuit (mirroring the existing empty-list guard) so the preflight `/validate` path catches it too. The now-unreachable empty-skip `continue` in the history loop is removed.
- **Load time (`cmd/eunox-mcp/manifest.go`)** — new `validateSequenceBlock` rejects the same condition at manifest load (`validate` subcommand and proxy startup), so the mistake is caught before the first request rather than shipping as a silently inert rule. A new exported `enforcement.StripEnginePrefix` (following the existing `MatchesResource`/`ResourceSpecificity` exported-form pattern) gives the validator the engine's *exact* stripping semantics, so the two checks agree on what counts as empty.

## Tests

- `pkg/enforcement/sequence_block_test.go`: all-entries-empty, one-entry-empty (mixed with a valid entry), and empty-entry-under-dry-run cases — all assert fail-closed `CONDITION_FAILED`.
- `cmd/eunox-mcp/manifest_validation_test.go`: valid load + table-driven rejection for empty list, bare `tool:`, bare `resource:`, and empty string.

## Docs

`docs/capability-manifest-guide.md`, the `SequenceBlockCondition` doc comment, and `CHANGELOG.md` (Security) updated. No manifest grammar change — `afterTools` is still `[]string`; only clearly-invalid values are now rejected.

## Verification

`go build ./...`, `go vet`, full `go test -race -count=1 ./...`, `make check-fmt`, and `make check-license` all pass locally.

https://claude.ai/code/session_01SjTC5yAMJBi1JrwzeXo6FJ

---
_Generated by [Claude Code](https://claude.ai/code/session_01SjTC5yAMJBi1JrwzeXo6FJ)_


pkg/enforcement/handlers.go
```diff
@@ -764,6 +764,31 @@ func (e *Engine) handleSequenceBlock(ctx context.Context, cond capability.Condit
...
+	// Fail closed when an entry names no tool once its namespace prefix is
+	// stripped (e.g. "" or a bare "tool:"). Such an entry can never match
+	// session history, so silently skipping it would let a list whose every
+	// entry strips to empty fall through to allow — fail-open for a declared
+	// security rule (issue #234). Surface it as a denial, and do so before the
+	// dry-run short-circuit (like the empty-list guard above) so the preflight
+	// /validate path catches the authoring mistake too.
+	var emptyAfterTools []string
+	for _, prior := range sb.AfterTools {
+		if stripEnginePrefix(prior) == "" {
```
eunolabs: Minor / optional (not blocking): the guard catches entries that strip to the *empty* string, but a whitespace-only entry slips through and is the same class of authoring error — it names no real tool yet reads as "satisfied".

```
"tool:"   -> ""    flagged
"tool: "  -> " "   NOT flagged   (matches session key for a tool literally named " ")
"  "      -> "  "  NOT flagged
"tool:\t" -> "\t"  NOT flagged
```

Per the PR's own framing ("an `afterTools` entry that names no tool is an unambiguous authoring error and must surface as a denial"), a whitespace-only name qualifies. Consider `strings.TrimSpace(stripEnginePrefix(prior)) == ""` here and in `validateSequenceBlock` so the two stay in lockstep. It's an unlikely typo and arguably outside issue #234's exact scope, so feel free to defer — but since this is a fail-closed security check, tightening it costs one call and removes the remaining "names nothing but passes" case.

---

eunolabs (commented):
Reviewed the fix for #234 — silent fail-open in `sequenceBlock` when `afterTools` entries strip to empty. This is a clean, correct fix and I'm comfortable with it. Build + the new tests pass locally (`go build ./...`, `go test -race` on `pkg/enforcement` and `cmd/eunox-mcp`).

**What's good**
- **Correct fail-closed semantics.** The runtime guard denies with `CONDITION_FAILED` the moment any entry strips to empty, and it's placed *before* the dry-run short-circuit (mirroring the existing empty-list guard) so the preflight `/validate` path catches it too. Removing the now-unreachable `if priorTool == "" { continue }` is the right cleanup, and the "guaranteed non-empty by the guard above" comment documents the invariant.
- **Defense in depth is justified, not redundant.** Validating at manifest load *and* in the engine matters because `pkg/enforcement` is importable standalone and JWT-issued capability claims bypass the cmd-layer manifest validation — the runtime guard is the only thing protecting those paths. Sharing `enforcement.StripEnginePrefix` between both layers (following the existing `MatchesResource`/`ResourceSpecificity` exported-form pattern) keeps load-time and runtime agreeing on what counts as empty.
- **No grammar/back-compat baggage.** `afterTools` stays `[]string`; only clearly-invalid values are rejected. Pre-1.0, turning a silently-inert rule into a hard denial is the right call, and it's documented in CHANGELOG under Security.
- **Tests and docs match the conventions.** All-empty, one-empty (mixed with a valid entry), and empty-under-dry-run at the engine layer; valid-load + table-driven rejection (empty list, bare `tool:`, bare `resource:`, empty string) at the manifest layer. Guide, doc comment, and CHANGELOG all updated. No manifest JSON schema exists (only the gateway-config schema), so nothing to update there.
- `Details.emptyAfterTools` is a sub-field of the condition error, consistent with how other handlers populate `Details` — no top-level audit-record field added, so no threat-model change needed.

**One minor, optional point** (inline on `handlers.go`): the guard checks for the empty string, but a whitespace-only entry (`"tool: "`, `"  "`) strips to non-empty whitespace and passes both guards while still naming no real tool. Same class of authoring error; a `strings.TrimSpace` would close it. Unlikely typo and arguably outside #234's scope — defer if you prefer.

Nothing here is blocking. Nice, tightly-scoped security fix.

---

pkg/enforcement/handlers.go
eunolabs: Good catch — agreed it's the same "names no tool" class, so I tightened both layers rather than defer. The engine guard and `validateSequenceBlock` now use `strings.TrimSpace(stripEnginePrefix(entry)) == ""` (in lockstep, as suggested), and the runtime history lookup keys on the same trimmed name so a whitespace-padded *real* name (`" read_credentials "`) still matches the recorded, whitespace-free tool name. Added engine + manifest-load tests for `"tool: "` / `"  "` and the padded-name-still-matches case. Done in 73ff392.

---

PR #260, opened 2026-06-13
sec: bound in-memory call-counter key set to prevent heap exhaustion (#241)

### LINKED ISSUE #241: bug: InMemory counter map is unbounded between Cleanup cycles — sustained creation of unique session keys exhausts heap
## Summary

`InMemory.entries` is an unbounded `map[string]*entry`. `Cleanup` removes stale entries, but it runs only periodically (default every 5 minutes via `DefaultCleanupInterval`). Between Cleanup cycles, the map grows monotonically with every new unique key.

The two session-scoped key prefixes are:
- `maxcalls:{sessionID}:{toolName}` — one entry per (session, tool) pair with a `maxCalls` condition
- `seq:{sessionID}:{toolName}` — one entry per (session, tool) pair recorded by `recordSessionCall` for `sequenceBlock`

In a deployment where sessions are short-lived (e.g. HTTP connections with Mcp-Session-Id UUIDs), many unique sessions can accumulate between Cleanup cycles.

## Worst-case scenario

Consider an HTTP deployment (or a test harness) that creates a new session for every request:
- Sessions are UUIDs (128 bits) — all unique
- `maxCalls` condition on a tool called once per session
- Cleanup fires every 5 minutes

In 5 minutes at 1000 req/s:
- 300 000 new session keys created
- Each `entry` = pointer + `[]time.Time` (24 bytes header) + `time.Time` (24 bytes for 1 call) + `int` = ~72 bytes
- Map overhead (bucket array, hash metadata) ≈ 2× entry size

Peak heap growth before cleanup: ~300 000 × 144 bytes ≈ **43 MB** for the entry objects plus map overhead. At 10 000 req/s, this is **430 MB per 5-minute window**.

After Cleanup fires, all 1-call entries with windowSec=60 are stale (they're >2× window old) and freed. But the **peak** occurs at the Cleanup boundary.

## Attack scenario

An attacker who can generate arbitrary session IDs (or cause the proxy to create many sessions) can exhaust heap between Cleanup cycles, causing OOM. Even without a deliberate attack, a high-throughput deployment with short sessions hits this organically.

## Comparison with Redis backend

The Redis backend stores each key in Redis (external memory) with a TTL, so the Go process heap is unaffected by key count. The InMemory backend is intended for single-replica or testing use, but the heap unboundedness should be documented or bounded.

## Suggested fix

**Option A — add a max-key limit to `InMemory`:**
```go
type InMemory struct {
    mu       sync.Mutex
    entries  map[string]*entry
    now      func() time.Time
    maxKeys  int // 0 = unlimited
}

func (m *InMemory) IncrementAndGet(ctx context.Context, key string, windowSec int) (int64, error) {
    ...
    m.mu.Lock()
    defer m.mu.Unlock()

    if m.maxKeys > 0 && len(m.entries) >= m.maxKeys && m.entries[key] == nil {
        return 0, fmt.Errorf("callcounter: key limit reached (%d)", m.maxKeys)
    }
    ...
}
```
A `WithMaxKeys(n int)` option lets operators set the limit. Exceeding it fails closed (deny, consistent with the fail-closed invariant).

**Option B — shorter default Cleanup interval:**
Reduce `DefaultCleanupInterval` from 5 minutes to 30 seconds. This limits the peak accumulation at the cost of more frequent lock contention. Not sufficient for extreme rates but reduces the blast radius.

**Option C — document as a known limitation:**
Add a note to the `NewInMemory` godoc warning that it is not suitable for high-session-rate deployments, and recommend the Redis backend.

Option A + C together is the most complete fix: bound the key count and document the limitation.

## Affected code

- `pkg/callcounter/memory.go` — `InMemory` struct and `IncrementAndGet` (lines 25–101)
- `pkg/callcounter/callcounter.go` — `DefaultCleanupInterval` constant

eunolabs:
## Summary

Fixes #241.

`callcounter.InMemory` held an unbounded `map[string]*entry` — one entry per distinct `(session, tool)` pair, for both `maxcalls:` keys (maxCalls) and `seq:` markers (sequenceBlock). `Cleanup` reclaims stale entries but runs only periodically (`DefaultCleanupInterval`, 5 minutes). A deployment that mints a fresh session per request (short-lived HTTP connections with UUID `Mcp-Session-Id` values) accumulates one entry per unique session between cleanup cycles, so the map — and the process heap — peaks at the cleanup boundary. A client able to drive unique session IDs can push that peak high enough to exhaust the heap and OOM the proxy.

This implements the issue's recommended **Option A + C**: bound the key count (fail closed) and document the limitation.

## Changes

- **`pkg/callcounter/memory.go`**
  - New `WithMaxKeys(n int)` option + `maxKeys` field. Once the map holds `n` entries, recording a call under a **new** key is refused via `admitNewKey()` with `callcounter: key limit reached (n)`. The check is gated on the new-key path (a key already present grows the map by nothing and is always admitted) and runs under the existing lock in both `IncrementAndGet` and `IncrementIfBelow`.
  - **Fail-closed semantics:** the error propagates so `maxCalls` denies the call (`handleMaxCalls` already maps any counter error to a `CONDITION_FAILED` denial); the best-effort `sequenceBlock` recorder swallows it and skips the marker. A key freed by `Cleanup` frees a slot.
  - `WithMaxKeys(0)` (the default) stays **unbounded**, preserving today's behavior for library callers and tests.
  - `NewInMemory` godoc now documents the footprint and points high-session-rate deployments at Redis or `WithMaxKeys` (Option C).
- **`cmd/eunox-mcp/main.go`** — new `--max-call-counter-keys` flag (default `1,000,000`; `0` disables) applied to the in-memory counter, so the binary is protected out of the box. Ignored when `--redis-addr` is set, since the Redis backend keeps counter state off the Go heap with per-key TTLs. The default is a memory backstop (~few hundred MB worst case), not a per-session quota — far above any realistic live-session count a single in-memory replica serves within one cleanup window.
- **`docs/threat-model-mcp.md`** — §3.6 "In-memory store footprint" paragraph + change-log row 1.10.

## What I deliberately did *not* do

Option B (shortening `DefaultCleanupInterval`) — the issue notes it is "not sufficient" and only reduces blast radius, so the interval is left at 5 minutes.

## Tests

- `pkg/callcounter/maxkeys_internal_test.go` (new, in-package so the map size is observable): the bound refuses a new key without growing the map and surfaces the right error on both `IncrementAndGet` and `IncrementIfBelow`; existing keys keep counting; the default/`0`/negative cases stay unbounded; a slot freed by `Cleanup` is reusable; `Peek` is not gated (no entry creation, no error at capacity).
- New functions at 100% coverage. Full suite green: `go test -race -count=1 ./...`, `make check-fmt`, `make check-license`, `golangci-lint` (0 issues).

https://claude.ai/code/session_01QHuCsoiNKJPAnYvh8JmzUY

---
_Generated by [Claude Code](https://claude.ai/code/session_01QHuCsoiNKJPAnYvh8JmzUY)_


docs/threat-model-mcp.md
```diff
@@ -275,6 +275,8 @@ The audit key is generated once at first startup using `crypto/rand` (Go's CSPRN
...
...olds one map entry per distinct `(session, tool)` pair — for both `maxcalls:` keys and the `seq:` markers `sequenceBlock` records — reclaimed only when the periodic cleanup fires (`DefaultCleanupInterval`, 5 minutes). A deployment that mints a fresh session per request (short-lived HTTP connections with UUID `Mcp-Session-Id` values are the common case) accumulates one entry per unique session between cleanup cycles, so the map — and the process heap — peaks at the cleanup boundary; a client able to drive unique session IDs can push that peak high enough to exhaust the heap and OOM the process (issue #241). The proxy bounds this by default: `--max-call-counter-keys` (1,000,000; set 0 to disable) caps the in-memory key count, and a call under a *new* key past the ceiling fails closed — `maxCalls` denies the call, the best-effort `sequenceBlock` marker is skipped — rather than growing the map. A key freed by cleanup frees a slot. The bound is a memory backstop, not a per-session quota; for sustained high-session-rate deployments prefer the Redis backend (`--redis-addr`), which keeps counter state in Redis with a per-key TTL rather than on the Go heap and so is unaffected by key count.
```
eunolabs: The bound is uniformly described as "fails closed" here (and in the `WithMaxKeys` / `defaultMaxCallCounterKeys` godoc), but the two key namespaces behave asymmetrically at saturation and only one is actually fail-closed:

- **`maxcalls:`** — a new key past the ceiling returns an error, `handleMaxCalls` maps it to `CONDITION_FAILED` and denies (`handlers.go:310`). Fail-closed. ✅
- **`seq:`** — `recordSessionCall` swallows that error and skips the antecedent marker (`engine.go:227`, `_, _ = ...`). The next `sequenceBlock` check then `Peek`s an absent marker, reads "A never ran", and does **not** block the dependent tool. That's fail-**open**.

Concretely: a `sequenceBlock(after: A)` guard on tool B can be bypassed under sustained saturation whenever A is not itself `maxCalls`-gated (if it were, A's own new-key call would fail closed and never run). Flood the map to capacity, and the victim session's A call records no `seq:` marker, so B is admitted.

This is narrow — it requires holding ~1M live keys, i.e. the very flood this PR defends against — and it's consistent with the pre-existing best-effort contract on `recordSessionCall`, so I don't think it blocks the PR. But the "fails closed" framing papers over it. Worth one explicit sentence in §3.6 noting that the bound degrades `sequenceBlock` to **fail-open** at saturation (the skipped marker is a security-relevant gap, not just a lost optimization), so an operator who relies on sequence ordering under load knows that's a reason to choose the Redis backend.

---

eunolabs (commented):
## Review — bound in-memory call-counter key set (#241)

Solid, well-scoped fix that does exactly what issue #241's Option A + C recommend. I checked it out, built it, and ran the suite: `go build ./cmd/eunox-mcp/` clean, `go test -race ./pkg/callcounter/` green (the new in-package `maxkeys_internal_test.go` included). The fail-closed claims hold up against the real handlers — `handleMaxCalls` maps any `IncrementIfBelow` error to `CONDITION_FAILED`/deny (`handlers.go:310`), and `recordSessionCall` swallows the `IncrementAndGet` error by design (`engine.go:227`).

**What's good**
- `admitNewKey()` is gated correctly on the `!ok` (new-key) branch in both `IncrementAndGet` and `IncrementIfBelow`, under the existing lock — so an already-present key never trips the bound and the map provably never exceeds `maxKeys`. Tests pin the "map must not grow" invariant on both paths.
- `Peek` is deliberately **not** gated — correct: a read at capacity must not error, or `sequenceBlock` evaluation would itself break at the ceiling. Nice that there's an explicit test for it.
- Defaulting in `main.go` so the binary is protected out of the box, and harmlessly discarding the option when `--redis-addr` overrides `counter` — `WithMaxKeys(0)` keeps the library default unbounded, so no behavior change for existing callers/tests.
- Docs + threat-model §3.6 + changelog row 1.10 all updated in-PR, per the conventions.

**One substantive note** (left inline on `docs/threat-model-mcp.md:278`): the "fails closed" framing is accurate for `maxCalls` but glosses over `sequenceBlock`, which degrades to **fail-open** at saturation (skipped antecedent marker → dependent tool not blocked). Narrow and arguably acceptable, but worth one explicit sentence so operators relying on sequence ordering under load know it's a reason to prefer Redis.

**Minor**
- The two namespaces share one `maxKeys` budget, so a flood of `seq:` keys can push `maxcalls:` to fail closed for unrelated sessions (and vice-versa). It's the right trade-off vs OOM and is implied by the docs, but the bound is a global memory backstop, not a per-condition guarantee — fine to leave as-is.
- No new metric/signal when the ceiling is hit; the `maxCalls` path is at least visible in the audit log (the `CONDITION_FAILED` message carries "key limit reached (n)"), while the `seq:` skip is silent. Acceptable given the project's no-telemetry stance.

Nothing blocking from me — just the doc-wording suggestion. Nice work on the test coverage and the threat-model write-up.

---

docs/threat-model-mcp.md
eunolabs: Good catch — fixed in 70fb3f7. §3.6 now spells out the asymmetry explicitly: `maxcalls:` fails closed, but `seq:` fails **open** at the ceiling (the swallowed `recordSessionCall` error skips the antecedent marker, so a later `sequenceBlock` check doesn't block the dependent tool), called out as a security-relevant gap with the worst case (`sequenceBlock(afterTools:[A])` on B bypassable when A isn't `maxCalls`-gated) and the Redis recommendation for sequence-ordering under load. I also qualified the same "fails closed" wording in the `WithMaxKeys` / `NewInMemory` godoc and the `defaultMaxCallCounterKeys` comment so the code and the threat model agree. No behavior change.

---

PR #261, opened 2026-06-13
fix: chain first post-upgrade audit record to legacy tail with empty prev_hmac (#228)

### LINKED ISSUE #228: bug: pre-chain legacy records (empty `_hmac`) produce a false `sha256:genesis` sentinel on restart, causing a spurious chain break in `audit-verify`
## Summary

When `openAuditSink` resumes from an existing log whose last record is in the **pre-chain format** (no `_hmac` field, written before signing support was added), it correctly sets:

```go
// audit.go:237-238
s.seq = prev.Seq       // e.g. 0 for a pre-seq record, or N if seq was added earlier
s.prevHMAC = prev.HMAC // "" — zero value, pre-chain record has no _hmac
```

Then in `writeRecord`, an empty `prevHMAC` is interpreted as **"no prior record — use the genesis sentinel"**:

```go
// audit.go:457-460
if s.prevHMAC == "" {
    rec.PrevHMAC = auditGenesisPrev  // "sha256:genesis"
} else {
    rec.PrevHMAC = s.prevHMAC
}
```

So the first post-restart record carries `prev_hmac: "sha256:genesis"`.

Meanwhile, `verifyAuditLog` tracks the chain by recording the `_hmac` of each record it processes:

```go
prevHMAC = rec.HMAC  // "" for the last pre-chain record
```

When it reaches the first post-restart record, it compares:

```
rec.PrevHMAC ("sha256:genesis") != prevHMAC ("") → chain break
```

**A chain break is reported for an untampered log.** This is a false positive.

## Impact

Users upgrading from a version without chain/signing support will see a spurious `CHAIN BREAK` from `eunox-mcp audit-verify` on the first log line written after the upgrade. This erodes trust in the tool's output and obscures real chain breaks that might follow.

The code comment at `audit.go:219` even documents the intended behaviour:

> "A log written by the pre-chain format has no seq (parses as 0) and the new records simply begin numbering from 1, **chained to the last legacy record's `_hmac`**."

Chaining to the legacy record's `_hmac` means the new record's `prev_hmac` should be `""` (the actual value). Instead, the code substitutes the genesis sentinel, contradicting its own documented intent.

## Root cause

The genesis-sentinel condition (`prevHMAC == ""`) conflates two distinct states:

1. **Fresh log** — no prior record exists; the sentinel correctly marks the chain's origin.
2. **Legacy transition** — a prior record exists but carries no HMAC; the sentinel is wrong.

There is no way to distinguish these states from `prevHMAC` alone.

## Suggested fix

Track whether a prior record was loaded from disk during startup. If `s.seq > 0` (a prior record existed) but `s.prevHMAC == ""`, the chain is continuing from a legacy record, not starting fresh. One approach:

```go
// In writeRecord, distinguish "no prior record" from "prior record had no HMAC":
if s.prevHMAC == "" && s.seq == 0 {
    rec.PrevHMAC = auditGenesisPrev  // genuine first record
} else if s.prevHMAC == "" {
    rec.PrevHMAC = ""  // continuing from a pre-chain record; leave as empty string
} else {
    rec.PrevHMAC = s.prevHMAC
}
```

And update `verifyAuditLog` to treat `prev_hmac: ""` on the first record (or after a `_hmac: ""` predecessor) as a legitimate chain start — not a break.

Alternatively, add a boolean flag `s.hasLegacyTail bool` set during startup when a pre-chain tail is detected, and check it in `writeRecord`.

## Affected code

`cmd/eunox-mcp/audit.go`, lines 219–222 (startup resume comment), 457–460 (`writeRecord` genesis check).

eunolabs:
## Summary

Fixes #228 — a spurious `CHAIN BREAK` reported by `eunox-mcp audit-verify` on the first log line written after upgrading from a pre-chain (unsigned) log.

When `openAuditSink` resumes from a pre-chain tail (a record with no `_hmac`, written before signing support), `s.prevHMAC` is set to that record's empty `_hmac` (`""`). `writeRecord` then read `prevHMAC == ""` as "fresh log → use the `sha256:genesis` sentinel", so the first appended record carried `prev_hmac: "sha256:genesis"`. But `verifyAuditLog` reconstructs the predecessor link from the legacy record's `_hmac`, which is `""`, so the comparison `"sha256:genesis" != ""` was flagged as a chain break for an untampered log.

## Fix

- **`writeRecord`** now distinguishes three states explicitly: continuing a signed chain (`prevHMAC != ""`), continuing from a resumed legacy tail (stamp `prev_hmac: ""`, matching what verify reconstructs), or the genuine first record of a fresh log (`sha256:genesis`).
- A resumed pre-chain tail is tracked with an explicit `resumedLegacyTail` flag set at startup, **not** a `seq == 0` heuristic — the issue's primary suggestion. A pre-seq legacy record also parses as `seq 0`, so `seq == 0` cannot tell it apart from a brand-new log and would leave the bug in place for that case. The flag is correct regardless of whether the legacy tail carried a seq.

## Why `verifyAuditLog` is unchanged (deliberate)

With the write-side fix, the first appended record carries `prev_hmac: ""`, which equals the legacy predecessor's reconstructed `""` → no break. No verify change is needed, and adding one would *weaken* tamper detection: a genuinely stripped `prev_hmac` mid-chain (whose predecessor has a real, non-empty HMAC) must still mismatch and be reported — which it does. Empty `prev_hmac` is only a valid link when it follows an unsigned (empty `_hmac`) record.

Tamper detection is otherwise unchanged: the genesis sentinel still marks the first record of a fresh log, and `seq`/`prev_hmac` are still covered by the signature.

## Tests

- New `TestAuditChain_ResumesFromLegacyTail` (table-driven) covers both a **pre-seq** legacy tail (seq 0 — the case `seq == 0` would mishandle) and a **seq-bearing** legacy tail. It asserts the first appended record carries `prev_hmac: ""` (not the genesis sentinel), that full-log verification reports zero chain breaks across the upgrade boundary, and that the appended records form a clean, individually verifiable chain.
- Full suite (`go test -race -count=1 ./...`), `make check-fmt`, `make check-license`, and `make lint` all pass.

## Docs

- `docs/threat-model-mcp.md` §3.4 — new "Upgrading from a pre-chain log" note.
- `docs/architecture.md` — chain description now mentions the `""` link for the first post-upgrade record.

Closes #228

https://claude.ai/code/session_01W5eZWJ56fKS7w9RS5phmh8

---
_Generated by [Claude Code](https://claude.ai/code/session_01W5eZWJ56fKS7w9RS5phmh8)_


docs/threat-model-mcp.md
```diff
@@ -248,6 +248,8 @@ The audit key is generated once at first startup using `crypto/rand` (Go's CSPRN
...
 
+**Upgrading from a pre-chain log.** A log written before signing/chaining support has no `_hmac` on its records. When the proxy resumes from such a tail, the first record it appends carries `prev_hmac: ""` — the empty `_hmac` of the legacy predecessor it is chaining onto — rather than the `sha256:genesis` sentinel (which marks only the first record of a brand-new log). `audit-verify` reconstructs the same empty link from the legacy record, so the upgrade boundary verifies as a continuous chain with no spurious break. The legacy records themselves remain individually unverifiable (they were never signed); only the signed records appended after the upgrade are covered by HMAC and chain checks.
```
eunolabs: The chain-break fix is correct, but "verifies as a continuous chain with no spurious break" may set the expectation that `audit-verify` *passes* after an upgrade — it won't, and that surprise is worth one sentence here.

A legacy record has no `_hmac`, so `VerifyRecord` counts it as `invalid`. `auditVerifyResult.ok()` is `invalid == 0 && chainBreaks == 0`, and `cmdAuditVerify` does `os.Exit(1)` on `!ok()` (`main.go:1411`). So post-upgrade, `audit-verify` prints `... 1 invalid ... 0 chain break(s)` and still **exits non-zero** until the pre-chain records rotate out — your own test encodes this by asserting only `chainBreaks == 0` on the full log rather than `ok()`.

I'm not suggesting a code change: treating an unsigned record as `skipped` instead of `invalid` would let an attacker dodge verification by stripping `_hmac`, so `invalid` is the right fail-closed call. The ask is just to make the exit-code consequence explicit here, so an operator who upgrades and watches the chain break disappear isn't then puzzled that `audit-verify` keeps failing until the legacy tail ages out.

---

eunolabs (commented):
## Review: chain first post-upgrade record to legacy tail (#228)

Reviewed the diff and traced it against the verify path on the PR branch. This is a correct, well-scoped, and well-documented fix — nice work, especially the reasoning for *not* touching `verifyAuditLog`.

**The fix is right.** `verifyAuditLog` reconstructs each record's expected predecessor link from the *preceding record's* `rec.HMAC` (`audit.go:673,681`). A resumed pre-chain tail has `HMAC == ""`, so the first appended record must carry `prev_hmac == ""` to match; the old genesis sentinel (`"sha256:genesis" != ""`) is exactly the spurious break in #228. The new three-way `switch` in `writeRecord` produces the right link, and `openAuditSink` derives `resumedLegacyTail` fresh from `prev.HMAC == ""` on every open, so it self-clears once a signed record becomes the tail.

**`seq == 0` heuristic correctly rejected.** A pre-seq legacy tail parses as `seq 0`, indistinguishable from a fresh log — the explicit flag is the right call, and the table-driven test covers both the pre-seq and seq-bearing tails.

**Tamper detection is not weakened** (I checked each angle):
- Fresh logs still get the genesis sentinel, so a stripped `prev_hmac` on a record following a *signed* predecessor still mismatches → break. Empty `prev_hmac` is only a valid link behind an unsigned (`_hmac == ""`) record.
- Deleting the first appended record still trips both a `prev_hmac` mismatch and a seq gap.
- Forged unsigned records inserted at the boundary still fail per-record HMAC (`invalid`), so `ok()` is false regardless of the chain check.
- Leading removal of the legacy record is undetectable from the file alone — but that's the pre-existing, already-documented limitation (`audit.go:641-645`), not a regression.

**Verified locally on the PR branch** (Go 1.25.11):
- `go build ./...` — clean
- `go test -race -count=1 -run TestAuditChain ./cmd/eunox-mcp/` — pass
- `gofmt -l` on both changed `.go` files — clean; `go vet ./cmd/eunox-mcp/` — clean

Docs (threat-model §3.4, architecture.md) updated in the same PR per CONTRIBUTING. No new audit-record field, so the sign-and-verify round-trip requirement isn't triggered.

One **non-blocking** note inline on the threat-model paragraph: after the fix the upgrade boundary no longer reports a *chain break*, but `audit-verify` still exits non-zero (legacy records count as `invalid`) until they rotate out — worth making explicit in the doc so operators aren't surprised. Not a code concern.

LGTM. CI is still running (`mergeable_state: unstable`) — I'll keep watching and re-review if anything changes.

---

docs/threat-model-mcp.md
eunolabs: Good catch — clarified in 4d37bce. The paragraph now states that `audit-verify` exits non-zero after an upgrade (legacy records count as `invalid`, printing `N invalid … 0 chain break(s)`) until the pre-chain records rotate out, and notes why `invalid` (not `skipped`) is the fail-closed call. No code change.

---

PR #262, opened 2026-06-13
sec: fail sequenceBlock closed when a call-counter write faults

### LINKED ISSUE #232: security: recordSessionCall silently swallows counter errors — sequenceBlock fails open on counter faults
## Summary

In `pkg/enforcement/engine.go:203`, `recordSessionCall` silently discards call-counter errors:

```go
```

The comment justifies this as "best-effort": a missing history marker can only cause `sequenceBlock` to fail open for one antecedent. But "fail open" is exactly what the fail-closed invariant prohibits.

## Failure scenario

1. Redis backend has a transient error (network partition, Redis restart, etc.)
2. Tool A is allowed and `recordSessionCall` is called — but the `IncrementAndGet` fails silently, so the call is **not recorded** in session history
3. The client calls Tool B, which has `sequenceBlock: afterTools: [A]`
4. `handleSequenceBlock` calls `history.Peek(...)` for A → returns 0 (no record) due to the missed write
5. B is **allowed** even though A ran — the block condition silently fails open

## Why this matters

From CLAUDE.md:
> **Fail closed.** On any ambiguity — missing manifest, unmapped method, malformed JWT, unset condition argument — deny.

A counter fault during `recordSessionCall` is exactly such ambiguity for any subsequent `sequenceBlock` check in the same session. The current code treats it as "no problem, best effort" but the consequence is a security-relevant block rule silently becoming inoperative.

## Suggested fix

Two options (in preference order):

**Option A — Deny any sequenceBlock in the same session after a failed record:**  
Store a flag on the context or in the counter that marks the session's history as potentially incomplete; `handleSequenceBlock` fails closed if the flag is set.

**Option B — At minimum, emit an audit record:**  
Log/audit the error so an operator can detect counter degradation. The current silent drop makes the failure invisible.

## Affected code

- `pkg/enforcement/engine.go:190–204` (`recordSessionCall`)
- `pkg/enforcement/handlers.go:750–785` (`handleSequenceBlock`)

eunolabs:
## Summary

Fixes #232. `recordSessionCall` swallowed call-counter errors as "best-effort", which let `sequenceBlock` **fail open** — a direct violation of the fail-closed invariant.

**Failure scenario (from the issue):** a transient counter fault (Redis partition/restart) during `recordSessionCall` silently drops the per-session history marker for an allowed antecedent (e.g. `read_credentials`). A later `sequenceBlock` check (e.g. `write_external` with `afterTools: [read_credentials]`) then `Peek`s a *falsely-absent* marker, returns 0, and **allows** the call the rule was meant to block.

## Fix (issue Option A)

When `recordSessionCall`'s `IncrementAndGet` write fails, the engine now **taints** the affected `(session, tool)` history key. `handleSequenceBlock` fails closed (`CONDITION_FAILED`, `reason: history_incomplete`) on any subsequent check against a tainted antecedent, rather than trusting a history it knows is incomplete.

Design points:
- **A true positive, not a conservative guess.** `recordSessionCall` runs only *after* a call is allowed and forwarded, so a tainted antecedent genuinely executed. The resulting deny is therefore never a false over-block.
- **Per-tool, not per-session** — an unrelated antecedent recorded successfully in the same session is unaffected.
- **Self-healing** — the taint expires after the 24h retention window (`sequenceHistoryWindowSec`), at which point the unwritten marker would itself have aged out, so its absence is no longer ambiguous and normal `Peek` evaluation resumes. The expired-entry sweep runs on **both** the write and read paths (shared `sweepExpiredTaintLocked`), so the map stays bounded even under a flap-once-then-recover burst across many keys.
- **Process-local** is sufficient because `sequenceBlock` is session-scoped and a session is served end-to-end by a single engine instance; the shared Redis counter aggregates independent sessions, never one session's ordered history across instances.
- **Observability (issue Option B):** the degraded counter surfaces as audited denials **when a gated call later occurs in the same session**. This is partial by design — a fault on an antecedent that is never subsequently gated leaves no later block to audit. The security property (no bypass) holds regardless; only counter-health visibility is best-effort (the engine carries no logger today, so a dedicated debug/audit log on the failed write would be a separate change).
- The record path still **never turns an otherwise-allowed call into a deny** — the antecedent call itself is forwarded; only later sequence checks are affected.

## Tests

New cases in `pkg/enforcement/sequence_block_test.go`, driven by a `recordFailCounter` that fails the write but serves `Peek` (reproducing the exact "write faults, then Peek recovers" window):
- `TestSequenceBlock_RecordFailureFailsClosed` — the core regression: dropped write → `write_external` denied instead of failing open.
- `TestSequenceBlock_RecordFailureTaintIsPerTool` — an unrelated, successfully-recorded antecedent is **not** over-blocked.
- `TestSequenceBlock_RecordFailureTaintExpiresAfterWindow` — taint clears once the retention window elapses.

Plus a white-box `TestHistoryTaint_ReadPathSweepsExpiredEntries` (`internal_coverage_test.go`) pinning that the read-path sweep reclaims faulted-then-expired keys that are never re-queried.

`make test` (race), `make check-fmt`, `make check-license`, `go vet`, and `golangci-lint` all pass.

## Docs

- `docs/capability-manifest-guide.md` and `docs/threat-model-mcp.md` (§3.2.1) updated to describe the both-directions fail-closed behavior.
- `CHANGELOG.md` Security entry added.

https://claude.ai/code/session_014awBTYiGoYiBzgoLQVCpCq


pkg/enforcement/engine.go
```diff
@@ -203,10 +212,22 @@ var _ capability.Enforcer = (*Engine)(nil)
...
+//
+// A counter fault here must not silently fail open, however. If the marker write
+// fails the call is NOT recorded, so a later sequenceBlock Peek for this
+// (session, tool) would read a falsely-absent marker and allow a call the rule
+// was meant to block (issue #232). To preserve the fail-closed invariant the
+// failed key is tainted: handleSequenceBlock denies any subsequent check against
+// a tainted key until its retention window elapses, rather than trusting history
+// it knows is incomplete. Those denials are themselves audited by the caller, so
+// the operator sees the counter degradation rather than a silent bypass.
+//
```
eunolabs, line 233: Worth calling out a strength here: because `recordSessionCall` runs only *after* the call has been allowed and forwarded, a tainted antecedent genuinely executed. So the later `sequenceBlock` deny isn't just a conservative fail-closed guess — it's a true positive (the antecedent really ran; we only lost the marker). The taint can therefore never cause a false over-block, which is a nicer property than "deny on ambiguity."

One caveat on "the operator sees the counter degradation" (and the PR description's claim that issue #232's Option B is satisfied for free): the degradation is surfaced **only if a gated call later occurs in the same session**. A write fault on an antecedent that is never subsequently gated — e.g. `read_credentials` runs but the session never attempts `write_external` — leaves no audit trace of the counter fault at all. The security property (no bypass) is fully met by Option A regardless; this is just to note the observability is partial vs. Option B, which suggested auditing the fault itself. Not blocking — but I'd soften the "for free" framing in the PR body, or consider a debug/audit log on the failed write if counter-health visibility is a goal.

---

pkg/enforcement/engine.go
```diff
@@ -224,7 +245,55 @@ func (e *Engine) recordSessionCall(ctx context.Context, req *capability.EnforceR
...
+// ambiguous and the taint is moot.
+func (e *Engine) markHistoryTaint(key string) {
+	now := e.clock.Now()
+	e.taintMu.Lock()
+	defer e.taintMu.Unlock()
+	if e.taintedHistory == nil {
+		e.taintedHistory = make(map[string]time.Time)
+	}
+	// Sweep entries past the retention window on this (rare) error path so the map
+	// cannot grow without bound under a persistently flapping counter.
```
eunolabs: Minor: the sweep runs only on this error path, so the "cannot grow without bound" guarantee holds for a *persistently* flapping counter (every new fault triggers a sweep). The other shape isn't covered: a burst of write faults across many distinct `(session, tool)` keys followed by full recovery. Those entries then stay resident until each is lazily cleared by a `historyTainted` lookup of that exact key — and if a key is never re-queried (no future fault ⇒ no future sweep), it lingers until process exit even though it's logically expired.

It's harmless (expired taints read as `false`) and outage-bounded (a string key + a `time.Time` each), so not a blocker. Just flagging that the comment slightly overstates the bound. If you want it tight, `historyTainted` already does opportunistic cleanup; a periodic/lazy sweep on the read path (or capping by count) would close the flap-once-then-recover case.

---

eunolabs (commented):
## Review — `sec: fail sequenceBlock closed when a call-counter write faults`

Solid, correct fix for #232. It implements the issue's preferred Option A (taint the history key on a failed write, fail closed on later lookups) cleanly, and I verified the build, the new race tests, `gofmt`, and `go vet` all pass locally.

**What's right:**
- **Window math is precise.** `markHistoryTaint` stores the time of the failed write, and the taint is honoured for exactly `sequenceHistoryWindowSec`. Since the unrecorded antecedent actually ran at that instant, the block lifts at exactly the point the marker would itself have aged out — so the taint window matches the real sequence window rather than approximating it. The `>= windowSec` expiry in `historyTainted` and the `!ts.After(cutoff)` sweep in `markHistoryTaint` agree at the boundary.
- **Not just conservative — precise.** `recordSessionCall` runs only after a call is allowed and forwarded, so a tainted antecedent genuinely executed. The resulting deny is a true positive, so the fix adds no false over-blocking. (Detail noted inline.)
- **Per-tool scoping** avoids penalising unrelated antecedents; the per-tool test pins that down.
- **Thread-safe** via a dedicated `taintMu`; clock/counter are set-once at construction. Map growth is bounded on the (rare) error path.
- Docs (`capability-manifest-guide.md`, `threat-model-mcp.md` §3.2.1), `CHANGELOG.md`, and tests all updated per repo conventions.

**Non-blocking notes (inline):**
1. The "operator sees the counter degradation / Option B for free" claim is only partially true — a write fault on an antecedent that's never subsequently gated in that session produces no audit trace. The bypass is still fully prevented; just the observability is partial.
2. The `markHistoryTaint` sweep runs only on the error path, so the "cannot grow without bound" comment holds for a persistently flapping counter but not for a flap-once-then-recover burst (entries linger, logically harmless, until lazily cleared or process exit).

Neither blocks merge — both are about a comment/PR-description nuance and a negligible transient memory footprint, not the security property. Nice work.

---

pkg/enforcement/engine.go
eunolabs: Good points on both. Softened the framing in 87cbb03: the `recordSessionCall` comment, CHANGELOG, manifest guide, and threat model now say the degradation surfaces as audited denials *only when a gated call later occurs in that session*, and explicitly note that a fault on a never-gated antecedent leaves no later block to audit (bypass still prevented; counter-health visibility is best-effort). I went with reframing rather than a debug/audit log on the failed write, since the engine carries no logger today and adding one is a larger, separate change.

Also folded your "true positive, not a conservative guess" observation into the `recordSessionCall` doc comment — since the marker is only written after the call was allowed and forwarded, a tainted antecedent genuinely ran, so the taint can never over-block.

---

pkg/enforcement/engine.go
eunolabs: Tightened it for real in 87cbb03 rather than just fixing the comment: extracted `sweepExpiredTaintLocked` and now run it on the `historyTainted` (read) path as well as the write path. So the flap-once-then-recover burst you describe — faults across many distinct keys, then full recovery with no same-key re-query — is reclaimed by whichever path next takes the lock, instead of lingering until process exit. The map is nil on the fault-free path, so the read-path sweep is a no-op in the common case. Added `TestHistoryTaint_ReadPathSweepsExpiredEntries` pinning that two faulted-then-expired keys are reclaimed by a read of a third, unrelated key.

---

eunolabs: Re-reviewed `87cbb03` — this resolves both notes. The `sweepExpiredTaintLocked` extraction is correct: the queried-key semantics are unchanged (an expired entry is swept, so a surviving entry is in-window ⇒ reported tainted), and the boundary still matches `historyTainted`'s old `>= window` check (`!ts.After(now - window)`). The nil/empty-map fast path keeps the read-path sweep a no-op when healthy. Verified locally on the PR head: `gofmt`/`go vet` clean and the taint + sequenceBlock tests (incl. the new `TestHistoryTaint_ReadPathSweepsExpiredEntries`) pass under `-race`. Agree that a real audit/debug log on the failed write is a larger, separate change given the engine has no logger today. No further comments from me.

---

PR #263, opened 2026-06-13
sec: reject argumentSchema on non-tool capability targets

### LINKED ISSUE #222: bug: argumentSchema on resource: and prompt: constraints is silently not enforced at runtime
## Summary

`argumentSchema` on `resource:` and `prompt:` capability constraints is accepted at manifest load time (patterns compiled, keywords checked) but **silently skipped at enforcement time**. Only `tool:` constraints have their `argumentSchema` enforced.

## Root cause

`ManifestPDP.Decide` (for `tools/call`) manually runs `enforcement.ValidateArgumentSchema` before calling `engine.EvaluateConditions` (`pdp.go:218-229`).

`DecideResourceRead` and `DecidePromptGet` call `engine.EvaluateConditions` **directly, without any `ValidateArgumentSchema` call**:

- `pdp.go:412` — `DecideResourceRead` jumps straight to `EvaluateConditions`
- `pdp.go:499` — `DecidePromptGet` jumps straight to `EvaluateConditions`

The `engine.EvaluateConditions` method itself (`engine.go:410-476`) evaluates conditions but does not perform argument schema validation.

## Impact

An operator who writes this manifest:
```yaml
capabilities:
  - target: resource:file:///sensitive/**
    actions: [read]
    argumentSchema:
      properties:
        uri:
          pattern: "^file:///sensitive/allowed/"
```
will find the `argumentSchema` compiled and cached at load time (no error), but **never evaluated** when `resources/read` calls arrive. The structural guard is silently bypassed.

Contrast: the manifest loader already rejects `directives` on non-tool targets (`manifest.go:197-199`) because they can't be applied. The same principle should apply to `argumentSchema` on resource/prompt targets if it genuinely cannot be enforced, OR `DecideResourceRead`/`DecidePromptGet` should run the validation (passing the synthesised `Arguments` map they already build).

## Suggested fix

Either:
1. **Enforce it**: add a `ValidateArgumentSchema` call to `DecideResourceRead` and `DecidePromptGet` before calling `EvaluateConditions`, matching the pattern in `Decide` (lines 218-229). The `req.Arguments` maps are already constructed.
2. **Reject it**: add a load-time error in `LoadManifest` for `argumentSchema` on `resource:`/`prompt:`/`system:` targets, analogous to the `directives` check at `manifest.go:197-199`, so operators learn at validate/startup time rather than via silent failure.

Option 1 is preferred if `argumentSchema` is intended to validate the synthesised argument maps; option 2 is cleaner if it is tool-only by design. Either way, the current silent discard violates the fail-closed invariant: a declared guard that has no runtime effect is equivalent to no guard.

## Affected files

- `cmd/eunox-mcp/pdp.go` — `DecideResourceRead` (line 412), `DecidePromptGet` (line 499)
- `pkg/enforcement/engine.go` — `EvaluateConditions` (line 410)
- `cmd/eunox-mcp/manifest.go` — manifest loader (line 187)

eunolabs:
## Summary

Fixes #222.

`argumentSchema` on `resource:` / `prompt:` (and `system:`) capability constraints was accepted at manifest load — patterns compiled, keywords checked — but **never evaluated at enforcement time**. Only `tools/call` (`ManifestPDP.Decide`) ran `enforcement.ValidateArgumentSchema`; `DecideResourceRead` and `DecidePromptGet` went straight to `EvaluateConditions`, which does not do schema validation. A declared structural guard with no runtime effect is equivalent to no guard — a silent fail-open that violates the fail-closed invariant.

## Decision: reject at load (the issue's Option 2)

The issue offered two fixes and noted *"option 2 is cleaner if it is tool-only by design."* The normative spec is decisive here:

> **SPEC § 3.2.1** — `argumentSchema`: *"…for validating tool arguments at the structural level… **Applies to `tool:` constraints only.**"*
> **SPEC § 3.2.2** — *"`argumentSchema` validates the shape of a `tool:` request's arguments…"*

`argumentSchema` is **tool-only by design**: `resources/read` and `prompts/get` are matched on a single synthesised `{"uri"}` / `{"name"}` map, and the sampling opt-in is argument-less, so there is no tool-argument map for it to validate. Enforcing it on those targets (Option 1) would *contradict* the spec; rejecting it at load makes the implementation conform to the spec. This exactly mirrors the existing `directives`-on-non-tool rejection (`manifest.go`), so operators learn at `validate`/startup time rather than via silent discard.

## Changes

- **`cmd/eunox-mcp/manifest.go`** — `validateLocalManifest` now rejects an `argumentSchema` on any non-`tool:` target with a descriptive error (compilation of `pattern`s still runs for valid `tool:` entries).
- **`cmd/eunox-mcp/pdp.go`** — added comments in `DecideResourceRead` / `DecidePromptGet` documenting *why* they intentionally do not run `argumentSchema` validation, so the asymmetry with `Decide` isn't mistaken for the bug again.
- **`docs/capability-manifest-guide.md`** — state `argumentSchema` is `tool:`-only (§ 1) with notes in the resource (§ 2a) and prompt (§ 2d) sections pointing operators to an `allowedValues` condition on `uri` / `name` instead.
- **`cmd/eunox-mcp/manifest_validation_test.go`** — table-driven load-rejection test across `resource:` / `prompt:` / `system:` targets.

No spec change required — the spec already specifies tool-only; this aligns the implementation to it. No existing test loaded `argumentSchema` on a non-tool target.

## Verification

- `go build ./...` — ok
- `gofmt -l` on changed files — clean
- `go vet ./cmd/eunox-mcp/ ./pkg/enforcement/` — clean
- `go test -race -count=1 ./cmd/eunox-mcp/ ./pkg/enforcement/ ./pkg/capability/` — all pass

https://claude.ai/code/session_01LnzLNKVAsi8hvvMMkYtNZU

---
_Generated by [Claude Code](https://claude.ai/code/session_01LnzLNKVAsi8hvvMMkYtNZU)_


eunolabs (commented):
## Review: reject `argumentSchema` on non-tool capability targets

**Verdict: correct, well-scoped, and complete. No blocking issues.**

### What it does
Closes the #222 fail-open: `argumentSchema` on `resource:`/`prompt:`/`system:` constraints was compiled and cached at load but never evaluated at enforcement time (only `ManifestPDP.Decide` runs `ValidateArgumentSchema`; `DecideResourceRead`/`DecidePromptGet` go straight to `EvaluateConditions`). The PR takes the issue's **Option 2 — reject at load** — which matches the spec scoping `argumentSchema` to `tool:` only (SPEC § 3.2.1/3.2.2) and mirrors the existing `directives`-on-non-tool rejection right below it. Good call choosing Option 2 over Option 1: enforcing a tool-argument schema against a synthesised `{"uri"}`/`{"name"}` map would contradict the spec, and the docs correctly redirect operators to an `allowedValues` condition on `uri`/`name`.

### Correctness — verified
- `targetType` is the value returned by `capability.ParseTarget(c.Target)` earlier in the loop, so the new `targetType != capability.TargetTypeTool` guard is correctly scoped per-capability.
- Traced control flow for all three test targets (`resource:` / `prompt:` / `system:`): each clears the earlier `ParseTarget` / breadth / `ValidateResourcePattern` / `ValidateActionForTargetType` checks and reaches the new guard, so the rejection (and the test's `argumentSchema` + `tool:` substring assertions) fires for the intended reason rather than tripping an earlier validator.
- The guard precedes `CompileSchemaPatterns`, so a non-tool target now surfaces the tool-only error instead of a regex-compile error — the more fundamental reason, which is the right ordering.
- **Positive path stays covered**: `TestLoadManifest_..._accepted` (~L1330) loads a `tool:read_file` target carrying an `argumentSchema`, so the refactor that nested `CompileSchemaPatterns` under the new check is still exercised.

### Conventions / invariants
- Fail-closed preserved and strengthened; `sec:` prefix is apt.
- Back-compat break (a previously-loading manifest now errors) is fine pre-1.0 and is documented in the same PR (guide §1 / §2a / §2d), consistent with CONTRIBUTING.
- No `schemas/` change required: the repo's only JSON Schema is `eunox-gateway-config.schema.json`; capability manifests aren't schema-validated and `argumentSchema` target-scoping isn't represented there — same as `directives`.
- The pdp.go comments accurately describe the synthesised `{"uri"}`/`{"name"}` maps and why no schema validation runs on those paths; this nicely prevents the asymmetry from being re-mistaken for the bug.

### Minor / optional (non-blocking)
- This `sec:` change closes a silent fail-open. If `docs/threat-model-mcp.md` enumerates the "declared guard with no runtime effect" class, a one-line note there could be worth it — though the sibling `directives` rejection set the precedent of not requiring one, so this is a judgment call, not a request.

CI was still queued at review time; nothing in the diff suggests a failure (the PR reports build/vet/gofmt/tests clean locally).

---

PR #264, opened 2026-06-13
fix(enforcement): allowedValues "*" matches any value; add "**" subtree glob

### LINKED ISSUE #238: usability: allowedValues glob `*` does not match argument values containing `/` — operators expecting a "allow anything" wildcard get silent denials
## Summary

`handleAllowedValues` uses `path.Match` for string glob matching:

```go
// handlers.go:682-685
if matched, err := path.Match(pattern, str); err == nil && matched {
    return nil
}
```

`path.Match` treats `/` as a path separator: `*` matches any sequence of **non-`/`** characters. A pattern like `"*"` will therefore **not** match argument values containing `/`, even though an operator writing `values: ["*"]` in a manifest typically means "allow any value."

## Concrete cases that fail silently

| Manifest `values` entry | Argument value | Expected | Actual |
|------------------------|----------------|----------|--------|
| `"*"` | `"/reports/q3.pdf"` | allow | **deny** |
| `"*"` | `"SELECT * FROM ..."` | allow | allow (no `/`) |
| `"/reports/*"` | `"/reports/sub/q3.pdf"` | allow (maybe) | **deny** (`*` doesn't cross `/`) |
| `"/reports/**"` | `"/reports/sub/q3.pdf"` | allow | **deny** (`**` not supported by `path.Match`) |
| `"file:*"` | `"file:///etc/passwd"` | deny | **allow** (no `/` in the `*` segment... wait no, `///etc/passwd` has `/`) → actually deny |

The most dangerous case is the `"*"` wildcard: an operator who adds it intending to "allow all values" silently restricts to values with no `/` — passing for short identifiers and SQL verbs but failing for file paths and URIs.

## Why this matters for the fail-closed invariant

This failure mode is **more restrictive** than intended (extra denials, not extra allows), so it does not create a security bypass. However:

1. It erodes operator trust: a manifest that should be permissive unexpectedly denies valid calls, and the audit log gives no hint that the glob mismatch is due to the `/` separator — it just shows `CONDITION_FAILED`.
2. It conflicts with the `allowedValues` documentation, which gives the example `"/reports/*"` matching `"/reports/q3.pdf"` — this example actually works, but only because there is no second `/` after the wildcard. An operator testing with `"/reports/q3.pdf"` but deploying with `"/reports/sub/q3.pdf"` gets an unexpected deny with no diagnostic.
3. The behaviour differs from what most shell/filesystem globbing users expect from `*`.

## Current documentation

The godoc comment on `AllowedValuesCondition` says:

> String values are matched both by exact equality and by path.Match glob patterns, so `"/reports/*"` matches `"/reports/q3.pdf"` but not `"/internal/secret.txt"`.

The comment is technically accurate but does not warn that `*` does not cross `/`, that `"*"` is NOT "match everything", and that `"**"` is not supported.

## Suggested fix

Two options:

**Option A — document the limitation clearly:**
Expand the `AllowedValuesCondition` godoc and the capability-manifest guide to explicitly state:
- `*` matches any sequence without `/`; it is NOT a "match everything" wildcard.
- To match `"anything"` (including paths), use `reflect.DeepEqual`-style exact matching or list the allowed prefixes explicitly.
- `"**"` is not supported.
- For URL-style paths, use prefix patterns like `"/reports/*"` (matches one level) and enumerate deeper patterns as needed.

**Option B — use a `/`-agnostic glob (e.g. `doublestar` library or a simple prefix+suffix match):**
Replace `path.Match` with a glob implementation where `*` matches anything including `/`, and `**` matches across directory levels. This is a more ergonomic semantics for non-path values (SQL, identifiers, URIs). The trade-off is adding a dependency or implementing a custom matcher.

Option A is low-risk and consistent with the existing code; Option B is more powerful but requires careful security review (a glob that crosses `/` can be broader than intended for path arguments).

At minimum, the manifest validator should detect when a `values` entry is `"*"` and warn (or document) that it does not match values containing `/`.

## Affected code

- `pkg/enforcement/handlers.go` — `handleAllowedValues` (lines 682–685)
- `pkg/capability/condition.go` — `AllowedValuesCondition` godoc (line 106–113)
- `docs/capability-manifest-guide.md` — `allowedValues` section

eunolabs:
## Summary

Fixes #238. `handleAllowedValues` matched string values with `path.Match`, where `*` matches only runs of non-`/` characters. An operator who wrote `values: ["*"]` intending "allow any value" silently got "allow any value **with no `/`**" — denying file paths and URIs (e.g. `/reports/q3.pdf`) with no diagnostic beyond `CONDITION_FAILED`.

This is the **hybrid** of the two options in the issue: it fixes the dangerous `["*"]` footgun and adds a `**` subtree wildcard, **without** silently broadening existing `/`-containing single-`*` patterns.

## What changed

New `enforcement.MatchValueGlob` — the single matcher for argument-**value** globbing:

| Pattern | Old (`path.Match`) | New |
| --- | --- | --- |
| `"*"` / `"**"` (alone) | matches values **without** `/` only | matches **any** value, incl. `/` |
| `/reports/*` | matches `/reports/q3.pdf`, not `/reports/sub/q3.pdf` | **unchanged** |
| `/reports/**` | behaves like `*` (no cross-`/`) | matches the whole `/reports` subtree |
| `report*`, `a[1]`, `prod-?` | per-segment glob | **unchanged** |
| malformed (`[bad`) | non-match, rejected at load | **unchanged** (fail-closed + load reject) |

Only two behaviours change, both per the issue's intent: a bare `"*"`/`"**"` becomes allow-all, and a `"**"` path segment crosses `/`. Single-`*` patterns are byte-for-byte identical.

### JWT consistency (pre-existing divergence fixed)

The JWT capability-claim shorthand (`?path=…`) went through `matchResource` → `enforcement.MatchesResource`, which **already** short-circuits a bare `"*"` to allow-all — so the manifest engine and the JWT path *already disagreed* on `["*"]`. `matchesAllowedValues` now routes through the same `MatchValueGlob` (via `matchArgumentValue`), so the two can no longer diverge. JWT claims still intersect with the manifest and can only restrict, never expand.

## Docs
- `AllowedValuesCondition` godoc and the capability-manifest guide now state the precise glob rules plus a pre-1.0 migration note.
- `CHANGELOG.md`: new **Changed** entry.
- Companion spec PR **eunolabs/mcp-capability-manifest#5** makes the `allowedValues` glob semantics normative in `SPEC.md` §3.3.

## Testing
- `pkg/enforcement/glob_test.go` — table-driven `MatchValueGlob` (bare-`*`, `**` cross-`/`, single-`*` unchanged, `?`/`[...]`, malformed fail-closed). `MatchValueGlob` 100% covered.
- `enforcement_test.go` — engine-level allow (bare `*` on a slash value, `**` subtree) and deny (`**` respects its prefix) cases; existing `/reports/*`-denies-subdirectory case preserved.
- `pdp_jwt_test.go` — `?path=/reports/**` shorthand allows a nested path, denies a sibling tree.
- `make lint`, `gofmt`, `make check-license`, and `go test -race ./...` all pass.

Fixes #238

https://claude.ai/code/session_01DUqQ3U7abPCbyRvRLht699


docs/capability-manifest-guide.md
```diff
@@ -817,18 +817,34 @@ literal values or glob patterns:
...
+>   contains `/`. Write `values: ["*"]` to mean "allow any value" — that is now
+>   exactly what it does. (Before, a bare `*` matched only values with no `/`,
+>   so it silently denied file paths and URIs; see the migration note below.)
+> - A **`*`** inside a pattern matches a run of characters that does **not**
+>   cross `/`. So `/reports/*` matches `/reports/q3.pdf` but **not**
+>   `/reports/sub/q3.pdf` or `/internal/secret.txt` — it matches exactly one
+>   path segment after `/reports/`.
+> - A **`**` path segment** matches zero or more `/`-delimited segments, so
+>   `/reports/**` matches both `/reports/q3.pdf` *and* `/reports/sub/q3.pdf`.
+>   Use it when you want to allow a whole subtree.
```
eunolabs, line 838: Scope note worth adding here: these `**` subtree rules apply to `allowedValues` argument **values** only. Resource-URI and tool/target name patterns go through `matchesResource` (`pkg/enforcement/engine.go`), which is plain `path.Match` — there `**` is just two single-`*` stars and does **not** cross `/`. So an operator who learns `**` from this section and then writes a `resources` target like `file:///reports/**` expecting a subtree will silently get single-segment matching — a footgun in the opposite direction from #238.

Consider one line in this block, e.g.: "These `**` rules apply to `allowedValues` only; resource and tool target patterns still match a single `/`-segment per `*`." Not blocking — just preventing the inverse surprise.

---

pkg/enforcement/glob.go
```diff
@@ -0,0 +1,85 @@
...
+			// the value.
+			rest := pat[1:]
+			for len(rest) > 0 && rest[0] == "**" {
+				rest = rest[1:]
+			}
+			if len(rest) == 0 {
+				// A trailing "**" matches whatever value segments remain.
+				return true
+			}
+			for i := 0; i <= len(val); i++ {
```
eunolabs, line 84: Non-blocking / for awareness: each `**` group backtracks over every suffix of the value (`val[i:]`), so a pattern with _k_ non-adjacent `**` groups matched against a value of _n_ `/`-segments is worst-case ~O(n^k). It's bounded in practice — patterns come from the trusted manifest (k is almost always 1 → O(n²)) and argument values are size-limited — and collapsing adjacent `**` (lines 60-63) already removes the cheapest blowup, so this isn't a merge blocker.

But the value side is attacker-influenced and this runs in the hot enforcement path, so it might be worth a brief comment noting the bound is on trusted pattern complexity (or a small cap on `**` count / segment depth) so a future multi-`**` manifest pattern can't be turned into a CPU sink. Fail-closed behavior on malformed segments is correct.

---

eunolabs (commented):
## Review — `allowedValues` `*`/`**` glob fix (#238)

Reviewed the matcher, the JWT/manifest unification, the tests, and the docs. This is a clean, well-scoped fix and the implementation is **correct**.

**What I verified**
- Traced `matchGlobSegments` through the tricky paths — `**` in the middle with a non-`**` tail (`/a/**/c` correctly rejects `/a/b/c/d`, accepts `/a/b/c`), zero-segment `**` (`/reports/**` matches `/reports`), prefix enforcement (`/reports/**` rejects `/internal/...`), and malformed-segment fail-closed. All behave as documented.
- Confirmed `**` patterns still pass the pre-existing load-time validator (`stdpath.Match(s, "")` accepts consecutive stars), so the "rejected at load" claim holds, and malformed classes after a `*` are still caught (checked empirically).
- `go test ./pkg/enforcement/ ./cmd/eunox-mcp/` for the glob/allowedValues/JWT tests passes on the branch.

**Strengths**
- Routing both the manifest engine (`handleAllowedValues`) and the JWT shorthand (`matchesAllowedValues`) through one `MatchValueGlob` so they can't silently diverge is the right move, and the PR correctly notes they *already* disagreed on bare `*` before this.
- The migration note honestly flags the one direction that **broadens** an existing policy on upgrade (bare `*` → allow-all), which is the failure mode that matters for a fail-closed proxy.

**Non-blocking notes** (two inline):
1. Doc scope note — the new `**` subtree semantics apply to `allowedValues` values only; resource/target-name patterns still use single-segment `path.Match`. Worth one line so operators don't expect `**` to work on `resources:` patterns (inverse footgun to #238).
2. Awareness — `**` matching backtracks over value suffixes; worst case is polynomial in the (trusted) `**` count, fine in practice but the value side is attacker-influenced and it's in the hot path.

One more minor doc-precision point: `values: ["*"]` is described as "allow any value", but the glob branch only fires when the argument value is a **string** — non-string args (numbers/booleans) are still matched by exact equality, so `["*"]` is really "allow any *string* value". Half a sentence in the godoc/guide if you want to be precise.

Nothing here blocks merge.

---

eunolabs:
Thanks for the review. Addressed all three non-blocking notes in f7d04cb (comment/doc-only, no behaviour change):

1. **Scope vs. resource patterns** — added a callout to the guide's glob block: the `**` subtree rules apply to `allowedValues` *values* only; `tool:`/`resource:` target names (and JWT target names) go through single-segment `path.Match`, where `**` is just two single-`*` stars and does not cross `/`. Same note in the `MatchValueGlob` godoc, so the inverse footgun (`file:///reports/**` expecting a subtree) is called out in both places.

2. **`["*"]` is "any *string* value"** — clarified in the godoc, the guide, and the changelog that the glob only fires on string args; a non-string argument (number/bool/null) is matched solely by exact equality, so `["*"]` does not allow it.

3. **`**` backtracking cost** — documented the bound on `matchGlobSegments`: O(n^k) where k is the trusted-manifest `**` count (almost always 1 → O(n²)) and n is the request-size-bounded value depth, with adjacent `**` collapsed. I went with the documented bound rather than a hard cap, since k is operator-controlled (trusted input) and a cap adds an arbitrary threshold + a new fail-closed path on something that isn't directly attacker-driven. The comment explicitly points a future maintainer at "cap the `**` count at load time" if a pathological multi-`**` pattern ever needs defending — happy to add that guard now instead if you'd prefer it be proactive.

---

PR #265, opened 2026-06-13
fix!: bound call-counter memory with IncrementAndGet retention cap

### LINKED ISSUE #233: bug: InMemory counter has no per-key timestamp cap — sustained high-rate calls within a long window cause unbounded slice growth
## Summary

`InMemory.IncrementAndGet` appends one `time.Time` value (24 bytes) per call, and the pruning step only removes timestamps that have fallen outside the current window. For a long window with sustained traffic, the slice grows without bound:

```go
// memory.go:85-100
valid := e.timestamps[:0]
    if ts.After(cutoff) {
        valid = append(valid, ts)
    }
}
valid = append(valid, now)   // always grows by 1 for every call within the window
```

`compactTimestamps` (introduced for issue #207) only reclaims the **backing array** after a burst drains to below 25% utilisation — it does not cap the live count. For sustained traffic it never fires.

## Concrete sizing

| Window | Rate | Timestamps/key | Memory/key |
|--------|------|----------------|------------|
| 3600 s (1 h) | 100 req/s | 360 000 | ~8.6 MB |
| 86400 s (24 h) | 100 req/s | 8 640 000 | ~207 MB |
| 86400 s (24 h) | 1000 req/s | 86 400 000 | ~2.1 GB |

The in-memory backend is intended for single-replica or test deployments, but even modest real-world rates against a 24 h `maxCalls` window can exhaust heap.

## Why this matters

A `maxCalls` condition designed to enforce a daily cap — e.g. "an agent may call `dangerous_tool` at most 1000 times per 24 hours" — needs only ONE counter value plus the timestamps of those 1000 calls. The current implementation stores ALL timestamps within the window, regardless of how many calls were made. A manifest author setting a 24-hour window with a high call rate (or no rate) is effectively enabling a heap-growth sink.

## Comparison with Redis backend

The Redis backend stores a sorted set with one entry per call, but entries are pruned by `ZRemRangeByScore` on every `IncrementAndGet`. There is no compaction delay — expired entries are removed immediately. The InMemory backend is asymmetric: it prunes on every call (correct) but never caps the live count, so under sustained load it diverges.

## Suggested fix

Two non-exclusive approaches:

**Option A — document as a known limitation for high-rate long-window quotas:**
Add a note in the `NewInMemory` godoc (or in the `maxCalls` condition docs) that the InMemory backend stores all in-window timestamps and is not suitable for rates above `maxCalls/windowSeconds * N` at steady state, where `N` is a memory budget. Recommend the Redis backend for high-rate long-window quotas.

**Option B — trim to the last `quota` timestamps on every call:**
Since the enforcement engine only needs to know `count > mc.Count`, once the count reaches `mc.Count + 1` no additional timestamps matter for the current decision. The slice could be trimmed to at most `mc.Count + 1` entries (keeping the most recent). This requires passing the quota limit into `IncrementAndGet`, which is a breaking change to the `Store` interface.

Option A is low-risk; Option B changes the interface. Either way the behaviour should be documented.

## Affected code

`pkg/callcounter/memory.go` — `IncrementAndGet` (lines 55–101) and `compactTimestamps` (lines 130–137).

eunolabs:
## Summary

Fixes #233.

`InMemory.IncrementAndGet` appended one `time.Time` per call and only pruned **expired** timestamps, so a key under sustained traffic on a long window grew without bound. A 24h history key at 100 req/s held ~8.6M timestamps (~207 MB); the #207 compaction never reclaimed it because the live count never dropped. The Redis backend had the same growth for its per-call sorted set.

Crucially, the **only production caller** of `IncrementAndGet` is the per-session **sequenceBlock history** recorded on *every allowed call* (`recordSessionCall`), and it only ever asks "did this tool run at least once?" (`Peek > 0`) — it discards the count entirely. So the heap was growing one entry per call to answer a single bit. (`maxCalls` already moved to the quota-bounded `IncrementIfBelow` in #204 and is unaffected.)

## Fix (issue's Option B)

`IncrementAndGet` gains a `maxEntries` retention cap and keeps only the **most-recent N** in-window timestamps; the returned count is capped at N. The Redis backend trims identically with `ZREMRANGEBYRANK` inside the existing MULTI/EXEC.

- **Keep the newest, not the oldest** — this is what makes the cap correct for a sliding window: the newest timestamp ages out last, so a later `Peek`/`IncrementAndGet` still reports presence (or a count up to `maxEntries`) for as long as any call remains in the window. Keeping the oldest would report absence prematurely and fail `sequenceBlock` *open*.
- **The cap is mandatory and fails closed** on a non-positive value (consistent with `windowSec` and the fail-closed invariant) rather than treating it as "unbounded" — an unbounded count is the sink itself.
- `recordSessionCall` passes `1` (presence), bounding the history marker to a single timestamp regardless of call rate.

With `maxEntries=1` the in-memory slice stabilises at length 1 / capacity 2; the Redis set holds one member.

## Breaking (pre-1.0)

The `callcounter.Store` / `capability.CallCounter` `IncrementAndGet` signature gains the `maxEntries` parameter. This is the interface change anticipated in the issue's Option B; per `CONTRIBUTING.md` pre-1.0 breaks cleanly with no shim.

## Tests

- `TestInMemory_IncrementAndGet_CapsToMaxEntries` — 10k in-window calls keep count **and** the stored slice/backing array at the cap.
- `TestInMemory/Redis_IncrementAndGet_CapKeepsNewest` — proves the cap retains the newest marker so presence survives until the last call ages out (the keep-newest correctness property).
- `TestRedis_IncrementAndGet_CapsToMaxEntries` — `ZREMRANGEBYRANK` trim verified via an independent `Peek` read path.
- `Test*_IncrementAndGet_RejectsNonPositiveMaxEntries` — fail-closed on both backends.
- `TestSequenceBlock_BlocksAfterManyAntecedentCalls` — end-to-end: a high-rate antecedent still arms the block.

`go test -race -count=1 ./...`, `go vet`, `gofmt`, and `golangci-lint` are all clean. The CHANGELOG has a `Fixed` entry.

https://claude.ai/code/session_012nxLALcR9avq4WFnNhoHoX

---
_Generated by [Claude Code](https://claude.ai/code/session_012nxLALcR9avq4WFnNhoHoX)_


eunolabs (commented):
## Review: bound call-counter memory with `IncrementAndGet` retention cap

Reviewed the full diff plus the surrounding `memory.go` / `redis.go` / `engine.go` / `handlers.go` context against issue #233. This is a correct, well-scoped fix — the issue's Option B implemented as a **mandatory, fail-closed** retention cap. No blocking issues found.

### Verified correct

- **Keep-newest invariant (the crux).** Both backends retain the most-recent `maxEntries`, so a sliding-window presence/threshold decision stays correct until the last in-window call ages out.
  - InMemory's front-shift `copy(valid, valid[len(valid)-maxEntries:])` is overlap-safe (dst index < src index, so no element is clobbered before it's read).
  - Redis's `ZRemRangeByRank(key, 0, -(maxEntries+1))` drops the oldest surplus and is a no-op at/under the cap, independent of set size; the negative-rank boundary checks out (`maxEntries=1` → stop `-2` keeps only rank `-1`, the newest).
  - The two `*_CapKeepsNewest` tests pin exactly this property — keeping the *oldest* would report absence prematurely and fail `sequenceBlock` **open**, which these tests guard against.

- **Fail closed.** `checkMaxEntries(<1)` rejects rather than treating `0` as "unbounded" — consistent with `checkWindowSec` and the repo's fail-closed invariant (an unbounded count is the sink itself). Covered on both backends by the `RejectsNonPositiveMaxEntries` tests.

- **Cap × compaction.** After the trim, `compactTimestamps` still reclaims an oversized backing array; `CapsToMaxEntries` asserts both `len` **and** `cap` stay bounded (cap ≤ 2 at `maxEntries=1`), so a past burst's array can't be pinned.

- **Scope is right.** `recordSessionCall` is the only production caller and only ever needs `Peek > 0` (confirmed in `handlers.go`), so `sequenceHistoryMaxEntries = 1` is sufficient. `maxCalls` goes through `IncrementIfBelow` (the `RateLimiter` path) and is untouched — confirmed. All four implementers (InMemory, Redis, `errorCounter`, `counterWithoutHistory`) and both interfaces (`callcounter.Store`, `capability.CallCounter`) are updated, and `engine.counter` is typed `capability.CallCounter`, so the build stays consistent.

### Optional (non-blocking)

- `docs/threat-model-mcp.md` already documents the analogous `IncrementIfBelow` state-growth fix (#204 — *"a rate-limited key's storage shrinks rather than growing"*). Since this PR closes the parallel unbounded-growth sink on the `IncrementAndGet` / `sequenceBlock`-history path, a one-line parallel note there would keep the threat model's "unbounded state growth" story complete. The CHANGELOG + godoc coverage is already solid, so this is purely doc-consistency.

CI is still queued on this PR. I'll re-review automatically on any new push.

---

eunolabs:
Thanks for the review. Two follow-ups pushed (force-push after a rebase):

- **Addressed the optional doc suggestion.** Added a parallel note to `docs/threat-model-mcp.md` §3.6: the per-key *size* bound this PR adds complements the #241 per-key *count* bound, and mirrors the #204 `IncrementIfBelow` "storage shrinks rather than growing" note. Added a matching `v1.12` version-history row.
- **Rebased onto `main` to resolve the merge conflict** that had developed. Clean integration with the two overlapping changes that landed on the same path: #262's history-taint fail-closed logic in `recordSessionCall` and #241's `WithMaxKeys`/`admitNewKey` key-count bound — both now thread through the new `maxEntries` parameter (the per-key count and per-key size bounds are orthogonal and coexist). All of main's new 3-arg `IncrementAndGet` call sites (#239 Peek tests, #241 maxkeys tests, #262's `recordFailCounter` mock) were updated to the 4-arg signature.

`go test -race -count=1 ./...`, `go vet`, `gofmt`, and `golangci-lint` are all green locally on the rebased tree.

---

PR #266, opened 2026-06-13
sec: fail closed when sequenceBlock session-history recording fails

### LINKED ISSUE #213: security: recordSessionCall swallows counter errors, silently bypassing sequenceBlock on backend failure
## Summary

`recordSessionCall` (`pkg/enforcement/engine.go:203`) silently discards call-counter errors:

```go
```

The code comment explains the trade-off:

> a counter error is swallowed because a missing history marker can only ever cause sequenceBlock to fail OPEN for that one antecedent, which is no worse than not having recorded it — and the record path must never turn an otherwise-allowed call into a deny.

However, this design conflicts with the load-bearing **fail-closed** invariant stated in CONTRIBUTING.md:

> **Fail closed.** On any ambiguity — missing manifest, unmapped method, malformed JWT, unset condition argument — deny, with a structured error code, and log it. **New code paths must preserve this.**

## Impact

If the call-counter backend (Redis or in-memory) fails at the moment an antecedent tool is called:

1. The antecedent tool call is **allowed** (correct — the condition is on the consequent tool).
2. `recordSessionCall` fails to write the history marker and swallows the error.
3. A subsequent call to the consequent tool (protected by `sequenceBlock`) does a `Peek` on a key with count 0, concludes the antecedent never ran, and **allows the call** — failing open.

The practical attack scenario: if an attacker can induce brief transient errors in the Redis backend (e.g. a short network partition or a targeted key eviction), they can bypass `sequenceBlock` constraints for the duration of the fault window. No audit record distinguishes this from a legitimate session that did not call the antecedent.

## Contrast with `handleMaxCalls`

The `handleMaxCalls` handler (line 278-298) correctly fails **closed** on counter errors:

```go
count, err := e.counter.IncrementAndGet(ctx, key, mc.WindowSeconds)
if err != nil {
    return &ConditionError{Code: ErrCodeConditionFailed, ...}
}
```

`recordSessionCall` is the only counter operation that deviates from this pattern.

## Considered alternatives

| Approach | Trade-off |
|---|---|
| Deny the antecedent call on record failure | Turns counter backend failure into a blanket deny for all tool calls — may be too broad |
| Return an error from `recordSessionCall` and propagate it to `ValidateAction` / `EvaluateConditions` as a deny | Targeted: only affects sessions where the antecedent fired; consistent with fail-closed; adds one error check in two callers |
| Write a distinct audit event on recording failure | Defense-in-depth — at minimum, make degradation visible |

The middle option is likely the right fix: return `error` from `recordSessionCall`, check it in both `ValidateAction` (line 301) and `EvaluateConditions` (line 457), and if non-nil, return a denial with code `ErrCodeConditionFailed` and message "session history recording failed; sequenceBlock state is unreliable". This is consistent with how every other stateful condition treats backend errors.

## Location

- `pkg/enforcement/engine.go:180-204` — `recordSessionCall`
- `pkg/enforcement/engine.go:299-301` — caller in `ValidateAction`
- `pkg/enforcement/engine.go:455-457` — caller in `EvaluateConditions`

eunolabs:
## Summary

Fixes #213.

`recordSessionCall` (`pkg/enforcement/engine.go`) silently discarded call-counter write errors:

```go
```

This conflicted with the **fail-closed** invariant. If the counter backend (Redis or in-memory) failed at the moment an allowed *antecedent* tool ran, the session-history marker was never written. A later `sequenceBlock`-protected call would `Peek` a key with count 0, conclude the antecedent never ran, and **allow the call — failing open**. An attacker able to induce a brief transient counter fault (a short network partition, a targeted key eviction) could bypass `sequenceBlock` for the duration of the fault, with no audit record distinguishing it from a legitimate session that never called the antecedent.

## Fix

Adopts the issue's recommended "middle option":

- `recordSessionCall` now returns an `error`.
- Both callers — `ValidateAction` and `EvaluateConditions` — propagate a non-nil error as a deny with code `CONDITION_FAILED`, `conditionType: sequenceBlock`, and message `session history recording failed; sequenceBlock state is unreliable`.
- This mirrors how `handleMaxCalls` and `handleSequenceBlock` already fail closed on counter-backend errors.

The early-return guard cases (dry-run, no counter, no session ID, a counter without `CallHistory`, empty tool name) remain legitimate "nothing to record" states and return `nil` — only an actual failed write denies. The change is **targeted**: it only affects sessions that reach the record path on an allowed call when the backend is genuinely faulting.

## Tests

`pkg/enforcement/sequence_block_test.go`:

- `TestSequenceBlock_RecordFailureFailsClosed` — `ValidateAction` denies when the record write fails (new `recordingErrorCounter` implements `CallHistory` so the write is reached).
- `TestSequenceBlock_RecordFailureFailsClosed_EvaluateConditions` — same for the `EvaluateConditions` entry point.
- `TestSequenceBlock_RecordSkipWithoutHistoryStillAllows` — boundary control: a counter without `CallHistory` is never written, so no recording failure can arise and the call still proceeds.

## Docs

`docs/capability-manifest-guide.md` — the `sequenceBlock` section now documents that recording an allowed antecedent is fail-closed.

## Verification

- `make test` (`go test -race -count=1 ./...`) — all packages pass.
- `make lint` (go vet + golangci-lint) — 0 issues.
- `make check-fmt`, `make check-license` — clean.

https://claude.ai/code/session_01VLFwACULrwiAdVRZbhMLc5

---
_Generated by [Claude Code](https://claude.ai/code/session_01VLFwACULrwiAdVRZbhMLc5)_


pkg/enforcement/engine.go
```diff
@@ -203,28 +203,42 @@ var _ capability.Enforcer = (*Engine)(nil)
...
 	if _, ok := e.counter.(CallHistory); !ok {
-		return
+		return nil
 	}
 	tool := stripEnginePrefix(req.ToolName)
 	if tool == "" {
-		return
+		return nil
+	}
+	if _, err := e.counter.IncrementAndGet(ctx, sequenceHistoryKey(req.SessionID, tool), sequenceHistoryWindowSec); err != nil {
```
eunolabs: **Recording is indiscriminate — the new fail-closed deny covers every allowed call, not just sequence antecedents.**

`recordSessionCall` runs after *every* allowed call and writes a marker for the called tool regardless of whether that tool is ever named in a `sequenceBlock.AfterTools` set. So once a `CallHistory` counter is configured (both built-in backends implement it), a write fault here denies **every** allowed call in any session carrying a `sessionId` — not only calls to tools that actually gate a sequence.

That is the correct fail-closed behavior and consistent with the invariant, so I'm not asking to change it. But worth noting: issue #213's decision table framed this "middle option" as *"targeted: only affects sessions where the antecedent fired."* As implemented, every allowed call is a potential antecedent, so the blast radius under (say) a brief Redis partition is effectively the same as the "blanket deny for all tool calls" option it was contrasted against — and broader than `maxCalls`, which only denies the specific calls carrying that condition. Two ways to square it:
- Accept it, and make the PR/issue wording explicit that the deny scope is "any allowed call while the counter backend is faulting"; or
- Bound it by recording only when the called tool actually appears in some `sequenceBlock.AfterTools` in the active manifest. Feasible in `ValidateAction` (it has the full constraint list) but not in `EvaluateConditions` (only the matched constraint), so it's a non-trivial change.

Nit on the same lines: `if _, err := …; err != nil { return err }` + `return nil` collapses to `_, err := …; return err`.

---

pkg/enforcement/engine.go
```diff
@@ -321,8 +335,21 @@ func (e *Engine) ValidateAction(ctx context.Context, req *capability.EnforceRequ
...
+	// closed (see recordSessionCall): a missing marker would let a later
+	// sequenceBlock Peek fail open, so the otherwise-allowed call is denied.
+	if err := e.recordSessionCall(ctx, req); err != nil {
+		return capability.EnforceResponse{
+			RequestID: requestID,
+			Decision:  capability.DecisionDeny,
+			DecidedAt: now,
+			Denial: &capability.DenialInfo{
+				Code:          capability.ErrCodeConditionFailed,
+				ConditionType: capability.ConditionTypeSequenceBlock,
```
eunolabs: The denial is labeled `conditionType: sequenceBlock` even when the call that failed to record carries no `sequenceBlock` condition (e.g. `read_credentials` in the new test has zero conditions). An operator triaging the audit log will then see a `sequenceBlock` denial on a tool that isn't gated by any sequence.

The distinct message disambiguates it and `sequenceBlock` is the most informative label available, so this is defensible. Just flagging it may surprise audit triage — if you want it unambiguous, a detail field (e.g. `Details: {"phase": "record"}`) would separate "recording failed" from an actual sequence hit without adding a top-level audit field.

---

pkg/enforcement/engine.go
```diff
@@ -477,8 +504,21 @@ func (e *Engine) EvaluateConditions(ctx context.Context, req *capability.Enforce
 	}
 
 	// Every condition passed: record the allowed call in session history for
-	// later sequenceBlock lookups (see recordSessionCall).
-	e.recordSessionCall(ctx, req)
+	// later sequenceBlock lookups (see recordSessionCall). A write failure fails
+	// closed for the same reason as in ValidateAction — a missing marker would
+	// let a later sequenceBlock Peek fail open.
+	if err := e.recordSessionCall(ctx, req); err != nil {
```
eunolabs, line 587: Optional: this deny `EnforceResponse` is now byte-for-byte identical to the one just added in `ValidateAction`. A tiny helper — `func (e *Engine) denyRecordFailure(requestID, now string) capability.EnforceResponse` — would keep the two record-failure paths (same code, same message) from drifting. Consistent with the existing inline-deny style in these two functions, so purely a judgment call.

---

eunolabs (commented):
Reviewed the fix for #213 — making `sequenceBlock` session-history recording fail closed.

**This is correct and well-executed.** It closes the real fail-open: a swallowed counter-write error left the antecedent marker unwritten, and a later `Peek` on that empty key concluded the antecedent never ran and allowed the gated call. The fix follows the issue's recommended middle option, mirrors the existing `handleMaxCalls` / `handleSequenceBlock` fail-closed precedent, covers both entry points (`ValidateAction` + `EvaluateConditions`), pins the no-`CallHistory` boundary with a test so the change can't over-deny, and updates the manifest guide. The rewritten `recordSessionCall` doc comment is a real improvement — it explains the *why*.

No blocking issues. A few advisory notes inline:
1. The recording (and therefore the new deny) fires for **every** allowed call, not just sequence antecedents — so the availability blast radius under a counter-backend outage is broader than the issue's "targeted" framing implies. Worth being explicit about, and possibly worth bounding. (Not a correctness problem — fail-closed is the invariant.)
2. `conditionType: sequenceBlock` on a tool with no sequence condition may surprise audit triage.
3. Two small cleanups: a one-line simplification in `recordSessionCall`, and optional de-duplication of the now-identical deny response across the two callers.

CI is still queued; I'll re-review once it reports and on any new pushes.

---

eunolabs:
Thanks for the review — all three advisory notes addressed:

**1. Indiscriminate recording / blast radius + nit (`b318ca8`)** — Took option 1: accepted the fail-closed scope and made it explicit. The `recordSessionCall` doc comment and the manifest guide now state that recording is indiscriminate, so a counter-backend outage denies **every** allowed call in a session carrying a `sessionId`, not only sequence antecedents. Left bounding out deliberately: it isn't reachable from `EvaluateConditions` (which sees only the matched constraint), so it would either diverge between the two entry points or require plumbing the antecedent set through the PDP — against the project's preference for one simple, consistent fail-closed rule. Nit applied (`_, err := …; return err`).

**2. `conditionType: sequenceBlock` surprising audit triage (`b936d76`)** — Added `Details: {"phase": "record"}` so a recording-path failure is machine-distinguishable from an actual sequenceBlock hit (which instead populates `afterTool`/`blockedTool`). Reuses the existing `DenialInfo.Details` field, already carried into the audit record on every deny path — no new top-level audit field. Kept `conditionType: sequenceBlock` as the most informative label.

**3. De-dup the identical deny response (`b318ca8`)** — Extracted `recordFailureDenial(requestID, now)`; both `ValidateAction` and `EvaluateConditions` return it verbatim, so the two record-failure paths can't drift.

No behavior change vs. the original fix — all gates green locally (`go test -race ./...`, `go vet`, golangci-lint, gofmt, license).

---

PR #267, opened 2026-06-13
fix: namespace sequenceBlock history key by target type (#221)

### LINKED ISSUE #221: bug: sequenceHistoryKey not namespaced by target type — tool and prompt with the same bare name share one seq: bucket
## Summary

This is the `sequenceBlock` analogue of #205 (which fixed the same bug shape in `maxCalls`). `sequenceHistoryKey` in `pkg/enforcement/engine.go` builds its per-session call-history key from the **bare** target name only:

```go
// pkg/enforcement/engine.go
}
```

`tool` is `stripEnginePrefix(req.ToolName)`, and `req.ToolName` is the bare name of the target — the tool name for `tools/call`, the prompt name for `prompts/get`, the URI for `resources/read`. So a **tool named `export`** and a **prompt named `export`** both record into the same `seq:<session>:export` bucket.

## Impact

`recordSessionCall` (`engine.go`) writes the marker on every allowed call, and `handleSequenceBlock` (`pkg/enforcement/handlers.go`) reads it back by bare name from the condition's `afterTools` list. Consequently:

- A `sequenceBlock` keyed on prior tool `export` is tripped by a `prompts/get` for a prompt *also* named `export` (false positive — a call is blocked that shouldn't be).
- Conversely, the marker carries no type, so the two namespaces cannot be distinguished when authoring `afterTools`.

Resource URIs are effectively immune for the same reason as in #205 — their scheme prefix (`file://`, `https://`) never collides with a bare tool/prompt name. The realistic collision is tool-vs-prompt.

## Fix sketch

Mirror the #205 fix: include the target type in the recorded key and in the lookup. The wrinkle is that `afterTools` entries name *tools* by bare name (or engine-prefixed form), so the lookup side needs the antecedent's type. Options:

1. Record under `seq:<session>:<type>:<bare>` and resolve each `afterTools` entry's type from its prefix (`tool:` / `prompt:` / `resource:`), defaulting to `tool` for an unprefixed entry to preserve today's authoring ergonomics.
2. Record one marker per `(type, bare)` and have `handleSequenceBlock` peek the type implied by the prefix on each `afterTools` entry.

Either way this is a breaking change to the runtime `seq:` key schema (in-flight session history resets on upgrade), acceptable under the pre-1.0 no-back-compat-shim policy — a CHANGELOG note suffices, same as #205.

## Context

Split out of #205 / PR #209, where this was deliberately scoped out to keep that fix focused on `maxCalls`. Filed so the second half of the same bug shape isn't lost.

eunolabs:
## Summary

Fixes #221 — the `sequenceBlock` analogue of the `maxCalls` fix in #205/#209.

`recordSessionCall` built the per-session call-history key from the **bare** target name only (`seq:<session>:<name>`). That name is the tool name for `tools/call`, the prompt name for `prompts/get`, and the URI for `resources/read`, so a tool named `export` and a prompt named `export` both recorded into `seq:<session>:export` and shared one history bucket. Consequences:

- A `sequenceBlock` keyed on the tool `export` was tripped by a `prompts/get` for a same-named prompt — a false-positive deny.
- The marker carried no type, so the two namespaces could not be told apart when authoring `afterTools`.

## Fix

Include the target type in the key (`seq:<session>:<type>:<name>`), mirroring #205:

- **Record side** (`recordSessionCall`): key the antecedent under `req.Target.Type`, which every `ManifestPDP` entry point (`Decide` / `DecideResourceRead` / `DecidePromptGet`) populates. A direct `ValidateAction` caller that leaves `req.Target` nil falls back to the type implied by any engine prefix on `req.ToolName`, then to `tool` — keeping the record and lookup sides symmetric for existing bare-name callers.
- **Lookup side** (`handleSequenceBlock`): resolve each `afterTools` entry's namespace from its prefix, defaulting a bare entry to the tool namespace. So `afterTools: [export]` is armed only by the **tool** `export` and `afterTools: [prompt:export]` only by the **prompt**. Resource URIs stay immune as before — their scheme prefix (`file://`, `https://`) never matches a bare name.
- A new `splitEnginePrefix` returns `(type, bare)` and `stripEnginePrefix` now delegates to it, keeping the recognized-prefix list in one place.

This is a **breaking change to the runtime `seq:` key schema**: in-flight session history resets on upgrade (in-memory / short-TTL Redis entries), acceptable under the pre-1.0 no-back-compat-shim policy — same as #205. A CHANGELOG note documents it.

## Tests / docs

- New four-way regression test in `pkg/enforcement/sequence_block_test.go` covering each `{tool, prompt}` antecedent against each `{tool-keyed, prompt-keyed}` block — both the false-positive that's now fixed and the still-correct matching-namespace blocks.
- All existing `sequenceBlock` tests pass unchanged (they exercise the nil-`Target` / bare-name path).
- CHANGELOG entry alongside the `maxCalls` one; `docs/capability-manifest-guide.md` updated to document the `afterTools` namespace-selection semantics.
- `make check-fmt`, `make check-license`, `make lint` (`go vet`), and `go test -race ./pkg/... ./cmd/...` all green locally.

Fixes #221

https://claude.ai/code/session_012BNJbwrJku1iA1cu9AsJ3o

---
_Generated by [Claude Code](https://claude.ai/code/session_012BNJbwrJku1iA1cu9AsJ3o)_


pkg/enforcement/handlers.go
```diff
@@ -815,11 +815,16 @@ func (e *Engine) handleSequenceBlock(ctx context.Context, cond capability.Condit
 
 	blockedTool := stripEnginePrefix(req.ToolName)
 	for _, prior := range sb.AfterTools {
-		priorTool := stripEnginePrefix(prior)
+		// Resolve each antecedent's namespace from its prefix (bare entries
+		// default to "tool"), mirroring how recordSessionCall keyed the call.
+		// This is what keeps a tool and a same-named prompt in separate buckets:
+		// afterTools: [export] matches only the tool, afterTools: [prompt:export]
+		// only the prompt (#221).
+		priorType, priorTool := splitEnginePrefix(prior)
```
eunolabs, line 902: Minor (non-blocking, audit clarity): you now have `priorType` here, but the denial just below (the `count > 0` branch, outside this diff) never uses it — now that an antecedent can legitimately be a prompt or resource, that denial loses the namespace that made the match:

- The message hard-codes `"tool %q was already called in this session…"`, which is inaccurate when `priorType != "tool"` (a `prompt:export` antecedent reads as `tool "export"`).
- `Details["afterTool"]` is the bare `priorTool` (prefix stripped), so a block that fired on `prompt:export` records `afterTool: "export"` with no type. The raw `afterTools` list is still in `Details`, so it isn't fully lost — but the one field identifying *which* antecedent matched is ambiguous if a manifest ever lists both `export` and `prompt:export`.

Consider threading `priorType` into the denial — e.g. report the namespaced form in `afterTool` and/or phrase the message by target type. Keeps audit records unambiguous about which namespace tripped the block, which is the point of #221. The keying itself is correct; this is purely about the denial's self-description.

---

eunolabs (commented):
## Review — namespace `sequenceBlock` history by target type (#221)

Reviewed against the actual codebase (checked out `claude/nice-pascal-z5mce4`). The fix is correct and the approach mirrors the `maxCalls` fix in #205/#209 cleanly.

**Core invariant verified — record↔lookup symmetry holds for every production namespace.** The new key is `seq:<session>:<type>:<name>`, and for it to be sound the type token recorded by `recordSessionCall` must equal the type `splitEnginePrefix` derives from each `afterTools` entry:

- `splitEnginePrefix` recognizes exactly `tool`/`resource`/`prompt`/`system` and defaults bare → `tool`.
- `EnforceRequestTarget.Type` is set from `string(TargetType)` and the constants (`constraint.go:18-23`) are exactly those four tokens.
- Every `ManifestPDP` entry point sets `Target.Type`: `Decide` → `"tool"` with `ToolName = target.Name` (`pdp.go:237-244`), `DecidePromptGet` → `"prompt"` (`pdp.go:502-504`), `DecideResourceRead` → `"resource"` with `ToolName = uri` (`pdp.go:407-413`). So a bare `afterTools: [export]` matches only the tool, `prompt:export` only the prompt, and the documented `resource:file://…` antecedent records/matches symmetrically (URI scheme prefix is never a recognized type, so the bare half stays intact).

**Fallback path is test-only.** The prefix-derived-type fallback (when `req.Target == nil`) is the only place behavior changes for a prefixed `ToolName`, and the only production callers of `EvaluateConditions` — `pdp.go:256/422/514` — all set `Target`. `ValidateAction` with a nil `Target` is exercised solely by tests. So no production regression.

**Other checks:** fail-closed preserved (empty bare name still skips record and `continue`s on lookup); the runtime-key break is documented in CHANGELOG + guide, acceptable under the pre-1.0 no-shim policy; `go test -race -run TestSequenceBlock ./pkg/enforcement/` passes and `go vet` is clean locally. Regression test covers the full {tool,prompt}×{tool-keyed,prompt-keyed} matrix including the false-positive that's now fixed.

One minor, non-blocking audit-clarity note left inline on `handlers.go` (the denial message/`afterTool` detail still hard-code "tool" / drop the namespace). Not a correctness issue — the keying is right.

LGTM aside from that nit.

---

pkg/enforcement/handlers.go
eunolabs: Done in d86d386. The `count > 0` denial now threads `priorType` through: the message is phrased by target type (`%s %q was already called …`) and `Details["afterTool"]` carries the namespaced form (`tool:export`, `prompt:export`). Added message + `afterTool` assertions to the regression test for both the tool-keyed and prompt-keyed cases, and noted the namespaced denial in the CHANGELOG entry.

---

PR #281, opened 2026-06-13
docs: correct resourceScoreWeight comments to match implementation

### LINKED ISSUE #275: docs: resourceScoreWeight comment references unimplemented "action specificity" scoring
## Summary

`resourceScoreWeight` in `pkg/enforcement/engine.go` is documented as:

```go
// resourceScoreWeight weights resource specificity 10× more than action specificity
// so that a narrower resource pattern always beats a broader one, regardless of
// how specific the action match is.
const resourceScoreWeight = 10
```

Action specificity is **never scored** in the implementation.  `actionMatchScore` returns a plain boolean (permit/deny); its result is not multiplied into any numeric score:

```go
score := resourceSpecificity(bare, bareToolName) * resourceScoreWeight
// actionMatchScore contributes 0 to the score — it is purely a filter
```

The phrase "10× more than action specificity" implies there is an action score that `resourceScoreWeight` dominates.  There is not.  The correct explanation is that `resourceScoreWeight` amplifies the resource score so that two resources with different specificity never tie even if future action scoring is added with a small weight.

The same stale language appears in the `findMatchingCapability` function comment: "Tie-breaking is stable: if two capabilities have identical resource **and action** specificity scores…" — action specificity is not currently computed.

## Suggested fix

Update both comments to describe what the code actually does:

```go
// resourceScoreWeight scales the resource-specificity score before comparison.
// It is large enough that any two resources with different specificity will
// never tie, regardless of any future per-action score that might be added.
const resourceScoreWeight = 10
```

And in `findMatchingCapability`:

```go
// Tie-breaking is stable: if two capabilities have identical resource
// specificity scores, the one that appears first in the list wins.
```

No behavior change; comment-only fix.

eunolabs:
## Summary

Fixes the stale comments flagged in #275. The `resourceScoreWeight` constant and the `findMatchingCapability` function in `pkg/enforcement/engine.go` were documented as if "action specificity" were scored and dominated by a 10× weight. It is not — `actionMatchScore` returns a plain permit/deny boolean and contributes nothing to the numeric score; only resource specificity is scored.

This updates both comments to describe what the code actually does:

- `resourceScoreWeight` **scales** the resource-specificity score so that two resources with different specificity never tie, regardless of any future per-action score that might be added.
- Tie-breaking falls back to list order when two capabilities have identical **resource** specificity scores (dropping the inaccurate "and action" phrasing).

Comment-only change; **no behavior change**.

## Verification

- `gofmt -l` clean
- `go build ./pkg/enforcement/` OK
- `go vet ./pkg/enforcement/` OK

Fixes #275

https://claude.ai/code/session_015dqHomQ2trMqtNoEuU9X5k

---
_Generated by [Claude Code](https://claude.ai/code/session_015dqHomQ2trMqtNoEuU9X5k)_


pkg/enforcement/engine.go
```diff
@@ -440,12 +440,12 @@ func (e *Engine) FindMatchingCapability(req *capability.EnforceRequest, capabili
 // noMatchScore is the sentinel value for "no matching capability found yet".
 const noMatchScore = -1 << 30
 
-// resourceScoreWeight weights resource specificity 10× more than action specificity
-// so that a narrower resource pattern always beats a broader one, regardless of
-// how specific the action match is.
+// resourceScoreWeight scales the resource-specificity score before comparison.
+// It is large enough that any two resources with different specificity will
+// never tie, regardless of any future per-action score that might be added.
```
eunolabs: Non-blocking nit on precision (the comment is already a clear improvement over the old text):

Scaling the resource score by a positive constant doesn't by itself make two resources "never tie" — `s1 != s2` implies `s1*10 != s2*10`, so unequal resource scores already differ at any weight (even `weight = 1`), and equal ones stay equal. The weight's real job is to reserve the low-order range `[0, weight)` as headroom for a *future additive* per-action term, so such a term could break ties **only** among equal-resource capabilities and never reorder unequal-resource ones.

Relatedly, "regardless of any future per-action score that might be added" slightly over-promises: it only holds while that future score stays below the weight. The smallest gap between two resource-specificity scores is 1 (e.g. `prefixLen*10 - wildcardCount` for `"a*"` = 9 vs `"a**"` = 8), which becomes 10 after `* resourceScoreWeight`, so an action term in `[0, 9]` is safe but `>= 10` could overtake a one-step resource gap.

Suggested tightening:

```go
// resourceScoreWeight scales the resource-specificity score, reserving the
// low-order range [0, resourceScoreWeight) as headroom for a possible future
// additive per-action score. That keeps any such action term able to break
// ties only between equal-resource capabilities, never to reorder two
// resources whose specificity already differs (the smallest such gap is 1,
// i.e. resourceScoreWeight after scaling).
```

Purely a doc refinement — happy to see this merge as-is.

---

eunolabs (commented):
## Review: accurate fix for the stale comments in #275

I verified both new comments against the implementation in `pkg/enforcement/engine.go`:

- **`actionMatchScore` contributes nothing to the score.** It returns a plain `bool` and is used purely as a filter (`if !actionMatchScore(constraint.Actions, req) { continue }`). The score is computed as `resourceSpecificity(bare, bareToolName) * resourceScoreWeight` — only resource specificity is scored. So the old "10× more than action specificity" / "identical resource **and action** specificity scores" language described scoring that does not exist. ✅
- The two replacement comments now match the code. ✅
- Comment-only, no behavior change; title is correctly scoped (`docs:`), and `gofmt`/`vet`/`build` are clean per the description. CI is currently queued (expected to pass for a comment-only diff).

**Verdict: LGTM** — fix is correct and faithful to the issue. (I can't formally Approve since this PR is authored by the same account I'm acting as.) One non-blocking precision nit left inline on the `resourceScoreWeight` comment — the "never tie regardless of any future per-action score" framing slightly over-promises (it holds only while a future action term stays below the weight). Fine to merge as-is; the suggestion is optional.

---

pkg/enforcement/engine.go
eunolabs: Good catch — you're right on both points. Scaling by a positive constant doesn't create the non-tie property (unequal resource scores already differ at any weight, even 1); the weight's actual job is to reserve `[0, resourceScoreWeight)` as headroom for a future additive per-action term, and "never tie regardless of any future score" over-promised since it only holds while that term stays below the weight.

I verified the smallest-gap claim against `resourceSpecificity`: the wildcard branch returns `prefixLen*10 - wildcardCount`, so `"a*"` = 9 and `"a**"` = 8 — a gap of 1, which becomes `resourceScoreWeight` (10) after scaling. Adopted your suggested wording verbatim in bfcf8b8.

---

PR #282, opened 2026-06-13
fix: sort property names in schemaValidateObject for reproducible validation errors

### LINKED ISSUE #269: bug: schemaValidateObject iterates Properties map non-deterministically — validation errors are not reproducible
## Summary

`schemaValidateObject` in `pkg/enforcement/schema.go` iterates `s.Properties` (a `map[string]*ArgumentSchema`) using a plain `range` loop.  Go map iteration order is non-deterministic, so when an input object violates multiple property schemas at once, the error that surfaces depends on which key the runtime happens to visit first — it can differ between two runs of the exact same request.

`compileSchemaPatterns` in the same file already handles this correctly: it builds a sorted `names` slice before iterating.  `schemaValidateObject` should do the same.

## Relevant code

```go
// pkg/enforcement/schema.go – schemaValidateObject (line ~228)
for name, propSchema := range s.Properties {   // non-deterministic order
    propVal, ok := v[name]
    if !ok {
        continue
    }
    if err := schemaValidateValue(p+"."+name, propVal, propSchema); err != nil {
        return err   // first error wins — but "first" is random
    }
}
```

Compare with `compileSchemaPatterns` (line ~74) which sorts first:
```go
names := make([]string, 0, len(s.Properties))
for name := range s.Properties { names = append(names, name) }
sort.Strings(names)
for _, name := range names { ... }
```

## Impact

* Operators debugging a rejected call may see different `argumentSchema` denial messages on repeated attempts against the same manifests and payloads.
* Integration tests that assert exact error text become flaky if a test schema has more than one violated property.
* Audit records for the same bad input look different across runs, complicating forensic analysis.

## Fix

Sort property names (as `compileSchemaPatterns` already does) before ranging:

```go
names := make([]string, 0, len(s.Properties))
for name := range s.Properties { names = append(names, name) }
sort.Strings(names)
for _, name := range names {
    propVal, ok := v[name]
    if !ok { continue }
    if err := schemaValidateValue(p+"."+name, propVal, s.Properties[name]); err != nil {
        return err
    }
}
```

A new table-driven test should verify that two properties that both fail produce the same (lexicographically-first-named) error on every run.

eunolabs:
## Summary

Fixes #269.

`schemaValidateObject` in `pkg/enforcement/schema.go` iterated `s.Properties` (a `map[string]*ArgumentSchema`) with a plain `range` loop. Go map iteration order is non-deterministic, so when an input object violated more than one property schema at once, the error that surfaced depended on which key the runtime happened to visit first — it could differ between two runs of the exact same request. `compileSchemaPatterns` in the same file already sorts before iterating; this brings `schemaValidateObject` in line.

While here, the `additionalProperties` check had the same root cause: it ranged over the input object's keys, so with several disallowed extra properties the reported one was also non-deterministic. Both loops now sort first.

## Why it matters

- Operators debugging a rejected call could see different `argumentSchema` denial messages on repeated attempts against the same manifest and payload.
- Integration tests asserting exact error text became flaky when a schema had more than one violated property.
- Audit records for the same bad input looked different across runs, complicating forensic analysis.

## Changes

- `pkg/enforcement/schema.go` — sort property names before validating each property; sort input keys before the `additionalProperties` check. Reports the lexicographically-first violation every run.
- `pkg/enforcement/argument_schema_test.go` — table-driven `TestValidateArgumentSchema_DeterministicMultiViolation` covering multiple `minLength` violations, mixed violation kinds (enum/maximum/pattern), and multiple additional properties. Each case runs 200 iterations and asserts identical error text every time, which would catch a regression to plain map ranging.

## Verification

```
gofmt -l pkg/enforcement/schema.go pkg/enforcement/argument_schema_test.go   # clean
go vet ./pkg/enforcement/                                                     # clean
go test -race -count=1 ./pkg/enforcement/                                     # ok
```

https://claude.ai/code/session_018gCLHrPYDxUiNKibxvZrN9

---
_Generated by [Claude Code](https://claude.ai/code/session_018gCLHrPYDxUiNKibxvZrN9)_


pkg/enforcement/schema.go
```diff
@@ -230,17 +230,34 @@ func schemaValidateObject(p string, v map[string]interface{}, s *capability.Argu
...
 		}
 	}
 	if s.AdditionalProperties != nil && !*s.AdditionalProperties {
+		// Sort the input keys for the same reason: with several disallowed extra
+		// properties, report the same (lexicographically-first) one every run.
+		extras := make([]string, 0, len(v))
 		for name := range v {
+			extras = append(extras, name)
+		}
+		sort.Strings(extras)
```
eunolabs: Optional nit (non-blocking): `extras` is built from *every* key in `v` and sorted in full, then the known-property keys are skipped only inside the loop below — so the name is slightly misleading (it holds all keys, not just extras) and the sort runs over more elements than necessary. Filtering to genuinely-extra keys before sorting makes the name accurate and shrinks the sort:

```go
var extras []string
for name := range v {
    if _, ok := s.Properties[name]; !ok {
        extras = append(extras, name)
    }
}
sort.Strings(extras)
if len(extras) > 0 {
    return fmt.Errorf("%s: additional property %q is not allowed", p, extras[0])
}
```

Functionally identical — still reports the lexicographically-first disallowed key every run. Purely a readability/efficiency tidy; fine to leave as-is.

---

eunolabs (commented):
## Review — correct, complete, and well-tested

This is a clean, well-targeted fix that matches the approach in #269 exactly, and correctly extends it to the second non-deterministic map iteration in the same function — the `additionalProperties` check — which has the identical root cause. Good catch including it; the issue only called out the `Properties` loop.

Determinism is now complete for `schemaValidateObject`: it walks `required` (a slice, already ordered) → sorted `Properties` → sorted extra keys, and because validation recurses through `schemaValidateValue`, nested objects are deterministic too. Note the ordering across categories is a fixed precedence (required → properties → additional) rather than globally lexicographic — that's deterministic and sensible, just worth being explicit that "lexicographically-first" applies within each loop.

I checked out the PR branch and verified locally:

- `gofmt -l pkg/enforcement/schema.go pkg/enforcement/argument_schema_test.go` — clean
- `go vet ./pkg/enforcement/` — clean
- `go test -race -count=1 ./pkg/enforcement/` — passes, including the new `TestValidateArgumentSchema_DeterministicMultiViolation`

The test is solid: mixed violation kinds (minLength / enum / maximum / pattern) plus multiple additional properties, each asserting identical error text across 200 iterations — that would reliably surface a regression to plain map ranging.

One optional, non-blocking nit left inline on the `additionalProperties` block. Nothing here blocks merge.

Non-blocking observation: unlike `compileSchemaPatterns` (which sorts once at manifest-load time), `schemaValidateObject` runs on the per-request enforcement hot path, so this adds a small slice allocation + sort per validated object. For operator-defined schemas with small property counts that's negligible — just flagging the context difference from the cited precedent.

---

pkg/enforcement/schema.go
eunolabs: Good call — applied in 47c4208. Now filters to genuinely-extra keys before sorting, so the name is accurate and the sort runs over fewer elements on the hot path (which also addresses the perf observation in the summary). Reports the lexicographically-first disallowed property every run, as before.

---

eunolabs: Re-reviewed at 47c4208 — looks good. Verified locally on the updated commit: `gofmt -l` clean, `go vet ./pkg/enforcement/` clean, and `go test -race -count=1 ./pkg/enforcement/` passes. The refactor reads cleanly and preserves the lexicographically-first-disallowed-key behavior. No further comments from me; this is good to merge.

---

PR #283, opened 2026-06-13
chore: remove unreachable score-900 branch in resourceSpecificity

### LINKED ISSUE #274: chore: score-900 branch in resourceSpecificity is unreachable dead code
## Summary

`resourceSpecificity` in `pkg/enforcement/engine.go` has a branch that returns `900` for patterns that contain no wildcards but are not identical to the tool name:

```go
func resourceSpecificity(resource, toolName string) int {
    if resource == toolName {
        return 1000
    }
    if !strings.ContainsAny(resource, "*?[") {
        return 900   // ← unreachable
    }
    ...
}
```

This branch is unreachable because:

1. `resourceSpecificity` is only called from `findMatchingCapability` for resources that already passed `matchesResource(bare, bareToolName)`.
2. `matchesResource` fast-paths on `resource == toolName` (already handled above) and `resource == "*"` (contains `*`, so wildcard), then falls through to `path.Match`.
3. When `resource` has no wildcards (`*`, `?`, `[`), `path.Match(resource, toolName)` is true **only** when `resource == toolName` — which is already handled by the `return 1000` case.
4. Therefore `resourceSpecificity` is never called with a no-wildcard resource that is not equal to the tool name.

## Suggested fix

Remove the dead branch and add a comment explaining why the only non-wildcard case is the exact-match case above:

```go
func resourceSpecificity(resource, toolName string) int {
    if resource == toolName {
        return 1000
    }
    // Any resource that reaches here contains a wildcard (matchesResource
    // filtered out non-matching literals above).
    prefixLen := ...
    wildcardCount := ...
}
```

A code-comment-only change and a test asserting that `resourceSpecificity("read_file", "write_file")` is never reached (via `FindMatchingCapability`) would guard against regression.

eunolabs:
## Summary

Fixes the dead-code branch flagged in #274. `resourceSpecificity` in `pkg/enforcement/engine.go` had a branch returning `900` for a resource that contains no `*?[` wildcard but is not identical to the tool name. Through the engine's selection path that input is never scored:

- `resourceSpecificity` is only scored after a `matchesResource` check at every call site — `findMatchingCapability` (engine), and the proxy mirrors `bareSpecificity`/`resSpecificity` in `cmd/eunox-mcp/pdp.go` and `drift.go`, which are all guarded by `matchBare`/`matchResource` first.
- `matchesResource` admits a wildcard-free literal only when `path.Match` reports a match; for an *unescaped* literal that happens only when it equals the tool name, which the `return 1000` exact-match branch above already handles.

So the `900` branch was unreachable for any literal a manifest actually writes.

## Edge case (per review)

`path.Match` honors `\` escapes (it is OS-independent, unlike `filepath.Match`), so a backslash-bearing literal such as the pattern `a\b` matches the *distinct* string `ab` and reaches `resourceSpecificity` without containing `*?[`. The removed `900` branch was reachable for that degenerate input, so this is **ranking-preserving** rather than strictly behavior-preserving: such an input now scores `prefixLen*10` instead of `900`. It cannot reorder a real ranking — an escaped literal's rune count (the `\` included) is `>=` the matched name's length, so it still outscores any glob matching the same name. `TestResourceSpecificity_BackslashEscapedLiteralReachesScoring` pins this.

## Changes

- **`pkg/enforcement/engine.go`** — remove the dead `return 900` branch and document why only the exact-match case reaches scoring as a literal, explicitly noting the backslash-escape edge case.
- **`pkg/enforcement/coverage_test.go`** — update the `specificity scoring` assertion that pinned the old `900` value to the general `prefixLen*10 - wildcardCount` result; add `TestFindMatchingCapability_NonWildcardLiteralFilteredBeforeScoring` (the regression guard #274 requested) and `TestResourceSpecificity_BackslashEscapedLiteralReachesScoring` (the escape-case invariant).
- **`cmd/eunox-mcp/coverage_gaps_test.go`** — update `TestResSpecificity_NonWildcardNoMatch`, which also pinned `900`, to the general-formula value (`90`).

`TestResSpecificity_MatchesEngine` continues to pass because the proxy wrapper and engine still compute the identical value for every case.

## Testing

- `go build ./...`
- `go test -race -count=1 ./pkg/enforcement/ ./cmd/eunox-mcp/` — pass
- `go vet` and `gofmt` — clean

Closes #274

https://claude.ai/code/session_013S1qDqyGCXQ9jdzbyNPYyh


pkg/enforcement/engine.go
```diff
@@ -664,9 +664,15 @@ func resourceSpecificity(resource, toolName string) int {
...
 		return 1000
 	}
-	if !strings.ContainsAny(resource, "*?[") {
-		return 900
-	}
+	// Every resource that reaches this point through the engine's selection path
+	// (findMatchingCapability, mirrored by the proxy helpers in cmd/eunox-mcp)
+	// contains a glob metacharacter: scoring is always guarded by a prior
+	// matchesResource check, and a bare literal satisfies matchesResource only
+	// when it equals toolName — which the exact-match case above already scores
```
eunolabs: Minor precision nit (non-blocking): the "unreachable" claim isn't quite airtight on POSIX. `path.Match` treats `\` as an escape character (only Windows disables this), so a bare pattern containing a backslash but none of `*?[` can match a *different* string and still pass `matchesResource`:

```go
path.Match("a\\b", "ab") // => true, nil
// "a\b" != "ab", and strings.ContainsAny("a\b", "*?[") == false
```

So a constraint target like `tool:a\b` matching tool `ab` does reach `resourceSpecificity` with a non-wildcard resource that isn't equal to `toolName` — the old `return 900` branch was reachable for it (just never exercised by inputs anyone writes). After this change such an input scores `prefixLen*10` instead of `900`, so strictly speaking this is a (tiny) behavior change rather than fully behavior-preserving.

Practical impact is negligible: an escaped literal's rune count (including the `\`) is `>=` the matched name's length, so it still outscores any glob that matches the same name and ranking never flips. But since the comment makes a strong "a bare literal satisfies matchesResource only when it equals toolName" claim, consider softening it to acknowledge the backslash-escape case (e.g. "…except for backslash-escaped patterns, which no manifest uses in practice"). The removal itself is sound and the regression test is a good guard.

---

eunolabs (commented):
Reviewed this dead-code cleanup (closes #274). The core reasoning holds: every ranking call site — `findMatchingCapability` (engine), `findConstraint`/`bareSpecificity` (pdp.go), and `bestManifestConstraint`/`resSpecificity` (drift.go) — scores only after a `matchesResource`/`matchBare`/`matchResource` guard, and a non-wildcard literal passes that guard only when it equals the tool name, which the `return 1000` case above already handles. I confirmed there are no other callers of `ResourceSpecificity` that bypass a match check.

Verification:
- CI is green (42/42 checks passing).
- Test updates correctly track the general `prefixLen*10 - wildcardCount` formula: `90` (`read_file`), `5*10` (`other`), `4*10-1` (`file*`).
- The new `TestFindMatchingCapability_NonWildcardLiteralFilteredBeforeScoring` is a reasonable regression guard for the invariant behind #274.
- `arch_review_fixes_test.go` only pins wildcard/exact cases, so it's unaffected by removing the `900` branch.

One minor, non-blocking precision nit left inline: the "unreachable" / "behavior-preserving" framing is slightly overstated because `path.Match` honors `\` escapes on POSIX, so a backslash-bearing literal can reach `resourceSpecificity`. It has no realistic ranking impact, but the in-code comment could be softened to match. LGTM otherwise.

---

pkg/enforcement/engine.go
eunolabs: Good catch — fixed in 3f8e043. `path.Match` honors `\` escapes (OS-independent, unlike `filepath.Match`), so a backslash-bearing literal like `a\b` matches the distinct name `ab` and does reach `resourceSpecificity` without a `*?[` char.

- Softened the in-code comment to acknowledge the escape case instead of claiming a bare literal only matches when equal to `toolName`.
- Updated the PR description's "behavior-preserving" framing to "ranking-preserving" with the same caveat.
- Added `TestResourceSpecificity_BackslashEscapedLiteralReachesScoring` pinning your ranking argument: ``a\b`` (3 runes → score 30) still outscores every glob that matches `ab` (`*` → -1, `a?`/`a*` → 9, `ab*` → 19), so the `900`→`prefixLen*10` shift never flips a decision.

---

eunolabs:
**Re-reviewed `3f8e043`** — the fix correctly resolves my earlier nit, and I have no further review comments.

- The `engine.go` comment now accurately covers both ways a literal-but-not-equal input reaches `resourceSpecificity` (a backslash-escaped pattern that `path.Match` accepts, and a direct call to the exported `ResourceSpecificity`), instead of claiming a bare literal matches only when equal to `toolName`. The `read\_file` → `read_file` example is correct — the `\` escapes the `_`, matching the literal.
- `TestResourceSpecificity_BackslashEscapedLiteralReachesScoring` pins the ranking argument precisely: `a\b` scores `30` and still outranks every glob that matches `ab` (`*`→-1, `a?`/`a*`→9, `ab*`→19), so the `900`→`prefixLen*10` change never reorders a decision. (`require` is already imported in `coverage_test.go`, so it compiles.)

No outstanding review comments — the PR looks good to me. CI is re-running on the new commit; I'll flag here only if something fails.

---

PR #284, opened 2026-06-13
sec: match allowedTables table/column names case-insensitively (#271)

### LINKED ISSUE #271: bug: allowedTables performs case-sensitive comparison while allowedExtensions, allowedOperations, and recipientDomain all normalize case
## Summary

Every other condition handler that compares string values normalizes case before matching:

| Condition | Normalization |
|-----------|--------------|
| `allowedExtensions` | Extensions lowercased on both sides |
| `allowedOperations` | Operation uppercased, `strings.EqualFold` for comparison |
| `recipientDomain` | Domain lowercased on both sides |
| **`allowedTables`** | **No normalization — exact byte match** |

`handleAllowedTables` builds `allowedTableSet` and `colSet` directly from the raw manifest strings and compares them directly against the raw argument values:

```go
allowedTableSet := make(map[string]bool, len(at.Tables))
for _, t := range at.Tables { allowedTableSet[t] = true }   // no ToLower
...
if !allowedTableSet[access.Table] { ... }  // exact match only

colSet := make(map[string]bool, len(allowedCols))
for _, c := range allowedCols { colSet[c] = true }   // no ToLower
...
if !colSet[col] { ... }  // exact match only
```

## Impact

**False denials:** A tool that sends `"Users"` when the manifest says `tables: ["users"]` (or vice versa) is denied even though the operator's intent was clear.

**Policy bypass in case-insensitive databases:** MySQL (default), SQL Server, and many other databases treat table and column names as case-insensitive.  An operator writes `columns: { users: ["name", "email"] }` to block access to `password_hash`.  A tool request using `"Password_Hash"` or `"PASSWORD_HASH"` bypasses the column restriction because the colSet lookup is an exact-case comparison.

## Suggested fix

Normalize both sides consistently.  Table names: lowercase (matching `recipientDomain`).  Column names: lowercase as well.

```go
// normalizing allowedTableSet
for _, t := range at.Tables {
    allowedTableSet[strings.ToLower(t)] = true
}
// checking
if !allowedTableSet[strings.ToLower(access.Table)] { ... }

// normalizing colSet
for _, c := range allowedCols {
    colSet[strings.ToLower(c)] = true
}
// checking
if !colSet[strings.ToLower(col)] { ... }
```

If there is a use case where case-sensitive table/column matching is intentional (e.g., quoted PostgreSQL identifiers), that should be documented explicitly and an opt-in flag should be added rather than making the current behavior silently inconsistent.

The fix needs accompanying table-driven tests covering:
- `tables: ["users"]` matches argument `"USERS"` 
- `columns: { users: ["Name"] }` matches argument column `"name"`
- Mixed-case manifest entries still work

eunolabs:
## Summary

Fixes #271.

`handleAllowedTables` compared table names, the per-table `columns` map keys, and column names as **exact byte matches**, while every other string-comparing condition (`allowedExtensions`, `allowedOperations`, `recipientDomain`) already normalizes case. This was both a source of false denials and — on case-insensitive databases (MySQL default, SQL Server, and many others) — a **policy bypass**.

Two distinct bypass classes are fixed:

1. **Column-name comparison** — a column allowlist written `columns: { users: ["name"] }` was defeated by a request spelling the column `Password_Hash` or `PASSWORD_HASH`.
2. **Column-map lookup (the more dangerous, second-order bypass)** — the per-table restriction lookup `at.Columns[access.Table]` was *also* case-sensitive, so a manifest keyed `users` did not match an argument table `USERS`. The column ACL was then **skipped entirely**, granting full-column access. (The issue's suggested snippet only addressed class 1; this PR closes both.)

## Fix

- Lowercase both sides of every table and column comparison.
- Index the column-restriction map by lowercased table name, so a case-mismatched argument table still locates its column ACL.
- Original (un-lowercased) column lists are preserved as the map values, so denial `details` still echo the manifest exactly as the operator wrote it.

Case-sensitive identifier matching (e.g. quoted PostgreSQL identifiers) is not supported; noted in the guide as a follow-up if a use case arises.

## Tests

New table-driven test `TestAllowedTables_CaseInsensitive` in `pkg/enforcement/coverage_test.go` covering:
- lowercase/uppercase/mixed-case manifest ↔ argument table matches (allow)
- column allowlist matching in the opposite case (allow)
- mixed-case `columns` map key matching a case-mismatched argument table (allow)
- **disallowed column denied despite table-name case mismatch** (the second-order bypass guard)
- disallowed column denied regardless of column case
- the "column list required" empty-columns guard still firing under table-name case mismatch
- a genuinely disallowed table still denied regardless of case

`go test -race ./...` passes; `gofmt`/`go vet` clean; Apache-2.0 headers present.

## Docs

- `docs/capability-manifest-guide.md` — documents case-insensitive matching across the string conditions, with the SQL-identifier bypass rationale.
- `docs/threat-model-mcp.md` — condition table updated.
- `CHANGELOG.md` — Security entry under `[Unreleased]`.

This is a matching-semantics fix, not a manifest grammar change, so the spec repo (which is silent on case for all three string conditions) is left unchanged.

https://claude.ai/code/session_01LD1UVtrPs9pb7tBmbhNAkH

---
_Generated by [Claude Code](https://claude.ai/code/session_01LD1UVtrPs9pb7tBmbhNAkH)_


pkg/enforcement/handlers.go
```diff
@@ -591,13 +591,36 @@ func (e *Engine) handleAllowedTables(_ context.Context, cond capability.Conditio
...
+	// lookup below is case-insensitive too — without this, a restriction keyed
+	// "users" would not be found for an argument table "USERS", skipping the
+	// column ACL entirely. The original (un-lowercased) column lists are kept as
+	// the map values so denial details still echo the manifest as the operator
+	// wrote it.
+	var columnsByTable map[string][]string
+	if at.Columns != nil {
+		columnsByTable = make(map[string][]string, len(at.Columns))
+		for table, cols := range at.Columns {
+			columnsByTable[strings.ToLower(table)] = cols
```
eunolabs, line 616: Non-blocking consideration: lowercasing the `columns` keys into `columnsByTable` makes case-variant keys collide. If a manifest contains both `columns: { users: [...], Users: [...] }`, the second `columnsByTable[strings.ToLower(table)] = cols` silently overwrites the first — and since Go map iteration order is randomized, *which* column ACL survives is non-deterministic across runs, so a stricter list could be the one dropped.

This collision couldn't occur before the PR (the two keys addressed distinct exact-case tables). Now that `users` and `Users` are by definition the same table, having both as keys is contradictory. Given the fail-closed invariant, it'd be worth either rejecting case-colliding `columns` keys at manifest-load/validation time (surfaces the operator error — preferable to a union, which would loosen the ACL), or at minimum documenting the last-write-wins behavior.

The same theoretical collision exists for `at.Tables` above, but that's a plain set membership test so it's harmless there — no action needed. None of this blocks the bypass fix itself, which is correct.

---

eunolabs (commented):
## Review — `sec: match allowedTables table/column names case-insensitively (#271)`

The core fix is correct and addresses **both** bypass classes precisely:

1. **Comparison casing** — table names, column-map keys, and column names are now lowercased on both sides (`allowedTableSet`, `colSet`, and the per-table lookup key).
2. **Second-order bypass** — re-indexing the column-restriction map by lowercased table name (`columnsByTable`) closes the more dangerous hole where `at.Columns[access.Table]` missed on a case-mismatched table (`USERS` vs a `users` key) and skipped the column ACL *entirely*. The original (un-lowercased) lists are preserved as the map values, so denial `details` still echo the operator's casing — verified in the diff.

**Consistency claim checks out.** `allowedExtensions` and `recipientDomain` already lowercase via `strings.ToLower`, and `allowedOperations` folds case with `ToUpper`+`EqualFold`, so this aligns `allowedTables` with the rest of the string-comparing conditions rather than introducing a new convention.

**Tests are thorough** — `TestAllowedTables_CaseInsensitive` covers lowercase/uppercase/mixed-case table matches, opposite-case column matches, the load-bearing "disallowed column denied despite table-name case mismatch" guard (the second-order bypass), disallowed-column-regardless-of-case, the empty-columns guard firing under a case-mismatched table (proving the ACL is *located*, not skipped), and a genuinely-disallowed table. Allow + deny + error-code assertions throughout.

**Docs** — `docs/capability-manifest-guide.md`, `docs/threat-model-mcp.md`, and `CHANGELOG.md` (Security) are all updated, matching CONTRIBUTING's "behavior change needs docs in the same PR" rule. Treating this as a matching-semantics fix (not a grammar change) and leaving the spec repo untouched is reasonable.

One **non-blocking** consideration left inline: case-colliding `columns` map keys now silently collide (last-write-wins, non-deterministic). Worth a validation-time guard or a doc note, but it doesn't block the bypass fix.

Nothing else flagged — fail-closed posture is preserved on every branch, no new top-level audit fields, no new dependencies. Looks good to merge once you've weighed the inline note.

---

pkg/enforcement/handlers.go
eunolabs: Good catch — agreed the non-determinism (randomized map iteration deciding which ACL survives) is the real problem, not just the collision. Fixed in a3c2624 by rejecting case-colliding `columns` keys at manifest load, per your preferred option (better than a union, which would only loosen the ACL).

New `validateAllowedTables` in `cmd/eunox-mcp/manifest.go` errors at `validate`/startup when two `columns` keys fold to the same lowercased table name. As you noted, `tables` is left alone — it's a plain membership set, so a case-variant duplicate folds harmlessly. Added load-time tests (reject collision, accept distinct keys) and documented the rule in the manifest guide + changelog.

---

cmd/eunox-mcp/manifest.go
```diff
@@ -378,6 +378,34 @@ func resourceWildcardsSchemeOrAuthority(pattern string) bool {
 // almost certainly a typo (e.g. `arguments:`/`value:` instead of
 // `argument:`/`values:`). Caught here, the operator learns at validate-time
 // rather than on the first denied call.
+// validateAllowedTables rejects an allowedTables condition that fails closed or
```
eunolabs, line 381: Nit (non-blocking, introduced by this commit): inserting `validateAllowedTables` here split the `validateAllowedValues` doc comment from its function.

The block directly above this line — `// validateAllowedValues rejects an allowedValues condition that would never match... `argument:`/`values:`... rather than on the first denied call.` — has no blank line before this `// validateAllowedTables ...` line, so Go now treats the two comment blocks as a single doc comment attached to `validateAllowedTables`. As a result `validateAllowedTables`'s godoc opens with an explanation of `allowedValues`, and `func validateAllowedValues` below is left with no doc comment of its own.

Fix is purely cosmetic: move the `validateAllowedValues` comment block down so it sits directly above `func validateAllowedValues` (i.e. place the new `validateAllowedTables` comment+function either above that comment or after the `validateAllowedValues` function). No behavior impact — the collision-rejection logic itself is correct, and the load-time tests + guide/changelog updates look great.

---

eunolabs (commented):
## Re-review of a3c2624 — case-colliding `columns` keys rejected at load

The follow-up correctly resolves my earlier comment. `validateAllowedTables` folds each `columns` key to lowercase and rejects two keys that collide, with a clear, actionable error ("merge them under a single key"). Rejecting (rather than unioning, which would only loosen the ACL) is the right fail-closed choice, and leaving the `tables` membership set alone is correct since a case-variant duplicate there folds harmlessly. Load-time tests (reject collision + accept distinct keys) and the guide/CHANGELOG updates are all in place. 

One **non-blocking nit** left inline: the new function was inserted between `validateAllowedValues`'s doc comment and its declaration, which orphans that comment onto `validateAllowedTables` and leaves `validateAllowedValues` undocumented. Cosmetic only — won't be caught by CI since both functions are unexported.

CI is re-running on the new head; I'll continue to monitor. Aside from the doc-comment placement, this is good to merge.

---

PR #285, opened 2026-06-13
fix: match compound file extensions by suffix in allowedExtensions (#270)

### LINKED ISSUE #270: bug: allowedExtensions checks only the final component of compound extensions — .tar.gz is treated as .gz
## Summary

`handleAllowedExtensions` in `pkg/enforcement/handlers.go` uses `filepath.Ext(filePath)` to extract the file extension.  `filepath.Ext` returns only the **last** dot-separated component: `filepath.Ext("archive.tar.gz")` returns `".gz"`, not `".tar.gz"`.

This creates two categories of real-world policy errors:

**False deny** — An operator allows `[".tar.gz"]`.  A file named `backup.tar.gz` is extracted as `".gz"`, which is not in the allowlist, so the call is denied unexpectedly.

**Silent bypass** — An operator allows `[".gz"]` intending to permit only gzip-compressed files.  A file named `malware.exe.gz` also ends in `.gz`, so it passes.  More subtly, a policy that allows `[".tar.gz"]` actually allows any `.gz` file (as explained above), widening the intended scope.

## Relevant code

```go
// pkg/enforcement/handlers.go – handleAllowedExtensions
for _, filePath := range paths {
    ext := strings.ToLower(filepath.Ext(filePath))  // returns only final component
    if ext == "" { ... }
    if !allowed[ext[1:]] { ... }
}
```

The allowlist normalization also strips only a single leading dot, so `".tar.gz"` in `ae.Extensions` is stored as `"tar.gz"`, which will never match `"gz"` or `".gz"` — so even if an operator writes the right thing the comparison still fails.

## Notes

* The manifest-guide currently does not document whether compound extensions are supported, so this is also a documentation gap.
* There is no test covering `"archive.tar.gz"` against `extensions: [".tar.gz"]`.

## Suggested fix options

**Option A (minimal — document the limitation):** Document that only the last extension component is checked, rename the word "extension" to "file suffix" in error messages/docs to set the correct expectation, and validate that allowlist entries do not contain a second dot at manifest load time (e.g. `".tar.gz"` is rejected with a clear error message).

**Option B (support compound extensions):** Change the extraction to return the full compound suffix by scanning backward for the first dot that follows a non-dot character:
```go
idx := strings.Index(base, ".")
if idx < 0 { /* no extension */ }
ext := strings.ToLower(base[idx:])  // ".tar.gz" for "archive.tar.gz"
```
The allowlist comparison and normalization would need to be updated consistently.

Option A is safer for a pre-1.0 codebase; Option B matches operator intent but is a breaking change to anyone relying on the current single-component behaviour.

eunolabs:
## Summary

Fixes #270.

`handleAllowedExtensions` extracted the extension with `filepath.Ext`, which returns only the **final** dot-separated component. A compound allowlist entry like `.tar.gz` therefore never matched anything:

- The allowlist normalization stored `.tar.gz` as `tar.gz`, while `filepath.Ext("backup.tar.gz")` yielded `.gz` → `gz`. The two could never be equal, so `extensions: [".tar.gz"]` **denied every file** (false deny).
- Conversely, the only single-component value that *could* match a `*.tar.gz` file was a bare `.gz` entry, so writing `.tar.gz` silently behaved like `.gz` — a misleading scope change.

## Fix

Switch from "last component, exact match" to **suffix matching on the file name**:

- Each allowlist entry is normalized to a single leading dot, lowercased (`tar.gz`, `.TAR.GZ` → `.tar.gz`).
- A path clears the allowlist when its file name ends with a listed entry **on a dot boundary** (`strings.HasSuffix`). The leading dot is the boundary, so `.gz` matches `data.gz` but not `datagz`/`x.tgz`.
- Compound extensions are honored in full: `.tar.gz` admits `backup.tar.gz` and **not** a bare `data.gz`.
- Matching uses `filepath.Base`, so a dotted *directory* can't smuggle an extension.
- An empty / all-blank allowlist denies fail-closed.

For single-component entries this is **identical to the previous behavior** (including the issue #243 dotfile cases: `.env` still has extension `.env`, `.env.local` still matches `.local`), so existing valid policies are unaffected — only the previously-broken compound case changes.

### Behavioral note (documented)

Matching is "ends-with", so `.gz` admits any name ending in `.gz`, including double extensions like `archive.tar.gz` and `payload.exe.gz`. To admit a true single-extension file and reject double extensions, constrain the path with `allowedValues` instead — extension matching can't express that. This is now spelled out in the manifest guide.

## Tests

- New table-driven `TestAllowedExtensions_CompoundExtensions` in `pkg/enforcement/coverage_test.go`: the false-deny fix, the no-longer-widening case, normalization (leading-dot-optional / case-insensitive), dot-boundary rejection, double-extension suffix semantics, blank-entry handling, and the array (`read_multiple_files`) form.
- Existing `TestAllowedExtensions_Dotfiles` (#243) and the other allowedExtensions tests continue to pass.
- Full suite green: `go test -race -count=1 ./...`.

## Docs

- `docs/capability-manifest-guide.md`: documents suffix/compound matching semantics and the double-extension caveat.
- A companion clarification for `eunolabs/mcp-capability-manifest` `SPEC.md` was prepared but not pushed (it wasn't an established destination for this session) — happy to open that separately if wanted.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JXC6N6WpaQ2tj33PXpyfS5)_


pkg/enforcement/handlers.go
```diff
@@ -510,13 +510,29 @@ func (e *Engine) handleAllowedExtensions(_ context.Context, cond capability.Cond
...
 
-	// Normalize the allowlist once: lowercased, leading dot removed.
-	allowed := make(map[string]bool, len(ae.Extensions))
+	// Normalize the allowlist once: lowercased, with exactly one leading dot so
+	// every entry is a dotted suffix matched with strings.HasSuffix. The leading
+	// dot is the segment boundary -- ".gz" matches "data.gz" but not "datagz".
+	// Compound entries are preserved in full: ".tar.gz" stays ".tar.gz", so it
+	// matches "backup.tar.gz" and not a bare "data.gz" (issue #270). Blank
+	// entries are skipped; an empty allowlist denies every path fail-closed.
+	allowed := make([]string, 0, len(ae.Extensions))
```
eunolabs, line 520: Nit (non-blocking): the allowlist changed from a `map[string]bool` to a `[]string`, so duplicate/overlapping entries (`[".gz", ".gz"]`) are no longer de-duplicated and get iterated redundantly. Negligible for realistic allowlist sizes and not a correctness issue — just noting the behavioral change from the previous set semantics.

---

pkg/enforcement/handlers.go
```diff
@@ -529,7 +545,20 @@ func (e *Engine) handleAllowedExtensions(_ context.Context, cond capability.Cond
...
+		// its name ends with any listed entry on a dot boundary.
+		base := strings.ToLower(filepath.Base(filePath))
+		matched := false
+		for _, suffix := range allowed {
+			if strings.HasSuffix(base, suffix) {
+				matched = true
+				break
+			}
+		}
+		if !matched {
```
eunolabs, line 567: Optional (cosmetic): on a compound-allowlist deny, the error message below still reports `ext` (the final-segment extension from `filepath.Ext`). E.g. denying `backup.zip.gz` against `[".tar.gz"]` yields `file extension ".gz" is not allowed`. It's technically accurate (`.gz` alone isn't in the allowlist), but it doesn't convey that it was the *compound* suffix that failed to match. `Details` already carries `filePath`, so this is debuggable — feel free to leave as-is, or include `base` in the message for compound cases.

---

docs/capability-manifest-guide.md
```diff
@@ -918,13 +918,27 @@ required `table` entry).
...
+its name ends with one of the listed entries on a dot boundary. Entries may be
+written with or without the leading dot, and case is ignored (`".env"`, `"env"`,
+and `".ENV"` are equivalent). A file with no dot at all (`Makefile`, `id_rsa`) has
+no extension and is denied — it can never satisfy an extension allowlist.
+
+**Compound extensions are honored in full.** An entry like `.tar.gz` matches
+`backup.tar.gz` and **not** a bare `data.gz` — listing the compound suffix means
+the compound suffix, never just its last segment. (Earlier builds took only the
+final dot, so `.tar.gz` silently denied every file *and* a `.gz` entry secretly
+admitted `*.tar.gz`; both are fixed.) Conversely, a single-component entry matches
```
eunolabs, line 931: The double-extension caveat is well stated. Worth surfacing as an explicit **security note**: an `allowedExtensions` allowlist is *not* a defense against double-extension smuggling — a `.csv` entry admits `payload.exe.csv`, and a `.gz` entry admits `payload.exe.gz`. This was already true before this PR (old `filepath.Ext` had the same property), so it's not a regression, but readers reaching for `allowedExtensions` as a content-type gate should be steered firmly to `allowedValues` for that use case. The PR body notes a companion clarification was prepared for `eunolabs/mcp-capability-manifest` `SPEC.md` but not pushed — per CONTRIBUTING, the matching semantics are a behavioral contract, so it'd be good to land that spec update too (separate PR is fine) so the guide and the spec don't drift.

---

eunolabs (commented):
## Review: compound extension matching in `allowedExtensions` (#270)

This is a correct and well-executed fix. I traced the logic against the existing behavior and the related issues (#270, #243) and it holds up.

**Correctness — verified:**
- **The core fix is right.** Switching the allowlist from de-dotted exact-match (`allowed[ext[1:]]`) to dotted-suffix `strings.HasSuffix(base, suffix)` correctly resolves both halves of #270: the false deny (`.tar.gz` entry vs `filepath.Ext` returning `.gz`) and the silent widening (`.tar.gz` degrading to `.gz`).
- **The "identical for single-component entries" claim checks out.** For a single-segment entry `.x`, old matched iff the final segment equals `x`; `HasSuffix(base, ".x")` is true iff `base` ends in `.x`, which is exactly "final segment is `x`". The implication holds both directions, so no single-extension policy changes behavior.
- **#243 dotfiles preserved.** `.env` still matches `.env`/`config.env`; `.env.local` against `[".env"]` still denies (final-segment logic unchanged), and `.env.local` now *also* matches an `.env.local` entry — the same class of compound fix.
- **Dot boundary is sound.** Every normalized entry begins with `.`, so a match can only start at a dot in the file name — `.gz` matches `data.gz` but not `datagz` or `x.tgz`. There's no way to match mid-segment.
- **Fail-closed preserved.** Blank entries are skipped and an all-blank/empty allowlist denies every path; a non-string array item and a no-extension file still deny. Matching on `filepath.Base` keeps a dotted directory from leaking an extension.

**Tests & docs:** the new table-driven `TestAllowedExtensions_CompoundExtensions` covers the false-deny, no-longer-widening, normalization (leading-dot-optional / case-insensitive), dot-boundary rejection, double-extension semantics, blank handling, and the array form. The manifest guide is updated in the same PR. All 42 CI checks are green (Test, Lint, gofmt, License Headers, CodeQL, govulncheck, demo + OPA-comparison stacks).

**Inline notes — all non-blocking:**
1. The allowlist is now a slice rather than a set, so duplicate entries are no longer de-duplicated (negligible).
2. The compound-deny error message reports the final-segment `ext` rather than the compound suffix — accurate but slightly less informative (cosmetic).
3. The double-extension behavior is worth calling out as a security note, and the companion `mcp-capability-manifest` `SPEC.md` update flagged in the PR description should be landed so the spec doesn't drift from the guide.

Nothing here blocks merge — the fix is solid. The points above are optional polish.

---

docs/capability-manifest-guide.md
eunolabs: Added an explicit **Security note** in the manifest guide (this push) spelling out that `allowedExtensions` is not a defense against double-extension smuggling and steering content-type gating to `allowedValues`.

On the spec: the companion `mcp-capability-manifest` `SPEC.md` change is written and committed locally, but the push to that repo was blocked by a session guardrail (it wasn't an established push destination for this session), so I can't land it from here. Happy to open it as a separate PR if you enable that repo as a destination. Leaving this thread open to track that follow-up.

---

eunolabs: Spec update is now open as a separate PR: eunolabs/mcp-capability-manifest#6. Resolving this thread — the in-guide security note landed here, and the spec contract is tracked there.

---

PR #286, opened 2026-06-13
sec: fail closed on stray argumentSchema in resource/prompt paths (#268)

### LINKED ISSUE #268: argumentSchema is silently skipped on resource:/prompt: constraints (fail-open)
## Summary

`argumentSchema` declared on a `resource:` or `prompt:` capability constraint is **silently ignored** at decision time. A manifest author who adds an `argumentSchema` to such a constraint expecting structural validation gets no enforcement and no error — a fail-open outcome that runs against the project's "fail closed" invariant.

This was surfaced while reviewing #247, which documents that `Engine.EvaluateConditions` deliberately does **not** run `argumentSchema` validation and that the caller is responsible for running `ValidateArgumentSchema` first. Two of the three in-tree callers don't do that.

## Details

Of the three `ManifestPDP` callers of `EvaluateConditions`:

| Caller | Runs `ValidateArgumentSchema` first? |
|---|---|
| `Decide` (tool calls) — `cmd/eunox-mcp/pdp.go:218-251` | **Yes** |
| `DecideResourceRead` — `cmd/eunox-mcp/pdp.go:412` | **No** |
| `DecidePromptGet` — `cmd/eunox-mcp/pdp.go:499` | **No** |

`ValidateAction` (`pkg/enforcement/engine.go:269`) and `Decide` both validate the schema before evaluating conditions, but `DecideResourceRead` and `DecidePromptGet` call `EvaluateConditions` directly.

`ArgumentSchema` is a field on the generic `Constraint` (`pkg/capability/constraint.go:105`), and manifest validation (`pkg/capability/validate.go`) does **not** restrict it to tool targets. So a `resource:`/`prompt:` constraint can carry an `argumentSchema` that is accepted at load time and then never enforced.

## Proposed fix (pick one)

1. **Validate in those paths** — run `ValidateArgumentSchema` against the matched constraint's `ArgumentSchema` in `DecideResourceRead` and `DecidePromptGet` before calling `EvaluateConditions`, mirroring `Decide`. For these targets the request args are `{"uri": uri}` / `{"name": promptName}` respectively.
2. **Reject at load time** — if `argumentSchema` is intended to be tool-only, have manifest validation reject it on non-tool targets so the silent skip can't occur.

Option 1 keeps `argumentSchema` available everywhere and consistently enforced; option 2 narrows the grammar. Either closes the fail-open gap. A regression test should cover a `resource:`/`prompt:` constraint with an `argumentSchema` (deny on malformed args, or load-time rejection).

## References

- Review thread: #247 (https://github.com/eunolabs/eunox/pull/247#discussion_r3408565336)
- `cmd/eunox-mcp/pdp.go:412`, `cmd/eunox-mcp/pdp.go:499`
- `pkg/enforcement/engine.go:434` (`EvaluateConditions`), `pkg/enforcement/engine.go:269` (`ValidateAction`)

eunolabs:
## Summary

Closes #268.

`argumentSchema` is tool-only and already rejected at manifest **load** (#263), but the runtime **decision path** still relied entirely on that loader guarantee. `DecideResourceRead` and `DecidePromptGet` evaluate conditions through `Engine.EvaluateConditions`, whose contract deliberately leaves `ValidateArgumentSchema` to the caller and does **not** run it (only `Decide` / `ValidateAction` do). So a `resource:`/`prompt:` constraint that reached those paths carrying an `argumentSchema` — a manifest assembled in-process rather than through `LoadManifest`, or a regression in the load-time check — had its declared schema **silently skipped**: a declared-but-unenforced structural guard, i.e. a fail-open against the project's fail-closed invariant.

This is the decision-path half of #268. #263 closed it at load time (Option 2 in the issue); this hardens the enforcement point itself so the tool-only guarantee no longer rests on load-time validation alone.

## Change

- `cmd/eunox-mcp/pdp.go` — `DecideResourceRead` and `DecidePromptGet` now fail closed: if a matched constraint carries an `argumentSchema`, the request is denied with `ENFORCEMENT_ERROR` (a distinct, matchable, fail-closed code — JSON-RPC `-32603`) instead of being forwarded with the schema ignored. The deny fires on the schema's *presence*, not on a validation outcome, because the schema is unenforceable on the synthesised `{"uri": uri}` / `{"name": promptName}` maps and forwarding it would be the leak.
- `pkg/capability/errors.go` — generalised the `ErrCodeEnforcementError` doc comment to record this second defense-in-depth use.

## Why a deny (and not Option 1 / validate-on-resource-prompt)

#263 already committed the grammar to **tool-only** (`argumentSchema` rejected on non-tool targets at load). Re-introducing validation on `resource:`/`prompt:` (the issue's Option 1) would contradict that and re-open the grammar. Denying is the coherent, minimal, fail-closed hardening: it closes the gap at the enforcement point while staying consistent with the shipped Option 2, matching this codebase's house style of keeping defensive fail-closed guards even on "unreachable today" branches.

## Tests

`cmd/eunox-mcp/pdp_argument_schema_test.go` — builds the offending constraint directly via `newTestManifestPDP` (which bypasses `LoadManifest`, the only way to even reach this state):

- `TestArgumentSchema_ResourceConstraint_FailsClosed` — resource read with a stray schema → `ENFORCEMENT_ERROR` deny (previously a silent **allow**).
- `TestArgumentSchema_PromptConstraint_FailsClosed` — prompts/get analogue.
- `TestArgumentSchema_ResourcePrompt_NoSchema_Allows` — boundary control: a schema-less resource/prompt constraint still allows, so the guard fires only on a stray schema.

Full suite: `go test -race -count=1 ./...` green; `go vet`, `gofmt`, and license headers clean.

## Docs

Updated `docs/capability-manifest-guide.md`, `docs/threat-model-mcp.md`, and `CHANGELOG.md` (Security) to note the decision path fails closed independently of load-time validation.

https://claude.ai/code/session_01KkqvSk5ZL33biz4KLFutwY

---
_Generated by [Claude Code](https://claude.ai/code/session_01KkqvSk5ZL33biz4KLFutwY)_


CHANGELOG.md
```diff
@@ -27,6 +27,18 @@ Section conventions:
 
 ### Security
 
+- **`resources/read` and `prompts/get` fail closed on a stray `argumentSchema`.**
```
eunolabs, line 30: Merge conflict here (the PR shows `mergeable_state: dirty`). After this branch was cut, `main` added a new entry at the top of `### Security` — `**\`sequenceBlock\` no longer silently passes when every \`afterTools\` entry strips to empty.**` — directly above the `sequenceBlock ... fails open when a call-counter write faults` entry that this hunk uses as its anchor context. Both branches insert at the same spot, so this won't apply cleanly.

Resolution is trivial: merge/rebase `main` and re-place this new entry among the others under `### Security`. Just calling it out since it blocks merge.

---

cmd/eunox-mcp/pdp.go
```diff
@@ -402,11 +402,29 @@ func (p *ManifestPDP) DecideResourceRead(ctx context.Context, sessionID, uri, so
...
-	// against the synthesised {"uri": uri} map.
+	// resource: targets, so a loader-built manifest never reaches here carrying one.
+	// But unlike Decide (tools/call), this path evaluates conditions through
+	// EvaluateConditions, whose contract deliberately leaves ValidateArgumentSchema
+	// to the caller and does not run it. A constraint that nonetheless carries an
+	// argumentSchema — a manifest built in-process rather than through LoadManifest,
+	// or a regression in the load-time check — would therefore have it silently
+	// skipped: a declared structural guard with no runtime effect is a fail-open.
+	// Refuse the request rather than trust the loader's invariant from inside the
+	// enforcement point (issue #268).
```
eunolabs, line 425: Worth being explicit (here and in the prompt analogue): like the other denials in this function, this early `return` is captured by the named return value and the `defer func() { resp.AuditOnly = true }()` set a few lines up for `matched.IsAuditOnly()` constraints. So for an audit-mode constraint, a stray-`argumentSchema` request is still forwarded with the schema skipped — just logged as a deny — rather than blocked.

That's consistent with audit mode's "observe, don't enforce" contract and with how `AUTHORIZATION_FAILED` / `CAPABILITY_DENIED` already behave here, so I think it's intended; flagging only so the "fail closed" framing in the comment/CHANGELOG is understood to be modulo audit mode. Not asking for a change.

---

eunolabs (commented):
## Review: `sec:` fail closed on stray `argumentSchema` in resource/prompt paths

Traced the full diff plus the surrounding decision paths. This is a clean, well-scoped defense-in-depth hardening and I'm happy with the approach.

**What's right**
- The guard fires on the schema's **presence**, not on a validation outcome — correct, since running `ValidateArgumentSchema` against the synthesised `{"uri": …}` / `{"name": …}` map is meaningless, and forwarding-with-skip is the actual leak. Denying with a distinct, matchable `ENFORCEMENT_ERROR` (`-32603`) keeps it from masquerading as a policy decision.
- Placement is right: after the action check, before `req` is built and conditions are evaluated, so the deny can't be shadowed by a condition outcome.
- Tests cover deny (resource), deny (prompt), and the schema-less boundary control that must still allow — and they correctly build the offending constraint via `newTestManifestPDP`, the only way to reach this state. Docs (manifest guide, threat model, CHANGELOG, `errors.go` comment) are all updated in-PR per the repo's "behavior change ⇒ docs in same PR" rule.
- Choosing deny over re-introducing validation (the issue's Option 1) is the coherent call given #263 committed the grammar to tool-only; reusing `ENFORCEMENT_ERROR` rather than minting a new code is well justified in the doc comment.

**Scope check — JWT path (no gap; confirming for the record)**
`JWTPDP.DecideResourceRead` / `DecidePromptGet` route through `JWTPDP.Decide` → `inner.Decide` (the tools/call path), which **does** run `ValidateArgumentSchema` (`pdp.go:218`), so the silent-skip this PR closes never existed on the JWT path — the manifest `DecideResourceRead`/`DecidePromptGet` direct paths were the only ones affected, and this PR covers both. One marginal asymmetry, **non-blocking**: on that JWT path the schema is enforced by validation *outcome* against `args=nil`, so a hypothetical *permissive* stray schema (no `required`, `additionalProperties` not `false`) would pass rather than deny-on-presence. Reachability is near-zero (JWT mode + in-process manifest + permissive stray schema on a non-tool target), so I'd leave it — just noting the presence-based guard added here isn't shared by that path.

**Action item before merge**
- `mergeable_state` is `dirty` — merge conflict in `CHANGELOG.md` (`main`'s `### Security` gained a new top entry after this branch was cut). Inline note left; resolution is a trivial rebase + re-placement of the new entry.

Plus one inline, non-blocking observation about the audit-mode interaction. No correctness concerns in the code itself.

---

PR #287, opened 2026-06-13
fix: isolate maxCalls counter buckets by window (#278)

### LINKED ISSUE #278: bug: multiple maxCalls conditions on the same tool inflate the timestamp slice — N conditions add N timestamps per admitted call
## Summary

When a capability has **two or more `maxCalls` conditions on the same tool sharing the same `windowSeconds`**, the `InMemory` backend adds **one timestamp per condition** for a single admitted call instead of one timestamp total. The effective rate limit becomes `configured_limit / N` where N is the number of conditions evaluated.

This is a secondary consequence of the same root cause as the cross-window prune bug (#reported separately): `InMemory` indexes its entry map by the raw caller-supplied key, so all `maxCalls` conditions that produce the same key write into the same `*entry`.

## Reproduction

```yaml
capabilities:
  - target: tool:export
    actions: [call]
    conditions:
      - type: maxCalls
        count: 10
        windowSeconds: 60   # condition A
      - type: maxCalls
        count: 20
        windowSeconds: 60   # condition B — same window, different limit
```

With `InMemory`, a single admitted call triggers:
1. Condition A: `IncrementIfBelow(key, 60, 10)` → appends timestamp → entry has 1 entry
2. Condition B: `IncrementIfBelow(key, 60, 20)` → appends timestamp → entry has 2 entries

After 5 admitted calls the entry holds **10** timestamps. On the 6th call:
- Condition A sees count 10 ≥ 10 → **DENIED**, even though only 5 real calls occurred.

The effective limit for condition A is **5** (10 / 2 conditions), not the configured 10.

## Root Cause

```go
// memory.go – IncrementIfBelow
e, ok := m.entries[key]   // key = "maxcalls:<session>:<type>:<tool>"
```

All conditions for the same tool share one `*entry`. Redis avoids this by appending `:<windowSec>` to the key, but even that does not fully isolate conditions with the same `windowSec`.

## Impact

Any deployment where a capability defines multiple `maxCalls` conditions on the same tool, regardless of whether their `windowSeconds` differ. The InMemory backend silently halves (or thirds, etc.) the effective quota.

The Redis backend is unaffected because each `(key, windowSec)` pair maps to a completely independent sorted set, and the enforcement engine evaluates conditions sequentially, so each condition only ever ZADDs to its own set once per admitted call.

## Proposed Fix

Include a condition-unique discriminator in the counter key. The simplest option is to include `windowSeconds` in the key at the `handleMaxCalls` call site (which also fixes the cross-window prune issue):

```go
```

For full isolation (two conditions with the same window but different limits), the condition's index within the conditions slice would need to be threaded through to the handler. A simpler policy: document that two `maxCalls` conditions with identical `windowSeconds` on the same capability are equivalent to taking the lower of the two `count` values, and validate/reject that combination at manifest load time.

## Notes

- The double-counting makes the `InMemory` backend **more restrictive** than the Redis backend (fails closed rather than open for this specific misconfiguration), but it misrepresents the actual call count in audit logs and in the `retryAfter` hint returned to callers.
- The cross-window variant (different `windowSeconds`) can be **less restrictive** (fail-open) for the longer window, which is the more serious security concern filed separately.

eunolabs:
## Summary

Fixes #278.

Two or more `maxCalls` conditions on the same capability shared a single counter `*entry` on the InMemory backend, because the key was built from `(session, targetType, name)` only. Every admitted call then appended **one timestamp per condition** into that shared entry, so:

- the in-window count was inflated — N conditions collapse the effective limit to roughly the lower `count` over N (the literal reproduction in the issue: two `windowSeconds: 60` conditions made a configured limit of 10 fire at 5); and
- worse, the shorter window's timestamps were pruned against the **longer** window, silently failing the longer-window quota **open** (the cross-window prune variant).

## Fix

Two complementary changes:

1. **Runtime (`pkg/enforcement/handlers.go`)** — fold `windowSeconds` into the `maxCalls` counter key (`maxcalls:<session>:<targetType>:<name>:<windowSeconds>`). Conditions with *different* windows (e.g. 30/minute **and** 500/hour) now address independent sliding-window buckets and are counted correctly; both must pass. The Redis backend already suffixed `windowSec` onto its sorted-set key, so this brings both backends into step.

2. **Load-time (`cmd/eunox-mcp/manifest.go`)** — reject a capability that carries two `maxCalls` conditions sharing the **same** `windowSeconds`. That is the only remaining collision after the key fix, and it carries no meaning a single condition lacks (two equal windows are satisfied exactly when the lower `count` is). The error points the author at the equivalent single condition, mirroring the other load-time guards.

## Why reject same-window rather than isolate at runtime

A same-window pair `{count: a}` + `{count: b}` is semantically just `{count: min(a, b)}` — collapsing it onto one bucket loses nothing, and an exact duplicate still double-counts even with per-window keying. Rejecting it at load is fail-closed, idiomatic for this codebase (it has many "turn a silently-wrong policy into a config error" validators), and is the resolution the issue endorses as the "simpler policy."

## Tests

- **Engine regression** (`enforcement_test.go`): two distinct-window conditions now count independently — the tighter limit binds at exactly its configured count (10th call), not at half (6th) as before — and the `RATE_LIMITED` denial reports the real count, not an inflated one.
- **Manifest validation** (`manifest_validation_test.go`): duplicate-window pair rejected (different counts and exact duplicate), distinct-window pair accepted.

## Docs

- `docs/capability-manifest-guide.md` — layering multiple `maxCalls` windows, and the same-window rejection.
- `docs/threat-model-mcp.md` — updated counter-key shape.
- `CHANGELOG.md` — Fixed entry under Unreleased.

## Verification

`go build ./...`, `go test -race -count=1 ./...`, `go vet ./...`, `golangci-lint run`, `make check-fmt`, and `make check-license` all pass.

https://claude.ai/code/session_011XDX9riY6osXYqqRDPa1Sy

---
_Generated by [Claude Code](https://claude.ai/code/session_011XDX9riY6osXYqqRDPa1Sy)_


pkg/enforcement/enforcement_test.go
```diff
@@ -568,6 +568,64 @@ func TestEngine_MaxCalls_NilTargetUsesEmptyTypeSegment(t *testing.T) {
...
+
+	// A tight short window (10 per minute) and a loose long window (1000 per hour).
+	// The long window's high count means it never binds within this test; the only
+	// limit that should fire is the short window's, at exactly its 10th call.
+	caps := []capability.Constraint{{
+		Target:  "tool:export",
+		Actions: []string{"call"},
+		Conditions: []capability.Condition{
+			&capability.MaxCallsCondition{Count: 10, WindowSeconds: 60},
+			&capability.MaxCallsCondition{Count: 1000, WindowSeconds: 3600},
```
eunolabs, line 615: Solid regression test for the inflation / double-count direction. One gap worth closing: issue #278 calls the cross-window **prune** variant — the shorter window's 60s prune erasing timestamps that the longer window still needs, so the longer-window quota silently fails **open** — the "more serious security concern," and this test doesn't exercise it. Here the long window (`count: 1000`) never binds, so a regression that reintroduced cross-window pruning would still pass this test.

Consider adding a case where the **longer** window is the binding limit and calls are spread in time:

- short window loose / never binds (e.g. `count: 100, windowSeconds: 60`)
- long window tight (e.g. `count: 5, windowSeconds: 3600`)
- inject a mutable clock via `callcounter.NewInMemory(callcounter.WithTimeFunc(...))` and advance it ~70s between calls

Pre-fix, the short window's prune would have deleted the long window's history every 60s, so the long-window limit of 5 would never trip even after many calls in the hour — the fail-open. Post-fix the two buckets are independent and the 6th call denies. That pins the fail-open direction explicitly, not just the fail-closed inflation that this test already covers.

---

pkg/enforcement/handlers.go
```diff
@@ -341,7 +355,7 @@ func (e *Engine) handleMaxCalls(ctx context.Context, cond capability.Condition,
 			Message:       "tool name is required for maxCalls condition",
 		}
 	}
-	key := fmt.Sprintf("maxcalls:%s:%s:%s", req.SessionID, targetType, toolName)
+	key := fmt.Sprintf("maxcalls:%s:%s:%s:%d", req.SessionID, targetType, toolName, mc.WindowSeconds)
```
eunolabs, line 361: Minor / FYI (no change strictly required): now that `windowSeconds` is folded into `key` here, the Redis backend doubles it. `Redis.IncrementIfBelow` builds `windowKey := fmt.Sprintf("callcounter:%s:%d", key, windowSec)` (redis.go:185), so the physical sorted-set key becomes `callcounter:maxcalls:<sess>:<type>:<name>:<window>:<window>` — the window now appears twice.

It's harmless for correctness (each `(key, window)` is still unique, and isolation is preserved on both backends), but it's redundant, and the physical Redis key no longer matches the single-`<windowSeconds>`-suffix logical shape the threat model now documents. Worth a one-line note near this `Sprintf` (or an eventual cleanup) so a future reader debugging raw Redis keys isn't surprised by the doubled segment.

---

cmd/eunox-mcp/manifest.go
```diff
@@ -210,6 +210,13 @@ func validateLocalManifest(m *LocalManifest) error {
 		if len(c.Directives) > 0 && targetType != capability.TargetTypeTool {
 			return fmt.Errorf("capability at index %d: constraint %q carries directives, which apply only to tool: targets (the proxy redacts tools/call results, not %s responses)", i, c.Target, targetType)
 		}
+		// Reject two maxCalls conditions on one capability that share a
+		// windowSeconds before the per-condition pass, so the cross-condition
+		// collision is reported as the misconfiguration it is rather than passing
+		// every per-condition check and then silently halving the quota at runtime.
+		if err := validateMaxCallsWindowsDistinct(i, c.Conditions); err != nil {
```
eunolabs, line 217: Optional nit on ordering: this runs *before* the per-condition `validateMaxCalls` (which enforces `windowSeconds >= 1`). So a manifest with two `maxCalls` that both omit `windowSeconds` (zero value) is rejected with `...are both maxCalls with the same windowSeconds (0); ... combine them into a single maxCalls with the lower 'count'` rather than the more pointed `maxCalls requires 'windowSeconds' >= 1` — slightly misleading, since the real fix there is to add the window, not merge the conditions.

Both paths fail closed, so this is purely error-message precedence. If you want the sharper message to win, either run the per-condition pass first, or have `validateMaxCallsWindowsDistinct` `continue` past conditions whose `window < 1` and leave them to `validateMaxCalls`.

---

eunolabs (commented):
## Review: fix `maxCalls` counter bucket isolation by window (#278)

Reviewed the diff and the surrounding code (handlers, both call-counter backends, manifest load/merge path, and the engine's constraint selection). The fix is **correct, complete, and well-targeted.** Verified locally on the branch: `go build ./...`, `go vet`, `gofmt -l` (clean), and `go test ./pkg/enforcement/ ./pkg/callcounter/ ./cmd/eunox-mcp/` all pass.

**What's right:**
- Root cause correctly identified and fixed at the right layer — folding `windowSeconds` into the counter key in `handleMaxCalls` gives each window its own sliding-window bucket and brings the InMemory backend into step with Redis (which already keyed by window).
- The remaining same-window collision is rejected at manifest load (fail-closed, idiomatic for this codebase), with a clear, actionable error pointing the author at the equivalent single condition.
- Coverage of all manifest-load paths is sound: I checked that the gateway/multi-file route also runs `validateLocalManifest` (via `LoadManifest` in `route.go`), and that `MergeManifests` only concatenates capability lists while `findMatchingCapability` evaluates a single most-specific constraint — so the cross-*constraint* and post-merge variants aren't reachable, and the within-constraint validator is sufficient.
- Tests, `docs/capability-manifest-guide.md`, `docs/threat-model-mcp.md` (key shape + new version-history row), and `CHANGELOG.md` are all updated in the same PR, matching the repo's conventions.

**Findings (all minor / optional — none are blockers):**
1. **Test coverage gap (most worth addressing)** — the new engine regression test covers the inflation/double-count (fail-*closed*) direction, but not the cross-window *prune* variant that the issue flags as the more serious, fail-*open* concern. Inline suggestion on the test with a concrete construction.
2. **Redis key now doubles the window** — benign redundancy; the physical Redis sorted-set key carries `<window>` twice and no longer matches the documented single-suffix logical shape. Inline note on `handlers.go`.
3. **Validation ordering nit** — `validateMaxCallsWindowsDistinct` runs before `validateMaxCalls`, so two `maxCalls` that both omit `windowSeconds` get the "duplicate window (0)" message instead of the sharper "windowSeconds >= 1." Inline note on `manifest.go`.

**Also (out of this PR's diff, optional cleanup):** a few test comments still describe the `maxcalls:` key without the new `<windowSeconds>` segment — `enforcement_test.go:380` and `:454` — and `adversarial_test.go:690` describes it as `maxcalls:<sessionID>:<toolName>`, missing both `<targetType>` (added earlier) and now `<windowSeconds>`. Not introduced by this PR, but easy to refresh while here.

Nice work overall — this is a clean fix with the fail-closed posture the codebase favors.

---

pkg/enforcement/enforcement_test.go
eunolabs: Added `TestEngine_MaxCalls_LongWindowNotPrunedByShortWindow` exactly along these lines: loose short window (`count: 100, windowSeconds: 60`), tight long window (`count: 5, windowSeconds: 3600`), and a mutable clock (`callcounter.WithTimeFunc`) advanced 70s between calls. Verified it fails against the old shared key (the 6th call is admitted — the fail-open) and passes with per-window buckets (6th denies on the long window). Commit 4a6409c.

---

pkg/enforcement/handlers.go
eunolabs: Added a note at the key construction in `handleMaxCalls` flagging that the Redis backend re-appends `windowSec`, so its physical sorted-set key carries the window twice — harmless (each `(key, window)` is still unique) but called out so a raw-Redis debugger isn't surprised. Commit 4a6409c.

---

cmd/eunox-mcp/manifest.go
eunolabs: Went with the `continue` option: `validateMaxCallsWindowsDistinct` now skips conditions whose `windowSeconds < 1`, leaving them to `validateMaxCalls` so two windowless `maxCalls` get the sharper `windowSeconds >= 1` message instead of the misdirecting duplicate-window (0) conflict. Commit 4a6409c.

---

eunolabs:
## Re-review of 4a6409c — all addressed, no further comments

Re-reviewed after the follow-up commit. All three findings from my earlier review are resolved correctly, and I have no new comments:

1. **Fail-open coverage** — `TestEngine_MaxCalls_LongWindowNotPrunedByShortWindow` pins exactly the cross-window prune variant the issue flags as the more serious concern (loose short window + tight long window, calls spaced 70s via `WithTimeFunc`, 6th call denies on the long window's count of 5). Traced it through and it's sound; verified it passes under `-race`, and the commit notes it fails against the old shared key.
2. **Redis double-window key** — note added at the key construction in `handleMaxCalls`.
3. **Validation ordering** — `validateMaxCallsWindowsDistinct` now `continue`s past `windowSeconds < 1`, so two windowless `maxCalls` get the sharper `windowSeconds >= 1` error. Plus the stale `maxcalls:` key comments in `enforcement_test.go` and `adversarial_test.go` were refreshed to the current shape.

Local verification on the branch is clean: `go build ./...`, `go vet`, `gofmt -l` (clean), and `go test ./pkg/enforcement/ ./pkg/callcounter/ ./cmd/eunox-mcp/` all pass. CI on this commit is green (CodeQL neutral, no failures).

LGTM — nothing left from my side.

---

PR #288, opened 2026-06-13
sec: make counter keys collision-resistant for colon-bearing session IDs

### LINKED ISSUE #273: bug: sequenceHistoryKey is not collision-resistant for non-UUID session IDs containing colons
## Summary

`sequenceHistoryKey` in `pkg/enforcement/engine.go` is:

```go
}
```

If `sessionID` contains a colon, the key is ambiguous — it cannot be uniquely decomposed back into its parts.  For example:

| sessionID | tool | key |
|-----------|------|-----|
| `"a:b"` | `"c"` | `"seq:a:b:c"` |
| `"a"` | `"b:c"` | `"seq:a:b:c"` |

These two inputs produce the same counter key.

The `maxCalls` key (`"maxcalls:%s:%s:%s"`) shares this design.

## Impact

The comment in `handleMaxCalls` states: "session IDs are server-assigned UUIDs (one per connection) and the HTTP transport rejects requests whose session belongs to a different route, making cross-route key collision architecturally impossible."  This is correct for the shipped HTTP transport, where session IDs come from `uuid.New()`.

However:
1. The `EnforceRequest.SessionID` field is a plain `string` with no format constraint enforced by the engine.  A direct caller of `ValidateAction` or `EvaluateConditions` (library use) can supply an arbitrary session ID.
2. `StdioProxy` derives a session ID from the upstream subprocess argument list (search `sessionID` in `stdio.go`), which might differ.
3. The key format is not documented as requiring UUID-format session IDs, so library users have no signal that colons are unsafe.

A crafted session ID like `"legitimate-session:read_file"` paired with a tool `"dummy"` would collide with session `"legitimate-session"` + tool `"read_file:dummy"` (after prefix stripping, which would leave `"read_file:dummy"` unchanged since `"read_file"` is not a recognized prefix).

## Suggested fix

Use a separator that cannot appear in tool names or session IDs, or percent-encode the components.  The simplest option is a multi-character separator that is not valid in either field:

```go
    return "seq\x00" + sessionID + "\x00" + tool
}
```

Null bytes cannot appear in either session IDs (UUIDs use only hex, hyphens) or bare tool names.  Update `maxCalls` key construction consistently.

Alternatively, document the constraint ("session IDs must not contain colons") in the `EnforceRequest` type and validate it at the engine boundary, denying with `MISSING_CONTEXT` if it does.

A regression test should confirm that `sessionID = "a:b"` + `tool = "c"` does not share a counter bucket with `sessionID = "a"` + `tool = "b:c"`.

eunolabs:
Fixes #273.

## Problem

The `sequenceBlock` history key and the `maxCalls` quota key joined their components with a bare `:` delimiter:

```go
}
// ...
key := fmt.Sprintf("maxcalls:%s:%s:%s", req.SessionID, targetType, toolName)
```

When a component contains a colon the key cannot be uniquely decomposed, so two different inputs collapse to one bucket:

| sessionID | tool | key |
|-----------|------|-----|
| `a:b` | `c` | `seq:a:b:c` |
| `a` | `b:c` | `seq:a:b:c` |

The shipped HTTP transport assigns colon-free UUID session IDs, but `EnforceRequest.SessionID` is a plain `string` with **no engine-enforced format**. A direct library caller of `ValidateAction` (or any non-UUID derivation) can supply a colliding value, yielding:

- **maxCalls quota bypass / cross-tenant accounting** — two distinct `(session, tool)` pairs draining one budget.
- **false `sequenceBlock` denial** — one session's recorded history marker aliasing another session's lookup, blocking a tool that was never preceded by its antecedent in that session.

## Fix

Introduce `compositeCounterKey`, which **length-prefixes** every component (`":" + decimal byte length + ":" + raw bytes`):

```go
compositeCounterKey("seq", "a:b", "c") // "seq:3:a:b:1:c"
compositeCounterKey("seq", "a", "b:c") // "seq:1:a:3:b:c"
```

This is injective for *arbitrary* byte content. I chose length-prefixing over the single-NUL-separator option floated in the issue because a NUL byte can appear in a caller-supplied component just as a colon can — only an encoding that doesn't assume anything about the input alphabet fully closes the class. The verbatim `seq:` / `maxcalls:` prefixes are preserved, so the disjoint key namespaces the callcounter backends rely on (`redis.go`) are untouched. Both `sequenceHistoryKey` and `handleMaxCalls` now build keys through this helper.

## Tests

- `counter_key_internal_test.go` — white-box injectivity for the colliding tuples from the issue (sequence key, all three maxCalls boundaries, a NUL-bearing case, and prefix preservation).
- `TestSequenceBlock_ColonSessionDoesNotAliasMarker` — end-to-end regression proving a colon in the session ID no longer aliases another session's marker.

Both new behavioral/unit tests were confirmed to **fail on the pre-fix colon-join** and pass after the fix.

## Docs / comments

- Updated `docs/threat-model-mcp.md` with the new length-prefixed key format and the rationale.
- Updated stale literal-format references in the surrounding code comments.

## Verification

`make check-fmt`, `make check-license`, `go vet ./...`, and `go test -race -count=1 ./...` all pass.

https://claude.ai/code/session_01A7n9RuuU6P65prgh3YSNUk

---
_Generated by [Claude Code](https://claude.ai/code/session_01A7n9RuuU6P65prgh3YSNUk)_


pkg/enforcement/engine.go
```diff
@@ -98,10 +99,43 @@ type RateLimiter interface {
...
+// from the host — so the encoding, not an assumption about the input alphabet,
+// has to carry the separation. Length-prefixing is injective for arbitrary byte
+// content (including a NUL, which a single sentinel separator would not be):
+// reading the length tells the decoder exactly where each part ends, leaving no
+// freedom for a part's contents to forge a different tuple's key.
+//
+// The prefix is emitted verbatim as the leading token, so the disjoint "seq:" /
+// "maxcalls:" key namespaces the callcounter backends rely on (see redis.go) are
+// preserved.
+func compositeCounterKey(prefix string, parts ...string) string {
```
eunolabs, line 134: Nit (non-blocking, doc precision): the encoding is injective for the *parts* given a fixed `prefix`, but the `prefix` itself is emitted verbatim without a length tag, so the doc claim "maps distinct `(prefix, parts...)` tuples to distinct keys regardless of the bytes the parts hold" isn't quite true across *varying* prefixes — e.g. `compositeCounterKey("x")` and `compositeCounterKey("x", "")` both render `"x:0:"`-style ambiguity if a prefix were allowed to carry a colon/digit. This is purely theoretical here: the only two call sites pass the constant, colon-free, distinct prefixes `"seq"` and `"maxcalls"`, each at a fixed arity, so the security property holds. Might be worth tightening the comment to say the guarantee is "for a fixed prefix" so a future caller doesn't pass a dynamic prefix and assume more than the encoding delivers.

Optional micro-opt: this runs on every recorded allowed call and every `maxCalls` check; a `b.Grow(...)` hint up front would save the Builder's reallocations. Marginal, mentioning only because the repo tracks a benchmark-sanity gate.

---

eunolabs (commented):
## Review — `sec: make counter keys collision-resistant for colon-bearing session IDs`

Reviewed the full diff against issue #273, the surrounding `handleMaxCalls`/`sequenceHistoryKey` code, and the callcounter backends. **This is a solid, correct fix.** Summary below; one non-blocking inline nit and one suggestion.

### Correctness — looks right
- **Length-prefixing is the right call** over the NUL-separator the issue floated. `":" + len(p) + ":" + p` per component is injective for arbitrary byte content (a NUL can appear in a caller-supplied component just as a colon can), and the PR's `TestCompositeCounterKey_NulSeparatorWouldNotSuffice` pins exactly that rationale.
- **The `seq:` / `maxcalls:` namespace disjointness the backends rely on is preserved.** The prefix is emitted verbatim as the leading token, and `seq` vs `maxcalls` differ in their first byte, so a `seq` key can never equal a `maxcalls` key. Confirmed the backends treat the key opaquely — `redis.go` wraps it as `callcounter:<key>:<windowSec>` and in-memory uses it as a map key; neither splits on `:`, so the encoding change is transparent to them. The `redis.go` comment at L188-189 about the two encodings "never meeting on one key" still holds.
- **Fail-closed posture intact.** The empty-`SessionID` and empty-`toolName` guards in `handleMaxCalls` are unchanged and still deny before key construction.

### Tests — thorough
White-box injectivity for all three adjacent-boundary collision cases + the empty-type case, the NUL case, prefix preservation, and the end-to-end `TestSequenceBlock_ColonSessionDoesNotAliasMarker` regression. The PR notes the behavioral tests were confirmed to fail pre-fix and pass post-fix, which is the right bar for a security regression. The `recordFailCounter` helper's `HasSuffix(key, ":"+tool)` still matches under the new format (key now ends `...:<len>:<tool>`), and its comment was correctly updated.

### CI
All 19 checks green (Test, Lint, Enforcement compliance, gofmt, License Headers, govulncheck, CodeQL, Trivy, all platform builds).

### Suggestion (non-blocking): add a CHANGELOG entry
This is a `sec:` fix for a `security`-labeled issue, and the curated `CHANGELOG.md` `[Unreleased] > Security` section documents the immediately-adjacent security fixes (#266, #232, …). Two things worth a one-line entry there:
1. The fix itself (counter-key collision for colon-bearing session IDs, issue #273).
2. **A migration note for the Redis backend:** this changes the runtime `seq:`/`maxcalls:` key schema, so persisted counters written under the old `seq:<session>:<tool>` / `maxcalls:<session>:<type>:<name>` format are orphaned on upgrade — in-flight `sequenceBlock` history and `maxCalls` quota state effectively reset. That's benign (fail-safe, and in-memory resets on restart regardless), but the repo documents this class of schema reset (cf. the existing threat-model note on the `seq:` schema resetting on upgrade), and Redis-backend operators would want the heads-up. Per CLAUDE.md's "document the migration" convention.

The threat-model doc update is good and covers the new key format; the CHANGELOG is the one curated doc that didn't get the corresponding entry.

Nice work — the reasoning in the `compositeCounterKey` doc comment and the threat-model prose is unusually clear about *why* length-prefixing rather than a sentinel.

---

pkg/enforcement/engine.go
eunolabs: All addressed in 5978e16:

- **Doc precision:** tightened the `compositeCounterKey` comment to scope injectivity to "a fixed prefix and arity," and noted the prefix is emitted without a length tag so the guarantee holds per constant prefix — which both call sites (`"seq"`, `"maxcalls"`) satisfy.
- **Micro-opt:** added a `b.Grow(...)` pre-size up front (sized from `len(prefix)` + per-part bytes + a small fixed envelope).
- **CHANGELOG (from the top-level review):** added an `[Unreleased] > Security` entry for the fix plus the Redis-backend migration note (old `seq:`/`maxcalls:` key state is orphaned once on upgrade, fail-safe).

---

eunolabs:
Re-reviewed after `5978e16`. All three non-blocking points from the earlier review are addressed:

- **Doc precision** — the `compositeCounterKey` comment now scopes injectivity to "a fixed prefix and arity" and explains that the prefix is emitted verbatim without a length tag, so the guarantee holds per constant prefix (both call sites pass `"seq"` / `"maxcalls"`).
- **Pre-size** — `b.Grow(size)` added; the over-estimate is safe and the rendered key is byte-identical.
- **CHANGELOG** — `[Unreleased] > Security` entry added for the fix, including the Redis-backend migration note (old `seq:`/`maxcalls:` state orphaned once on upgrade, fail-safe).

The underlying fix remains sound: length-prefixed components are injective for arbitrary bytes, the disjoint `seq:`/`maxcalls:` namespaces the backends rely on are preserved, and the fail-closed guards are intact. All 19 CI checks are green on the head commit.

**No remaining review comments — this looks good to me.**

---

PR #289, opened 2026-06-13
refactor!: document and consolidate the call-counter interface contract

### LINKED ISSUE #280: design: CallCounter / Store interfaces underspecify requirements for maxCalls and sequenceBlock conditions
## Summary

`pkg/capability.CallCounter` (the interface accepted by `enforcement.WithCallCounter`) declares only `IncrementAndGet`. At runtime the enforcement engine discovers two additional capabilities — `RateLimiter` (`IncrementIfBelow`) and `CallHistory` (`Peek`) — via type assertion. A third-party backend that only satisfies `CallCounter` compiles cleanly but silently disables:

- **`maxCalls`**: handler falls back to `"configured call counter does not support the atomic rate-limit check required by maxCalls"` (`CONDITION_FAILED`) on every call — i.e., every `maxCalls` condition denies unconditionally.
- **`sequenceBlock`**: handler returns `"call counter does not support call history queries"` — i.e., every `sequenceBlock` condition denies unconditionally.

This is fail-closed (correct per the invariant), but it is extremely surprising: an operator adds a custom counter backend, everything compiles, and then every tool with `maxCalls` or `sequenceBlock` stops working with opaque error messages.

## Current interface topology

```
pkg/capability
  CallCounter { IncrementAndGet }         ← what callers must implement

pkg/enforcement/engine.go
  RateLimiter { IncrementIfBelow }        ← needed for maxCalls
  CallHistory { Peek }                    ← needed for sequenceBlock

pkg/callcounter
  Store { IncrementAndGet }               ← mirrors CallCounter; not used by engine
  InMemory  implements all three
  Redis     implements all three
```

`Store` in `pkg/callcounter` is defined separately from `capability.CallCounter` but has the same shape — a second minimal interface that is also never type-asserted to.

## Problems

1. **Silent feature degradation**: The fail-closed errors in condition handlers are indistinguishable from a policy decision in the audit log. An operator will not know their custom counter is incompatible until they trace the `CONDITION_FAILED` log entries.

2. **Discoverable contract gap**: Neither `pkg/capability.CallCounter` nor `pkg/callcounter.Store` lists `IncrementIfBelow` or `Peek`. A reader of either interface has no indication that additional methods are required for the two most-used condition types.

3. **Redundant `Store` interface**: `pkg/callcounter.Store` duplicates `capability.CallCounter` without adding value. It is not referenced by the engine and is not the interface one would implement to plug in a custom backend.

### Option A — widen `capability.CallCounter` (breaking for existing minimal implementations)

```go
type CallCounter interface {
    // Required for maxCalls.
    IncrementIfBelow(ctx context.Context, key string, windowSec int, limit int64) (count int64, admitted bool, retryAfter time.Duration, err error)
    // Required for sequenceBlock.
}
```

This is a pre-1.0 breaking change and aligns with the stated pre-1.0 policy of not carrying shims.

### Option B — keep the narrow interface, add compile-time documentation

Add a doc comment to `capability.CallCounter` listing which methods each condition type requires and pointing to `RateLimiter` / `CallHistory` in `pkg/enforcement`. Add a `var _ RateLimiter = (*callcounter.InMemory)(nil)` style assertion in `cmd/eunox-mcp` or `pkg/enforcement` to keep the relationship pinned across refactors.

### Option C — expose a combined interface

Define a `FullCallCounter` interface in `pkg/capability` or `pkg/callcounter` that embeds all three, and document `WithCallCounter` as accepting any backend but `maxCalls`/`sequenceBlock` conditions as requiring a `FullCallCounter`:

```go
// FullCallCounter is the complete call-counter interface required for all
// condition types. maxCalls requires IncrementIfBelow; sequenceBlock requires Peek.
// A backend implementing only CallCounter compiles but disables those conditions.
type FullCallCounter interface {
    CallCounter
    IncrementIfBelow(ctx context.Context, key string, windowSec int, limit int64) (count int64, admitted bool, retryAfter time.Duration, err error)
}
```

## Notes

- The `pkg/callcounter.Store` interface should either be removed or aliased to `capability.CallCounter` to eliminate the duplicate minimal interface.
- `enforcement.RateLimiter` and `enforcement.CallHistory` can remain as internal extension interfaces to preserve the optional-capability pattern in the engine; the documentation gap is the actionable issue here.

eunolabs:
## Summary

Closes #280.

`enforcement.WithCallCounter` accepts `capability.CallCounter` (just `IncrementAndGet`), but at runtime the engine discovers two more capabilities by type assertion — `RateLimiter` (`IncrementIfBelow`, required by `maxCalls`) and `CallHistory` (`Peek`, required by `sequenceBlock`) — and **fails closed** when they are absent. A third-party backend that implements only the base interface compiles cleanly and then denies *every* `maxCalls`/`sequenceBlock` call with an opaque `CONDITION_FAILED` message. This is correct (fail-closed) but extremely surprising, and neither minimal interface advertises the additional methods.

The issue offers three options and concludes that **"the documentation gap is the actionable issue here"** and that `RateLimiter`/`CallHistory` should remain optional extension interfaces. This PR takes that path (Option B) and folds in the combined-interface idea (Option C) for discoverability, plus the suggested `Store` cleanup — without forcing the breaking widening of Option A.

## Changes

- **Document the contract on `capability.CallCounter`** — which method each condition needs, that a missing extension fails closed with `CONDITION_FAILED`, and where the extensions live.
- **Add `enforcement.FullCallCounter`** (`capability.CallCounter` + `RateLimiter` + `CallHistory`) — a single discoverable name for "backs every condition type". Both built-in backends are pinned to it with compile-time assertions, so a refactor that drops a method becomes a build failure instead of a silently disabled condition.
- **Name the missing method + extension interface** in the two fail-closed denial messages (`maxCalls`/`sequenceBlock`), so an operator tracing the audit log can immediately see *what* their custom backend is missing.
- **Remove the redundant `callcounter.Store` interface** — an exact duplicate of `capability.CallCounter` (single `IncrementAndGet`) that the engine never asserted against. The proxy wiring (`main.go`, `route.go`) now takes `capability.CallCounter` directly. Pre-1.0, no compatibility shim.
- **Docs + changelog** — `docs/architecture.md` `WithCallCounter` bullet and a `Changed`/`Removed` entry in `CHANGELOG.md`.

## Behavior

No runtime behavior change for the built-in `callcounter.InMemory`/`callcounter.Redis` backends — they already satisfy the full contract. The only observable difference is clearer text in the two existing fail-closed denial messages.

## Breaking change

`callcounter.Store` is removed from the exported API (it was an unused duplicate). Embedders that referenced it should use `capability.CallCounter`, or `enforcement.FullCallCounter` for a backend supporting every condition type.

## Testing

- `go build ./...`, `go vet ./...`, `gofmt -l` clean, `golangci-lint` reports 0 issues.
- `go test -race -count=1` passes for `pkg/enforcement`, `pkg/callcounter`, `pkg/capability`, and `cmd/eunox-mcp`.
- The existing `TestEngine_MaxCalls_CounterWithoutRateLimiterFailsClosed` and `TestSequenceBlock_CounterWithoutHistoryFailsClosed` continue to cover the fail-closed paths; the new `FullCallCounter` is pinned by the extended compile-time assertion block.

https://claude.ai/code/session_01QXBM2sK8hhUXdiNrh5WiY9

---
_Generated by [Claude Code](https://claude.ai/code/session_01QXBM2sK8hhUXdiNrh5WiY9)_


pkg/enforcement/handlers.go
```diff
@@ -298,7 +298,7 @@ func (e *Engine) handleMaxCalls(ctx context.Context, cond capability.Condition,
 		return &ConditionError{
 			Code:          capability.ErrCodeConditionFailed,
 			ConditionType: capability.ConditionTypeMaxCalls,
-			Message:       "configured call counter does not support the atomic rate-limit check required by maxCalls",
+			Message:       "configured call counter does not implement IncrementIfBelow, the enforcement.RateLimiter extension required by maxCalls (the built-in callcounter.InMemory and callcounter.Redis backends implement it)",
```
eunolabs: Non-blocking, but worth a thought: this `Message` is returned to the MCP **host** (it flows `ConditionError.Message` → `DenialInfo.Message` → JSON-RPC error), not only into the audit log. Embedding Go-internal identifiers (`IncrementIfBelow`, `enforcement.RateLimiter`, `callcounter.InMemory`/`Redis`) is exactly what an operator tracing the audit trail wants — but the remote host can't act on internal type names and arguably shouldn't see them.

Since this only fires on operator misconfiguration (a custom backend missing the extension), one option is to keep the implementation-level detail in the audit `Details`/operator-facing log and let the host-facing `Message` stay capability-oriented. Either way the new text is a clear improvement over the old; flagging only the audience question. Same applies to the `sequenceBlock` message below (line ~865).

---

pkg/capability/enforce.go
```diff
@@ -102,10 +102,32 @@ type Enforcer interface {
...
-// Implementations must be safe for concurrent use.
+// Implementations must be safe for concurrent use. It is the interface that
+// enforcement.WithCallCounter accepts.
 //
 // Defined here alongside Enforcer so that both consumer-facing interfaces live
 // in pkg/capability rather than in the producer package.
+//
+// IncrementAndGet alone is the BASE contract; on its own it backs no condition
+// type's full check. The engine widens what a backend can enforce by probing for
+// two optional extension interfaces (defined in pkg/enforcement) via type
```
eunolabs, line 113: Flagging an intentional coupling, not requesting a change: this makes `pkg/capability` (the lower layer) document types that live in `pkg/enforcement` (which imports `capability`, not the reverse). It's prose-only so there's no import cycle, and since the engine discovers these extensions by type assertion, naming them here is the right call for discoverability. Just noting the upward doc reference so it's a deliberate decision on the record.

---

pkg/enforcement/enforcement_test.go
```diff
@@ -16,11 +16,18 @@ import (
...
+// the base interface; pinning it here turns a refactor that drops a method on
+// either backend into a compile error rather than a silently disabled condition.
 var (
-	_ enforcement.RateLimiter = (*callcounter.InMemory)(nil)
-	_ enforcement.RateLimiter = (*callcounter.Redis)(nil)
+	_ enforcement.RateLimiter     = (*callcounter.InMemory)(nil)
+	_ enforcement.RateLimiter     = (*callcounter.Redis)(nil)
+	_ enforcement.CallHistory     = (*callcounter.InMemory)(nil)
+	_ enforcement.CallHistory     = (*callcounter.Redis)(nil)
+	_ enforcement.FullCallCounter = (*callcounter.InMemory)(nil)
```
eunolabs, line 33: Minor/optional: the two `FullCallCounter` assertions already subsume the four `RateLimiter`/`CallHistory` ones above — `FullCallCounter` embeds both, so a backend dropping `Peek` or `IncrementIfBelow` would fail to compile on the `FullCallCounter` line regardless. Keeping the granular assertions does buy a more localized compile error that names the exact missing extension, which is a reasonable reason to leave them. Just noting they're logically redundant if you'd rather trim to the two `FullCallCounter` lines.

---

eunolabs (commented):
## Review: call-counter interface contract (#280)

Reviewed the full diff plus the post-change state of `pkg/callcounter`, `pkg/enforcement/engine.go`, and the gateway wiring on the PR head. This is a clean, low-risk change that does what the description says — documentation + discoverability + clearer fail-closed messages, with one safe public-API removal. No correctness issues found.

**What holds up well**
- `enforcement.FullCallCounter` correctly bundles `capability.CallCounter` + `RateLimiter` + `CallHistory`, and the new compile-time assertions pin both built-in backends to it, so a dropped method becomes a build failure instead of a silently disabled condition. Good fail-closed-preserving design, and it directly addresses the discoverability gap from #280.
- The `callcounter.Store` removal is genuinely safe on this base: the diff touches only the type annotation in `main.go`/`route.go` and drops the now-unused `callcounter` import from `route.go`. `memory.go`, `redis.go`, and `callcounter_test.go` never named the `Store` interface (they aren't in the diff), so nothing dangles — the backends satisfy `capability.CallCounter` structurally.
- `recordSessionCall` still gates marker writes on the `CallHistory` type assertion, so the new doc claim that `IncrementAndGet` alone backs no condition's full check stays accurate.
- The breaking removal is correctly typed (`refactor!`) and documented under `CHANGELOG.md` → Removed, consistent with the repo's pre-1.0 no-shim policy. `docs/architecture.md` and the `WithCallCounter` godoc match the code.

**Notes (all non-blocking)** — three inline. The only one I'd actually weigh is the first: the two improved denial `Message`s are returned to the MCP host, not only the audit log, and they now embed Go-internal type names — worth a quick decision on the intended audience. The other two are an intentional-coupling FYI and a redundant-assertion nit.

Nice cleanup overall.

---

pkg/enforcement/handlers.go
eunolabs: Good catch — agreed the host shouldn't see Go-internal type names. Fixed in 786134f: both `Message`s are back to their capability-oriented wording, and the precise missing method + extension now go into structured `Details` (`reason: counter_missing_extension`, `requiredMethod`, `requiredExtension`), which reaches the audit log. This mirrors how these handlers already populate `Details` (`retryAfter`, `afterTool`, `reason`) and keeps prose out of the structured field. Tests assert the new `Details` and that the host `Message` no longer leaks the extension type name. Applies to both maxCalls and sequenceBlock.

---

pkg/enforcement/enforcement_test.go
eunolabs: Keeping the granular assertions deliberately. Two reasons: they buy a localized compile error that names the exact missing extension (worthwhile for a security-relevant contract pin), and they're not fully subsumed in the other direction — `FullCallCounter` additionally pins the base `IncrementAndGet`, which the four granular lines don't cover, so the two sets are complementary rather than purely redundant. Tightened the comment to say so.

---

eunolabs:
## Re-review of `4e4d3de` (commit `786134f` + test follow-up)

Re-reviewed after the push. All three of my earlier comments are resolved — **no further review comments from me.**

- **handlers.go denial messages** — verified the fix is complete, not just relocated. Both `Message`s are back to capability-oriented wording, and the missing method/extension now go into structured `Details` (`reason`/`requiredMethod`/`requiredExtension`). I traced the host path to confirm the type name no longer reaches the MCP host: `denialResult` only surfaces `code`/`conditionType`/`target`/`argument` into the JSON-RPC error (and `denialArgument` whitelists *only* `Details["argument"]`), while the full `Details` map flows solely to the audit log via `sink.Record(..., denial.Details, ...)`. That matches the existing "audit gets full Details, host gets a sanitized whitelist" pattern, and the new test assertions (`NotContains(Message, ...)` + the `Details` keys) pin it. Good.
- **enforcement_test.go assertions** — agreed with keeping the granular + bundled assertions; the tightened comment captures it well, and the point that the `FullCallCounter` lines additionally pin the base `IncrementAndGet` (which the four granular lines don't) is a fair complement I'd undercredited.
- **capability → enforcement doc reference** — was an FYI only; no change needed.

One non-review heads-up: the PR currently shows a merge conflict against `main` (`mergeable_state: dirty`), so it'll need a rebase before it can merge. CI is re-running on this head; I'll flag any failures.

Nice work on the turnaround.

---

Issue #168, opened 2026-06-13
bug: in-memory call counter Cleanup wipes live maxCalls counters for windows > 24h (fail-open rate-limit bypass)


eunolabs:
## Summary

`InMemory.Cleanup` uses a **hardcoded 24h staleness threshold** to decide when to drop a counter entry. For any `maxCalls` condition whose `windowSeconds` exceeds 24h (e.g. a weekly or monthly quota), a session that goes idle for more than 24h has its still-live counter deleted by the background cleanup goroutine. The next call then restarts the count at 1, **silently resetting the quota mid-window** — a fail-open rate-limit bypass that contradicts the project's fail-closed invariant.

This affects the **default** counter backend (in-memory). The Redis backend is **not** affected, because it scales its TTL with the window.

## Where

`pkg/callcounter/memory.go` — `Cleanup`:

https://github.com/eunolabs/eunox/blob/main/pkg/callcounter/memory.go#L126-L142

```go
// Cleanup removes expired entries. Call periodically to prevent memory growth.
func (m *InMemory) Cleanup() {
	m.mu.Lock()
	defer m.mu.Unlock()

	now := m.now()
	for key, e := range m.entries {
		if len(e.timestamps) == 0 {
			delete(m.entries, key)
			continue
		}
		// If the most recent timestamp is old enough, the whole entry is stale
		if e.timestamps[len(e.timestamps)-1].Before(now.Add(-24 * time.Hour)) {
			delete(m.entries, key)
		}
	}
}
```

The `-24 * time.Hour` constant is independent of the window the entry was counting. `IncrementAndGet` accepts an arbitrary `windowSec` per call, but `Cleanup` assumes every entry's window is ≤ 24h.

## Why it is reachable in the shipped binary

1. **In-memory is the default backend.** The proxy uses `callcounter.NewInMemory()` unless `--redis-addr` is set (`cmd/eunox-mcp/main.go:234`).
2. **The cleanup goroutine is wired up.** `main.go` starts it on a 5-minute tick for the in-memory counter (`cmd/eunox-mcp/main.go:299-301`, `DefaultCleanupInterval = 5 * time.Minute`).
3. **`windowSeconds` has no upper bound.** `MaxCallsCondition` (`pkg/capability/condition.go:76-80`) is not range-checked anywhere in `pkg/capability/validate.go`, so a manifest may legitimately specify `windowSeconds: 604800` (7-day quota), `2592000` (30-day), etc.
[...]

PR #290, opened 2026-06-13
sec: isolate in-memory call-counter windows to fix maxCalls bypass (#277)

### LINKED ISSUE #277: sec: InMemory backend prunes cross-condition timestamps, enabling maxCalls bypass for multi-window policies
## Summary

The `InMemory` call-counter backend shares a single `*entry` (timestamp slice) across all `windowSec` values for the same logical key. When a capability has **two `maxCalls` conditions with different `windowSeconds`** on the same tool, the shorter-window condition's prune step destroys timestamps that are still live under the longer window, causing the longer-window condition to systematically **undercount admitted calls** — a fail-open rate-limit bypass.

The `Redis` backend does **not** have this bug because it namespaces each sorted set by `windowSec` (key format `callcounter:<key>:<windowSec>`), giving each window its own isolated set. `InMemory` uses the raw caller-supplied key (`m.entries[key]`), so all windows share one slice.

## Reproduction

```yaml
capabilities:
  - target: tool:dangerous_export
    actions: [call]
    conditions:
      - type: maxCalls
        count: 5
        windowSeconds: 60    # burst: 5/min
      - type: maxCalls
        count: 10
        windowSeconds: 3600  # sustained: 10/hour
```

With `InMemory`:

1. **Minute 1 (T=0–59):** Client makes 5 calls. Both conditions admit each call. Because both conditions call `IncrementIfBelow` on the same `*entry`, each call appends **2 timestamps** (one per condition). After 5 calls, `e.timestamps` has **10** entries at T≈0.

2. **Minute 2 (T=61):** Client makes a 6th call.
   - Condition 1 (60s window) prunes all entries older than 60s — that is, **all 10** T=0 timestamps. Entry is now empty.
   - Condition 1 admits the 6th call (count 0 < 5) and appends T=61.
   - Condition 2 (3600s window) sees only **1** timestamp (T=61). It admits the call (1 < 10).

   The correct count for the 3600s window is **6** (5 from minute 1 + this call). `InMemory` reports **1**, allowing 9 more calls before the hour limit is reached. The actual number of admitted calls can reach **5 × N/min × 60min = 300** instead of the configured 10/hour.

## Root Cause

`InMemory.IncrementIfBelow` indexes `m.entries` by the raw `key` argument:

```go
// memory.go
e, ok := m.entries[key]
```

`IncrementAndGet` and `Peek` do the same. The `windowSec` parameter is **only** used to compute a cutoff for filtering, not as part of the storage key.

`Redis.IncrementIfBelow` (and `IncrementAndGet`, `Peek`) derives a window-specific key:

```go
// redis.go
windowKey := fmt.Sprintf("callcounter:%s:%d", key, windowSec)
```

## Impact

Any deployment using `InMemory` (the default for single-replica deployments and the only backend available without Redis) where a single capability defines **multiple `maxCalls` conditions with different `windowSeconds`** is affected. The shorter-window condition prunes timestamps for all longer windows, making sustained-rate limits ineffective.

## Proposed Fix

Two complementary options (both needed for full correctness):

### Option A — Fix in `InMemory` (mirrors Redis)

Namespace `m.entries` by `windowSec` so each window gets an isolated slice, matching Redis's behaviour:

```go
// memory.go
func (m *InMemory) IncrementIfBelow(_ context.Context, key string, windowSec int, limit int64) ... {
    internalKey := fmt.Sprintf("%s:%d", key, windowSec)
    e, ok := m.entries[internalKey]
    ...
}
```

Apply the same to `IncrementAndGet` and `Peek`. The `e.windowSec` field becomes redundant (windowSec is now encoded in the key); Cleanup can use the encoded windowSec or a fixed margin.

### Option B — Fix in the caller (`handleMaxCalls`)

Include `windowSeconds` in the key passed to `IncrementIfBelow`:

```go
```

This is the minimal targeted fix and produces behaviour consistent with Redis's existing namespacing.

**Option A is the preferred long-term fix** as it makes `InMemory` self-consistently correct regardless of caller key conventions.

## Related

- The double-counting side-effect (each admitted call appends N timestamps for N maxCalls conditions sharing an entry) is a secondary symptom of the same root cause and is fixed by either option above.
- `IncrementAndGet` and `Peek` in `InMemory` have the same key-namespacing gap but their callers (sequenceBlock history) always use a fixed `windowSec` (`sequenceHistoryWindowSec = 86400`), so no multi-window interference arises there in practice.

eunolabs:
## Summary

Fixes a fail-open rate-limit bypass (#277) in the in-memory call-counter backend.

The `InMemory` backend indexed `m.entries` by the raw caller key, so **all `windowSec` values for one key shared a single timestamp slice**. When a capability carried two `maxCalls` conditions with different `windowSeconds` on the same tool (e.g. a `5/min` burst limit alongside a `10/hour` sustained limit), the shorter window's prune destroyed timestamps still live under the longer window. The longer window then **undercounted admitted calls** and its sustained rate limit failed open — a client could drive far more calls per hour than configured.

The `Redis` backend was never affected: it already keys each window in its own sorted set (`callcounter:<key>:<windowSec>`).

## Fix (Option A from the issue — the preferred long-term fix)

- Namespace `InMemory` entries by `(key, windowSec)` via a new `entryKey` helper, mirroring Redis so the two backends count identically.
- `windowSec` is now fixed per entry (encoded in the map key), so it is set once at creation for `Cleanup`'s staleness threshold — the long-window-quota preservation (issue #168) is retained.
- This also removes the related **double-counting** side effect, where each admitted call appended one timestamp per condition sharing the entry.
- `sequenceBlock` history and single-window `maxCalls` are unchanged (their callers use a single fixed window, so `Peek`/record still agree).

## Tests

- `TestInMemory_IncrementIfBelow_MultiWindowIsolation` — direct regression for the issue's reproduction (burst + sustained windows count independently; fails on the old shared-slice code at both the double-count and the post-prune steps).
- `TestRedis_IncrementIfBelow_MultiWindowIsolation` — parity test pinning the two backends together.
- Rewrote `TestInMemory_Peek_RespectsWindow` (cross-window visibility is intentionally removed — now matches Redis's window-namespaced `Peek` contract) and updated the in-package tests that index the entries map by raw key.
- Full suite green under `go test -race -count=1 ./...`; `gofmt`/`go vet` clean.

## Docs

- `CHANGELOG.md` — Security entry.
- `docs/threat-model-mcp.md` — §3.6 multi-window note, in-memory footprint now `(session, tool, window)`, revision row 1.12.
- `docs/capability-manifest-guide.md` — note that multiple `maxCalls` windows on one target are enforced independently.

Fixes #277

https://claude.ai/code/session_015j8ervVP6td7rD1Wuo6Nwf

---
_Generated by [Claude Code](https://claude.ai/code/session_015j8ervVP6td7rD1Wuo6Nwf)_


pkg/callcounter/memory.go
```diff
@@ -113,6 +116,18 @@ func (m *InMemory) admitNewKey() error {
...
 
+// entryKey namespaces the caller-supplied key by windowSec so each window gets
+// its own isolated timestamp slice. Without it, two maxCalls conditions on the
+// same tool with different windowSeconds shared a single slice, and the
+// shorter-window prune dropped timestamps still live under the longer window —
+// the longer window then undercounted admitted calls and its rate limit failed
+// open (issue #277). This mirrors the Redis backend, which already isolates each
+// window in its own sorted set keyed "callcounter:<key>:<windowSec>" (redis.go),
+// so the two backends count identically.
+func entryKey(key string, windowSec int) string {
```
eunolabs, line 128: The `entryKey` namespacing is the right fix and correctly mirrors the Redis backend's `callcounter:<key>:<windowSec>` keying. Two things to be aware of when this gets rebased (see the summary comment for the full picture):

- Current `main`'s `IncrementAndGet` is the **4-arg** `(ctx, key, windowSec, maxEntries int)` form (retention cap, #265), and `Cleanup` is the **two-phase batched** version with `shouldDeleteEntry`/`cleanupDeleteBatch` (#256). This branch predates both, so `entryKey` needs to be re-applied on top of them. The good news: it composes cleanly — each `(key, windowSec)` entry independently caps at `maxEntries`, and `shouldDeleteEntry` just reads the now-fixed `e.windowSec`, so no logic in either feature has to change.

- Minor / non-blocking (and pre-existing parity with Redis): the key is built with an unescaped `:` delimiter (`fmt.Sprintf("%s:%d", key, windowSec)`), so in principle a caller-supplied `key` that already ends in `:<int>` could alias another `(key, windowSec)` pair's bucket. In practice the keys are server-controlled (`maxcalls:<uuid>:<type>:<name>`, `seq:...`) with the window always appended last, and Redis uses the identical construction, so this is fine to leave as-is — just flagging the unescaped delimiter.

---

eunolabs (commented):
## Review: isolate in-memory call-counter windows (#277)

**The core fix is correct and well-targeted.** Namespacing in-memory entries by `(key, windowSec)` is the right approach: it makes the in-memory backend count identically to Redis (which already keys each window in its own sorted set, and was never affected), it removes both the post-prune undercount of the longer window *and* the per-condition double-count, and the regression tests (`TestInMemory_IncrementIfBelow_MultiWindowIsolation` plus the `TestRedis_..._MultiWindowIsolation` parity test) pin the behavior on both backends so they can't drift.

I also verified the load-bearing safety argument behind the new "`Peek`/record must use the same window" contract that replaces the old cross-window visibility:
- The only `IncrementAndGet`/`Peek` caller is `sequenceBlock` history, which records (`engine.go` `recordSessionCall`) and reads (`handlers.go` `handleSequenceBlock`) both at the fixed `sequenceHistoryWindowSec` (24h) → same `entryKey`, so they still agree.
- `maxCalls` uses `IncrementIfBelow` with each condition's own `windowSeconds` → which is exactly the isolation this fixes.

So no production caller relied on the cross-window `Peek` visibility that the rewritten `TestInMemory_Peek_RespectsWindow` intentionally drops. The doc/threat-model updates are accurate for the change.

---

### 🔴 Blocking: the branch has fallen behind `main` and now conflicts (`mergeable_state: dirty`)

This PR forked from `main` *before* several call-counter changes merged, so `pkg/callcounter/memory.go` here is an older revision with the window-namespacing fix layered on top. The 3-dot PR diff is computed against that earlier fork point, which is why it doesn't surface the divergence — but comparing the actual files at the head SHA (`c69cf00a`) against current `main` (`44708da`):

- **`IncrementAndGet` signature reverts #265.** Current `main` — and even this PR's *recorded* base `c1ca6458` (which is #265 itself) — has the **4-arg** `IncrementAndGet(ctx, key, windowSec, maxEntries int)` retention cap (`fix!: bound call-counter memory with IncrementAndGet retention cap`, #233). This head has the **3-arg** version with no `maxEntries` / `checkMaxEntries` / cap logic.
- **`Cleanup` reverts #256.** Current `main` has the **two-phase batched** `Cleanup` + `shouldDeleteEntry`/`cleanupDeleteBatch` (`perf: split InMemory.Cleanup into scan + batched-delete phases`, #237). This head has the older single-phase `Cleanup`.

Merging or naively rebasing as-is would therefore **revert #265 and #256**, and it wouldn't even compile against the current `engine.go`, which calls `IncrementAndGet(..., sequenceHistoryWindowSec, sequenceHistoryMaxEntries)` with the 4th argument. The all-green CI on this PR ran against the branch's old fork point, so it does **not** reflect a merge with current `main` and can't catch this.

**Suggested fix:** rebase onto current `main` and re-apply only the window-namespacing change on top of the 4-arg `IncrementAndGet` and the two-phase `Cleanup`. As noted in the inline comment, the fix is orthogonal to both features and should compose cleanly, but the touched test files (`cleanup_internal_test.go`, `compact_internal_test.go`) and docs (`docs/threat-model-mcp.md` revision rows — `main` likely already has rows past 1.11 from #266 etc. — and `CHANGELOG.md`) will need conflict resolution against the newer `main` too.

---

### Non-blocking
- See the inline note on `entryKey` re: the unescaped `:` delimiter (pre-existing parity with Redis, fine to leave).
- After the rebase onto the 4-arg `IncrementAndGet`, it's worth a quick re-confirm that `sequenceHistoryMaxEntries = 1` still behaves as intended now that the `seq:` history is window-namespaced (it should — record and Peek both use 24h, so they land in the same capped entry).

Once the rebase is done I'll re-review the updated diff.

---

eunolabs:
Rebased onto current `main` (`3b83102`), which now carries the 4-arg `IncrementAndGet` retention cap (#265), the two-phase batched `Cleanup` + `shouldDeleteEntry` (#256), and the more recent #268/#271/#283. As you predicted, the window-namespacing composes cleanly on top:

- `entryKey(key, windowSec)` is re-applied over the 4-arg `IncrementAndGet`, so each `(key, windowSec)` entry caps at `maxEntries` independently.
- `shouldDeleteEntry` just reads the now-fixed `e.windowSec` — no change to the cap or two-phase Cleanup logic.
- The in-package tests that index the entries map by raw key (`compact_internal_test.go`, `cleanup_internal_test.go`) are updated to the namespaced key. Notably, the `TestInMemory_Cleanup_ConcurrentIncrementNotDropped` stress test now seeds each stale entry under `entryKey(key, windowSec)` so `IncrementAndGet` refreshes that same bucket — otherwise it would have stopped exercising the phase-2 re-check race it exists for.

Left the unescaped `:` delimiter as-is per your non-blocking note (Redis parity; keys are server-controlled with the window appended last).

One correction to the PR description: the threat-model revision row is **1.13** (1.12 was already taken by the #233 per-key-size bound now on `main`), and the §3.6 footprint note reads `(session, tool, window)`.

Full suite is green locally under `go test -race -count=1 ./...`; CI is re-running on the rebased commit. Ready for re-review.

---

PR #333, opened 2026-06-14
sec: namespace InMemory call-counter entries by window; document/pin counter contract (#277–#280)

### LINKED ISSUE #277: sec: InMemory backend prunes cross-condition timestamps, enabling maxCalls bypass for multi-window policies

### LINKED ISSUE #279: bug: StartCleanup is not idempotent — multiple calls spawn redundant background goroutines
## Summary

`InMemory.StartCleanup` has no guard against being called more than once on the same instance. Each call unconditionally spawns a new cleanup goroutine. In a codebase where the initialization path is exercised more than once (e.g., tests that reconstruct the counter, or a server that calls `StartCleanup` after a config reload), redundant goroutines accumulate and compete under the shared mutex on every tick.

## Current behaviour

```go
// memory.go
func (m *InMemory) StartCleanup(ctx context.Context, interval time.Duration) {
    if interval <= 0 {
        interval = DefaultCleanupInterval
    }
    go func() {          // no guard — launches unconditionally
        t := time.NewTicker(interval)
        defer t.Stop()
        for { ... }
    }()
}
```

Calling `StartCleanup` twice on the same `*InMemory` with the same context produces two goroutines that both call `m.Cleanup()` at every tick, holding `m.mu` twice as often as needed. With N calls there are N goroutines, all doing identical work.

## Impact

- **Goroutine accumulation**: In long-running binaries or test suites that repeatedly initialise the counter, goroutines pile up until the context is cancelled. Under `-race` the race detector can also flag the doubled concurrent Cleanup calls if the timing aligns.
- **Mutex contention**: Every extra goroutine holds `m.mu.Lock()` for the duration of `Cleanup`, blocking `IncrementAndGet` / `IncrementIfBelow` / `Peek` callers unnecessarily.
- **Misleading behaviour in tests**: A test that calls `StartCleanup` in a `TestMain` or `setup` function and then again in a subtest will observe cleanup running at double (or more) the expected rate, making timing-sensitive assertions flaky.

## Proposed Fix

Add a `sync.Once` (or a simple atomic flag) to make the first call effective and all subsequent calls no-ops:

```go
type InMemory struct {
    mu           sync.Mutex
    entries      map[string]*entry
    now          func() time.Time
    cleanupOnce  sync.Once   // guards StartCleanup
}

func (m *InMemory) StartCleanup(ctx context.Context, interval time.Duration) {
    if interval <= 0 {
        interval = DefaultCleanupInterval
    }
    m.cleanupOnce.Do(func() {
        go func() {
            t := time.NewTicker(interval)
            defer t.Stop()
            for {
                select {
                case <-t.C:
                    m.Cleanup()
                case <-ctx.Done():
                    return
                }
            }
        }()
    })
}
```

Alternatively, document the once-only contract more prominently and add a check that panics (similar to `sync.Once`-based patterns in the stdlib) so misuse surfaces immediately rather than silently.

## Notes

- The `interval` argument is only meaningful on the first call; subsequent calls with a different interval are silently ignored under either fix — that is acceptable given the once-only semantics.
- `Cleanup` itself is safe to call concurrently (it takes `m.mu`), so the goroutine accumulation is a correctness/performance concern rather than a data-race.

### LINKED ISSUE #280: design: CallCounter / Store interfaces underspecify requirements for maxCalls and sequenceBlock conditions

eunolabs:
## Summary

Addresses the call-counter issue batch #277–#280. The substantive change is the #277 security hardening; the rest are documentation, compile-time pins, and cross-references, since #278/#279 and the `pkg/callcounter` side of #280 already landed in earlier work.

### #277 — InMemory cross-window prune → `maxCalls` fail-open (security) — **fixed**

The `InMemory` backend indexed `m.entries` by the **raw** caller key, so two `maxCalls` conditions on one capability with **different `windowSeconds`** shared a single timestamp slice. The shorter window's prune (which fires the moment its window elapses) destroyed timestamps the longer window still counted, so the sustained-rate quota silently failed **open**.

The Redis backend already isolates windows via its `callcounter:<key>:<windowSec>` sorted set. `InMemory.IncrementAndGet` / `IncrementIfBelow` / `Peek` now derive the same per-window storage key (`storageKey(key, windowSec)`), so each window keeps its own history regardless of how a caller composes the logical key — the issue's preferred **Option A**. This fixes the root cause at the backend layer; the caller-side window key from #278 (Option B) masked it only for the shipped caller.

- **No behaviour change for the shipped callers** (`maxCalls` already folds the window into its key; `sequenceBlock` uses a single fixed 24h window).
- New backend-level regression tests: `TestInMemory_IncrementIfBelow_CrossWindowPruneIsolated` (the 5/min + 10/hour scenario from the issue) and `TestStorageKey_NamespacesByWindow`.
- `TestInMemory_Peek_RespectsWindow` updated: a `Peek` at a different window than the recording call now reads a separate bucket (0), matching Redis and the documented "windowSec must match" contract — the old InMemory cross-window read was the very leakage #277 is about.

### #280 — CallCounter contract underspecified — **fixed**

`WithCallCounter` accepts the narrow `capability.CallCounter` (`IncrementAndGet`), while the engine reaches for `IncrementIfBelow` (`enforcement.RateLimiter`, for `maxCalls`) and `Peek` (`enforcement.CallHistory`, for `sequenceBlock`) by runtime type assertion — a bare backend compiles but makes both conditions fail closed with an opaque deny.

- `capability.CallCounter` now documents which optional capability each condition needs and how to implement/pin it.
- New `cmd/eunox-mcp/counter_contract.go` pins both built-in backends against the **consumer-side** `enforcement.RateLimiter` / `enforcement.CallHistory` (and `capability.CallCounter`), in the one binary that imports both packages — complementing the producer-side `LimiterStore`/`HistoryStore` pins from #236. (`pkg/enforcement` deliberately does not import `pkg/callcounter`, so the pins live in the binary.)

### #279 — StartCleanup idempotency — already resolved, cross-referenced

`StartCleanup` is already guarded by `sync.Once` (landed under #235) with idempotency tests. Cross-referenced the duplicate in the doc comment and CHANGELOG; no code change required.

### #278 — multiple-`maxCalls` bucket collision — already fixed, reinforced

Already resolved by the window-keyed counter buckets + manifest validation. The #277 backend fix now defends it in depth (the in-memory double-count and cross-window prune both stem from the shared-entry root cause this change removes).

## Verification

- `go vet ./...` / `make lint` — 0 issues
- `make check-fmt`, `make check-license` — clean

Fixes #277
Fixes #279
Fixes #280
Relates to #278

https://claude.ai/code/session_01B7znbzckMNEtsUYuvhDotg

---
_Generated by [Claude Code](https://claude.ai/code/session_01B7znbzckMNEtsUYuvhDotg)_


pkg/callcounter/memory.go
```diff
@@ -14,15 +15,38 @@ import (
...
+// every caller to fold the window into the key itself.
+//
+// The window is appended as a trailing ":<decimal>" segment. windowSec is
+// rendered as digits that contain no colon, so the final colon in the result
+// always separates key from window: distinct (key, windowSec) pairs always map
+// to distinct strings even when key itself contains colons. The internal key is
+// only ever constructed here and used for map lookup — never parsed back — so no
+// caller can forge a colliding pair.
+func storageKey(key string, windowSec int) string {
+	return key + ":" + strconv.Itoa(windowSec)
```
eunolabs, line 47: Core fix looks correct. Namespacing the entry by `(key, windowSec)` is exactly what removes the cross-window prune, and it brings `InMemory` into line with the Redis backend's `callcounter:<key>:<windowSec>` buckets (redis.go:92/185/226).

The tightened `Peek` semantics this introduces (a `Peek` at a window other than the recording window now reads a *separate* bucket and returns 0, instead of the old "wider window still sees it") are safe only under one invariant: **for any given logical key, the record window and the read window always match.** I checked both consumers and it holds:

- `sequenceBlock` — recorded via `IncrementAndGet(..., sequenceHistoryWindowSec)` (engine.go:329) and read via `Peek(..., sequenceHistoryWindowSec)` (handlers.go:1024), the same `86400` constant for both.
- `maxCalls` — the check-and-record is a single atomic `IncrementIfBelow`, so the window can't diverge.

So no production caller is affected, and this actually *fixes* a latent InMemory/Redis divergence (InMemory's old wider-window `Peek` leaked across buckets; Redis never did — see its `Peek` doc, redis.go:217). Worth calling out for any future caller that does an out-of-band `Peek`: it must pass the same `windowSec` the call was recorded under, which is now a hard requirement rather than a best-effort.

---

pkg/callcounter/memory.go
```diff
@@ -144,18 +168,21 @@ func (m *InMemory) IncrementAndGet(_ context.Context, key string, windowSec, max
...
+		m.entries[sk] = e
 	}
 
-	// Record the window so Cleanup can scale its staleness threshold to it.
-	// Take the max in case the same key is ever counted with differing windows,
-	// so cleanup never drops an entry that is still live under any of them.
+	// Record the window so Cleanup can scale its staleness threshold to it. Every
+	// call that reaches this entry carries the same window (the map is keyed by
+	// it), so this assigns once on creation and is stable thereafter.
 	if windowSec > e.windowSec {
```
eunolabs: Minor / optional: now that the map is keyed by `(key, windowSec)`, every call reaching a given `e` carries the same `windowSec`, so this `if windowSec > e.windowSec` can only ever fire on first insert — the original "take the max in case the same key is counted with differing windows" rationale no longer exists. You could fold it into the `if !ok` branch above (`e = &entry{windowSec: windowSec}`) and drop the conditional here (same in `IncrementIfBelow`). Not blocking — the new comment already notes it's set-once; just flagging that the `>` guard is now vestigial and slightly misleading to a future reader.

---

eunolabs (commented):
## Review — looks solid, security fix is correct

### What it does
Fixes the #277 fail-open in the `InMemory` call counter: the backend indexed `m.entries` by the raw caller key, so two `maxCalls` conditions on one capability with *different* windows (e.g. 5/min + 10/hour) shared one timestamp slice. The shorter window's prune wiped timestamps the longer window still counted, silently failing the sustained-rate quota **open**. The fix namespaces each entry by `(key, windowSec)` via a new `storageKey`, matching the Redis backend's per-window sorted set. The rest is docs/pins: a `capability.CallCounter` doc spelling out the optional `IncrementIfBelow`/`Peek` capabilities `maxCalls`/`sequenceBlock` need (#280), consumer-side compile-time pins in `cmd/eunox-mcp/counter_contract.go`, and cross-references for #278/#279.

### Correctness — verified
- **Record/read window invariant holds** (the property the tightened `Peek` rests on): `sequenceBlock` records and reads both under `sequenceHistoryWindowSec` (engine.go:329 / handlers.go:1024); `maxCalls` is atomic via `IncrementIfBelow`. No production caller mixes windows on one key, so the behaviour change is invisible to them and actually removes a prior InMemory↔Redis divergence. (Inline note on `storageKey`.)
- **Supporting claims check out**: Redis really does key by `callcounter:<key>:<windowSec>` (redis.go:92/185/226); `enforcement.RateLimiter`/`CallHistory` and `callcounter.LimiterStore`/`HistoryStore` exist, so the new pins and doc references are valid.
- **No new memory/DoS surface**: shipped callers already fold the window into the logical key (`maxCalls`, #278) or use one fixed window (`sequenceBlock`), so physical entry count is unchanged; window count is manifest-bounded, not attacker-controlled.

### Tests
Good coverage. `TestInMemory_IncrementIfBelow_CrossWindowPruneIsolated` reproduces the issue's 5/min + 10/hour scenario and drives the hourly quota to its boundary (admit 7–10, deny 11), `TestStorageKey_NamespacesByWindow` pins the structural guarantee, and the internal `_test.go` map lookups were correctly migrated to `storageKey(...)`. The `TestInMemory_Peek_RespectsWindow` rewrite correctly encodes the new same-window-bucket semantics; non-mutation coverage is preserved by the separate `TestInMemory_Peek_DoesNotMutate`.

### Conventions / CI
SPDX headers present, comment style consistent, no emojis, fail-closed preserved, CHANGELOG updated. All 42 checks green (Test, Lint, gofmt, License Headers, govulncheck, CodeQL, Trivy, Enforcement compliance, demo stdio/JWT/gateway).

### Nits (non-blocking)
1. The `if windowSec > e.windowSec` max-tracking is now vestigial under window-keyed entries — see inline on memory.go:186.
2. Observation, not a request: with the backend fix in place, the #278 caller-side window-folding in `handleMaxCalls` now doubles the window segment in the physical key (`...:<win>:<win>`). The PR documents this as harmless defense-in-depth, which is fair; just noting the caller-side fold could eventually be retired now that isolation is guaranteed at the backend layer.

Overall this is a clean root-cause fix with the latent fail-open closed at the right layer. Happy to see it merge once the nits are considered.

---

eunolabs:
Thanks for the thorough review. Addressed the actionable nit in `345fb2b`:

- **`entry.windowSec` max-tracking guard (memory.go nit #1):** removed. Since the map is now keyed by `(key, windowSec)`, the window is set once at entry creation (`e = &entry{windowSec: windowSec}`) in both `IncrementAndGet` and `IncrementIfBelow`, and the vestigial `if windowSec > e.windowSec` conditional is dropped. Behaviour-identical, and no longer misleading to a future reader.

On nit #2 (the caller-side window fold in `handleMaxCalls` now double-encoding the window in the physical key): agreed it's now redundant with backend-level isolation. Leaving the caller-side fold in place for this PR as documented defense-in-depth so the change stays scoped to the fail-open fix; retiring it is a clean follow-up once we're comfortable relying solely on the backend guarantee.

---

Issue #296, opened 2026-06-13
callcounter: implementing Store alone is insufficient for maxCalls/sequenceBlock — RateLimiter and CallHistory not visible from the package


eunolabs:
## Summary

`callcounter.Store` (and `capability.CallCounter`) declares only `IncrementAndGet`. The additional `RateLimiter` (`IncrementIfBelow`) and `CallHistory` (`Peek`) interfaces — required for `maxCalls` and `sequenceBlock` conditions respectively — are defined in `pkg/enforcement` and detected by runtime type assertion. A third-party implementation that satisfies only `Store` compiles cleanly but fails closed at runtime on every `maxCalls` and `sequenceBlock` condition, with errors that point to the enforcement engine rather than the counter implementation.

## Current behavior

```
"configured call counter does not support the atomic rate-limit check required by maxCalls"
"configured call counter does not support the session-history lookups required by sequenceBlock"
```

These are surfaced as `CONDITION_FAILED` denials — correctly fail-closed — but operators and library authors reading `callcounter.Store` have no way to discover that these methods are also required until they hit those denials in production.

## Affected files

- `pkg/callcounter/callcounter.go` — `Store` interface (only `IncrementAndGet`)
- `pkg/enforcement/engine.go` — `RateLimiter` and `CallHistory` interfaces
- `capability.CallCounter` in `pkg/capability/enforce.go` — also just `IncrementAndGet`

## Proposal

Three options in increasing strength:

**A. Godoc only**: Add a package-level comment in `pkg/callcounter` and on the `Store` interface warning that `maxCalls` requires `IncrementIfBelow` and `sequenceBlock` requires `Peek`, with a pointer to `pkg/enforcement.RateLimiter` and `pkg/enforcement.CallHistory`.

**B. Re-export the optional interfaces**: Re-export `RateLimiter` and `CallHistory` from `pkg/callcounter` (or `pkg/capability`) so implementors can find them without diving into the enforcement engine.

**C. Extend the interface**: Merge all three methods into a single `Store` interface (or introduce a `FullStore` embedding all three) so a compile-time check (`var _ callcounter.FullStore = (*MyStore)(nil)`) can catch a missing method at build time.

Option A is the minimum and has zero risk. Option C would require every existing stub to grow two new methods, which is acceptable pre-1.0 (no backwards-compat shims needed per CONTRIBUTING.md).

The discoverability gap is the real issue: the package should not require reading the enforcement engine source to know what the built-in conditions need.

PR #334, opened 2026-06-14
fix(callcounter): correct over-limit retryAfter, unify Redis score encoding, cap window bound (#292–#298)

### LINKED ISSUE #292: callcounter: Redis incrIfBelowScript reports wrong retryAfter when count > limit
## Summary

`incrIfBelowScript` in `pkg/callcounter/redis.go` always fetches the rank-0 (oldest) entry to estimate `retryAfter`. This is correct when `count == limit` (exactly one slot needs to free), but underestimates when `count > limit` — it says "retry when the oldest entry expires" when actually `count - limit + 1` entries must expire before any slot opens.

## When does `count > limit` occur?

When `maxCalls.count` in the manifest is **reduced** (e.g. from 10 to 5) while entries from the previous, more permissive limit are still in the sorted set. After the reload, `count` (after `ZREMRANGEBYSCORE` prunes expired entries) can be 6-10 while `limit = 5`. Every subsequent call is correctly denied but receives an inaccurate retryAfter hint.

## Broken code

`pkg/callcounter/redis.go` — `incrIfBelowScript`:

```lua
local oldest = redis.call('ZRANGE', KEYS[1], 0, 0, 'WITHSCORES')
if limit >= 1 and oldest[2] then
  retry = (tonumber(oldest[2]) + tonumber(ARGV[6])) - tonumber(ARGV[2])
```

This is rank 0 (oldest overall). When `count > limit`, the blocking entry is at rank `count - limit`, not rank 0. Rank 0 expires sooner, so the hint underestimates: a well-behaved caller retries, is denied again (count is still ≥ limit), and retries again — extra churn until the actual blocker ages out.

## Correct fix

Replace the `ZRANGE 0 0` with `ZRANGE (count - limit) (count - limit)`:

```lua
local targetRank = count - limit
local blocker = redis.call('ZRANGE', KEYS[1], targetRank, targetRank, 'WITHSCORES')
if limit >= 1 and blocker[2] then
  retry = (tonumber(blocker[2]) + tonumber(ARGV[6])) - tonumber(ARGV[2])
  if retry < 0 then retry = 0 end
end
```

When `count == limit`, `targetRank = 0` — same as today. The in-memory `IncrementIfBelow` already does this correctly:

```go
// pkg/callcounter/memory.go
if idx := int(cur - limit); idx >= 0 && idx < len(valid) {
    if r := valid[idx].Add(window).Sub(now); r > 0 {
        retry = r
    }
}
```

## Impact

- **Security**: none — the deny decision is correct regardless. retryAfter is purely advisory.
- **Correctness**: callers receive a too-short hint, retry early, are denied again. For limit reductions with many excess entries the gap can be significant (e.g. count=10, limit=5: hint says "retry in 30s" but the slot doesn't open for 180s if the oldest entries were spread over 3 minutes).
- **Scope**: only affects deployments that reduce `maxCalls.count` mid-window. Steady-state `count == limit` behavior is unaffected.

## Tests to add

A table-driven test for `IncrementIfBelow` that reduces the limit mid-window (by pre-seeding entries via `IncrementAndGet` on the same key — or via miniredis direct ZADD) so `count > limit`, then verifies retryAfter equals `valid[count-limit].Add(window).Sub(now)`.

### LINKED ISSUE #293: bug: Redis `IncrementIfBelow` computes wrong `retryAfter` when in-window count exceeds limit
## Summary

`Redis.IncrementIfBelow` always uses `ZRANGE KEYS[1] 0 0 WITHSCORES` (the **oldest** in-window entry) to compute the `retryAfter` hint. This is correct when `count == limit` (exactly one entry must expire before a slot frees up), but gives an underestimated hint when `count > limit` — the situation that arises when a manifest reload reduces `maxCalls.count` while there are already more than the new limit's worth of in-window calls.

## Root cause

`incrIfBelowScript` (redis.go, `var incrIfBelowScript`):

```lua
local oldest = redis.call('ZRANGE', KEYS[1], 0, 0, 'WITHSCORES')
if limit >= 1 and oldest[2] then
  retry = (tonumber(oldest[2]) + tonumber(ARGV[6])) - tonumber(ARGV[2])
```

This always fetches rank 0 (the oldest member). When `count > limit`, slot recovery requires `count - limit + 1` entries to expire, not 1. The index of the entry whose expiry first drops `count` below `limit` is `count - limit` (0-indexed), i.e., `ZRANGE KEYS[1] count-limit count-limit WITHSCORES`.

## Comparison with InMemory

`InMemory.IncrementIfBelow` (memory.go:303-307) correctly uses `valid[cur-limit]`:

```go
if idx := int(cur - limit); idx >= 0 && idx < len(valid) {
    if r := valid[idx].Add(window).Sub(now); r > 0 {
        retry = r
    }
}
```

This is the `(count - limit)`-th oldest entry — the one whose expiry drops the in-window count to `limit - 1 < limit`. The Redis backend diverges from this.

## When does `count > limit` occur?

The `IncrementIfBelow` key does **not** include the limit value — only session, target type, and tool name. If an operator reloads the manifest with a lower `maxCalls.count` while the old (higher) call set is still in the sliding window, the next denied call will see `count > limit`. The `retryAfter` hint tells the caller it can retry sooner than it actually can; the caller retries and is immediately denied again (correct fail-closed behaviour) but with a worse user experience than necessary.

## Proposed fix

In `incrIfBelowScript`, replace the oldest-entry lookup:

```lua
-- Before (wrong when count > limit):
local oldest = redis.call('ZRANGE', KEYS[1], 0, 0, 'WITHSCORES')
if limit >= 1 and oldest[2] then
  retry = (tonumber(oldest[2]) + tonumber(ARGV[6])) - tonumber(ARGV[2])

-- After (consistent with InMemory):
local idx = count - limit   -- 0-based index of the entry that, on expiry, drops count below limit
local pivot = redis.call('ZRANGE', KEYS[1], idx, idx, 'WITHSCORES')
if limit >= 1 and pivot[2] then
  retry = (tonumber(pivot[2]) + tonumber(ARGV[6])) - tonumber(ARGV[2])
```

`idx >= 0` is guaranteed because the deny path requires `count >= limit >= 1`. When `count == limit` (the common case), `idx = 0` and behaviour is unchanged.

## Missing test coverage

There is no test that seeds a counter above the limit (simulating a manifest reload) and verifies that `retryAfter` from both backends agrees. A cross-backend consistency test should be added alongside the fix.

### LINKED ISSUE #294: callcounter: InMemory IncrementIfBelow has no maxEntries cap — grows to limit-sized slice for large quotas
## Summary

`IncrementAndGet` in the InMemory backend takes a `maxEntries` parameter that caps the retained timestamp slice (fix for issue #233). `IncrementIfBelow` has no equivalent cap: it stores one timestamp per **admitted** call, so the in-window slice can grow to `limit` entries. For large quotas this means proportionally large heap use per active key.

## Details

Issue #233 observed that a high-rate key over a long window could accumulate one timestamp per call in the slice, causing unbounded heap growth. The fix capped `IncrementAndGet` at `maxEntries` (typically 1 for sequenceBlock history). `IncrementIfBelow` avoids the worst case because:

1. Denied calls never write a timestamp (issue #204 fix).
2. Expired entries are pruned on every call.
3. The effective cap is `limit` itself — admitted calls can't exceed it.

But for a manifest like `maxCalls: { count: 1_000_000, windowSeconds: 86400 }` (1M calls/day per session), the in-window slice for one session/tool key can reach ~1M `time.Time` values (24 bytes each ≈ 24 MB per key). With many concurrent sessions this adds up.

The Redis backend is immune: sorted-set members are stored in Redis, not on the Go heap, with a per-key TTL.

## Relevant code

`pkg/callcounter/memory.go` — `IncrementIfBelow`:

```go
// Below the limit: record the call.
valid = append(valid, now)
valid = compactTimestamps(valid)
e.timestamps = valid
return int64(len(valid)), true, 0, nil
```

No trim step, so `len(e.timestamps) == count_in_window <= limit`.

Compare `IncrementAndGet`:

```go
// Cap to the most-recent maxEntries by dropping the oldest surplus from the front.
if len(valid) > maxEntries {
    n := copy(valid, valid[len(valid)-maxEntries:])
    valid = valid[:n]
}
```

## Proposed fix

`IncrementIfBelow` only needs presence (was this key admitted at all?) for the sequenceBlock integration. For `maxCalls` itself, the stored count is what matters, not the individual timestamps — the only purpose of storing them is to know when in-window entries age out so the count naturally decreases.

Two options:

**A. Implicit cap at `limit`**: document that `limit` IS the maxEntries cap and add a comment; no code change but make it explicit.

**B. Explicit compaction**: after pruning and before the limit check, if `len(valid) > someInternalCap` (e.g. `min(limit, sequenceHistoryMaxEntries * someMultiplier)`), trim to the most-recent entries. This mirrors the `maxEntries` cap in `IncrementAndGet`.

Option A is simpler and sufficient — the current behavior IS bounded by `limit`; the concern is only that very large limits can translate to large slices. Documenting the bound and pointing large-quota deployments to the Redis backend is probably the right call rather than silently capping and giving a wrong count.

At minimum: add a godoc note on `InMemory` and `NewInMemory` that `IncrementIfBelow` retains up to `limit` timestamps per key per window, and recommend the Redis backend for quotas above a few thousand calls per window.

### LINKED ISSUE #295: code quality: Redis backend score-encoding diverges between `IncrementAndGet` (float) and `IncrementIfBelow` (integer)
## Summary

`IncrementAndGet` and `IncrementIfBelow` in the Redis backend use different score representations for microsecond timestamps stored in sorted sets. The engine prevents them from ever sharing a key by convention (different call-site prefixes: `maxcalls:*` for `IncrementIfBelow`, `seq:*` for `IncrementAndGet`), but the asymmetry is undocumented and fragile.

## The divergence

**`IncrementAndGet`** (redis.go:93–101):
```go
nowUnixMicro := float64(now.UnixMicro())
cutoff := float64(now.Add(-time.Duration(windowSec) * time.Second).UnixMicro())
// ...
pipe.ZRemRangeByScore(ctx, windowKey, "-inf", fmt.Sprintf("%f", cutoff))
pipe.ZAdd(ctx, windowKey, redis.Z{Score: nowUnixMicro, Member: member})
```
Scores are stored as `float64`, and `ZRemRangeByScore` bounds are formatted with `%f` (6 decimal places).

**`IncrementIfBelow` Lua script** (redis.go:148–164):
```lua
redis.call('ZREMRANGEBYSCORE', KEYS[1], '-inf', ARGV[1])   -- ARGV[1] = cutoff (int64 micros)
redis.call('ZADD', KEYS[1], ARGV[2], ARGV[3])              -- ARGV[2] = now (int64 micros)
```
Scores arrive via `ARGV[2]` as `nowMicros := now.UnixMicro()` (int64), passed to Lua as an integer string.

**`Peek`** (redis.go:238):
```go
cutoff := float64(now.Add(-...).UnixMicro())
r.client.ZCount(ctx, windowKey, "("+fmt.Sprintf("%f", cutoff), "+inf")
```
Uses the float-format path, consistent with `IncrementAndGet` but **not** with `IncrementIfBelow`.

## Why it doesn't bite today

Current microsecond timestamps (~1.7 × 10¹⁵) fit exactly in a 64-bit float (2⁵³ ≈ 9 × 10¹⁵), so both representations round-trip without loss and Redis sees identical doubles. The key-prefix convention (`maxcalls:` vs `seq:`) further ensures the two methods never touch the same sorted set.

## Why it matters

1. **Library consumers.** A caller that uses both `IncrementAndGet` and `IncrementIfBelow` on the same key (possible via the `Store` and `RateLimiter` interfaces) would get scores written via two different code paths. Even though the current values are identical, future changes (e.g. nanosecond precision, sub-microsecond scoring) could silently introduce mismatch.

2. **`Peek` inconsistency.** If `Peek` is ever called on a key used by `IncrementIfBelow` (e.g. if the `CallHistory` interface is composed with a `RateLimiter`), the boundary format diverges: `Peek` uses `%f`-formatted float; `IncrementIfBelow`'s `ZREMRANGEBYSCORE` uses integer strings. With current values these are numerically equal, but the asymmetry is invisible in code review.

3. **Documentation gap.** The comment in `IncrementIfBelow` (redis.go:186–190) names the key-prefix separation as the reason the two representations "never meet on one key", but this is an engine-level convention not enforced by the `callcounter` package itself.

## Suggested fix

Standardise on one score representation across all three methods. Integer microseconds (`now.UnixMicro()`) are exact and simpler; updating `IncrementAndGet` and `Peek` to use integer strings removes the asymmetry. Alternatively, make the key-prefix separation a documented invariant in the package-level doc comment and add a package test that asserts calling both methods on the same key is idempotent/safe.

### LINKED ISSUE #297: bug: `MaxWindowSeconds` cannot be cast to `int` on 32-bit platforms; boundary test covers the wrong code path
## Summary

`MaxWindowSeconds` is an `int64` constant (~4.6 × 10⁹), but `IncrementAndGet`/`IncrementIfBelow`/`Peek` all take `windowSec int`. On a 32-bit platform `int` is 32-bit with a maximum of ~2.1 × 10⁹ — less than `MaxWindowSeconds`. This creates two interrelated problems:

1. The documented parameter range `[1, MaxWindowSeconds]` is unreachable on 32-bit (callers can only pass up to `MaxInt32`, which is well below `MaxWindowSeconds`).
2. Test code that computes `huge := int(callcounter.MaxWindowSeconds) + 1` gets a **small positive number** on 32-bit (the int64 value truncates modulo 2³²), which passes `checkWindowSec` without error — the opposite of what the test asserts.

## Details

```go
// callcounter.go
const MaxWindowSeconds = math.MaxInt64 / (cleanupMarginFactor * int64(time.Second))
// ≈ 4,611,686,018  (does not fit in int32 ≈ 2,147,483,647)
```

Test code in `callcounter_test.go:TestInMemory_RejectsOutOfRangeWindow` (and the Redis equivalent):

```go
huge := int(callcounter.MaxWindowSeconds) + 1
// On 32-bit: int(4_611_686_018) + 1
//            = int(4_611_686_018 mod 2^32) + 1
//            = 316_718_722 + 1
//            = 316_718_723   ← a valid positive int!
_, err := counter.IncrementAndGet(ctx, "huge", huge, noTrimCap)
require.Error(t, err, "overflowing windowSec must fail closed")
// FAILS on 32-bit: huge passes checkWindowSec (316_718_723 < MaxWindowSeconds),
// so err == nil and the require.Error assertion panics.
```

### Is this a real overflow hazard in production on 32-bit?

No, not in the usual time.Duration sense. The true risk `MaxWindowSeconds` guards is `time.Duration(windowSec) * cleanupMarginFactor * time.Second` overflowing `int64`. On a 32-bit system, `windowSec` is at most `MaxInt32 ≈ 2.1 × 10⁹`, and `2.1 × 10⁹ × 2 × 10⁹ ns ≈ 4.3 × 10¹⁸ ns < MaxInt64 ≈ 9.2 × 10¹⁸ ns` — so the multiplication is safe. The 32-bit system therefore never needs `MaxWindowSeconds` as an upper bound; the guard is simply misleading/untestable on that architecture.

## Impact

* Any CI run on a 32-bit host would silently skip the overflow code path that `TestInMemory_RejectsOutOfRangeWindow` (and `TestRedis_RejectsOutOfRangeWindow`) is meant to cover, and the test itself would fail with a misleading error.
* The package documentation claims the parameter range is `[1, MaxWindowSeconds]`, which is inaccurate on 32-bit.

## Options

**Option A** — Widen the parameter to `int64` across the `Store`, `RateLimiter`, and `CallHistory` interfaces. This is a breaking change pre-1.0, consistent with the no-shims policy.

**Option B** — Cap `MaxWindowSeconds` at `math.MaxInt` so the constant is always representable as `int` regardless of platform:

```go
const MaxWindowSeconds = min(
    math.MaxInt64/(cleanupMarginFactor*int64(time.Second)),
    math.MaxInt,
)
```

This makes `int(MaxWindowSeconds) == MaxWindowSeconds` on all platforms and fixes the test without an interface change.

**Option C** — Add a 32-bit guard to the test:

```go
if bits.UintSize < 64 {
    t.Skip("MaxWindowSeconds overflow test requires 64-bit int")
}
```

Option B is simplest and keeps the interface stable. Option A is more correct long-term.

### LINKED ISSUE #298: callcounter: InMemory and Redis backends disagree on per-windowSec isolation for the same key
## Summary

`InMemory` maintains a single `entry` struct per key, shared across all `windowSec` values. `Redis` uses a separate sorted set per `(key, windowSec)` pair (`callcounter:<key>:<windowSec>`). These are behaviorally different: calling the same key with two different windows produces mixed timestamps in InMemory but isolated counters in Redis.

## Reproduction

```go
ctx := context.Background()

// InMemory
mem := callcounter.NewInMemory()
mem.IncrementAndGet(ctx, "k", 60, 100)    // adds timestamp T0 to entry["k"]
n, _ := mem.IncrementAndGet(ctx, "k", 3600, 100)
// n == 2: the 60s-window call's timestamp is visible in the 3600s window

// Redis
r := callcounter.NewRedis(client)
r.IncrementAndGet(ctx, "k", 60, 100)     // writes to "callcounter:k:60"
n, _ = r.IncrementAndGet(ctx, "k", 3600, 100)
// n == 1: separate sorted set "callcounter:k:3600" starts empty
```

## Current safe due to invariant, but undocumented

The enforcement engine always uses a consistent `windowSec` per key:
- `sequenceBlock` always uses `sequenceHistoryWindowSec = 86400` for both recording and Peek.
- `maxCalls` uses the `windowSeconds` from the matched constraint, which is fixed per constraint.

So in practice both backends produce the same results. But this invariant is implicit — nothing in `callcounter.Store` or its documentation says "always pass the same `windowSec` for a given key", and a library caller who violates it gets different behavior depending on which backend is deployed.

## Why it matters

1. **Backend substitution**: switching a deployment from InMemory to Redis (or vice versa) can change observable behavior if any caller uses the same key with different windows (even across restarts).
2. **Testing**: tests that run against InMemory may pass but fail against Redis in staging or production.
3. **Future conditions**: a new condition added outside the engine core could inadvertently violate the single-windowSec-per-key rule and get different results on each backend.

## Proposed fix

Document the invariant explicitly in `callcounter.Store.IncrementAndGet` and `callcounter.InMemory`:

> The window size (`windowSec`) for a given key must be consistent across all calls. Mixing window sizes for the same key produces undefined behavior: the InMemory backend merges timestamps from all calls regardless of window, while the Redis backend uses separate sorted sets per window.

Optionally, InMemory could detect and error on inconsistent window sizes for the same key (returning an error when `windowSec != e.windowSec` for an existing entry, rather than silently taking the max), but this would be a breaking change for any caller that legitimately uses the same key with two windows (if any).

eunolabs:
Addresses the `pkg/callcounter` issues #292–#298 in a single change. Three are real fixes (with tests), three are documentation, and one was already resolved by prior work (verified).

### Redis `retryAfter` under-estimate when count > limit — Closes #292, Closes #293
`incrIfBelowScript` always read rank-0 (the oldest entry) to estimate `retryAfter`. That is correct only when `count == limit`. When `count > limit` — the situation a manifest reload that lowers `maxCalls.count` creates while earlier, more permissive calls are still in the window — a slot frees only once enough of the oldest entries expire to drop the count below the new limit, and the last of those to age out is the entry at rank `count - limit`. Rank 0 expires sooner, so the hint under-estimated and a well-behaved client retried early only to be denied again.

The script now reads `ZRANGE key (count-limit) (count-limit)`, matching `InMemory`'s `valid[cur-limit]`. When `count == limit` the rank is 0, so steady-state behaviour is unchanged.

### Uniform Redis score encoding — Closes #295
`IncrementAndGet`/`Peek` formatted sorted-set score bounds as `%f` floats while `IncrementIfBelow` used integer microseconds. The values round-trip to the same double today, but the asymmetry was undocumented and fragile. All three methods now use integer microseconds for scores and bounds, so the encoding is uniform regardless of which method writes or reads a key.

### `MaxWindowSeconds` representable as `int` on 32-bit — Closes #297
`MaxWindowSeconds` (~4.6e9) exceeds `MaxInt32`, so on a 32-bit build `int(MaxWindowSeconds)` truncated — making the documented `[1, MaxWindowSeconds]` range unreachable and the out-of-range tests (which compute `int(MaxWindowSeconds)+1`) exercise the wrong path. It is now `min(overflowBound, math.MaxInt)`. On 64-bit the overflow bound binds unchanged; on 32-bit the int width binds and is the true maximum (a 32-bit `windowSec` can never overflow the cleanup/TTL margin anyway).

### `IncrementIfBelow` retention bound — Closes #294
`IncrementIfBelow` has no `maxEntries` cap; admitted calls each record a timestamp, so an in-memory `maxCalls` key retains up to `limit` timestamps per window. Documented on `LimiterStore.IncrementIfBelow`, `InMemory.IncrementIfBelow`, `NewInMemory`, and in the threat model, with a Redis recommendation for large quotas.

### Per-window isolation invariant — Closes #298
The divergence the issue describes (InMemory merging timestamps across windows) was structurally fixed by #277, which namespaced the InMemory entry map by window (`storageKey`). Both backends now isolate the same key across windows, so backend substitution does not change observed counts. Documented this invariant on `Store.IncrementAndGet`.

## Already resolved — Refs #296
Implementing `Store` alone being insufficient for `maxCalls`/`sequenceBlock` was already addressed by #280: producer-side `callcounter.HistoryStore`/`LimiterStore` and consumer-side `enforcement.RateLimiter`/`CallHistory` both exist, with cross-referencing godoc and compile-time pins in `pkg/callcounter/callcounter_test.go` and `cmd/eunox-mcp/counter_contract.go`, plus implementor guidance in `docs/architecture.md`. Verified — no further change needed. (Left open for a maintainer to close, since this PR adds no code for it.)

## Tests
- `TestRedis_IncrementIfBelow_RetryAfterWhenCountExceedsLimit` and `TestInMemory_IncrementIfBelow_RetryAfterWhenCountExceedsLimit` — cross-backend agreement that, with 6 calls in-window and the limit lowered to 3, both report `retryAfter == 35s` (the rank-3/T=30 entry), not 5s (rank 0/T=0).
- `TestRedis_ScoreEncoding_ConsistentAcrossMethods` — `IncrementAndGet` + `IncrementIfBelow` + `Peek` compose on one key under the uniform encoding.
- `TestMaxWindowSeconds_BoundaryDoesNotOverflow` — now asserts `int` representability and is tight against whichever limit binds (overflow bound on 64-bit, int cap on 32-bit).

`make test`, `make check-fmt`, `make check-license`, and `make lint` (`go vet` + golangci-lint) all pass.

https://claude.ai/code/session_01KmvqeDyPQPimuZjKGTbB2R

---
_Generated by [Claude Code](https://claude.ai/code/session_01KmvqeDyPQPimuZjKGTbB2R)_


pkg/callcounter/redis.go
```diff
@@ -145,16 +151,27 @@ func (r *Redis) IncrementAndGet(ctx context.Context, key string, windowSec, maxE
...
 local limit = tonumber(ARGV[4])
 if limit < 1 or count >= limit then
   local retry = 0
-  local oldest = redis.call('ZRANGE', KEYS[1], 0, 0, 'WITHSCORES')
-  if limit >= 1 and oldest[2] then
-    retry = (tonumber(oldest[2]) + tonumber(ARGV[6])) - tonumber(ARGV[2])
-    if retry < 0 then retry = 0 end
+  if limit >= 1 then
+    local idx = count - limit
+    local pivot = redis.call('ZRANGE', KEYS[1], idx, idx, 'WITHSCORES')
```
eunolabs, line 170: Verified this rank fix is correct. Freeing one slot requires the in-window count to drop below `limit`, which needs `count - limit + 1` of the oldest entries to expire; the last of those to age out is the one at rank `count - limit`, so `idx = count - limit` is right. It also collapses to rank 0 when `count == limit`, so steady-state behaviour is unchanged. The #292/#293 test math (6 entries, limit lowered to 3 → rank 3 = T+30, freeing at T+90, 35s from T+55) checks out.

One subtlety worth recording: Redis selects this pivot by **score order** (`ZRANGE`), whereas `InMemory` selects `valid[cur-limit]` by **insertion order**. Those coincide only while the clock is monotonically non-decreasing — which `memory.go` itself flags (the `valid` slice can be unsorted under a backward-jumping `WithTimeFunc`). So the new cross-backend "both report the same 35s" assertion is guaranteed only under a monotonic clock (true in these tests and in production via `time.Now()`'s monotonic reading). Since `retryAfter` is purely advisory this is fine — just flagging so the equal-value invariant isn't leaned on under a non-monotonic clock.

---

pkg/callcounter/callcounter.go
```diff
@@ -96,7 +115,21 @@ const cleanupMarginFactor = 2
...
+// representable as the int that windowSec is passed as (issue #297). On a 64-bit
+// platform math.MaxInt (~9.2e18) dwarfs the overflow bound, so the cap is a no-op
+// and the overflow bound binds. On a 32-bit platform math.MaxInt (~2.1e9) is the
+// smaller of the two, so the int width binds instead — and there it is the true
+// maximum: a 32-bit windowSec can never exceed math.MaxInt anyway, and
+// 2.1e9 * cleanupMarginFactor * 1e9 ns stays under MaxInt64, so the margin still
+// cannot overflow. Without the cap, int(MaxWindowSeconds) would truncate on
+// 32-bit, making the documented [1, MaxWindowSeconds] range unreachable and the
+// overflow guard untestable.
+const MaxWindowSeconds = min(
```
eunolabs, line 129: The `min(overflowBound, math.MaxInt)` cap is correct, and `min` in a `const` is fine on this repo's Go 1.25; the result stays a typed `int64` constant, identical to before on 64-bit. Worth noting the natural consequence: on a 32-bit build `MaxWindowSeconds == math.MaxInt`, so the upper-bound check in `checkWindowSec` (`int64(windowSec) > MaxWindowSeconds`) can never fire — any representable `int` window passes. That's exactly the intent (the `int` width is itself the bound), and the boundary test handles both regimes, so no change needed; just confirming the behaviour is understood.

---

eunolabs (commented):
## Review: callcounter #292–#298

Reviewed the full diff against the head revision (`e4c5926`) plus the surrounding `pkg/callcounter` source. This is a clean, well-scoped, and unusually well-documented change. All 35 CI checks are green, and the `CONTRIBUTING.md`/`CLAUDE.md` conventions are respected (license headers, cross-backend table-style tests, threat-model prose for the new memory bound).

### Correctness — verified
- **`retryAfter` rank fix (#292/#293):** Independently checked the formula. Freeing one slot requires the count to fall below `limit`, i.e. `count - limit + 1` of the oldest entries must expire; the binding one is at rank `count - limit`. The Redis script's `idx = count - limit` and `InMemory`'s `valid[cur-limit]` agree, and both collapse to rank 0 when `count == limit`, so **steady-state behaviour is preserved**. The regression-test math (6 in-window, limit lowered to 3 → rank 3 = T+30 → 35s, not the old 5s) is right. See the inline note on the Lua script.
- **Score encoding (#295):** Agree this is a robustness/consistency change, not an active-bug fix — as the description states, the old `%f` bound and the integer bound both parse to the same `double` for sub-2^53 micros, and Redis stores all scores as doubles regardless, so cross-method reads on one key already agreed today. Formatting all three methods' bounds as integer micros removes the latent asymmetry cleanly.
- **`MaxWindowSeconds` cap (#297):** Correct; see the inline note on `callcounter.go`. The boundary test now adapts to whichever limit binds, which keeps it meaningful on both 32- and 64-bit.

### Suggestions (non-blocking)
1. **`CHANGELOG.md`.** The curated `[Unreleased]` section records `pkg/callcounter` fixes at exactly this granularity (#277, #273, #213, …), but this PR adds no entry. The `retryAfter` correction (#292/#293) is a user-visible behaviour change (the back-off hint a client receives changes value when `count > limit`), so a short **Fixed** entry would match the repo's documented "behavior changes need a docs update in the same PR" practice. #295/#297 are internal enough to fold in or omit.
2. **Cross-backend agreement is monotonic-clock-scoped.** See the inline comment on the Lua script — the two backends pick the pivot entry by different orderings (score vs. insertion), so the "identical 35s" invariant the new tests assert holds under a monotonic clock only. Not a defect (retryAfter is advisory), just worth not over-relying on.
3. **#296 left open for a maintainer** is a reasonable call given the PR adds no code for it; the verification write-up in the description is convincing.

Nice work — the inline godoc/threat-model narration makes the intent and the cross-backend parity easy to follow.

---

Issue #332, opened 2026-06-14
bug: Redis maxcalls sorted-set key carries the window duration twice


eunolabs:
## Summary

`handleMaxCalls` in `pkg/enforcement/handlers.go` builds the logical counter key with `windowSeconds` embedded:

```go
key := fmt.Sprintf("maxcalls:%s:%s:%s:%d", req.SessionID, targetType, toolName, mc.WindowSeconds)
```

The Redis backend in `pkg/callcounter/redis.go` then re-appends `windowSec` to form the physical sorted-set key:

```go
windowKey := fmt.Sprintf("callcounter:%s:%d", key, windowSec)
```

The resulting Redis key is:
```
callcounter:maxcalls:<session>:<type>:<tool>:<window>:<window>
```

The window duration appears **twice**. The in-memory backend uses the logical key verbatim:
```
callcounter:maxcalls:<session>:<type>:<tool>:<window>
```

## Impact

1. **Operational confusion** — anyone inspecting Redis with `SCAN callcounter:*` or debugging counters with `KEYS` sees an odd double-window suffix that matches no documented key format.
2. **Backend key divergence** — the in-memory and Redis backends produce structurally different keys for the same logical counter. Any future tooling that parses these keys will need to handle both formats.
3. **No functional correctness issue** — the doubling is consistent within the Redis backend, so counts are not shared with any other key namespace.

## Fix

Choose one canonical approach:

**Option A**: Remove `mc.WindowSeconds` from the logical key in `handleMaxCalls` and let each backend append it exactly once:

```go
key := fmt.Sprintf("maxcalls:%s:%s:%s", req.SessionID, targetType, toolName)
```

**Option B**: Remove the re-appending of `windowSec` in the Redis backend and use the logical key as the sorted-set key as-is.

Either option will change the physical Redis key format; existing counters (which reset within their window anyway) don't need migration.

Issue #331, opened 2026-06-14
docs/bug: argumentSchema `pattern` keyword uses unanchored (partial) regex matching — operators expect full-string matching


eunolabs:
## Summary

The `pattern` keyword in `argumentSchema` is validated using `re.MatchString(v)`:

```go
// pkg/enforcement/schema.go
func schemaValidateString(p, v string, s *capability.ArgumentSchema) error {
    if s.Pattern != "" {
        re, err := compilePattern(s.Pattern)
        // ...
        if !re.MatchString(v) {
            return fmt.Errorf("%s: value does not match pattern %q", p, s.Pattern)
        }
    }
```

`regexp.Regexp.MatchString` reports whether the pattern matches **anywhere** in the string — it does NOT require the pattern to match the entire value. This follows JSON Schema spec semantics (ECMA 262 partial matching), but it produces results that commonly surprise operators:

| Pattern | Value | Result | Operator expectation |
|---------|-------|--------|----------------------|
| `[0-9]+` | `abc123` | **PASS** | fail (contains non-digits) |
| `[A-Z][a-z]+` | `helloWorld` | **PASS** | fail (starts with lowercase) |
| `SELECT\|INSERT` | `DROP TABLE` | **PASS** (`TABLE` contains no match, but `SELECT` alone matches `SELECT`) | fail |
| `sql_[a-z]+` | `prefix_sql_table` | **PASS** | fail |

The third example is especially relevant to `allowedOperations`-style guards written as `argumentSchema` patterns.

## Impact

An operator who writes:

```yaml
argumentSchema:
  properties:
    operation:
      type: string
      pattern: "SELECT|INSERT|UPDATE"
```

to restrict SQL operations will be surprised that `"DROP TABLE users; SELECT 1"` passes — the pattern finds `SELECT` later in the string.

Similarly, `pattern: "[0-9]+"` intended as "only digits" allows any string containing a digit.

## Suggested fix

Two options:

1. **Document the behaviour prominently** in `docs/capability-manifest-guide.md`: add a warning that `pattern` uses partial matching (like JSON Schema) and operators must use `^` and `$` anchors for full-string matching. Provide examples showing the difference.

2. **Implicitly anchor patterns**: wrap user patterns as `^(?:<pattern>)$` before compiling, making full-string matching the default. This diverges from JSON Schema but is safer for a security policy context. A flag or keyword could opt out.

[...]

Issue #330, opened 2026-06-14
perf: resourceSpecificity calls strings.ContainsRune in a hot-path inner loop


eunolabs:
## Summary

`resourceSpecificity` in `pkg/enforcement/engine.go` is called on every enforcement request for every candidate constraint. It contains two loops that each call `strings.ContainsRune("*?[", r)` — scanning a 3-character string for each rune of the resource pattern:

```go
for _, r := range resource {
    if strings.ContainsRune("*?[", r) {  // scans 3-char string every iteration
        break
    }
    prefixLen++
}
wildcardCount := 0
for _, r := range resource {
    if strings.ContainsRune("*?[", r) {  // same, second pass
        wildcardCount++
    }
}
```

`strings.ContainsRune` for a 3-character string is implemented as a linear scan of the UTF-8 bytes. For a resource pattern with N characters and a capability list with M entries, this is called O(N × M) times per request.

## Fix

Replace with a direct rune comparison, which is a single branch at the CPU level:

```go
isWild := func(r rune) bool { return r == '*' || r == '?' || r == '[' }

for _, r := range resource {
    if isWild(r) {
        break
    }
    prefixLen++
}
wildcardCount := 0
for _, r := range resource {
    if isWild(r) {
        wildcardCount++
    }
}
```

Or combine into a single pass:

```go
prefixDone := false
for _, r := range resource {
    if r == '*' || r == '?' || r == '[' {
        if !prefixDone {
            prefixDone = true
        }
        wildcardCount++
    } else if !prefixDone {
        prefixLen++
    }
}
```

This eliminates the inner string scan and reduces to a single pass with simple comparisons.

Issue #329, opened 2026-06-14
bug: alwaysAllowPDP (wiretap mode) emits audit records without RequestID or DecidedAt


eunolabs:
## Summary

`alwaysAllowPDP.Decide` in `cmd/eunox-mcp/pdp.go` returns a zero-value `EnforceResponse`:

```go
func (alwaysAllowPDP) Decide(...) capability.EnforceResponse {
    return capability.EnforceResponse{Decision: capability.DecisionAllow}
}
```

The returned `RequestID` and `DecidedAt` fields are empty strings. The audit record written from this response will have no `request_id` and no `decided_at`.

## Contrast with ManifestPDP

`Engine.EvaluateConditions` (called by `ManifestPDP`) generates a UUID `RequestID` and timestamps the response before returning. Every audit record from the production PDP carries a `request_id` that can be used for correlation with request tracing.

In wiretap / audit-only mode (`alwaysAllowPDP`), the `request_id` field is absent entirely, making it impossible to correlate audit entries with specific requests or sessions in downstream processing.

## Fix

Generate a `RequestID` and `DecidedAt` in `alwaysAllowPDP.Decide`:

```go
func (alwaysAllowPDP) Decide(_ context.Context, req *capability.EnforceRequest, ...) capability.EnforceResponse {
    return capability.EnforceResponse{
        Decision:  capability.DecisionAllow,
        RequestID: uuid.New().String(),
        DecidedAt: time.Now().UTC().Format(time.RFC3339Nano),
    }
}
```

The same applies to `DecideResourceRead` and `DecidePromptGet` on `alwaysAllowPDP`.

## Why it matters

Wiretap mode is explicitly described as an audit/observation mode. Having structurally incomplete audit records in precisely the mode designed for auditing is a contradiction. It also means any audit-verification tooling that validates the presence of `request_id` will flag every wiretap record as malformed.

PR #335, opened 2026-06-14
fix: resolve enforcement, redaction, and audit issues (#317–#332)

eunolabs:
Fixes a batch of bug, security, performance, and docs issues across the policy
engine, list filtering, response redaction, and wiretap auditing. Each fix has
regression tests; the capability-manifest guide and threat model are updated where
behavior or security posture changed.

## Enforcement engine (`pkg/enforcement`)

- **#317** — `actionMatchScore` now recognizes each namespace's canonical verb
  (`read`/`get`/`allow`), not only `call`/`*`, via
  `capability.ValidateActionForTargetType` (single source of truth shared with the
  loader).
- **#320** (security) — `findMatchingCapability` requires the constraint's
  namespace type to match the request's. A wildcard `resource:*` constraint can no
  longer approve a tool call; a bare/untyped request defaults to `tool` and cannot
  cross into another namespace.
- **#321** — the engine threads its decision timestamp through `ctx`, so
  `input.context.timestamp` carries the request time instead of the always-empty
  `req.Context.Now`.
- **#322** — `ValidateAction`/`EvaluateConditions` no longer write
  `EnforceRequest.Directives`; directives reach `BuildRegoInput` via `ctx`. Removes
  the data race on a shared `*EnforceRequest` (new `-race` regression test).
- **#324** — `allowedExtensions`/`recipientDomain` return `CONDITION_FAILED` (not
  `MISSING_CONTEXT`) for a present-but-wrong-type argument.
- **#328 / #332** — `windowSeconds` is dropped from the **logical** `maxCalls` key
  so the **physical** counter key carries the window once (the backend appends it),
  not twice (`…:300:300`). Per-window isolation (#278) is unchanged.
  - **Operational/migration note:** this reshapes the physical Redis key
    (`…:300:300` → `…:300`), so existing `maxCalls` counters are orphaned on
    upgrade (in-flight windows reset), and during a rolling deploy old- and
    new-code instances address different keys, so a session can transiently get up
    to ~2× its limit until rollout completes. Acceptable pre-1.0; documented in the
    threat-model changelog (1.14).
- **#330** (perf) — single-pass `resourceSpecificity`, removing
  `strings.ContainsRune` from the hot path.

## Proxy / PDP (`cmd/eunox-mcp`)

- **#318** (security) — manifest and JWT `*/list` filters fail closed (empty list)
  on a re-marshal error instead of returning the unfiltered upstream bytes.
- **#319** — `redactFields` forwards a genuine free-form text content item (an
  error string, a `[ERROR] …` log line) unchanged instead of failing the whole
  response. **Review follow-up:** to keep the fail-closed invariant, a text block
  that *looks like* a JSON object (leading `{`) but fails to parse — a malformed or
  truncated object that may still carry the redaction target — fails the response
  closed; only non-object-shaped text is forwarded. Threat model §6.3 updated.
- **#325** — `isSQLVerb` is made genuinely case-insensitive by dropping the
  redundant `s == strings.ToUpper(s)` guard. Note this is **hardening, not a live
  bypass fix**: the only current caller (`evaluateJWTConditions`) already passes an
  upper-cased token, so the guard was a no-op there; this removes a footgun for any
  future caller that forwards a raw, mixed-case token.
- **#326** (security) — array-index `redactFields` paths (`users[0].ssn`) are
  rejected at manifest load; the runtime redactor would otherwise treat them as a
  literal key and silently no-op.
- **#329** — `alwaysAllowPDP` (wiretap mode) emits a `RequestID` and `DecidedAt`,
  so audit-only records carry the same correlation fields a `ManifestPDP` decision
  does.

## Docs

- **#323** (security) — the `allowedExtensions` guide note spells out the
  append-an-allowed-extension bypass (`malware.exe.txt`, `script.sh.csv`).
- **#331** (security) — documents that `argumentSchema` `pattern` matches a
  **substring** (JSON-Schema/ECMA-262); anchor with `^…$` for full-string matching.

## Already resolved

- **#327** — `validateMaxCalls` already rejects `count`/`windowSeconds < 1` at load
  and is exercised by existing tests; no code change needed.

## Verification

`go test -race ./...`, `go vet`, `golangci-lint`, `gofmt`, license headers, and the
`pkg/` coverage gate (97.9%) all pass.

https://claude.ai/code/session_01PA2uPqJqq1uwc5cJqZ9WeA


cmd/eunox-mcp/pdp.go
```diff
@@ -862,13 +899,20 @@ func applyRedactObligs(resultBytes []byte, obligs []capability.Obligation) ([]by
...
 	var val interface{}
 	if err := json.Unmarshal([]byte(text), &val); err != nil {
-		return "", fmt.Errorf("redactFields: text content is not parseable JSON; cannot verify redaction of %v (fail closed)", paths)
+		// Not JSON: a text content item that is intentionally plain text carries no
+		// JSON-structured field for a dot-path to address. Forward it unchanged
+		// rather than failing the whole response — the prior fail-closed posture made
+		// redactFields incompatible with any tool that ever returns a plain-text
+		// error message (e.g. "Error: file not found"), blocking the error instead of
+		// passing it through (issue #319).
+		return text, nil
```
eunolabs: **#319 — this is the one change I'd push back on.** It flips `redactFields` from fail-closed to fail-open for an entire class of inputs: anything `json.Unmarshal` rejects is now forwarded verbatim.

For genuine prose (`Error: file not found`) that's the right call and a real usability win. But a text item that is a **malformed or truncated JSON object still carrying the redaction target** also fails `Unmarshal` and is now forwarded with the secret intact, where it previously failed closed:

- `{"ssn":"123-45-6789",}` — trailing comma, invalid JSON in Go
- `{"ssn":"123-45-6789"` — truncated object, unexpected EOF

A secret addressed by a dot-path lives in a JSON object, which serializes starting with `{`. Consider gating the forward-unchanged path on the content not *looking* like a JSON object — e.g. fail closed when `strings.TrimSpace(text)` begins with `{`, forward otherwise (prose, log lines, `[ERROR] …`). That keeps the usability fix while preserving fail-closed for the realistic bypass. (I'd avoid keying on a leading `[` too, since `[ERROR] …`-style log lines are common.)

This also crosses the repo's load-bearing **fail-closed** invariant; per CONTRIBUTING a security-posture change like this should get a note in `docs/threat-model-mcp.md`, which this PR doesn't touch.

---

pkg/enforcement/handlers.go
```diff
@@ -371,12 +387,10 @@ func (e *Engine) handleMaxCalls(ctx context.Context, cond capability.Condition,
...
-	// The window length is part of the key so two maxCalls conditions on the same
-	// (session, target, tool) but different windows occupy isolated buckets rather
-	// than sharing one counter (#278). It is length-prefixed like every other
-	// component for the collision-resistance reason above (#273), though as a
-	// decimal integer it can carry no separator of its own.
-	key := compositeCounterKey("maxcalls", req.SessionID, targetType, toolName, strconv.Itoa(mc.WindowSeconds))
+	// The window is passed to the backend as a separate argument (below), not folded
+	// into this logical key — the backend appends it to the physical key exactly
+	// once, isolating distinct windows without doubling the segment (#328, #332).
+	key := compositeCounterKey("maxcalls", req.SessionID, targetType, toolName)
```
eunolabs, line 393: **#328/#332 — correct fix.** I confirmed both backends append `windowSec` to the physical key exactly once (Redis `callcounter:<key>:<windowSec>`, InMemory `storageKey`), so dropping it from the logical key removes the doubled `…:300:300` without re-opening the per-window isolation issue (#278).

One operational note worth a line in the PR/migration text: this changes the **physical** counter-key shape (`…:300:300` → `…:300`). On a Redis-backed multi-instance deployment, existing maxCalls counters are orphaned on upgrade (in-flight rate-limit windows effectively reset), and during a rolling deploy old- and new-code instances address different keys, so a session can transiently get up to ~2× its limit until rollout completes. Acceptable pre-1.0, but worth documenting.

---

cmd/eunox-mcp/pdp_jwt.go
```diff
@@ -1012,8 +1027,15 @@ var sqlVerbs = map[string]bool{
...
+// isSQLVerb reports whether s is a recognized SQL statement verb, case-
+// insensitively. The function name promises case-insensitive detection, so it
+// must not depend on the caller pre-uppercasing: the prior `&& s ==
+// strings.ToUpper(s)` guard returned false for any lowercase verb (e.g.
+// isSQLVerb("select") == false), so a future call site passing a raw,
+// mixed-case argument token would silently fail to detect smuggled lowercase SQL
+// (issue #325). ToUpper alone is the correct, footgun-free test.
 func isSQLVerb(s string) bool {
-	return sqlVerbs[strings.ToUpper(s)] && s == strings.ToUpper(s)
+	return sqlVerbs[strings.ToUpper(s)]
```
eunolabs, line 1038: **#325 — fine to take, but note it's a no-op at the only current caller.** `evaluateJWTConditions` passes `word := strings.ToUpper(parts[0])` into `isSQLVerb`, so the old `&& s == strings.ToUpper(s)` guard was already always true there — detection was effectively case-insensitive in practice. This change is pure hardening for a hypothetical future caller that forwards a raw, mixed-case token; it doesn't close a live bypass (the PR description reads slightly like it does). Worth keeping the dead guard removed regardless.

---

eunolabs (commented):
Reviewed the full diff (12 files, +833/−125) against the base, plus the surrounding code each fix depends on. Overall a high-quality, well-tested batch — every fix has a targeted regression test, the engine `findMatchingCapability` rework is sound, and CI is green across all 42 checks. I verified the load-bearing assumptions behind the riskier changes:

- **#320 namespace-type match** — all three `ManifestPDP` entry points (`Decide`/`DecideResourceRead`/`DecidePromptGet`) set `Target.Type`, and the engine defaults a bare/untyped request to `tool`, so resource/prompt enforcement doesn't regress while `resource:*` can no longer approve a tool call. Unprefixed manifest targets are already rejected at load, so this only tightens the programmatic path. ✔
- **#328/#332 maxCalls key** — confirmed both counter backends append `windowSec` to the physical key exactly once, so dropping it from the logical key removes the doubled segment without re-opening #278. ✔ (one operational note inline)
- **#317 canonical verbs / #330 single-pass specificity** — verified the rewrites preserve the existing scoring formula, and routing through `ValidateActionForTargetType` as the single source of truth shared with the loader is the right call. ✔
- **#321 / #322 ctx threading** — moving the timestamp and directives onto `ctx` instead of mutating the caller's `*EnforceRequest` is a clean fix for the data race; the `-race` regression test is a good addition. ✔

**One thing I'd push back on / want confirmed:**

- **#319 (redactFields now forwards non-JSON text unchanged)** — see the inline comment on `redactJSONText`. The usability fix is right, but forwarding *everything* `json.Unmarshal` rejects also forwards malformed/truncated JSON objects that still carry the redaction target, which previously failed closed. That's a fail-open for a class of inputs and crosses the repo's "fail closed" invariant without a `docs/threat-model-mcp.md` update. Suggest gating the forward path on the text not looking like a JSON object (leading `{`).

**Process nits (non-blocking):**

- Per CONTRIBUTING, the manifest-semantics clarifications here (#326 no array-index redact paths, #331 `pattern` matches a substring, #323 extension smuggling) ideally also land in the spec repo (`eunolabs/mcp-capability-manifest`); this PR only updates the in-repo guide.
- #329 reuses the existing `request_id`/`decided_at` audit fields (not new ones), so no new sign-and-verify round-trip is required — but a quick assertion that a wiretap *audit record* (not just the `EnforceResponse`) carries them would close the loop end-to-end.

Nothing here is blocking except the #319 question — nice cleanup overall.

---

cmd/eunox-mcp/pdp.go
eunolabs: Good catch — fixed in bb11743. `redactJSONText` now gates the forward-unchanged path on shape: a text block whose trimmed body begins with `{` but fails to parse (malformed/truncated object that may still carry the target) fails the response closed; genuine free-form text (error strings, `[ERROR] …` log lines, bare scalars) is forwarded. I left `[` as prose per your note. Added regression tests for both the malformed-object fail-closed and bracketed-log-line forward cases, and documented the posture change in `docs/threat-model-mcp.md` §6.3 + changelog 1.14.

---

pkg/enforcement/handlers.go
eunolabs: Agreed — added the operational/migration note to the PR description and to the threat-model changelog (1.14): the physical Redis key reshapes `…:300:300` → `…:300`, so existing counters are orphaned on upgrade and a rolling deploy can transiently admit up to ~2× the limit until rollout completes. Acceptable pre-1.0.

---

cmd/eunox-mcp/pdp_jwt.go
eunolabs: Correct — it's a no-op at the only current caller (`evaluateJWTConditions` pre-uppercases). I've reworded the PR description to call it out as hardening for a hypothetical future caller, not a live-bypass fix. Keeping the dead guard removed.

---

eunolabs (commented):
Re-reviewed the follow-up commit `bb11743` (delta since my last pass). It resolves the one concern I raised and cleans up the process nits:

- **#319 — fully addressed.** `redactJSONText` now fails closed when `strings.TrimSpace(text)` starts with `{` but doesn't parse (a malformed/truncated object that may still carry the redaction target), and forwards only genuine non-object text (prose, `[ERROR] …` log lines, bare scalars). That's exactly the shape-gated fail-closed I asked for. The new tests cover the right cases — trailing comma, truncated object, leading-whitespace object (fail closed) and the bracketed log line (forwarded). ✔
- **Threat model — updated.** §6.3 now documents the text-content posture, and changelog entry 1.14 records both the #319 redaction change and the #328/#332 physical-key migration note (orphaned counters on upgrade, transient ~2× during rolling deploy). Addresses the CONTRIBUTING "fail-closed change needs a threat-model note" point. ✔
- **#325 — reframed** in the description as hardening rather than a live-bypass fix, matching the fact that the only current caller passes an upper-cased token. ✔

One residual is explicitly documented and I agree with the call: a *malformed* top-level JSON **array** (leading `[`) that still carries secrets is forwarded unchanged, traded deliberately for `[ERROR] …` log-line ergonomics — with guidance to emit a well-formed object / `structuredContent` when that shape must be redacted. Reasonable, and path-based redaction can't address substrings of free-form text anyway.

CI: required checks are green on `bb11743` (Test, Lint, all Build targets, govulncheck, Trivy, enforcement-compliance, the OPA-comparison and demo suites); the lone non-green is a CodeQL "Analyze (go)" re-run still finishing — non-blocking.

No remaining blocking concerns — nice, responsive iteration.

---

Issue #323, opened 2026-06-14
security: allowedExtensions suffix matching allows bypass by appending an allowed extension


eunolabs:
## Summary

`handleAllowedExtensions` in `pkg/enforcement/handlers.go` matches file extensions using `strings.HasSuffix`:

```go
for _, suffix := range allowed {
    if strings.HasSuffix(base, suffix) {
        matched = true
        break
    }
}
```

This is intentional for compound extensions (`.tar.gz`), but it creates an exploitable bypass: **any disallowed file type passes the check when an allowed extension is appended to its name**.

## Example

Manifest:
```yaml
conditions:
  - type: allowedExtensions
    argument: path
    extensions: [".txt", ".csv"]
```

These paths would **pass** despite carrying disallowed primary types:
- `malware.exe.txt` (executable with `.txt` appended)
- `script.sh.csv`
- `payload.dll.txt`

These pass because `strings.HasSuffix("malware.exe.txt", ".txt") == true`.

## Impact

An LLM agent or malicious tool argument can bypass file-type restrictions by renaming files to append an allowed extension. Operators who write `extensions: [".txt"]` reasonably expect only files whose primary/sole extension is `.txt` to be permitted — not arbitrary files renamed to end with `.txt`.

The `filepath.Ext` check only validates that the file has *some* extension (not empty), but it does not gate on whether the *last* extension matches. The actual allowlist check then uses the full base name with `HasSuffix`, allowing any suffix that ends with an allowed extension.

## Suggested behaviour

Options:
1. **Change the default to single-extension matching** (`filepath.Ext(base)` must be in the allowlist). Add an explicit opt-in for compound extension matching (e.g. `compound: true`).
2. **Keep the current behaviour but document it prominently** in the manifest guide with a security warning: "allowing `.txt` also permits files named `anything.exe.txt`; use exact-match or allowedValues if you need strict type enforcement."
3. **Add a load-time warning** when an allowedExtensions list contains single-segment extensions (no dot in the pattern itself) while multiple-segment filenames are a known bypass vector.

At minimum, option 2 is needed. The current docs do not mention this behaviour.

Issue #322, opened 2026-06-14
bug: ValidateAction and EvaluateConditions mutate caller's EnforceRequest.Directives — data race under concurrent use


eunolabs:
## Summary

Both `Engine.ValidateAction` and `Engine.EvaluateConditions` in `pkg/enforcement/engine.go` write to the caller-supplied `*capability.EnforceRequest`:

```go
// engine.go:345
req.Directives = matched.Directives

// engine.go:557
req.Directives = matched.Directives
```

Mutating the caller's struct is a side-effect that violates normal Go API conventions. More critically, if the same `*EnforceRequest` is accessed concurrently (e.g. the caller reads it for logging while the engine processes it, or two goroutines race on the same pointer), the Go race detector will flag this as a data race on `req.Directives`.

## Why Directives Are Attached

The mutation exists so that a downstream `PolicyEvaluator` (Rego/OPA/Cedar) can see `req.Directives` in `BuildRegoInput`. But this goal can be achieved without mutating the original struct.

## Consequences

1. **Data race** — `go test -race` can detect it if any test creates a concurrent access pattern on the same `*EnforceRequest`.
2. **Surprising caller behavior** — A caller that supplies a zero-value `EnforceRequest`, calls `ValidateAction`, then logs the original struct will see `Directives` populated from the matched constraint, which it did not set.
3. **API contract violation** — The Go convention for methods accepting a pointer is that they may update the pointed-to value only if that is explicitly documented. The `EnforceRequest` godoc does not document that `Directives` is an output field.

## Fix

Pass directives through a local variable rather than storing them on the request. For example, thread them through `EvaluateConditions` as a parameter or return value, or via a separate `enforceContext` struct internal to the engine:

```go
// Instead of:
req.Directives = matched.Directives

// Pass to condition handlers via a separate context struct:
ctx := &enforceContext{directives: matched.Directives, request: req}
```

Alternatively, document `EnforceRequest.Directives` explicitly as "populated by the engine during evaluation — must not be set by callers; must not be read concurrently with an ongoing Decide/ValidateAction call."

Issue #317, opened 2026-06-14
bug: Enforcer.ValidateAction silently denies resource/prompt/system constraints with type-specific action verbs


eunolabs:
## Summary

`engine.ValidateAction` (which backs the public `capability.Enforcer` interface) uses `actionMatchScore` to match constraints. That function only recognises `"call"` and `"*"` as valid action verbs:

```go
// pkg/enforcement/engine.go
func actionMatchScore(actions []string, _ *capability.EnforceRequest) bool {
    for _, a := range actions {
        if a == "call" || a == "*" {
            return true
        }
    }
    return false
}
```

The manifest grammar and `capability.validActionsFor` define per-namespace verbs (`"read"` for `resource:`, `"get"` for `prompt:`, `"allow"` for `system:`), and `capability.ValidateActionForTargetType` enforces those at load time. But `actionMatchScore` ignores all of them.

## Impact

Any caller using `engine.ValidateAction` or `engine.FindMatchingCapability` (the `capability.Enforcer` interface) and passing resource/prompt/system constraints with their canonical verbs will always receive a "no matching capability" denial — even for perfectly valid manifests.

`ManifestPDP` works around this silently by **never calling `ValidateAction`**; it calls `engine.EvaluateConditions` after its own type-scoped `findConstraint`. But this makes the `Enforcer` interface misleading for embedded library users: importing `pkg/enforcement` and calling `ValidateAction` with `resource:` constraints produces wrong results without any warning.

## Reproduction

```go
engine := enforcement.New()
req := &capability.EnforceRequest{ToolName: "file:///data/report.csv"}
constraints := []capability.Constraint{{
    Target:  "resource:file:///data/*",
    Actions: []string{"read"},
}}
resp, _ := engine.ValidateAction(context.Background(), req, constraints)
// resp.Decision == "deny" — always, despite a valid manifest entry
```

## Expected behaviour

`ValidateAction` / `FindMatchingCapability` should recognise the canonical action verb for each namespace (`read`, `get`, `allow`) in addition to `"*"`.

## Fix options

1. Extend `actionMatchScore` to accept `"read"`, `"get"`, and `"allow"` in addition to `"call"` and `"*"` (but the function currently ignores `req`, so it has no target-type context to enforce correct pairing).
2. Pass the target type through to `actionMatchScore` and check against `validActionsFor`.
[...]

PR #336, opened 2026-06-14
fix: resolve enforcement, JWKS, and obligation issues (#306–#316)

### LINKED ISSUE #306: bug: Engine.ValidateAction silently fails to match resource:, prompt:, system: constraints — actionMatchScore only accepts "call" and "*"
## Summary

`actionMatchScore` in `pkg/enforcement/engine.go` only returns `true` for `"call"` and `"*"`:

```go
// engine.go:676
func actionMatchScore(actions []string, _ *capability.EnforceRequest) bool {
    for _, a := range actions {
        if a == "call" || a == "*" {
            return true
        }
    }
    return false
}
```

The manifest accepts — and `validateLocalManifest` actively validates — `"read"` for `resource:` targets, `"get"` for `prompt:` targets, and `"allow"` for `system:` targets. But `actionMatchScore` is used by `findMatchingCapability`, which is called by `ValidateAction`. Any constraint whose `actions` list contains only `"read"`, `"get"`, or `"allow"` is silently skipped by `findMatchingCapability`, causing `ValidateAction` to return "no matching capability" → deny for every request, even a legitimately covered one.

## Reproduction

```go
caps := []capability.Constraint{
    {
        Target:  "resource:file:///data/*",
        Actions: []string{"read"},
    },
}
req := &capability.EnforceRequest{ToolName: "file:///data/report.pdf"}
resp, _ := engine.ValidateAction(ctx, req, caps)
// resp.Decision == "deny"  ("no matching capability for requested action")
// Expected: "allow"
```

## Impact

`ManifestPDP` works around this by not routing non-tool operations through `Engine.ValidateAction`. It uses its own `findConstraint` (which scopes by namespace type) plus `containsAction` (which correctly checks `"read"`, `"get"`, `"allow"`), then calls `EvaluateConditions` directly. So the shipped proxy is not affected.

However:

1. **`capability.Enforcer.ValidateAction`** is the exported interface and is documented as a general-purpose enforcement function. A library caller who passes `resource:` or `prompt:` constraints through `ValidateAction` gets systematic silent denies with no error.
2. **`FindMatchingCapability`** (the exported wrapper around `findMatchingCapability`) has the same blind spot — it returns `nil` for any resource/prompt/system constraint, making the validate endpoint incorrect for non-tool capabilities.
3. The discrepancy between what `validateLocalManifest` accepts as a valid action and what `actionMatchScore` can match is confusing: a manifest is accepted as valid at load, then silently produces no-match denials at runtime when called through the engine directly.

## Fix direction

Either:
- Extend `actionMatchScore` to accept `"read"`, `"get"`, and `"allow"` in addition to `"call"` and `"*"` (the engine still ignores the request parameter, which is fine since action matching today is purely list-membership), or
- Have `actionMatchScore` consult `req.Target.Type` to determine which action keyword is required and check only that — which is exactly what `ManifestPDP.requiredActionFor` + `containsAction` does, and consolidating into the engine would remove the divergence.

The engine's comment that `"read"`, `"write"`, `"delete"` etc. are "former semantic categories" that are "not recognized" is now inaccurate: `"read"`, `"get"`, and `"allow"` are the current valid actions for non-tool namespaces per `constraint.go:validActionsFor`.

### LINKED ISSUE #307: bug: argumentSchema `pattern` uses unanchored regex — substring match may silently permit values that should be rejected
## Summary

`schemaValidateString` in `pkg/enforcement/schema.go` uses `re.MatchString(v)`, which returns `true` if the pattern matches **anywhere** in the string:

```go
// schema.go:202
if !re.MatchString(v) {
    return fmt.Errorf("%s: value does not match pattern %q", p, s.Pattern)
}
```

`regexp.MatchString` is unanchored: it succeeds whenever the regex matches a substring of the input. An operator who writes `pattern: "[0-9]{4}"` intending "the value must be exactly a 4-digit PIN" actually gets "the value must contain at least 4 consecutive digits" — so `"abc1234xyz"`, `"9999_extra"`, and `"prefix0000"` all pass.

## Concrete examples

| Pattern (operator intent) | Value | Expected | Actual |
|---|---|---|---|
| `[0-9]{4}` | `"abc1234"` | deny (not purely 4 digits) | **allow** |
| `[A-Z]{3}` | `"prefix_ABC_suffix"` | deny | **allow** |
| `https://example\.com/.*` | `"https://evil.com/foo https://example.com/bar"` | deny | **allow** |
| `^[a-z]+$` | `"UPPER"` | deny | deny (anchors work when written) |

## Why it matters

Operators coming from most validation libraries (OpenAPI validators, JSON Schema validators in other languages, form-validation frameworks) expect `pattern` to match the **whole string** unless they explicitly use `.*` around the pattern. Unanchored matching here means:

- A content-injection allowlist like `pattern: "safe_[a-z]+"` permits any argument containing the substring `safe_<letters>`, including `"SAFE_prefix safe_inject UNSAFE_suffix"`.
- A SQL-identifier guard like `pattern: "[a-zA-Z_][a-zA-Z0-9_]*"` permits `"1; DROP TABLE users"` because the substring `TABLE` matches the character-class pattern.

## JSON Schema spec note

JSON Schema (draft-07 §6.3.3) does specify `pattern` as unanchored — it matches if the regex is found anywhere in the string. So the current behavior is *spec-compliant*. However:

1. The capability-manifest guide (`docs/capability-manifest-guide.md`) does not document this, so operators are likely to be surprised.
2. `argumentSchema` is already a restricted subset of JSON Schema; the guide could document the unanchored behavior with a worked example and recommend `^...$` anchors whenever full-string matching is intended.
3. Alternatively, the engine could silently anchor patterns (prepend `^`, append `$`) for a more intuitive API — a breaking change, but acceptable pre-1.0.

## Recommendation

At minimum, add a note to the guide and to the `ArgumentSchema.Pattern` field doc:

> `pattern` matches if the regex is found **anywhere** in the string (JSON Schema behavior). Use `^` and `$` anchors to require a full-string match: `pattern: "^[a-z]+$"`.

Consider also adding a load-time warning (or error) when a `pattern` value lacks both `^` at the start and `$` at the end, to catch the common mistake proactively.

### LINKED ISSUE #308: bug: Engine.ValidateAction and EvaluateConditions never set EnforceResponse.AuditOnly — audit-mode constraints silently produce hard denials for direct engine callers
## Summary

`EnforceResponse.AuditOnly` is never set by `Engine.ValidateAction` or `Engine.EvaluateConditions`. Both methods evaluate conditions and return `allow` or `deny` without inspecting the matched constraint's `Enforcement` field. A constraint with `enforcement: "audit"` that fails a condition produces a plain `deny` response, indistinguishable from a hard-deny on a non-audit constraint.

## How it works today (and where the gap is)

`ManifestPDP.Decide`, `DecideResourceRead`, and `DecidePromptGet` each add audit-mode stamping via a `defer`:

```go
// pdp.go:198-200
if matched.IsAuditOnly() {
}
```

This works for the proxy's own enforcement paths. But:

1. **Direct `Engine.ValidateAction` callers** — the method is the sole implementation of the exported `capability.Enforcer` interface and is the contract other packages depend on. A caller who passes a constraint with `enforcement: "audit"` receives a deny with `AuditOnly: false`. Nothing in `ValidateAction`'s godoc warns that the caller must additionally check `matched.Enforcement` and set `AuditOnly` themselves.

2. **`Engine.EvaluateConditions`** — same issue. Its godoc says "after all conditions pass, obligations are collected from the constraint's Directives" but does not mention that `Enforcement` is ignored, leaving `AuditOnly` always `false`.

3. **`capability.Enforcer` interface** — the interface specifies `ValidateAction` but nowhere documents that `EnforceResponse.AuditOnly` will never be set. A new PDP implementation or wrapper that relies on the engine result to decide whether to forward a denied request will miss audit-mode constraints.

## Reproduction

```go
matched := &capability.Constraint{
    Target:      "tool:debug_exec",
    Actions:     []string{"call"},
    Enforcement: capability.EnforcementAudit,
    Conditions: []capability.Condition{
        &capability.TimeWindowCondition{NotBefore: "2099-01-01T00:00:00Z"},
    },
}
req := &capability.EnforceRequest{ToolName: "debug_exec"}
resp, _ := engine.EvaluateConditions(ctx, req, matched)
// resp.Decision  == "deny"
// resp.AuditOnly == false   ← should be true for an audit constraint
// A caller treating this as a hard deny would block the request instead of logging and forwarding.
```

## Fix

`ValidateAction` and `EvaluateConditions` should set `resp.AuditOnly = matched.IsAuditOnly()` (or `matched.Enforcement == capability.EnforcementAudit`) before returning any deny response whose source is the matched constraint. (Kill-switch, no-match, and JWT-intersection denials produced by ManifestPDP — which do not go through these engine methods — correctly stay as hard denials.)

The simplest change is to add `AuditOnly: matched.IsAuditOnly()` to every deny `EnforceResponse` literal returned by `ValidateAction` and `EvaluateConditions` when `matched != nil`. That removes the need for every caller to reimplement the `defer` stamping that ManifestPDP currently carries.

### LINKED ISSUE #309: bug: ValidateAction and EvaluateConditions mutate the caller's EnforceRequest.Directives without documenting the side effect
## Summary

Both `Engine.ValidateAction` and `Engine.EvaluateConditions` write to the caller-supplied `*EnforceRequest` before returning:

```go
// engine.go:345 (ValidateAction), engine.go:557 (EvaluateConditions)
req.Directives = matched.Directives
```

This mutation is intentional (it makes `input.directives` available to Rego policy conditions via `BuildRegoInput`), but it is undocumented on either function and violates the common Go expectation that a function accepting a pointer does not modify its caller-visible state unless the mutation is the function's stated purpose.

## Concrete consequences

**1. Caller reuse:** A caller that reuses the same `EnforceRequest` across multiple `ValidateAction` / `EvaluateConditions` calls will find `req.Directives` polluted from the previous call. If the second call matches a constraint with no directives, `matched.Directives` is nil and `req.Directives` is set to nil — overwriting whatever the caller had. Conversely, if the first constraint had directives and the second is a deny path that returns before the assignment (e.g. unknown condition type), `req.Directives` still holds the first constraint's directives.

**2. Audit / logging:** A caller that logs `req` after `ValidateAction` returns will see `Directives` set to the matched constraint's directives even on a deny — directives are post-allow obligations and should not appear on a denied request. The mutation happens unconditionally at the top of both functions, before condition evaluation:

```go
req.Directives = matched.Directives  // set even if every condition subsequently fails
```

**3. Library usability:** The `capability.Enforcer` interface documents `ValidateAction` as a read-evaluate-return operation; nothing signals that calling it modifies `req`. A library user who passes the same `req` to multiple backends in a chain or fan-out would get cross-contaminated `Directives`.

## Reproduction

```go
req := &capability.EnforceRequest{ToolName: "read_file"}

// First call — matched constraint has directives
resp1, _ := engine.ValidateAction(ctx, req, capsWithDirectives)
// req.Directives is now populated from the matched constraint

// Second call — matched constraint has no directives; condition fails
resp2, _ := engine.ValidateAction(ctx, req, capsNoDirectives)
// req.Directives is now nil (overwritten)
// But resp2.Decision == "deny" — caller inspects req thinking directives reflect the deny context
```

## Fix options

**A. Document the mutation** on `ValidateAction`, `EvaluateConditions`, and the `Directives` field of `EnforceRequest`, and add a note in `capability.Enforcer` that `ValidateAction` modifies `req.Directives`. Low risk, no behavior change.

**B. Move the assignment inside the allow branch only** — set `req.Directives = matched.Directives` only when all conditions pass (immediately before building obligations), so a deny never contaminates `req`. Directives on a deny are meaningless anyway; Rego policy conditions that reference `input.directives` during a deny path should see `[]`.

**C. Pass directives to handlers through the context** rather than mutating `req`, removing the side effect entirely. Larger refactor, cleanest API.

Option B is low-risk and fixes the primary correctness concern (directives on a denied request) while keeping the Rego-input use case working for the allow path.

### LINKED ISSUE #310: bug: resourceSpecificity ignores post-wildcard literal content — `*_export` and `*` score identically, causing wrong constraint to win on tie
## Summary

`resourceSpecificity` in `pkg/enforcement/engine.go` scores a pattern by counting the literal characters before the first wildcard (`prefixLen`) minus the number of wildcard characters (`wildcardCount`). It does not account for any literal content after the first wildcard.

```go
// engine.go:703-716
prefixLen := 0
for _, r := range resource {
    if strings.ContainsRune("*?[", r) {
        break
    }
    prefixLen++
}
wildcardCount := 0
for _, r := range resource {
    if strings.ContainsRune("*?[", r) {
        wildcardCount++
    }
}
```

Two patterns that both begin with a wildcard score identically regardless of how much literal content follows:

| Pattern | prefixLen | wildcardCount | Score |
|---|---|---|---|
| `*` | 0 | 1 | -1 |
| `*_export` | 0 | 1 | -1 |
| `*_export_v2` | 0 | 1 | -1 |
| `tool_*` | 5 | 1 | 49 |
| `tool_*_v2` | 5 | 1 | 49 |

## When this matters

A manifest with two overlapping constraints where the more specific one starts with a wildcard:

```yaml
capabilities:
  - target: tool:*          # catch-all: deny with conditions
    actions: [call]
    conditions:
      - type: maxCalls
        count: 10
        windowSeconds: 3600

  - target: tool:*_admin    # admin tools: deny entirely (no conditions → always allow, but stricter intent)
    actions: [call]
    conditions:
      - type: allowedValues
        argument: role
        values: ["admin"]
```

For a request to `tool:create_admin`, both patterns match. `resourceSpecificity("*_admin", "create_admin")` returns -1 and `resourceSpecificity("*", "create_admin")` also returns -1. The tie resolves to first-in-list order, so the winner depends on manifest ordering rather than semantic specificity. An operator who puts the catch-all first (natural order: general before specific) would have `*_admin` silently shadowed.

The correct score for `*_admin` should be higher than for `*`, since `_admin` is literal content that uniquely qualifies the wildcard match. A pattern with more total literal characters is unambiguously more specific.

## Fix

After computing `prefixLen` and `wildcardCount`, also count the literal characters **after** all wildcard characters and add them to the score. A simple approximation: count total non-wildcard rune characters in the pattern.

```go
literalCount := 0
for _, r := range resource {
    if !strings.ContainsRune("*?[", r) {
        literalCount++
    }
}
return literalCount*10 - wildcardCount
```

This gives `*_admin` a score of `6*10 - 1 = 59` and `*` a score of `0*10 - 1 = -1`, correctly ranking the more specific pattern above the catch-all. Exact matches (`prefixLen == len(toolName)`) already short-circuit to score 1000 and are unaffected.

The existing behavior is documented as "Tie-breaking is stable: if two capabilities have identical resource specificity scores, the one that appears first in the list wins" — but the tie only arises because post-wildcard literal content is not scored. Fixing the scoring removes the unstable tie entirely.

### LINKED ISSUE #311: usability: sequenceBlock denial Details reports blockedTool as bare name but afterTool as namespaced — inconsistent audit log format
## Summary

When `handleSequenceBlock` produces a denial, the `Details` map has an inconsistent format for the two tool name fields:

```go
// handlers.go:948-957
blockedTool := stripEnginePrefix(req.ToolName)   // bare: "write_external"
priorTarget := priorType + ":" + priorTool        // namespaced: "tool:read_credentials"

return &ConditionError{
        "afterTool":   priorTarget,   // "tool:read_credentials"
        "blockedTool": blockedTool,   // "write_external"  ← no namespace
        "afterTools":  sb.AfterTools,
    },
}
```

`afterTool` is in `namespace:name` form (`"tool:read_credentials"`), but `blockedTool` is the bare name without namespace (`"write_external"`). The denial `Message` also uses the bare form:

```
tool "read_credentials" was already called in this session; "write_external" is blocked after it
```

## Why it matters

**Audit log analysis:** A query filtering audit records by the blocked tool (e.g. "show me every sequenceBlock denial that blocked a write operation") must use `blockedTool: "write_external"` for the blocked side but `afterTool: "tool:read_credentials"` for the antecedent. A consistent namespace-qualified format on both sides would let a single filter pattern (`"tool:*"`) address all tool-namespace denials.

**Prompt/resource namespaces:** The fix that introduced `targetType` into the denial (#221) correctly namespaces `afterTool`, but left `blockedTool` bare. If both the blocked tool and the antecedent could be prompts or resources, the bare `blockedTool` value is ambiguous — `"export"` could be a tool or a prompt.

**Cross-namespace policies:** A `sequenceBlock` that fires because `prompt:export` was called should report `blockedTool: "tool:write_external"` (or whatever the blocked resource's namespace is), not just `"write_external"`. The namespace is currently available from `req.Target.Type` (set by ManifestPDP) or derivable from `splitEnginePrefix(req.ToolName)`.

## Fix

Compute `blockedTool` the same way `priorTarget` is computed — by joining the blocked target's namespace type and bare name:

```go
blockedType, blockedName := splitEnginePrefix(req.ToolName)
    blockedType = req.Target.Type
}
blockedTarget := blockedType + ":" + blockedName

// Details:
"blockedTool": blockedTarget,   // "tool:write_external"
```

And update the Message to use the qualified form:

```
tool "read_credentials" was already called in this session; tool "write_external" is blocked after it
```

This keeps the format consistent with `afterTool` and makes the audit record unambiguous when tools, prompts, and resources share names.

### LINKED ISSUE #312: bug: `engine.ValidateAction` and `FindMatchingCapability` silently fail for resource, prompt, and system constraints
## Summary

`engine.ValidateAction` and `engine.FindMatchingCapability` — the two methods exposed by the `capability.Enforcer` interface — use `actionMatchScore` to filter constraints. `actionMatchScore` only returns `true` for `"call"` or `"*"`. Actions valid for other target types (`"read"` for `resource:`, `"get"` for `prompt:`, `"allow"` for `system:`) are silently unrecognized, causing `findMatchingCapability` to skip every matching constraint.

As a result, a library consumer who calls `engine.ValidateAction` or `engine.FindMatchingCapability` with resource, prompt, or system constraints always gets a "no matching capability" denial — even for a perfectly valid manifest entry.

## Affected code

`pkg/enforcement/engine.go` — `actionMatchScore`:

```go
func actionMatchScore(actions []string, _ *capability.EnforceRequest) bool {
    for _, a := range actions {
        if a == "call" || a == "*" {
            return true
        }
    }
    return false
}
```

The comment explicitly documents the restriction ("The only accepted values are 'call' (the MCP tool-call verb) and '*' (wildcard)"), but the `Enforcer` interface exposes `ValidateAction` as the general enforcement entrypoint without surfacing this restriction:

```go
// pkg/capability/enforce.go
type Enforcer interface {
    ValidateAction(ctx context.Context, req *EnforceRequest, capabilities []Constraint) (EnforceResponse, error)
    FindMatchingCapability(req *EnforceRequest, capabilities []Constraint) *Constraint
}
```

## Why this is a problem

`ManifestPDP` works around the issue by NOT using `ValidateAction`. It calls `findConstraint` (its own per-type constraint finder) and then `EvaluateConditions` directly. But the exported `Enforcer` interface contract and `FindMatchingCapability` docstring make no mention of this restriction:

> "FindMatchingCapability returns the most specific capability that matches the request. It uses the same glob semantics and specificity scoring as ValidateAction, and is exported so that callers that need the matched constraint (e.g. the validate endpoint) can obtain it without re-implementing the selection logic."

A library consumer following the `Enforcer` interface will:

1. Write a manifest with `{target: "resource:file.txt", actions: ["read"], conditions: [...]}`.
2. Call `engine.ValidateAction(ctx, req, capabilities)` or `engine.FindMatchingCapability(req, capabilities)`.
3. Always receive a `"no matching capability"` denial, with no diagnostic distinguishing it from a genuine no-match.

## Reproduction

```go
engine := enforcement.New()
constraints := []capability.Constraint{
    {
        Target:  "resource:file.txt",
        Actions: []string{"read"},
    },
}
req := &capability.EnforceRequest{ToolName: "file.txt"}

// Returns "no matching capability" — constraint is never selected
resp, _ := engine.ValidateAction(context.Background(), req, constraints)
// resp.Decision == "deny", resp.Denial.Code == "AUTHORIZATION_FAILED"

// Also returns nil
matched := engine.FindMatchingCapability(req, constraints)
// matched == nil
```

## Impact

- Any library consumer relying on `capability.Enforcer` for non-tool targets gets silent fail-closed misbehavior that is indistinguishable from a policy denial in the audit log.
- The `validate --live` flow uses its own path that avoids this, as does `ManifestPDP`, but third-party integrations and external PDPs that embed the engine directly are affected.

## Proposed fix

Extend `actionMatchScore` (or replace it with per-target-type awareness) to recognize the canonical action for each namespace:

```go
func actionMatchScore(actions []string, req *capability.EnforceRequest) bool {
    for _, a := range actions {
        if a == "*" {
            return true
        }
        // Match the canonical action for the request's target type.
        if req.Target != nil {
            switch capability.TargetType(req.Target.Type) {
            case capability.TargetTypeResource:
                if a == "read" { return true }
            case capability.TargetTypePrompt:
                if a == "get" { return true }
            case capability.TargetTypeSystem:
                if a == "allow" { return true }
            default:
                if a == "call" { return true }
            }
        } else if a == "call" {
            return true // legacy: no Target set → default to tool/call
        }
    }
    return false
}
```

Alternatively, align the `Enforcer` interface docs and `FindMatchingCapability` godoc to make the tools-only limitation explicit, and expose a separate `FindConstraintForTarget(target EnforceTarget, capabilities []Constraint)` that applies the correct per-type action check — matching the logic that `ManifestPDP.findConstraint` already implements.

### LINKED ISSUE #313: bug: `obligation.go` "annotate" type is dead code — no directive generates it and the engine never emits it
## Summary

`pkg/capability/obligation.go` implements full JSON serialization and deserialization for an `"annotate"` obligation type (with `Key` and `Value` fields). However:

1. No `Directive` struct in `pkg/capability/directive.go` or `pkg/capability/directives.go` generates an annotate obligation.
2. The engine's obligation-collection loop in `pkg/enforcement/engine.go` only handles `*capability.RedactFieldsDirective`; any other directive type is silently dropped.
3. `marshalDirective` in `pkg/capability/directives.go` returns `"unsupported directive payload"` for any type other than `RedactFieldsDirective`.
4. `unmarshalDirective` returns `"unknown directive type"` for any type other than `"redactFields"`, so a manifest cannot declare an annotate directive.

The result is that the annotate type is reachable only via `Obligation.UnmarshalJSON` (e.g., reading an audit log) — it cannot be produced through any manifest or engine code path.

## Code locations

**`pkg/capability/obligation.go`** — handles "annotate" in marshal/unmarshal:
```go
case "annotate":
    return json.Marshal(struct {
        Type  string `json:"type"`
        Key   string `json:"key"`
        Value string `json:"value"`
    }{Type: o.Type, Key: o.Key, Value: o.Value})
```

**`pkg/enforcement/engine.go`** — obligation collection after allow (identical in `ValidateAction` and `EvaluateConditions`):
```go
for _, dir := range matched.Directives {
    if rd, ok := dir.(*capability.RedactFieldsDirective); ok {
        obligations = append(obligations, capability.Obligation{
            Type:  capability.DirectiveTypeRedactFields,
            Paths: rd.Fields,
        })
    }
    // No else branch — any future directive type is silently dropped
}
```

**`pkg/capability/directives.go`** — `unmarshalDirective` cannot produce an annotate directive:
```go
func newDirective(directiveType string) Directive {
    switch directiveType {
    case DirectiveTypeRedactFields:
        return &RedactFieldsDirective{}
    default:
        return nil  // "annotate" → nil → error returned upstream
    }
}
```

## Why it matters

1. **Misleading API surface**: A reader of `obligation.go` sees the annotate type and assumes it is supported. There is no `AnnotateDirective`, no path through the engine that produces an annotate obligation, and no proxy handler that applies one — so the type is phantom.

2. **Silent forward-compatibility trap**: When an `AnnotateDirective` is eventually added, the developer must remember to update the engine's obligation-collection loop. Nothing in the code signals this requirement; a well-written `AnnotateDirective` + `newDirective` update would load and round-trip correctly but produce zero runtime effect because the engine loop still ignores it.

3. **Audit-log inconsistency**: If an annotate obligation were somehow inserted into an audit record (e.g., by a future proxy version), `Obligation.UnmarshalJSON` would accept it without error, but the current proxy could never have produced it.

## Proposed fix

**Option A — Remove "annotate" from `obligation.go` until an `AnnotateDirective` is implemented.** Pre-1.0 policy allows clean removal. The `Key` and `Value` fields on `Obligation` can also be dropped until needed.

**Option B — Implement `AnnotateDirective` to close the gap.** Define the struct in `directive.go`, add it to `directives.go`'s marshal/unmarshal, and handle it in the engine loop:
```go
case *capability.AnnotateDirective:
    obligations = append(obligations, capability.Obligation{
        Type:  "annotate",
        Key:   ad.Key,
        Value: ad.Value,
    })
```

**Option C — Add a panicking or error-returning default in the engine loop** (compile-time exhaustiveness guard), so adding a new directive without updating the loop fails loudly:
```go
default:
    return capability.EnforceResponse{}, fmt.Errorf("unhandled directive type %T in obligation collection; update engine.go", dir)
```

Option A is the lowest-risk pre-1.0 cleanup. Option C is the minimum needed to prevent the silent-drop maintenance trap regardless of which option is chosen for the annotate type itself.

### LINKED ISSUE #314: bug: engine obligation-collection loop silently drops any directive type other than `redactFields`
## Summary

The obligation-collection loops in `pkg/enforcement/engine.go` (`ValidateAction` lines 414–422 and `EvaluateConditions` lines 604–612) use a bare `if` type assertion with no `else` branch. Any `Directive` type that is not `*capability.RedactFieldsDirective` is silently discarded. A directive that loads successfully from the manifest and is attached to the matched constraint produces no obligation at runtime, with no error, no log entry, and no audit record distinguishing this from a constraint with no directives.

## Current code (identical in both methods)

```go
var obligations []capability.Obligation
for _, dir := range matched.Directives {
    if rd, ok := dir.(*capability.RedactFieldsDirective); ok {
        obligations = append(obligations, capability.Obligation{
            Type:  capability.DirectiveTypeRedactFields,
            Paths: rd.Fields,
        })
    }
    // No else — unrecognized directive types are silently dropped
}
```

## Why this is a bug

A directive's contract is that it produces a runtime obligation. Silently dropping it is a silent fail-open: the proxy forwards the upstream response without applying a declared post-processing obligation. For security-relevant directives (redaction, annotation of sensitive content, etc.) a silent drop is indistinguishable from correct enforcement unless the operator compares every audit record against the manifest.

The bug is a maintenance trap rather than a current exploitable flaw (only `redactFields` exists today), but it is triggered automatically the moment a second directive type is added — which `obligation.go`'s "annotate" type signals is planned.

## Scenario

1. Developer adds `AnnotateDirective` to `pkg/capability/directive.go` and `directives.go`.
2. Manifest loader accepts `{type: annotate, key: "sensitivity", value: "high"}` in a directive array.
3. The engine's `EvaluateConditions` loops over directives, hits the `if rd, ok := dir.(*RedactFieldsDirective); ok` guard — which is false for an `AnnotateDirective` — and skips it silently.
4. The `EnforceResponse.Obligations` slice contains no annotate obligation.
5. The transport handler applies no annotation to the upstream response.
6. The audit record shows `obligations: []` — identical to a constraint with no directives.

There is no error, no warning, and no test failure unless the developer also adds a test that checks the obligation output. The existing `pdp_directives_test.go` tests cover `redactFields` only.

## Proposed fix

Replace the bare `if` with a type-switch that has an explicit `default` arm:

```go
var obligations []capability.Obligation
for _, dir := range matched.Directives {
    switch d := dir.(type) {
    case *capability.RedactFieldsDirective:
        obligations = append(obligations, capability.Obligation{
            Type:  capability.DirectiveTypeRedactFields,
            Paths: d.Fields,
        })
    default:
        // Fail closed: a directive the engine does not handle cannot be applied.
        // Return an error rather than silently forwarding without the obligation.
            RequestID: requestID,
            DecidedAt: now,
                Code:    capability.ErrCodeEnforcementError,
                Message: fmt.Sprintf("unhandled directive type %T; the enforcement engine must be updated to emit the corresponding obligation", dir),
            },
        }, nil
    }
}
```

This preserves the fail-closed invariant: a directive the engine cannot apply results in a deny rather than silent omission. Updating the engine when a new directive type is added is then a hard requirement (the default arm fires in tests) rather than an implicit expectation.

The same fix must be applied in both `ValidateAction` (around line 414) and `EvaluateConditions` (around line 604).

## Related

- `obligation.go`'s "annotate" type is already dead code for this exact reason (companion issue filed separately).
- `pdp_directives_test.go` only tests `redactFields`; add a test that covers the default/unknown-directive path.

### LINKED ISSUE #315: bug: `JWKSClient.VerifyToken` calls `time.Now()` directly — token expiry and leeway are untestable
## Summary

`pkg/capability/jwks.go` — `JWKSClient.VerifyToken` — calls `time.Now()` directly when building the `jwt.Expected` struct used for claim validation:

```go
// jwks.go:160-163
expected := jwt.Expected{
    Time:        time.Now(),   // ← hardcoded real time
}
if err := claims.ValidateWithLeeway(expected, time.Minute); err != nil {
```

`JWKSClientConfig` has no clock injection point. The 1-minute leeway, near-expiry handling, and the fail-closed behavior when a token's `exp` claim has just passed cannot be unit-tested without real time delays.

The same pattern is inconsistent with `TokenCacheConfig`, which does provide a `Now func() time.Time` field specifically so that cache-expiry logic can be tested without sleeps:

```go
// token_cache.go:32-33
// Now is an optional clock function used for testing. Defaults to time.Now.
Now func() time.Time
```

Additionally, `refreshKeys` hardcodes `time.Now()` when stamping the JWKS cache timestamp:

```go
// jwks.go:266
c.fetchedAt = time.Now()
```

This means the JWKS cache TTL (`CacheTTL`) cannot be advanced in tests either — a test that needs to force a JWKS re-fetch cannot do so without a real `CacheTTL` sleep.

## Impact

1. **Token expiry edge cases are untestable**: whether a token 59 seconds past expiry is rejected (it should be, within the leeway), or a token exactly at the leeway boundary behaves correctly, requires real-time delays — or cannot be tested at all.
2. **JWKS cache behavior is untestable**: a test that verifies JWKS is refreshed after `CacheTTL` must sleep for the full TTL duration (default 5 minutes), making it impractical.
3. **Inconsistency**: `TokenCache` is fully clock-injectable; `JWKSClient` is not. Both live in `pkg/capability` and are configured at the same call site in `cmd/eunox-mcp`.

## Proposed fix

Add a `Now func() time.Time` field to `JWKSClientConfig` and thread it into `JWKSClient`:

```go
type JWKSClientConfig struct {
    // ... existing fields ...
    // Now is an optional clock function used for testing. Defaults to time.Now.
    Now func() time.Time
}

func NewJWKSClient(cfg JWKSClientConfig) *JWKSClient {
    // ...
    if cfg.Now == nil {
        cfg.Now = time.Now
    }
    client := &JWKSClient{
        // ...
        now: cfg.Now,
    }
}
```

Then replace the two hardcoded calls:

```go
// VerifyToken — line 161
expected := jwt.Expected{
    Time:        c.now(),   // ← injectable
}

// refreshKeys — line 266
c.fetchedAt = c.now()   // ← injectable
```

And add test cases for:
- Token whose `exp` is `now - 30s` (within 1-minute leeway) → should allow
- Token whose `exp` is `now - 90s` (outside leeway) → should deny
- JWKS cache hit when `now - fetchedAt < CacheTTL` → no re-fetch
- JWKS cache miss when `now - fetchedAt >= CacheTTL` → re-fetch triggered

### LINKED ISSUE #316: usability: `allowedExtensions` suffix match — a single extension (e.g. `.gz`) implicitly allows compound extensions (e.g. `.tar.gz`)
## Summary

`handleAllowedExtensions` in `pkg/enforcement/handlers.go` matches file paths against the allowlist using `strings.HasSuffix(base, suffix)` on the full filename. This means that allowing `.gz` also implicitly allows any file whose name ends in `.gz`, including compound-extension archives like `backup.tar.gz`, `dump.sql.gz`, and `data.dump.gz`. An operator who writes `extensions: [".gz"]` intending to allow only gzip-compressed single files will be surprised to find that rename attacks and compound archives also pass.

## The matching logic

```go
// handlers.go:576-581
for _, suffix := range allowed {
    if strings.HasSuffix(base, suffix) {
        matched = true
        break
    }
}
```

The `allowed` slice contains normalized suffixes including a leading dot (e.g. `".gz"`, `".tar.gz"`). Since suffix matching operates on the entire lowercased base filename:

| File | AllowedExtensions | HasSuffix result | Decision |
|------|-------------------|-----------------|----------|
| `data.gz` | `[".gz"]` | `true` | allow |
| `backup.tar.gz` | `[".gz"]` | `true` | **allow (likely surprising)** |
| `backup.tar.gz` | `[".tar.gz"]` | `true` | allow |
| `report.pdf` | `[".gz"]` | `false` | deny |

## Why this is surprising

The "no extension" guard uses `filepath.Ext(filePath)` (last extension only):

```go
ext := strings.ToLower(filepath.Ext(filePath))
if ext == "" {
    return &ConditionError{...}  // "file has no extension"
}
```

For `backup.tar.gz`, `filepath.Ext` returns `.gz` (the last extension only). The "no extension" guard passes because `.gz` is non-empty. Then the allowlist check uses `strings.HasSuffix` on the full base name, so `.gz` in the allowlist matches `backup.tar.gz`.

The two code paths use different extension representations: `filepath.Ext` (last-only) for the empty-ext guard, but `strings.HasSuffix` (full base) for the allowlist match. The denial message uses `fileSuffix(base)` (full compound suffix), which can differ from what `filepath.Ext` would return — creating a third representation.

## Security relevance

An operator restricts a file-upload tool to `extensions: [".pdf"]` to prevent arbitrary file types. A client uploads `malicious.sh.pdf`. The extension check passes because `strings.HasSuffix("malicious.sh.pdf", ".pdf") == true`. The upstream server receives a shell script with a `.pdf` suffix. While rename attacks are hard to fully prevent with extension checks alone, the behavior is counterintuitive and undocumented.

More concretely, an `allowedExtensions: [".gz"]` list meant to allow compressed telemetry exports would also allow a `.tar.gz` archive containing arbitrary files.

## Proposed fix / options

**Option A — Document the behavior explicitly** in `docs/capability-manifest-guide.md`:
> Extensions are matched as a suffix of the full filename, not just the last dot-separated component. `extensions: [".gz"]` matches any file whose name ends in `.gz`, including compound extensions like `.tar.gz`. To restrict to a specific compound extension, list it explicitly: `extensions: [".tar.gz"]`.

**Option B — Match only the last extension (filepath.Ext) by default**, breaking compound extension matching. This is a behavior change: operators who depend on `[".tar.gz"]` matching the compound suffix would need to add `.gz` as well, or the handler could support both.

**Option C — Match against `filepath.Ext` for the not-found guard AND use `strings.HasSuffix` for the allowlist** (current behavior), but surface this in an explicit constant or comment explaining why the two are intentionally different and what the security implication is. Currently neither the code nor the docs explain this.

Option A is the minimum to prevent operator confusion. The capability-manifest guide currently says only:

> `allowedExtensions`: list of file extensions permitted (e.g. `.pdf`, `.csv`). The argument field must name the parameter that carries the path(s). Every path must match or the call is denied.

No mention of compound-extension behavior.

## Missing test

There is no test that:
1. Puts `.gz` in the allowlist and checks whether `backup.tar.gz` is allowed or denied.
2. Verifies the documented semantics of single vs. compound extension entries.

Add a table-driven test to `handlers_test.go` (or equivalent) that covers:
- `file.gz` against `[".gz"]` → allow
- `file.tar.gz` against `[".gz"]` → (currently) allow — document whether this is intended
- `file.tar.gz` against `[".tar.gz"]` → allow
- `file.gz` against `[".tar.gz"]` → deny (`.gz` does not end in `.tar.gz`)

eunolabs:
## Summary

Addresses issues **#306 through #316**. A few of these were already resolved by the earlier #317–#332 work; this PR confirms them with regression coverage and fixes the rest. All changes preserve the fail-closed invariant.

| Issue | Status | What this PR does |
|---|---|---|
| #306 | Already fixed (#317) | `actionMatchScore` already recognizes each namespace's canonical verb via `ValidateActionForTargetType`. Confirmed by existing `TestValidateAction_CanonicalVerbPerNamespace`. |
| #307 | Docs | `argumentSchema` `pattern` is unanchored (substring) — the guide already documents it (#331); added the same note to the `ArgumentSchema.Pattern` field godoc. |
| #308 | Fixed | `ValidateAction`/`EvaluateConditions` now stamp `EnforceResponse.AuditOnly` from the matched constraint on every matched-constraint deny, so a direct engine caller no longer needs the PDP's `defer` to distinguish an audit-mode deny from a hard deny. No-match and record-failure denials stay hard. |
| #309 | Already fixed (#322) | Directives are threaded via context, not by mutating the caller's request. Corrected the now-stale `EnforceRequest.Directives` field doc. |
| #310 | Fixed | `resourceSpecificity` now counts **every** literal rune (including those after a wildcard), so `*_admin` outranks the catch-all `*` regardless of manifest order instead of tying at `-1`. |
| #311 | Fixed | `sequenceBlock` denials report `blockedTool` in `namespace:name` form (consistent with `afterTool`) and name the blocked target's namespace in the message. |
| #312 | Already fixed (#317) | Same root cause as #306; covered by the same fix and test. |
| #313 | Fixed | Removed the dead `annotate` obligation type and its `Key`/`Value` fields — no directive ever produced it and the engine never emitted it. |
| #314 | Fixed | The obligation-collection loop is now a type switch with a fail-closed `default`: an unhandled directive type denies with `ENFORCEMENT_ERROR` rather than being silently dropped. |
| #315 | Fixed | `JWKSClientConfig` gains an injectable `Now` clock (mirroring `TokenCacheConfig`), used for token expiry/leeway and the JWKS cache-TTL check, making those edges unit-testable without real sleeps. |
| #316 | Docs | Compound-extension behavior (`.gz` admits `.tar.gz`) is already documented in the guide (#323); added regression coverage. |

## Notable behavior changes

- **Specificity ranking (#310):** the most-specific match now accounts for post-wildcard literal content, so an overlapping `tool:*_admin` is no longer shadowed by an earlier-listed `tool:*`. The user-facing guide's "most-specific match wins" note was updated to match.
- **Audit-log shape (#311):** the `blockedTool` value inside a `sequenceBlock` denial's `Details` map is now namespace-qualified. This is a `Details` entry, not a top-level audit field, so the threat model is unaffected.

## Tests

- New regression tests in `pkg/enforcement/issues_306_316_test.go` (#308, #310, #311, #314) and `pkg/capability/issues_306_316_test.go` (#313, #315).
- Updated existing tests affected by the specificity formula change (#310) and the `annotate` removal (#313).
- `go build`, `go vet`, `golangci-lint`, `gofmt`, and the full `-race` suite all pass.

Closes #306
Closes #307
Closes #308
Closes #309
Closes #310
Closes #311
Closes #312
Closes #313
Closes #314
Closes #315
Closes #316

https://claude.ai/code/session_013rJ6BhNKg2Uj4xrBrG3nwk

---
_Generated by [Claude Code](https://claude.ai/code/session_013rJ6BhNKg2Uj4xrBrG3nwk)_


pkg/enforcement/engine.go
```diff
@@ -801,31 +847,38 @@ func resourceSpecificity(resource, toolName string) int {
...
 	// every rune of every candidate pattern on the enforcement hot path (issue
-	// #330). prefixLen counts the leading non-wildcard runes (those before the first
-	// glob metacharacter); wildcardCount counts every metacharacter in the pattern.
-	prefixLen := 0
+	// #330). wildcardCount counts every metacharacter in the pattern.
+	literalCount := 0
 	wildcardCount := 0
-	inPrefix := true
 	for _, r := range resource {
 		if r == '*' || r == '?' || r == '[' {
```
eunolabs: **#310 — counting `[ ]` character-class internals as literals over-weights them.** The fix correctly solves the `*_admin` vs `*` ordering, but because only `*`, `?`, `[` are treated as wildcards, the runes *inside* a class — and the closing `]` — all fall through to `literalCount`. So a one-character class is scored as if it contributed several literal chars:

```
read_[abc]  => 89   // a,b,c,] all count as literal; only '[' is a wildcard
read_file   => 90
user_*      => 49
```

`read_[abc]` matches exactly one character in that position yet scores nearly as high as the 9-char literal `read_file`, and the `]` is pure syntax that matches nothing. It won't break the targeted fix (ties between a bracket pattern and another pattern matching the same name are rare), and exact matches still short-circuit to 1000 — so this is a low-severity heuristic wart, not a blocker. But since the formula is being reworked here anyway, it'd be worth either skipping class internals/`]` or treating a whole `[...]` group as a single wildcard unit. At minimum, a regression test with a `[...]` target would pin whatever behavior you intend.

---

pkg/enforcement/engine.go
```diff
@@ -709,11 +736,30 @@ func (e *Engine) EvaluateConditions(ctx context.Context, req *capability.Enforce
...
 	for _, dir := range matched.Directives {
-		if rd, ok := dir.(*capability.RedactFieldsDirective); ok {
+		switch d := dir.(type) {
+		case *capability.RedactFieldsDirective:
 			obligations = append(obligations, capability.Obligation{
 				Type:  capability.DirectiveTypeRedactFields,
-				Paths: rd.Fields,
+				Paths: d.Fields,
 			})
+		default:
```
eunolabs, line 760: **#314 — two things worth confirming on this fail-closed arm.**

1. **Audit-mode interaction (please confirm intended).** This loop runs only after all conditions passed and `recordSessionCall` succeeded — i.e. on the would-be *allow* path. For an **audit-mode** constraint, an unhandled directive here turns a request that audit mode contracts to "log-but-forward" into a **hard** deny with no `AuditOnly` flag. I think hard-deny is the right call (an unwired directive is an engine bug, and forwarding e.g. an un-applied redaction would leak — "fail closed on ambiguity" wins over "audit never blocks"). Just flagging it as a deliberate-looking choice that isn't called out: the deny is intentionally *not* `AuditOnly` even when `matched.IsAuditOnly()`. A one-line comment or test asserting that would lock the intent in.

2. **Duplication (optional).** This ~16-line `default` arm — and the whole `for _, dir := range matched.Directives` collection loop — is now byte-for-byte identical in `ValidateAction` and `EvaluateConditions`. Extracting a `collectObligations(matched, requestID, now) ([]Obligation, *EnforceResponse)` helper would keep the two from drifting (the next directive type would otherwise need the same edit in two places). Non-blocking.

---

pkg/capability/constraint.go
```diff
@@ -137,15 +137,22 @@ type ArgumentSchema struct {
...
-	MinLength            *int                       `json:"minLength,omitempty"`
-	MaxLength            *int                       `json:"maxLength,omitempty"`
-	Minimum              *float64                   `json:"minimum,omitempty"`
-	Maximum              *float64                   `json:"maximum,omitempty"`
-	Items                *ArgumentSchema            `json:"items,omitempty"`
-	MaxItems             *int                       `json:"maxItems,omitempty"`
-	MinItems             *int                       `json:"minItems,omitempty"`
-	Description          string                     `json:"description,omitempty"`
+	// Pattern is an RE2 regexp the string value must match. Matching follows
+	// JSON-Schema / ECMA-262 semantics and is UNANCHORED: the regex need only be
```
eunolabs: Doc-precision nit (#307): `Pattern` is compiled with Go's `regexp` (`compilePattern` → `regexp.Compile`) in `pkg/enforcement/schema.go`, i.e. **RE2**, not full ECMA-262. RE2 has no lookahead/lookbehind or backreferences, so a policy author who copies an ECMA-262 pattern using those will hit a load-time `invalid pattern` error rather than the unanchored-match surprise this comment is warning about. The substantive point (unanchored → anchor with `^...$`) is correct and well worth stating; consider saying "RE2 (Go `regexp`) semantics" instead of "JSON-Schema / ECMA-262 semantics" to avoid implying unsupported constructs work.

---

eunolabs (commented):
## Review — #306–#316 enforcement / JWKS / obligation fixes

Solid, well-scoped PR. Each fix is small, fails closed, and is backed by a focused regression test. I checked it out locally and confirmed `go build`, `go vet`, `gofmt -l`, and the `-race` suites for `pkg/capability`, `pkg/enforcement`, and the touched `cmd/eunox-mcp` test all pass.

**Verified the load-bearing bits:**
- **#308** — `matched` is non-nil at every new `AuditOnly` stamping site (guarded by the no-match return; `EvaluateConditions` already dereferences `matched` first), so no nil-deref risk. The engine's stamping and the PDP's `defer resp.AuditOnly = true` agree (both set `true` for audit constraints) — no double-stamp conflict — and the new `cmd` tests still assert the no-match / kill-switch denials stay hard.
- **#311** — `blockedTool` is now produced in exactly one place and the one pre-existing assertion in `sequence_block_test.go` only checks `afterTool`, so nothing breaks on the Details-shape change. Agreed this stays inside `Details` and doesn't touch a top-level audit field, so the threat model is unaffected.
- **#313** — no remaining references to the removed `annotate` / `Obligation.Key` / `Obligation.Value` anywhere in the tree (or in `schemas/`), and the `DirectiveTypeRedactFields` constant is now the single source for the discriminator. The doc comment pointing the way back if `annotate` is ever implemented is a nice touch.
- **#315** — clock injection is complete: no stray `time.Now()` / `time.Since` left in `jwks.go`; `VerifyToken` leeway, both cache-TTL checks, and the `fetchedAt` store all go through `c.now()`.

**Minor / non-blocking** (left inline):
1. `resourceSpecificity` now counts `[...]` character-class internals and the closing `]` as literals, over-weighting bracket patterns (`read_[abc]` → 89). Doesn't affect the `*_admin` fix; worth a follow-up + a `[...]` regression test.
2. The fail-closed `default` arm (#314) turns a would-be audit-mode *allow* into a hard, non-`AuditOnly` deny — I believe that's correct, just worth pinning intent; and the arm is now duplicated verbatim across both engine methods (optional helper).
3. `Pattern` godoc says "ECMA-262 semantics" but it's compiled as RE2 (Go `regexp`) — minor wording.

Nothing here blocks merge.

---

eunolabs:
Thanks for the thorough review. Addressed all three follow-ups in 8ee1758:

- **#310 (bracket scoring)** — `resourceSpecificity` now treats a whole `[...]` character class as a single wildcard (like `?`): its members and the closing `]` are skipped, so `read_[abc]` scores 49 instead of 89. Added `TestSpecificity_CharacterClassScoredAsSingleWildcard` pinning that a class scores like `?` and that the number of class members doesn't change the score.
- **#314 (fail-closed arm)** — (1) Confirmed intended: the unhandled-directive deny is a hard `ENFORCEMENT_ERROR` and deliberately **not** `AuditOnly` even when `matched.IsAuditOnly()`, since an unwired directive is an engine bug, not a policy verdict, and forwarding an un-post-processed response (e.g. an un-applied redaction) is the exact leak audit mode must not wave through. Pinned with a new `audit-mode constraint still hard-denies, not AuditOnly` subtest and a comment on the helper. (2) Extracted the duplicated loop into `Engine.collectObligations(matched, requestID, now) ([]Obligation, *EnforceResponse)`, so the next directive type is wired up in one place.
- **#307 (godoc wording)** — Reworded the `ArgumentSchema.Pattern` doc to say RE2 (Go `regexp`), not ECMA-262, and to note lookaround/backreferences are unsupported and rejected at load.

CI was green on the prior commit; will re-verify on the new one.

---

pkg/enforcement/engine.go
```diff
@@ -801,31 +848,66 @@ func resourceSpecificity(resource, toolName string) int {
...
-	for _, r := range resource {
-		if r == '*' || r == '?' || r == '[' {
+	runes := []rune(resource)
+	for i := 0; i < len(runes); i++ {
+		switch runes[i] {
+		case '*', '?':
+			wildcardCount++
+		case '[':
+			// One class = one wildcard; advance past it to the closing ']'. Per
+			// path.Match, a ']' immediately after '[' or '[^' is a literal member
```
eunolabs: Tiny follow-up nit on the (otherwise correct) class scanner — non-blocking.

I cross-checked this scoring against `path.Match` for valid class patterns (`[abc]`, `[^x]`, `[a-z]`, an escaped `[a\]b]`, `pre[abc]post`, `a*[0-9]?z`) and every one scores exactly as "literals-outside-the-class ×10 − wildcards", so the fix is correct for everything that can actually reach this function.

One inaccuracy in the comment, though: *"Per path.Match, a ']' immediately after '[' or '[^' is a literal member (not the terminator)"* isn't true for Go's `path.Match`. Unlike POSIX/`fnmatch`, Go's matcher rejects a leading `]` — `getEsc` returns `ErrBadPattern` for it, so `path.Match("[]a]", …)` yields `syntax error in pattern`, not a match. I confirmed this empirically. So a `[]…` class is never load-valid and (being unmatchable) never reaches scoring either way, which means the `if runes[i] == ']' { i++ }` leading-member branch (lines 897–899) is handling input that can't occur. Harmless, but the comment will mislead a future maintainer into thinking `[]abc]` is a usable eunox target. Worth either dropping that branch or rewording to "Go's path.Match rejects a leading ']' , so we never see one; this guard is purely defensive." Your call — doesn't affect behavior.

---

eunolabs (commented):
## Re-review — follow-up commit `8ee1758`

Re-reviewed after the `fix: address review on #310/#314/#307 follow-ups` commit. All three points from the previous round are properly resolved, and I re-verified locally (`go build`, `go vet`, `gofmt -l`, and the `-race` suites for `pkg/enforcement` + `pkg/capability` all pass on the new head):

- **#310 — character-class over-weighting: fixed.** A whole `[...]` now counts as one wildcard (like `?`), so `read_[abc]` scores 49 instead of 89. I cross-checked the new scanner against `path.Match` for ranges, negation, escaped `]`, and class-then-literal patterns — all score correctly, and the inner `i`-advance handles escapes and the closing `]` without off-by-one. New `TestSpecificity_CharacterClassScoredAsSingleWildcard` pins it (class scores like `?`, member count doesn't move the score).
- **#314 — duplication + intent: fixed.** The obligation loop is now a single `Engine.collectObligations`, so the two methods can't drift, and the new `audit-mode constraint still hard-denies, not AuditOnly` subtest locks in the deliberate choice that an unwired-directive `ENFORCEMENT_ERROR` stays a hard deny even under an audit-mode constraint. Both engine entry points wire `obligations`/`denyResp` through correctly.
- **#307 — doc precision: fixed.** The `Pattern` godoc now says RE2 (Go `regexp`) and explicitly notes lookaround/backreferences are unsupported and rejected at load.

One trivial leftover noted inline: the scanner's "leading `]` is a literal member" comment doesn't match Go's `path.Match` (which rejects `[]…`), so that branch is dead/defensive. Purely cosmetic — nothing blocks merge.

CI on this head is still in progress but the local equivalents (build, vet, gofmt, license headers, race tests) are green.

---

PR #339, opened 2026-06-14
feat(callcounter): add FullStore for a one-line backend contract check (#296)

### LINKED ISSUE #296: callcounter: implementing Store alone is insufficient for maxCalls/sequenceBlock — RateLimiter and CallHistory not visible from the package

eunolabs:
## What

Closes #296.

Adds `callcounter.FullStore` — an interface embedding `HistoryStore` and `LimiterStore` (which transitively include `Store`) — so a call-counter backend that means to support **every** built-in condition can pin the whole contract with a single compile-time assertion:

```go
var _ callcounter.FullStore = (*MyStore)(nil)
```

An omitted `Peek` or `IncrementIfBelow` then fails the build, instead of surfacing only at runtime as an opaque `CONDITION_FAILED` deny on the first `maxCalls` / `sequenceBlock` request.

## Why

Issue #296 lays out three options of increasing strength:

- **A — godoc.** Already landed (#277–#280): `Store`, `capability.CallCounter`, and the `enforcement.RateLimiter` / `enforcement.CallHistory` mirrors all carry godoc explaining the extra methods.
- **B — discoverable interfaces in the package.** Already landed (#236, PR #255): `callcounter.HistoryStore` (`Peek`) and `callcounter.LimiterStore` (`IncrementIfBelow`) live next to `Store`.
- **C — a single combined interface for one build-time check.** This was the remaining gap: a backend supporting both conditions still needed *two* separate assertions, and missing one left the runtime-deny trap open.

This PR implements option C. The narrower interfaces stay exactly as they are — the engine still type-asserts for only the one capability the condition in front of it needs, so a backend that deliberately supports just one condition implements just that interface. `FullStore` is the convenience contract for the common "support everything" case.

## Changes

- `pkg/callcounter/callcounter.go` — add the `FullStore` interface; cross-reference it from the `Store` godoc.
- `pkg/callcounter/callcounter_test.go` — replace the four per-backend `HistoryStore`/`LimiterStore` pins with a single `FullStore` assertion per built-in backend (it subsumes both, so no coverage is lost and it demonstrates the one-line pattern).
- `pkg/capability/enforce.go` — lead the `CallCounter` godoc's pin guidance with `callcounter.FullStore`, framing the single-capability `enforcement.RateLimiter` / `enforcement.CallHistory` assertions as the deliberate one-condition subset.
- `docs/architecture.md` — note `FullStore` in the custom-backend extension-point row.
- `CHANGELOG.md` — `[Unreleased]` entry as the capstone to the #236 / #280 entries.

## Testing

- `go build ./...`, `go vet ./...`, `gofmt` clean.
- `go test -race -count=1 ./pkg/callcounter/... ./pkg/capability/...` — pass. The new `FullStore` compile-time assertions hold (the test package compiles), and built-in `InMemory`/`Redis` behavior is unchanged.

No behavior change: this is an additive, pre-1.0 API/discoverability improvement. No new runtime dependencies.

https://claude.ai/code/session_01Cz2vKFnWJmVqdsdaw2CHNb


pkg/capability/enforce.go
```diff
@@ -131,8 +131,10 @@ type Enforcer interface {
 // #280). A custom backend that means to support those conditions should therefore
 // implement the extended callcounter.LimiterStore / callcounter.HistoryStore
 // contracts and pin the relationship with a compile-time assertion (e.g.
-// `var _ enforcement.RateLimiter = (*MyCounter)(nil)`), turning an omitted method
-// into a build error rather than a silent runtime degradation. Both built-in
+// `var _ enforcement.RateLimiter = (*MyCounter)(nil)`, or — for a backend that
+// backs both conditions — `var _ callcounter.FullStore = (*MyCounter)(nil)`,
+// which embeds both in one check), turning an omitted method into a build error
```
eunolabs: Optional (wording): the first example here — `var _ enforcement.RateLimiter = (*MyCounter)(nil)` — only guards the `maxCalls` capability (`IncrementIfBelow`). A backend author who *also* wants `sequenceBlock` and copies just this example still silently misses `Peek` — which is precisely the opaque runtime-deny trap #296 is closing. Since `FullStore` is the robust "support everything" guard, consider leading with it and framing the single-interface assertions (`enforcement.RateLimiter` / `enforcement.CallHistory`) as the deliberate one-condition subset. Non-blocking.

---

eunolabs (commented):
## Review: `feat(callcounter): add FullStore` (#339)

Clean, additive, no behavior change — and the PR's core claims hold up against the code:

- **The diamond embedding is valid.** `FullStore` embeds both `HistoryStore` and `LimiterStore`, which each embed `Store`; Go (repo is on 1.25) permits the overlapping `IncrementAndGet` method set, and CI's build/vet/test are green.
- **No test coverage is lost.** `FullStore` is the union of `HistoryStore` + `LimiterStore`, so the two new `var _ callcounter.FullStore = …` assertions still require `Peek`, `IncrementIfBelow`, *and* `IncrementAndGet` on each backend. A dropped method still breaks the build and the compiler names the missing one, so the only thing given up vs. the prior four assertions is per-interface localization — immaterial.
- **The narrow interfaces remain what the engine actually requires.** Confirmed the engine type-asserts only `RateLimiter`/`CallHistory` (`pkg/enforcement/handlers.go:317,1035`), and the structural identity between the `callcounter` and `enforcement` interface pairs is already pinned in `pkg/enforcement/enforcement_test.go:35-36` — so `FullStore` is transitively tied to the consumer-side mirrors without needing a new test.
- The doc updates (`docs/architecture.md` and the three godocs) are accurate.

### Should address (non-blocking)

1. **No `CHANGELOG.md` entry.** The curated `[Unreleased]` section already documents the two predecessors of this exact effort — the `HistoryStore`/`LimiterStore` addition (#236) and the consumer-side `enforcement.RateLimiter`/`CallHistory` mirrors (#280) — so `FullStore` (#296) is the natural capstone, and omitting it is inconsistent with that precedent. It isn't CI-gated, but worth a one-line **Added** entry, e.g.:
   > `callcounter.FullStore` — a combined `HistoryStore`+`LimiterStore` contract so a backend that means to support every built-in condition can pin the whole thing with one compile-time assertion instead of two (#296).

### Minor

2. The PR description credits **#255** for the `HistoryStore`/`LimiterStore` work, but `CHANGELOG.md` credits **#236** for the same change — worth reconciling the cross-reference.
3. Design note (no change requested): the one-line convenience exists only producer-side (`callcounter.FullStore`); the consumer side (`enforcement.RateLimiter` + `CallHistory`) still takes two assertions, which is why `cmd/eunox-mcp/counter_contract.go` is correctly left untouched. If a one-line consumer-side guard is ever wanted, that's a clean follow-up — not needed here.

Left one inline note on `pkg/capability/enforce.go`. Overall LGTM.

---

eunolabs:
Thanks for the review — addressed in 4942018:

1. **CHANGELOG** — added an `[Unreleased]` entry for `FullStore` immediately after the #236 / #280 entries, as the capstone to that trio.
2. **`enforce.go` wording (inline thread)** — reworked the `CallCounter` pin guidance to lead with `callcounter.FullStore` as the "support everything" guard, and reframed the single-capability assertions as the deliberate one-condition subset — now spelling out *both* (`enforcement.RateLimiter` for `maxCalls`, `enforcement.CallHistory` for `sequenceBlock`) so copying one no longer silently leaves the other condition's method unguarded. Resolved that thread.
3. **#236 vs #255 cross-reference** — reconciled the PR description to credit #236 (the issue, matching the CHANGELOG), keeping #255 as the PR pointer.

The design note about a consumer-side one-line guard is left as a clean follow-up, out of scope here — `cmd/eunox-mcp/counter_contract.go` stays untouched as you noted.

---

PR #340, opened 2026-06-14
sec: reject IdP JWTs that omit the exp claim

eunolabs:
## What

The active JWT validation path (`JWTPDP.ValidateToken`) relied solely on go-jose's `stdClaims.ValidateWithLeeway`, which only checks `exp` **when it is present**. A validly-signed token that omits `exp` therefore passed validation and **never expired** — defeating the bearer-token expiry and kill-switch revocation window.

This reintroduces an explicit `exp`-required guard the dormant `pkg/capability` path already had but the active path lacked. The token is now rejected (fail closed) before the leeway validation if `exp` is absent.

```go
// Require exp explicitly: go-jose only checks Expiry when present, so a
// token that omits exp would never expire.
if stdClaims.Expiry == nil {
    return ctx, fmt.Errorf("token has no exp claim; non-expiring tokens are rejected")
}
```

## Why this scope

This is one of the two follow-up findings selected for hardening after the security review (PR #171). The other selected item — **byte-exact audit HMAC verification** (legitimate records with large `details` integers could fail verification via float64 round-trip) — turned out to be **already fixed on `main`** (the maintainers' #306–#336 work added a `json.Decoder` with `UseNumber()` in `VerifyRecord`, which preserves number text exactly). So this PR contains only the `exp` fix; nothing further was needed for the audit path.

## Testing

- New `TestJWTPDP_ValidateToken_NoExpRejected`: a validly-signed token with no `exp` is rejected with a message naming the missing claim.
- Existing JWT validation tests (valid/expired/issuer/audience/version) still pass; tokens used across the suite and the demo IdP all carry `exp`, so no legitimate flow is affected.
- Full `go test -race ./...`, `gofmt`, `go vet`, and `golangci-lint` are clean.

## Docs

Threat model §2.2 (verification steps) and §3.1 (capability-claim-forgery mitigations) updated to state that `exp` is required and an `exp`-less token is rejected.

https://claude.ai/code/session_0178SuBF8wjhMAUid5qm6BDi

---
_Generated by [Claude Code](https://claude.ai/code/session_0178SuBF8wjhMAUid5qm6BDi)_


cmd/eunox-mcp/pdp_jwt.go
```diff
@@ -494,6 +494,13 @@ func (p *JWTPDP) ValidateToken(ctx context.Context, authHeader string) (context.
 		}
 
 		// Standard claim validation.
+		// Require exp explicitly: go-jose only checks Expiry when it is present, so
+		// a token that omits exp would never expire — defeating the bearer-token
+		// expiry and kill-switch revocation window. Reject it (fail closed) before
+		// the leeway validation rather than silently accept a non-expiring token.
+		if stdClaims.Expiry == nil {
```
eunolabs, line 501: Core fix is correct. `ValidateWithLeeway` only enforces a time claim when its pointer is non-nil, so an `exp`-less but validly-signed token would otherwise pass and never expire. Two things that make this placement right:

- It runs **after** `tok.Claims` (signature verified) and `return`s rather than `continue`-ing — matching the issuer/version checks immediately below, so a verified token with bad standard claims is rejected outright rather than retried against other keys.
- It brings the active path in line with the dormant `pkg/capability/jwks.go:158`, which already rejected `exp`-less tokens with the same rationale.

Non-blocking parity note: that dormant path also requires `iat` (`jwks.go:155`) and a non-empty `sub` (`jwks.go:161`); this active path requires neither. Neither omission creates a *non-expiring* token, so they're correctly out of scope for this `sec:` fix — flagging only in case the goal is to fully converge the two validators in a follow-up.

---

cmd/eunox-mcp/pdp_jwt_test.go
```diff
@@ -540,6 +540,56 @@ func TestJWTPDP_ValidateToken_ExpiredToken(t *testing.T) {
 	}
 }
 
+// makeIDPTokenNoExp signs an otherwise-valid IdP JWT with no exp claim, to
+// exercise the non-expiring-token rejection path.
+func makeIDPTokenNoExp(t *testing.T, key testKey, caps []string, iss, aud, sub string) string {
```
eunolabs, line 545: Minor (non-blocking): this duplicates almost the entire signer/serialize body of `makeIDPTokenVersion` (line 75) just to drop the `Expiry` field. Reusing it would mean changing that helper's `exp time.Time` parameter to a `*time.Time` (or adding a `withExp bool`) and threading it through every caller — arguably more churn than this focused helper, so it's fine as-is. Flagging only in case you'd prefer a single parameterized builder over a near-copy.

---

eunolabs (commented):
## Review: reject IdP JWTs that omit `exp`

**Solid, tightly-scoped security fix — correct and fails closed. Two non-blocking suggestions below.**

### Why it's right
`JWTPDP.ValidateToken` previously delegated `exp` enforcement entirely to go-jose's `ValidateWithLeeway`, which only checks `exp` *when the claim is present*. A validly-signed token with no `exp` therefore passed and never expired, defeating both the bearer-token expiry window and the kill-switch revocation window. The new `stdClaims.Expiry == nil` guard closes that.

- **Fail-closed** — returns an error; the HTTP caller (`http.go:490`) records a `deny` / `JWT_INVALID` audit entry and returns 401 without ever calling upstream.
- **Placement** — runs after signature verification and `return`s (not `continue`), consistent with the issuer/version checks below it.
- **Consistency** — matches the dormant `pkg/capability/jwks.go:158` guard that already rejected `exp`-less tokens.
- **Tests** — `TestJWTPDP_ValidateToken_NoExpRejected` is meaningful: it isolates the `exp` path (issuer/audience empty, but the new guard fires first). Existing valid/expired/issuer/audience tests all carry `exp`, so no legitimate flow regresses.
- **Docs** — threat-model §2.2 and §3.1 both updated to state `exp` is required.
- **Conventions** — `sec:` prefix, no new deps, no audit-shape change.
- **CI** — green at review time (Test, Lint, gofmt, License, demo/JWT mode all pass; only `Analyze (go)` still running, CodeQL `neutral`).

### Non-blocking suggestions
1. **CHANGELOG `[Unreleased] > Security`** — comparable mitigations (call-counter fail-open #277, counter-key collisions #273) are logged there; this one arguably warrants an entry too. The PR currently updates only the threat model.
2. **Validator parity (follow-up)** — see inline note: the dormant `pkg/capability/jwks.go` path also requires `iat` and `sub`, which the active path doesn't. Out of scope for this fix, but worth a follow-up if you want the two validators fully converged.
3. **Minor test duplication** — see inline note on `makeIDPTokenNoExp`.

---

cmd/eunox-mcp/pdp_jwt.go
eunolabs: Agreed on all points, thanks. Keeping the `iat`/`sub` parity out of this PR deliberately — neither creates a non-expiring token, and requiring `sub`/`iat` has IdP-compatibility implications (e.g. some client-credentials tokens) that deserve their own change rather than riding along on this focused `exp` fix. Happy to open a follow-up that fully converges the active path with `pkg/capability/jwks.go` (require `iat` + non-empty `sub`) if you'd like that hardening — just say the word.

---

cmd/eunox-mcp/pdp_jwt_test.go
eunolabs: Agreed — keeping the focused `makeIDPTokenNoExp` helper as-is per your read: parameterizing `makeIDPTokenVersion` (to `*time.Time` or a `withExp` flag) would thread churn through every caller for no real gain. Left unchanged.

---

PR #341, opened 2026-06-14
Fix session rev-20260614T0324Z-584163 bugs

### LINKED ISSUE #337: [rev-20260614T0324Z-584163] IncrementIfBelow inserts empty map entry before limit<1 denial, consuming a maxKeys slot
**Location:** `pkg/callcounter/memory.go:292-315`

**Reference:** https://github.com/eunolabs/eunox/blob/c3affce/pkg/callcounter/memory.go#L292-L315

**Severity:** Medium

**Description:**

`IncrementIfBelow` inserts a new `entry` into `m.entries` (lines 300-301) **before** it checks `limit < 1` (line 315). When `limit < 1` and the key has not been seen before, the code path is:

1. `admitNewKey()` runs and counts the new key against the `maxKeys` budget (line 294).
2. An empty `entry` with `timestamps: nil` is allocated and stored: `m.entries[sk] = e` (line 301).
3. The filter loop produces an empty `valid` slice, so `cur = 0`.
4. `limit < 1` is true → the deny branch is taken and `(0, false, 0, nil)` is returned.
5. The empty entry is now in the map, holding a `maxKeys` slot, until the next `Cleanup` cycle removes it (as a zero-timestamp entry, which `shouldDeleteEntry` immediately flags).

The Redis backend's Lua script (`incrIfBelowScript`) does **not** have this problem: `ZADD` and `EXPIRE` are only executed if the call is admitted, so no state is created for a denied call.

**Trigger / repro:**

```go
m := callcounter.NewInMemory(callcounter.WithMaxKeys(1))
ctx := context.Background()

// Fill the one available slot with a limit<1 call for a new key.
// admitNewKey() accepts it, an empty entry is inserted, then denied.
_, admitted, _, err := m.IncrementIfBelow(ctx, "key-A", 60, 0)
// admitted=false, err=nil — but "key-A" now occupies the sole maxKeys slot.

// A completely legitimate call for a different new key now fails at admitNewKey().
_, admitted, _, err = m.IncrementIfBelow(ctx, "key-B", 60, 5)
// err != nil — denied because the slot is taken by the phantom entry.
```

Without `WithMaxKeys`, the phantom entry is merely wasteful (occupies heap until `Cleanup`). With `WithMaxKeys`, it can permanently deny legitimate new keys until `Cleanup` reclaims the slot.

**Suggested fix:**

Check `limit < 1` **before** acquiring the lock and before any map mutation, matching how `IncrementAndGet` checks `maxEntries` upfront:

```go
func (m *InMemory) IncrementIfBelow(ctx context.Context, key string, windowSec int, limit int64) (count int64, admitted bool, retryAfter time.Duration, err error) {
        return 0, false, 0, err
    }
    // Fail closed before touching any map state: a limit<1 can never admit,
    // so no entry should be created or counted against maxKeys.
    if limit < 1 {
        return 0, false, 0, nil
    }

    m.mu.Lock()
    defer m.mu.Unlock()
    // ... rest unchanged
```

This makes the invariant explicit, matches the Redis backend's behavior (no state written on denial), and prevents a `limit<1` caller from burning a `maxKeys` slot.

**Session:** rev-20260614T0324Z-584163

### LINKED ISSUE #338: [rev-20260614T0324Z-584163] Redis IncrementAndGet silently returns 0 on per-command pipeline error, potentially fail-open
**Location:** `pkg/callcounter/redis.go:130-135`

**Reference:** https://github.com/eunolabs/eunox/blob/c3affce/pkg/callcounter/redis.go#L130-L135

**Severity:** Medium

**Description:**

`IncrementAndGet` executes a `TxPipeline` (MULTI/EXEC) and correctly checks the top-level `pipe.Exec` error for connection-level failures. However, it then calls `countCmd.Val()` without checking `countCmd.Err()`:

```go
_, err := pipe.Exec(ctx)
if err != nil {
    return 0, fmt.Errorf("redis pipeline: %w", err)
}
return countCmd.Val(), nil   // countCmd.Err() is never checked
```

In go-redis v9, `TxPipeline.Exec` returns a **non-nil top-level error** only for connection errors and `EXECABORT`. **Runtime command errors** that occur inside an `EXEC` block (e.g., `WRONGTYPE` because a non-sorted-set key exists at `windowKey`) are reported per-command: `pipe.Exec` returns `nil` but the individual `redis.Cmder`'s `.Err()` is non-nil and `.Val()` returns the zero value.

When this happens for the `ZCard` command, `countCmd.Val()` returns `0` and the function returns `(0, nil)` — silently reporting zero in-window calls. Every downstream check then sees a counter at zero:

- The `sequenceBlock` condition, which uses `IncrementAndGet` with `maxEntries=1`, interprets a `0` return as "this tool has not run yet" and allows a tool that should be blocked.
- Any caller that interprets a `0` count as under-limit will admit the call.

**Trigger / repro:**

1. Point eunox-mcp at a Redis instance where `callcounter:<key>:<windowSec>` happens to be a String or Hash (e.g., from a namespace collision with another application or a manual `SET` command).
2. Call a tool that the `sequenceBlock` or `maxCalls` condition is supposed to gate.
3. `ZCard` returns a `WRONGTYPE Operation against a key holding the wrong kind of value` error; `pipe.Exec` top-level error is nil; `countCmd.Val()` returns 0; the condition passes as if no calls have been made.

**Suggested fix:**

Check the individual command's error after `Exec`:

```go
_, err := pipe.Exec(ctx)
if err != nil {
    return 0, fmt.Errorf("redis pipeline: %w", err)
}
if err := countCmd.Err(); err != nil {
    return 0, fmt.Errorf("redis pipeline ZCard: %w", err)
}
return countCmd.Val(), nil
```

Similarly audit `IncrementAndGet`'s sibling methods (`IncrementIfBelow` uses a Lua script so `Run` error handling is sufficient there; `Peek` uses a direct `ZCount` call which is fine). A broader fix would check each queued command's `.Err()` after `Exec`, but checking `countCmd` is the minimum needed to fail closed on the return value used for enforcement.

**Session:** rev-20260614T0324Z-584163

eunolabs:
## Bugs addressed (session rev-20260614T0324Z-584163)

- **Closes #337** — `InMemory.IncrementIfBelow` inserted a new entry (counted against `WithMaxKeys` via `admitNewKey`) **before** checking `limit<1`, so a `limit<1` call on a previously-unseen key created a phantom entry that burned a `maxKeys` slot until the next `Cleanup`, permanently denying legitimate new keys. The `limit<1` fail-closed check now runs **before** the lock and any map mutation — matching `IncrementAndGet`'s upfront `maxEntries` check and the Redis backend, whose Lua script writes no state on a denied call. (This made the deny branch's own `limit<1` handling unreachable, so it was simplified to `cur >= limit` with no behavior change.) **Confirmed genuine:** the regression test fails without the fix (`entries = 1, want 0`).

- **Closes #338** — `Redis.IncrementAndGet` returned `countCmd.Val()` after checking only the top-level `pipe.Exec` error. **Self-review correction:** the issue's premise — that `pipe.Exec` returns `nil` while the per-command `ZCard` error is non-nil — does **not** hold for go-redis v9. `(*Pipeline).Exec` returns `cmdsFirstErr`, the first non-nil error across **all** queued commands (`ZCard` included), so a `WRONGTYPE` (or any per-command) error inside `EXEC` is already surfaced by the existing top-level check and the original code already failed closed. I kept the requested `countCmd.Err()` check as honest **defence in depth** — it localizes the fail-closed guarantee to the specific command whose `Val()` is returned, rather than relying on that driver-internal aggregation — and reworded the code comment, the regression test, and the CHANGELOG so none of them claim a fail-open that does not exist. Verified empirically (probe + go-redis source `cmdsFirstErr`).

Both changes preserve the repo's **fail-closed** invariant and bring the in-memory and Redis backends into structural step (no state written on denial; per-command error surfaced).

## Testing

- New regression tests:
  - `TestInMemory_IncrementIfBelow_LimitBelowOneWritesNoState` (`pkg/callcounter/maxkeys_internal_test.go`) — a `limit<1` call on a new key writes no entry (`numEntries == 0`) and leaves the sole `maxKeys` slot free. **Fails without the #337 fix.**
  - `TestRedis_IncrementAndGet_WrongType_ReturnsError` (`pkg/callcounter/redis_failure_test.go`) — a `WRONGTYPE` collision at the window key makes `IncrementAndGet` return an error rather than a silent `0` (end-to-end fail-closed guard).
- Audited the sibling Redis methods called out in #338: `IncrementIfBelow` (Lua `Run().Result()`) and `Peek` (`ZCount().Result()`) already check their errors and fail closed — no change needed.
- `gofmt -l` clean, `go vet ./...` clean, Apache-2.0 SPDX headers present on all changed files.
- CHANGELOG.md updated with curated Security entries for both fixes.


eunolabs:
✅ Ready to merge — all session `rev-20260614T0324Z-584163` bugs addressed (#337, #338), self-review complete with no outstanding comments, tests passing.

Self-review summary (3 rounds):
- Round 1 — caught that the #338 entry overclaimed: go-redis `(*Pipeline).Exec` returns `cmdsFirstErr` (first error across all queued commands, `ZCard` included), so the original code already failed closed. Kept the `countCmd.Err()` check as honest defence-in-depth and reworded the comment, test, and CHANGELOG to match reality.
- Round 2 — removed now-dead `limit<1` handling in `IncrementIfBelow`'s deny branch that the new upfront fail-closed return made unreachable (no behavior change).
- Round 3 — audited the sibling Redis methods (`IncrementIfBelow`, `Peek`); both already check their command errors. No new findings.

Local verification: `go test -race -count=1 ./...` green across all packages; `gofmt -l` and `go vet ./...` clean; SPDX headers present. CI is green except two long-running jobs (CodeQL "Analyze (go)" and "Enforcement compliance") still finishing; merging once they complete.

---

PR #351, opened 2026-06-14
feat: operational hardening for long-lived deployments

eunolabs:
## Summary

Adds configurable bounds for long-lived HTTP proxy deployments: concurrent-session caps, idle-session reaping, audit-log retention, and a deny-by-default mode. Also introduces nested-argument path selectors (`$.a.b`) for value-matching policy conditions, and loopback-only `/healthz` and `/metrics` endpoints for operational visibility.

## Why

The HTTP transport previously left several resource limits open:
- **Unbounded sessions**: each client session spawns an upstream subprocess or connection; a misbehaving or long-lived client could exhaust the proxy's resources.
- **Idle sessions**: a client that connects but goes silent pins its upstream indefinitely.
- **Audit-log growth**: rotated audit files accumulate without bound, eventually filling the disk.
- **Policyless routes**: the default fail-open posture allows routes with no policy, which is unsafe in production; operators need a way to enforce "deny by default."
- **No operational visibility**: operators had no real-time view of proxy health (session count, audit loss, kill-switch status) without parsing logs after the fact.

Additionally, policy conditions could only match flat top-level arguments; tools that nested their data required delegating to an external policy engine.

## Type of change

- New feature (CLI flags, config keys, operational endpoints)
- Security hardening

## How it was tested

```
make test    # All new tests pass; existing tests updated for new BuildRoutes signature
make lint    # No new findings
```

New test coverage:
- `TestResolveUpstreamTimeout` — timeout resolution logic
- `TestRequirePolicy_BuildRoutes` — deny-by-default and requirePolicy enforcement
- `TestConfigValidation_OperationalFields` — validation of new listen/audit fields
- `TestLoadConfig_OperationalKeys` — YAML parsing of new config keys
- `TestAuditPruneRotated` — rotated audit-file retention and pruning
- `TestSessionCap_*` — session cap enforcement and slot freeing
- `TestIdleReaper_*` — idle-session detection and reaping
- `TestNestedArgumentPath_AllowedValues` — nested-path argument resolution in conditions
- `TestArgumentPathSegments`, `TestArgumentRootKey`, `TestIsArgumentPath` — path parsing utilities

## Security considerations

**Fail-closed preserved:** 
- Policyless routes are denied when `requirePolicy` or `--deny-by-default` is set; audit mode is exempt.
- Nested-argument path traversal fails closed: a malformed path, missing key, or non-object segment denies (same as a missing flat argument).
- Session cap refuses new sessions with 503 rather than spawning unbounded upstreams.
- Idle reaper closes sessions, freeing resources; no data loss.

**Audit tape:**
- Audit retention (`retainRotated`) only deletes rotated siblings matching the timestamp suffix; the active log and unrelated files are never touched.
- No changes to record fields, signing, or rotation logic.

**Secrets:**
- No new secret handling; no env-var expansion or token logging.

**Authority:**
- Nested-argument paths are opt-in (`$.` prefix required); a literal top-level key with a dot keeps its exact-match meaning — the manifest never silently matches an alternative argument.
- `--deny-by-default` and `requirePolicy` only restrict (deny policyless routes); they cannot widen what a manifest permits.

## Checklist

- [x] Docs updated (`README.md`, `docs/capability-manifest-guide.md`, `CHANGELOG.md`)
- [x] Schema updated (`schemas/eunox-gateway-config.schema.json`) with matching docs
- [x] PR description explains *why* (resource exhaustion, audit growth, operational visibility, policy safety)

https://claude.ai/code/session_01KmmM4tW9rUBkpCJCUoP679


cmd/eunox-mcp/audit.go
```diff
@@ -26,6 +28,11 @@ import (
 	"github.com/eunolabs/eunox/pkg/capability"
 )
 
+// rotatedAuditRe matches the suffix rotate() appends to a rotated audit file:
+// a leading "." then the UTC timestamp layout "20060102T150405Z". Used by
+// pruneRotated to retain only genuine rotated siblings of the active log.
+var rotatedAuditRe = regexp.MustCompile(`^\.\d{8}T\d{6}Z$`)
```
eunolabs: **Blocking — audit retention is a no-op in production (format mismatch).**

This regex only matches *second*-precision suffixes (`.YYYYMMDDTHHMMSSZ`), but `rotatedPath()` emits *nanosecond*-precision names via `time.Now().UTC().Format("20060102T150405.000000000Z")` → e.g. `audit.jsonl.20260614T051842.123456789Z`, plus an optional `.N` uniquifier from `uniqueRotatedPath`. The literal `.` and 9 fractional digits before `Z` mean `pruneRotated` matches **zero** real rotated files, so `len(rotated)` is always 0, nothing is ever deleted, and the log directory still grows without bound — the exact problem `retainRotated` is advertised to fix.

`TestAuditPruneRotated` passes only because it fabricates second-precision filenames (`.20260101T000000Z`) that the rotater never actually produces, so CI gives false confidence here.

Suggested fix — match the real producer format (and the `.N` backstop):
```go
var rotatedAuditRe = regexp.MustCompile(`^\.\d{8}T\d{6}\.\d{9}Z(\.\d+)?$`)
```
and harden the test so it can't drift from the producer again — build the rotated names via `rotatedPath()` (or drive a real `rotate()` loop) rather than hand-writing the suffix. (The doc comments on this var and on `pruneRotated` also state the old second-precision format and should be updated.)

---

cmd/eunox-mcp/http.go
```diff
@@ -1212,9 +1303,15 @@ func (p *HTTPProxy) newSession(ctx context.Context, route *upstreamRoute) (*http
...
-	p.mu.Lock()
-	p.sessions[sess.id] = sess
-	p.mu.Unlock()
+	//
+	// registerSession also enforces the maxSessions cap. On a race that pushes us
+	// over the cap here, kill the subprocess we just spawned (readUpstream is not
+	// yet running, so sess.done would never close — kill directly rather than via
+	// sess.close) and surface the limit error.
+	if err := p.registerSession(sess); err != nil {
+		_ = cmd.Process.Kill()
```
eunolabs, line 1315: **Zombie subprocess leak on the session-cap race.**

This `Kill()` has no matching `Wait()`, so the killed subprocess is never reaped. Contrast the `initUpstream` failure path a few lines up, which deliberately calls `cmd.Wait()` "so a failed initialize never leaves a zombie." Here the per-session cleanup goroutine that would otherwise `Wait` is started only *after* `registerSession` succeeds, so on this path nothing reaps the process.

This path is reached exactly on the concurrent-`initialize` race the cap is designed to handle (the cheap `atSessionCap()` pre-check passed, then the authoritative check under lock rejects). A client flooding `initialize` against a full proxy could accumulate defunct processes — an unbounded resource leak inside the feature meant to bound resources. `initUpstream` has already joined its handshake goroutine, so `Wait` is safe to call here:

```go
if err := p.registerSession(sess); err != nil {
    _ = cmd.Process.Kill()
    _ = cmd.Wait()
    return nil, err
}
```

---

cmd/eunox-mcp/manifest.go
```diff
@@ -705,6 +712,43 @@ func validateSequenceBlock(i, j int, v *capability.SequenceBlockCondition) error
...
+		return nil
+	}
+	for claimName, patterns := range principal {
+		if !capability.SupportedPrincipalClaims[claimName] {
+			return fmt.Errorf("principal claim %q is not supported — valid claims are %s", claimName, strings.Join(sortedSupportedPrincipalClaims(), ", "))
+		}
+		if len(patterns) == 0 {
+			return fmt.Errorf("principal claim %q must list at least one allowed value", claimName)
+		}
+		for _, p := range patterns {
```
eunolabs, line 732: **Nit (defense-in-depth): validate glob well-formedness here too.**

`validatePrincipal` rejects unknown claim names and empty values precisely so a mistake is caught at load rather than silently never matching at runtime. But a *malformed glob* slips through: `matchesAnyPattern` swallows `path.Match`'s `ErrBadPattern` (`err == nil && matched`), so a pattern like `agent-[` becomes a value that can never match. If it's the only pattern for a claim, the principal-scoped constraint silently never applies → fail-closed deny with no diagnostic — the same footgun this function already guards against for claim names.

A cheap check closes it:
```go
if _, err := path.Match(p, ""); err != nil {
    return fmt.Errorf("principal claim %q has an invalid pattern %q: %w", claimName, p, err)
}
```

---

eunolabs (commented):
## Review: operational hardening + principal scoping + nested-argument paths

Thorough, well-documented PR — the fail-closed posture is preserved across all three features (nested-path traversal denies on missing/non-object/malformed; principal mismatch is skipped like a target mismatch; policyless routes deny under `requirePolicy`). The session-cap design (cheap pre-spawn `atSessionCap()` + authoritative re-check under lock in `registerSession`) is the right pattern, the idle reaper correctly spares sessions holding an open SSE stream, the reaper's `p.mu → notifMu` lock ordering is safe (no inversion), and list filtering now hides principal-scoped entries from non-matching identities. CHANGELOG/README/guide are all updated, and the spec-repo follow-up is honestly flagged in the changelog.

Two issues worth fixing before merge, plus some smaller notes (left inline).

### 🔴 Blocking
- **Audit retention (`retainRotated`) never deletes anything in production.** `pruneRotated`'s regex matches second-precision suffixes, but `rotatedPath()` writes nanosecond-precision names (`…T150405.000000000Z`, optional `.N`). Real rotated files never match → the disk-growth bound the feature promises doesn't take effect. The test passes only because it fabricates the wrong filename format. Details + suggested regex/test fix inline on `audit.go`.

### 🟠 Should fix
- **Zombie subprocess leak on the session-cap race** (`http.go`): the `registerSession`-failure path `Kill()`s without `Wait()`, unlike the `initUpstream` path right above it, and the reaping cleanup goroutine isn't started yet. A client flooding `initialize` at a full proxy can pile up defunct processes. One-line fix inline.

### 🟡 Notes / nits
- **Principal glob patterns aren't validated at load** (`manifest.go`): a malformed glob silently never matches (fail-closed deny, no diagnostic). Inline suggestion to validate with `path.Match`.
- **`path.Match` semantics for `iss`/`sub`:** `*` does not cross `/`, so a pattern like `https://idp.example.com/*` won't match a slash-bearing URL. Since `iss`/`sub` are commonly URLs, a one-line caveat in the manifest guide would save operators a surprise.
- **Tie-break: specificity dominates principal.** A broad per-principal `deny` (e.g. `tool:*` for `agent_id:X`) is overridden by a more-specific *general* `allow` (`tool:foo`). This is the documented design, but it's a sharp edge — operators may expect a per-principal deny to win everywhere. Worth an explicit callout in the guide/threat-model.
- **`--upstream-timeout` default change:** unset now means **30s** instead of "no timeout." Documented, but it's a behavior change that can bite deployments with legitimately long-running tools (opt out with `--upstream-timeout 0`).
- **Threat model not updated:** audit retention introduces *deletion* of signed rotated evidence (an availability/integrity surface). Once the retention bug above is fixed, a short note in `docs/threat-model-mcp.md` would be appropriate.
- **Spec-repo sync** (`eunolabs/mcp-capability-manifest`) for the `principal` field and `$.` path is still pending — the changelog already acknowledges this; flagging so it isn't lost.

CI is green (the two in-progress checks — CodeQL `Analyze (go)` and `Enforcement compliance` — were still running at review time, not failing). Nice work overall; the two top items are the gates.

---

eunolabs (commented):
## Re-review (second pass, after the review fixes)

Re-reviewed at `2112c1d`. The three items from the first review are addressed, and a deeper pass over the concurrency turned up nothing new.

### First-review items — resolved
- **🔴 Audit retention no-op** — `rotatedAuditRe` now matches the real nanosecond format (`^\.\d{8}T\d{6}\.\d{9}Z(\.\d+)?$`), and `TestRotatedAuditReMatchesProducer` ties the regex to `rotatedPath()` so they can't drift again; `TestAuditPruneRotated` was rebuilt on the real format with a `.N` and a legacy-format negative case.
- **🟠 Zombie on session-cap race** — `registerSession`-failure path now `cmd.Wait()`s after `Kill()`.
- **🟡 Principal glob validation** — `validatePrincipal` now rejects malformed globs at load (`path.Match` `ErrBadPattern`), with table coverage.

### Second-pass observations (no blockers)
- **Concurrency is sound.** `httpSession.close` is guarded by `sync.Once`, so the idle reaper, `closeAllSessions` (shutdown), and the per-session cleanup goroutine can all race to close the same session without a double-close panic. The reaper's `p.mu → notifMu` order has no inversion, and `lastActive` is atomic — the reaper reads it without the proxy lock.
- **Engine/PDP tie-break is consistent.** `Engine.findMatchingCapability` and `ManifestPDP.findConstraint` implement the same "specificity first, then principal at equal specificity, order-independent" rule, and both skip principal-scoped entries on nil claims (fail closed). List filtering hides principal-scoped targets from non-matching identities.
- **Minor (optional):** `reapOnce` closes reaped *local* sessions serially, each blocking up to `shutdownMs` waiting on `done`. It runs outside `p.mu` so it never blocks request handling, but a sweep that reaps many stale local sessions at once could take a while. Closing them concurrently (or with a short bounded wait) would tighten reap latency. Not worth holding the PR.

### Follow-ups from the review notes — done
- Guide updated with the `path.Match` URL caveat for `iss`/`sub` and the specificity-dominates-principal edge; threat-model (§3.4) now documents the retention deletion trade-off. (Both in this PR, commit `2112c1d`.)
- Spec-repo sync is no longer pending: `eunolabs/mcp-capability-manifest` has a branch (`claude/confident-turing-6no7il`) adding the `principal` field (SPEC §3.2.3 + JSON schema, schema-validated) and the `$.` selector (SPEC §3.3) to the 0.1 grammar.

CI is green on the latest commit (only the long-running CodeQL `Analyze (go)` job was still in progress at review time). LGTM once the optional reaper-latency nit is considered.

---

PR #352, opened 2026-06-14
Fix automated-review session bugs (rev-20260614T0404Z-d4f109, rev-20260614T0405Z-879400)

### LINKED ISSUE #344: [rev-20260614T0405Z-879400] audit-verify chain check silently disabled when --request-id or --since filter is active
**Location:** `cmd/eunox-mcp/audit.go:757-808`
**Reference:** https://github.com/eunolabs/eunox/blob/e3cbf867db3c524b824c5c50b42c5fe03abc7f6e/cmd/eunox-mcp/audit.go#L757-L808
**Severity:** High

**Description:**

`verifyAuditLog` sets `chainCheck = requestID == "" && since.IsZero()`. This means the tamper-evident chain is **not verified at all** when any filter is active. The comment directly above the guarded block contradicts this:

```go
// Chain verification runs over EVERY record (independent of filters) so a
// deleted or reordered record is caught regardless of --request-id/--since.
if res.chainCheck {          // ← only true when NO filter is active
    ...
    havePrev = true
    prevHMAC = rec.HMAC
    prevSeq  = rec.Seq
}
```

Because `havePrev`/`prevHMAC`/`prevSeq` are never updated when a filter is active, the chain break comparisons (`rec.PrevHMAC != prevHMAC`, `rec.Seq != prevSeq+1`) are never executed. A deletion of any record goes completely undetected when the caller passes a filter.

**Trigger / repro:**

```bash
# Write 5 records; delete record 3 from the log; then verify with a filter:
eunox-mcp audit-verify --request-id req-5   # no chain break reported — deletion hidden
eunox-mcp audit-verify                       # chain break correctly reported
```

An adversary who has tampered with the log only needs to ensure the forensic investigator uses a `--request-id` or `--since` filter (both common in incident response) to avoid detection entirely.

**Suggested fix:**

Decouple the chain-state update from `res.chainCheck`. Always advance `havePrev`/`prevHMAC`/`prevSeq` for every record, and always check chain breaks when `havePrev` is true. Apply the filter **only** to the per-record HMAC reporting (`valid`/`invalid`/`skipped` counts). The `chainCheck` bool in `auditVerifyResult` can still gate whether chain breaks are surfaced in the return value, but the chain state must be maintained unconditionally:

```go
// always update chain state
if havePrev {
    if rec.PrevHMAC != prevHMAC { res.chainBreaks++; fmt.Fprintf(...) }
    if rec.Seq != prevSeq+1     { res.chainBreaks++; fmt.Fprintf(...) }
} else {
    res.firstSeq = rec.Seq
}
havePrev = true
prevHMAC = rec.HMAC
prevSeq  = rec.Seq

// filter applies only to HMAC checking
if requestID != "" && rec.RequestID != requestID { res.skipped++; continue }
...
```

**Session:** rev-20260614T0405Z-879400

### LINKED ISSUE #345: [rev-20260614T0405Z-879400] audit log rotation skips fsync, risking record loss on power failure
**Location:** `cmd/eunox-mcp/audit.go:854`
**Reference:** https://github.com/eunolabs/eunox/blob/e3cbf867db3c524b824c5c50b42c5fe03abc7f6e/cmd/eunox-mcp/audit.go#L847-L856
**Severity:** Medium

**Description:**

`rotate()` closes the old file descriptor without first syncing it to disk:

```go
_ = s.f.Close()   // line 854 — no Sync() before this
s.f = f
s.written = 0
```

By contrast, `Close()` correctly syncs before closing:

```go
    closeErr = err
    return
}
closeErr = s.f.Close()
```

`os.File.Close` on Linux does **not** guarantee that all dirty page-cache pages are flushed to storage; only `fsync(2)` (via `f.Sync()`) provides that guarantee. If the host loses power or is forcibly killed between the `os.Rename` that archives the old log and the page-cache flush, the renamed (rotated) file could be missing the last N records written before the rotation threshold was reached.

For an append-only tamper-evident audit log this is a durability violation: records that are reported as written (the drainer advanced `seq` and `prevHMAC`) may not exist on disk after a crash, creating an undetectable gap in the chain (since the HMAC-linked successor record was written to the new file and references a predecessor HMAC that is no longer on disk).

**Trigger / repro:**

Any deployment where the audit log reaches the rotation threshold (`--audit-rotate-size`, default 100 MiB) on a host that subsequently crashes or is power-cycled. The rotated file may be truncated; the surviving chain in the new file references HMACs from the missing tail.

**Suggested fix:**

Add `f.Sync()` before `f.Close()` in `rotate()`, mirroring the pattern in `Close()`:

```go
    fmt.Fprintf(os.Stderr, "[eunox-mcp] audit rotate sync error: %v\n", err)
    // continue: sync failure is not fatal; falling back to writing is worse
}
_ = s.f.Close()
```

**Session:** rev-20260614T0405Z-879400

### LINKED ISSUE #346: [rev-20260614T0405Z-879400] request_id collides across process restarts — audit-verify --request-id returns records from multiple sessions
**Location:** `cmd/eunox-mcp/audit.go:70-74`
**Reference:** https://github.com/eunolabs/eunox/blob/e3cbf867db3c524b824c5c50b42c5fe03abc7f6e/cmd/eunox-mcp/audit.go#L70-L74
**Severity:** Medium

**Description:**

`reqCounter` is a package-level `atomic.Uint64` that starts at zero on every process start:

```go
var reqCounter atomic.Uint64

func nextRequestID() string {
    return "req-" + strconv.FormatUint(reqCounter.Add(1), 16)
}
```

Request IDs are therefore `req-1`, `req-2`, … in every process session. A long-running audit log — which spans multiple process restarts — will contain many records with identical `request_id` values coming from completely different sessions, tools, or callers.

**Impact:**

1. **`audit-verify --request-id req-5`** returns records from every process session that had a fifth request, not just the intended one. A forensic analyst searching for a specific event (e.g., an unauthorized `tools/call`) will receive a polluted result set.
2. **SIEM queries** joining on `request_id` will produce false joins across unrelated sessions.
3. The `RequestID` field is documented as a per-request correlator (it appears in every audit record and is the primary field for audit-verify filtering), so correctness matters for the tool's stated purpose.

**Trigger / repro:**

Restart `eunox-mcp` twice. The first request of each session both have `request_id: "req-1"`. Run:

```bash
eunox-mcp audit-verify --request-id req-1
```

Records from both sessions are returned.

**Suggested fix:**

Generate a per-process-session random nonce at startup and embed it in every request ID:

```go
var (
    sessionNonce = func() string {
        b := make([]byte, 4)
        _, _ = rand.Read(b)
        return hex.EncodeToString(b)
    }()
    reqCounter atomic.Uint64
)

func nextRequestID() string {
    return "req-" + sessionNonce + "-" + strconv.FormatUint(reqCounter.Add(1), 16)
}
```

This keeps the counter cheap (no crypto on the hot path) while making each session's IDs globally unique.

**Session:** rev-20260614T0405Z-879400

### LINKED ISSUE #347: [rev-20260614T0405Z-879400] chain-break switch masks seq gap — only one of two simultaneous violations is counted
**Location:** `cmd/eunox-mcp/audit.go:780-793`
**Reference:** https://github.com/eunolabs/eunox/blob/e3cbf867db3c524b824c5c50b42c5fe03abc7f6e/cmd/eunox-mcp/audit.go#L780-L793
**Severity:** Low

**Description:**

`verifyAuditLog` uses a `switch` to check the two chain-integrity conditions — `prev_hmac` mismatch and sequence gap — as **mutually exclusive** cases:

```go
switch {
case !havePrev:
    res.firstSeq = rec.Seq
case rec.PrevHMAC != prevHMAC:          // fires first
    res.chainBreaks++
    fmt.Fprintf(out, "CHAIN BREAK at seq %d: ...\n", rec.Seq)
case rec.Seq != prevSeq+1:              // never reached when case above is true
    res.chainBreaks++
    fmt.Fprintf(out, "SEQ GAP: record seq %d does not follow %d ...\n", rec.Seq, prevSeq)
}
```

When one or more records are deleted from the interior of the log, **both** conditions are simultaneously true for the first record after the gap:

- `rec.PrevHMAC` references the deleted record's HMAC, which ≠ the last seen `prevHMAC` → case 2 fires
- `rec.Seq` skips the deleted entries → case 3 would fire, but it is never reached

Result: `chainBreaks` is incremented by 1, but the `SEQ GAP` diagnostic (which tells an auditor *how many* records are missing) is never emitted. If an attacker deletes N consecutive records, only a single `CHAIN BREAK` line appears with no indication of the gap's size. The auditor cannot determine the extent of the tampering from the output alone.

**Trigger / repro:**

Write 5 records, delete records 2 and 3, run `audit-verify`:

```
CHAIN BREAK at seq 4: prev_hmac does not match the preceding record
```

No `SEQ GAP` line is printed and `chainBreaks == 1`. The missing-two-record gap is invisible to the output.

**Suggested fix:**

Replace the `switch` with separate `if` statements so both conditions are independently evaluated and counted:

```go
if !havePrev {
    res.firstSeq = rec.Seq
} else {
    if rec.PrevHMAC != prevHMAC {
        res.chainBreaks++
        fmt.Fprintf(out, "CHAIN BREAK at seq %d: prev_hmac does not match ...\n", rec.Seq)
    }
    if rec.Seq != prevSeq+1 {
        res.chainBreaks++
        fmt.Fprintf(out, "SEQ GAP: record seq %d does not follow %d ...\n", rec.Seq, prevSeq)
    }
}
```

This ensures both types of tampering are always detected, counted, and reported independently.

**Session:** rev-20260614T0405Z-879400

### LINKED ISSUE #348: [rev-20260614T0404Z-d4f109] recordFailureDenial omits AuditOnly, causing hard denies in audit-mode constraints on counter backend failure
**Location:** `pkg/enforcement/engine.go:379-391` (function), `pkg/enforcement/engine.go:503-505` and `pkg/enforcement/engine.go:709-711` (call sites)

**Reference:**
- https://github.com/eunolabs/eunox/blob/e3cbf867db3c524b824c5c50b42c5fe03abc7f6e/pkg/enforcement/engine.go#L379-L391
- https://github.com/eunolabs/eunox/blob/e3cbf867db3c524b824c5c50b42c5fe03abc7f6e/pkg/enforcement/engine.go#L503-L505
- https://github.com/eunolabs/eunox/blob/e3cbf867db3c524b824c5c50b42c5fe03abc7f6e/pkg/enforcement/engine.go#L709-L711

**Severity:** High

**Description:**

`recordFailureDenial` constructs the hard-deny response returned when `recordSessionCall` cannot persist a session-history marker. The function does **not** set `AuditOnly` on the response:

```go
func recordFailureDenial(requestID, now string) capability.EnforceResponse {
        RequestID: requestID,
        DecidedAt: now,
        // AuditOnly is absent
        Denial: &capability.DenialInfo{...},
    }
}
```

Both call sites (`ValidateAction` line 503–505 and `EvaluateConditions` line 709–711) have `matched` in scope, which carries the constraint's enforcement mode, but neither passes it to `recordFailureDenial`.

This creates a behavioral asymmetry:

| Failure kind | Audit-mode constraint |
|---|---|
| Condition failure (e.g. `ipRange` deny) | `AuditOnly: true` — logged, call forwarded |
| Counter backend failure (`recordSessionCall` error) | `AuditOnly: false` — hard deny, call **blocked** |

Operators use audit mode (`enforcement: audit`) for staged policy rollout: they want to observe what the policy would deny without actually blocking production traffic. A transient counter backend failure (brief Redis partition, key eviction) during this rollout phase unexpectedly converts the "observe only" posture into a full block for **all** calls to that tool — including calls that would have passed every declared condition — because `recordSessionCall` runs on the allow path after all conditions pass.

The analogous `collectObligations` path documents its deliberate choice not to set `AuditOnly` ("an unwired directive is an engine bug, not a policy verdict"), but `recordFailureDenial` has no such comment, and its situation is different: a transient infrastructure fault is not an engine bug.

**Trigger / repro:**

1. Create a manifest with a `sequenceBlock` or `maxCalls` condition on a constraint marked `enforcement: audit`.
2. Start the proxy with a call-counter backend (Redis).
3. While the proxy is running, cause a transient Redis failure (e.g., `DEBUG SLEEP 5` or a short partition).
4. Send a tool call that passes all conditions.
5. Observe: the call is hard-denied (HTTP 403 / JSON-RPC error) rather than logged-and-forwarded.

In audit mode the caller expects the call to be forwarded with a denial logged; instead it receives a hard error indistinguishable from a real enforcement block.

**Suggested fix:**

Thread the matched constraint's `IsAuditOnly()` into `recordFailureDenial` (or set it at the call site):

```go
// Option A: add a parameter
func recordFailureDenial(requestID, now string, auditOnly bool) capability.EnforceResponse {
        RequestID: requestID,
        AuditOnly: auditOnly,
        DecidedAt: now,
        Denial:    &capability.DenialInfo{...},
    }
}

// call sites:
return recordFailureDenial(requestID, now, matched.IsAuditOnly()), nil

// Option B: inline at call site
if err := e.recordSessionCall(ctx, req); err != nil {
    resp := recordFailureDenial(requestID, now)
    resp.AuditOnly = matched.IsAuditOnly()
    return resp, nil
}
```

If the hard-deny behavior is intentional even in audit mode (consistent with `collectObligations`), that decision should be documented with a comment analogous to the one on `collectObligations` explaining why "fail closed wins over audit never blocks" in this specific case, so it is clearly a choice rather than an oversight.

**Session:** rev-20260614T0404Z-d4f109

### LINKED ISSUE #349: [rev-20260614T0404Z-d4f109] Missing load-time validation for empty allowedOperations.operations, allowedExtensions.extensions, and allowedTables.tables lists
**Location:** `cmd/eunox-mcp/manifest.go:241-324` (`validateLocalManifest` condition switch) and `pkg/enforcement/handlers.go`

**Reference:**
- https://github.com/eunolabs/eunox/blob/e3cbf867db3c524b824c5c50b42c5fe03abc7f6e/cmd/eunox-mcp/manifest.go#L241-L324
- https://github.com/eunolabs/eunox/blob/e3cbf867db3c524b824c5c50b42c5fe03abc7f6e/cmd/eunox-mcp/manifest.go#L433-L458

**Severity:** Medium

**Description:**

The manifest loader enforces an important convention: an allowlist-style condition with an empty allowlist is almost certainly an authoring mistake and is rejected at load time with a clear error. The convention is applied consistently for several condition types:

| Condition | Empty-list field | Load-time check |
|---|---|---|
| `recipientDomain` | `domains` | ✓ rejected ("denies every call") |
| `ipRange` | `cidrs` | ✓ rejected ("matches no source IP") |
| `allowedValues` | `values` | ✓ rejected ("matches nothing and denies every call") |
| `allowedOperations` | `operations` | **✗ not checked** |
| `allowedExtensions` | `extensions` | **✗ not checked** |
| `allowedTables` | `tables` | **✗ not checked** |

The `validateLocalManifest` switch for each of these three condition types checks only that the `argument` field is non-empty, not that the allowlist itself is populated:

```go
// manifest.go — present for allowedOperations:
case capability.AllowedOperationsCondition:
    if v.Argument == "" {
        return fmt.Errorf(...)
    }
    // ← no check for len(v.Operations) == 0
```

At runtime, each handler loops over the allowlist and falls through to deny when it is empty:

```go
// handlers.go — handleAllowedOperations:
for _, allowed := range ao.Operations {      // loop body never executes
    if strings.EqualFold(allowed, operation) || allowed == "*" {
        return nil
    }
}
return &ConditionError{Code: ErrCodeConditionFailed, ...}  // every call denied
```

The result: a manifest with `allowedOperations: {argument: sql, operations: []}` (e.g., from a typo like `operation:` instead of `operations:`) passes load-time validation, starts the proxy, and then silently denies **every** tool call with a `CONDITION_FAILED` that reads like a real policy enforcement event rather than a misconfiguration. The operator only learns of the problem on the first denied live call, with no load-time warning.

The same logic applies to `allowedExtensions` (`extensions: []`) and `allowedTables` (`tables: []`). In the extensions case, the `allowed` slice may also become effectively empty after the normalization step strips whitespace-only entries — e.g., `extensions: ["", "  "]` — but only a fully-empty post-trim list is the concern here.

**Trigger / repro:**

```yaml
capabilities:
  - target: tool:run_query
    actions: [call]
    conditions:
      - type: allowedOperations
        argument: sql
        operations: []   # typo: forgot to list SELECT, INSERT, etc.
```

Load this manifest: `eunox-mcp proxy --policy manifest.yaml`. Expect an error at load ("empty operations list..."); actual behavior: proxy starts silently and every `run_query` call is denied at runtime.

**Suggested fix:**

Add empty-list guards in `validateLocalManifest` (or in the dedicated `validateAllowedOperations`, `validateAllowedExtensions`, `validateAllowedTables` helpers) mirroring the existing `validateAllowedValues` guard:

```go
// for allowedOperations:
if len(v.Operations) == 0 {
    return fmt.Errorf("capability at index %d, condition %d: allowedOperations requires a non-empty 'operations' list; an empty list matches no operation and denies every call", i, j)
}

// for allowedExtensions:
if len(v.Extensions) == 0 {
    return fmt.Errorf("capability at index %d, condition %d: allowedExtensions requires a non-empty 'extensions' list; an empty list matches no file extension and denies every call", i, j)
}

// for allowedTables:
if len(v.Tables) == 0 {
    return fmt.Errorf("capability at index %d, condition %d: allowedTables requires a non-empty 'tables' list; an empty list matches no table and denies every call", i, j)
}
```

These should live in the same case-matched switch where `Argument == ""` is already caught (lines 241–324), and should follow the same phrasing convention as the analogous `validateAllowedValues` guard (line 437–439).

**Session:** rev-20260614T0404Z-d4f109

### LINKED ISSUE #350: [rev-20260614T0404Z-d4f109] handleRecipientDomain misclassifies double-@ addresses as "domain not allowed" instead of "invalid recipient email"
**Location:** `pkg/enforcement/handlers.go:862-881`

**Reference:** https://github.com/eunolabs/eunox/blob/e3cbf867db3c524b824c5c50b42c5fe03abc7f6e/pkg/enforcement/handlers.go#L862-L881

**Severity:** Low

**Description:**

`handleRecipientDomain` splits each recipient on the **first** `@` using `strings.SplitN(recipient, "@", 2)` and then validates the two parts:

```go
if len(parts) != 2 || parts[0] == "" || parts[1] == "" ||
    strings.ContainsFunc(parts[0], unicode.IsSpace) ||
    strings.ContainsFunc(parts[1], unicode.IsSpace) {
    return &ConditionError{..., Message: "invalid recipient email: " + recipient}
}
```

For the address `user@@example.com`, `strings.SplitN` yields:
- `parts[0]` = `"user"` — non-empty, no whitespace
- `parts[1]` = `"@example.com"` — non-empty, no whitespace

All four guards pass. The extracted `domain` is `"@example.com"`, which is looked up in `domainSet`. Since no operator writes `"@example.com"` in their allowlist, the lookup fails and the call is denied — **but with error code `CONDITION_FAILED` and message "recipient domain '@example.com' is not allowed"** rather than "invalid recipient email".

The comment on line 864 explicitly lists `"@@example.com"` (empty local part) as a case the guard catches, but omits `"user@@example.com"` (non-empty local part before the first `@`, second `@` embedded in what is presented as the domain). The guard catches the former (local part is `""`) but not the latter.

**Security impact:** In production the call is correctly denied regardless. The security concern arises in the narrow case where an operator accidentally lists `"@example.com"` (with the `@`) in the `domains` allowlist — perhaps copying a full address instead of just the domain. In that scenario `user@@example.com` would be incorrectly admitted as if it were a valid address at `@example.com`. The `validateRecipientDomain` load-time validator (`manifest.go:595–599`) rejects empty and whitespace-only domain entries but does not reject domains starting with `@`, leaving this misconfiguration path open.

The secondary impact is **audit trail integrity**: the denial is recorded with code `CONDITION_FAILED` and detail `"domain": "@example.com"`, which mixes a structural format error with a policy-enforcement event. Operators reviewing audit logs cannot distinguish a malformed address from a real domain violation.

**Trigger / repro:**

```go
    Arguments: map[string]interface{}{
        "to": "user@@example.com",
    },
}
cond := &capability.RecipientDomainCondition{
    Argument: "to",
    Domains:  []string{"example.com"},
}
```

`handleRecipientDomain` returns `CONDITION_FAILED / "recipient domain '@example.com' is not allowed"` rather than `CONDITION_FAILED / "invalid recipient email: user@@example.com"`.

**Suggested fix:**

Validate that `parts[1]` (the domain side) contains no embedded `@` character, since a valid email domain never does:

```go
if len(parts) != 2 || parts[0] == "" || parts[1] == "" ||
    strings.ContainsFunc(parts[0], unicode.IsSpace) ||
    strings.ContainsFunc(parts[1], unicode.IsSpace) ||
    strings.ContainsRune(parts[1], '@') {   // ← new guard
    return &ConditionError{..., Message: "invalid recipient email: " + recipient}
}
```

Additionally, update `validateRecipientDomain` in `manifest.go` to reject domain entries starting with `@` (catching the accidental full-address entry in the allowlist):

```go
for k, d := range v.Domains {
    if strings.TrimSpace(d) == "" {
        return fmt.Errorf(..., "empty or whitespace-only domain entry")
    }
    if strings.HasPrefix(strings.TrimSpace(d), "@") {
        return fmt.Errorf(..., "domain entry %q must not start with '@'; use the bare domain (e.g. example.com)", d)
    }
}
```

**Session:** rev-20260614T0404Z-d4f109

eunolabs:
## Bugs addressed

Two automated-review sessions filed open bugs against the same `main` HEAD (`e3cbf86`); both are resolved here in one PR, one commit per bug.

### Session `rev-20260614T0405Z-879400` (audit log)
- Closes #344 — audit-verify silently skipped the tamper-evident chain check whenever a `--request-id`/`--since` filter was active, so a deletion went undetected if the investigator passed a filter. Chain state is now maintained over **every** record unconditionally; filters narrow only per-record HMAC reporting. Removed the now-dead `chainCheck` field and the "chain verification skipped" note.
- Closes #345 — `rotate()` closed the old fd without `Sync()`, so a crash between the `os.Rename` and the page-cache flush could truncate the rotated file (an undetectable chain gap). Now syncs before close, mirroring `Close()`.
- Closes #346 — `reqCounter` restarted at 0 each process start, so request IDs collided across sessions (`req-1`, `req-2`, …), polluting `audit-verify --request-id` and SIEM joins. A random per-session nonce is now mixed into each ID; the per-request path stays counter-only.
- Closes #347 — the chain-integrity `switch` treated prev_hmac mismatch and seq gap as mutually exclusive, so an interior deletion (which trips both) emitted only `CHAIN BREAK` and hid the gap size. Now uses independent `if`s so both are counted and printed.

### Session `rev-20260614T0404Z-d4f109` (enforcement / manifest)
- Closes #348 — `recordFailureDenial` omitted `AuditOnly`, so a transient counter-backend fault during a staged audit-mode rollout hard-blocked every otherwise-allowed call. The matched constraint's `IsAuditOnly()` is now threaded through, so the fault is downgraded to observe-only under audit mode while enforce mode still fails closed.
- Closes #349 — empty `operations`/`extensions`/`tables` allowlists passed load-time validation but deny every call at runtime (e.g. an `operation:` vs `operations:` typo shipped a silent deny-all). Added non-empty-list guards mirroring the existing `allowedValues`/`ipRange`/`recipientDomain` checks.
- Closes #350 — `handleRecipientDomain` split on the first `@`, so `user@@example.com` parsed to domain `@example.com` and denied with the misleading "domain not allowed" instead of "invalid recipient email". Now rejects a second `@` in the domain side, and `validateRecipientDomain` rejects allowlist entries beginning with `@` (the accidental full-address paste that was the only way such a recipient could ever be admitted).

## Testing
- `make test` — `go test -race -count=1 ./...`: all packages pass.
- `make lint` (`go vet` + golangci-lint), `make check-fmt`, `make check-license`: clean.
- New regression tests added with each fix: interior-deletion reports both break and gap (#347); chain check survives both `--request-id` and `--since` filters (#344); audit-mode vs enforce-mode recording-failure denial (#348); empty operations/extensions/tables (incl. empty-tables-with-columns) rejected at load (#349); `user@@`/embedded-`@` recipients classified invalid and `@`-prefixed domain entries rejected at load (#350).
- Behavior/docs in sync: `docs/capability-manifest-guide.md` updated for the new load-time guards (#349, #350).

---
_Generated by [Claude Code](https://claude.ai/code/session_01NoPR8VgnsSTF1UWBuEgo4x)_


eunolabs (commented):
Self-review round 1 — one substantive finding:

**`docs/threat-model-mcp.md:496` (#346):** The audit-record table documents `request_id` as `"d3b07384-d113-..."` / "UUID v4, unique per record". That was already inaccurate (the implementation has always used `req-<counter>`), and this PR's #346 change makes the format `req-<nonce>-<counter>`. Since #346 is specifically about `request_id` uniqueness and this PR touches that field's generation, leaving the doc claiming UUID v4 violates the "keep audit-shape docs in sync" invariant. Fixing the row to describe the real format and the new cross-restart-uniqueness property.

No other substantive issues found in this pass (fail-closed semantics preserved, no new nil-deref in the `matched.IsAuditOnly()` call sites since `matched` is already dereferenced upstream, no-filter `verifyAuditLog` behavior unchanged, all new guards mirror existing conventions).

---

eunolabs:
✅ Ready to merge — all session bugs addressed (#344, #345, #346, #347, #348, #349, #350), self-review complete with no outstanding comments (the round-1 finding on the stale `request_id` threat-model row was fixed in 094ce15), and all 42 CI checks are green (Test, Lint, gofmt, License Headers, Enforcement compliance, builds on all platforms, demos, OPA comparisons, CodeQL, Trivy, govulncheck).

---

PR #369, opened 2026-06-14
Fix session 7gz87n bugs

### LINKED ISSUE #353: [rev-20260614T0504Z-186321] MatchValueGlob("/**", "") returns true — empty string bypasses absolute-path allowedValues patterns
**Location:** `pkg/enforcement/glob.go:55` (call to `matchGlobSegments`) and `pkg/enforcement/glob.go:70-101` (`matchGlobSegments`)

**Reference:** https://github.com/eunolabs/eunox/blob/e3cbf867db3c524b824c5c50b42c5fe03abc7f6e/pkg/enforcement/glob.go#L39-L56

**Severity:** Medium

---

## Description

`MatchValueGlob` incorrectly returns `true` when `pattern = "/**"` and `value = ""` (empty string). A pattern of `"/**"` is intended to match any absolute path — that is, any value beginning with `/`. An empty string `""` is not an absolute path, but it passes the check.

The root cause is how `strings.Split` handles its inputs:

```go
strings.Split("/**", "/")  // → ["", "**"]   (leading "" from the root "/")
strings.Split("",    "/")  // → [""]          (ONE empty-string element)
```

Both produce a leading `""` segment. Inside `matchGlobSegments`, `path.Match("", "")` returns `true`, so the anchor segment matches. After consuming it, `pat = ["**"]` and `val = []`. Because `**` as a trailing segment unconditionally returns `true` (it matches zero or more remaining segments), the whole call returns `true`.

Reproduction (confirmed against commit `e3cbf86`):

```go
enforcement.MatchValueGlob("/**", "")  // got: true, want: false
enforcement.MatchValueGlob("/**", "/") // got: true, want: true  (correct)
enforcement.MatchValueGlob("/**", "/a")// got: true, want: true  (correct)
```

Note that `"/reports/**"` is NOT affected — after `""` matches `""`, the next segment `"reports"` fails to match the empty remainder of `val`, so it correctly returns `false`. Only patterns of the form `"/**"` (root anchor followed immediately by `**` with nothing fixed in between) hit this case.

The existing `glob_test.go` file has no test for `MatchValueGlob("/**", "")` and does not test the pattern `"/**"` at all.

---

## Trigger / Repro

A manifest capability with:

```yaml
conditions:
  - type: allowedValues
    argument: path
    values:
      - "/**"
```

An attacker (or confused client) sends `{"path": ""}` — an empty string — as the argument value. The `handleAllowedValues` handler calls `MatchValueGlob("/**", "")`, which returns `true`, allowing the call through. The upstream tool receives an empty `path` argument when only a non-empty absolute path was intended.

This also affects the JWT PDP (`cmd/eunox-mcp/pdp.go:360`) which delegates to `enforcement.MatchValueGlob` for capability shorthand matching.

---

## Suggested Fix

In `MatchValueGlob` (or equivalently, at the `matchGlobSegments` call site at line 55), add a pre-check that enforces the leading-slash contract before entering the segment-splitting path:

```go
// Before splitting by "/", verify that the pattern and value agree on whether
// they are "rooted" (start with "/"). Without this, strings.Split("/**", "/")
// and strings.Split("", "/") both produce a leading "" segment, causing
// path.Match("", "") to fire and matching the root anchor of any "/..." pattern
// against an empty value.
if strings.HasPrefix(pattern, "/") != strings.HasPrefix(value, "/") {
    return false
}
return matchGlobSegments(strings.Split(pattern, "/"), strings.Split(value, "/"))
```

This change is a no-op for patterns that do not start with `/` (e.g. `"reports/**"` vs `""` already returns false via the first non-"**" segment mismatch) and for values that do start with `/` (they correctly match `/`-prefixed patterns). It closes only the asymmetric-root case.

A corresponding regression test should be added to `glob_test.go`:

```go
{"doublestar-root does not match empty string", "/**", "", false},
{"doublestar-root matches root slash", "/**", "/", true},
{"doublestar-root matches absolute path", "/**", "/a/b", true},
```

**Session:** rev-20260614T0504Z-186321

### LINKED ISSUE #354: [rev-20260614T0505Z-dbabd3] sequenceBlock bypassed when antecedent tool runs in audit mode with a failing condition
**Location:** `pkg/enforcement/engine.go:655-726` (`EvaluateConditions`), `cmd/eunox-mcp/pdp.go:189-281` (`ManifestPDP.Decide`)

**Reference:**
- https://github.com/eunolabs/eunox/blob/e3cbf867db3c524b824c5c50b42c5fe03abc7f6e/pkg/enforcement/engine.go#L655-L726
- https://github.com/eunolabs/eunox/blob/e3cbf867db3c524b824c5c50b42c5fe03abc7f6e/cmd/eunox-mcp/pdp.go#L189-L281

**Severity:** High

**Description:**

`recordSessionCall` is only invoked after all conditions on the matched constraint pass (the allow path). When a constraint in `enforcement: audit` mode has a failing condition, `EvaluateConditions` returns a deny without ever calling `recordSessionCall`. In `ManifestPDP.Decide`, the deferred `resp.AuditOnly = true` (line 216–217) stamps that deny as audit-only, so the transport forwards the call to the upstream — the tool *actually runs*. But because `recordSessionCall` was skipped, no session-history entry is written.

A subsequent `sequenceBlock` condition on another tool that names the antecedent in `afterTools` calls `history.Peek(...)`, sees count 0, and allows the call — the block rule silently fails open even though the antecedent ran.

This is distinct from the previously fixed issues #213/#232 (which concerned `recordSessionCall` swallowing write errors) and #348 (which concerned `recordFailureDenial` omitting `AuditOnly`). Here, `recordSessionCall` is never reached at all because the code path exits at the first failing condition before the call is recorded.

**Trigger / repro:**

```yaml
capabilities:
  - target: tool:read_credentials
    actions: [call]
    enforcement: audit          # observe-only; denials are forwarded
    conditions:
      - type: timeWindow
        notBefore: "2099-01-01T00:00:00Z"  # always fails today

  - target: tool:write_external
    actions: [call]
    conditions:
      - type: sequenceBlock
        afterTools: [read_credentials]
```

1. Client calls `read_credentials`. `handleTimeWindow` fails (before 2099). `EvaluateConditions` returns deny. `AuditOnly: true` is stamped by the defer. Transport forwards the call — `read_credentials` runs in the upstream. `recordSessionCall` is **never called**.
2. Client calls `write_external`. `handleSequenceBlock` calls `Peek("read_credentials")` → 0 (no entry). Returns nil (condition passes). `write_external` is **allowed**.

The `sequenceBlock` on `write_external` should have blocked step 2, but it did not.

**Suggested fix:**

In `ManifestPDP.Decide`, after detecting `matched.IsAuditOnly()`, record the session call even when `EvaluateConditions` returns a deny. Because `recordSessionCall` is unexported on `*Engine`, one clean approach is to add an exported `RecordSessionCall(ctx, req)` method — or to add an `auditMode bool` parameter to `EvaluateConditions` so the engine itself knows to record on the deny path when audit mode is active:

```go
// Option A: PDP records the call after an audit-mode deny
if err != nil { ... }
if matched.IsAuditOnly() && resp.Decision == capability.DecisionDeny {
    _ = p.engine.RecordSessionCall(ctx, req)  // best-effort or fail-closed
}
return resp
```

Whether a recording failure on this path should itself produce a hard deny or be best-effort requires a policy decision (fail-closed argues for hard deny; audit-mode intent argues for best-effort). Either way the current behavior — the call runs but is invisible to `sequenceBlock` — is incorrect.

**Session:** rev-20260614T0505Z-dbabd3

### LINKED ISSUE #355: [rev-20260614T0505Z-dbabd3] sequenceBlock TOCTOU race: Peek and recordSessionCall are not atomic, allowing concurrent bypass
**Location:** `pkg/enforcement/handlers.go:1054-1096` (`handleSequenceBlock`), `pkg/enforcement/engine.go:340-364` (`recordSessionCall`)

**Reference:**
- https://github.com/eunolabs/eunox/blob/e3cbf867db3c524b824c5c50b42c5fe03abc7f6e/pkg/enforcement/handlers.go#L1054-L1096
- https://github.com/eunolabs/eunox/blob/e3cbf867db3c524b824c5c50b42c5fe03abc7f6e/pkg/enforcement/engine.go#L340-L364

**Severity:** Medium

**Description:**

`sequenceBlock` guards tool B by checking whether tool A was previously called in the session via `history.Peek(...)`. If the count is > 0, the call is denied. After tool B's own conditions all pass, `recordSessionCall` writes tool B's own history entry. The check (Peek on tool A) and the record (IncrementAndGet for tool A, done earlier) are two separate, non-atomic operations.

Under concurrent load in an HTTP-mode session (two requests on the same session at the same time), there is a narrow TOCTOU window:

1. Request R1 calls tool A. All conditions pass. `recordSessionCall` is about to execute but has not yet.
2. Request R2 calls tool B concurrently. `handleSequenceBlock` calls `Peek("tool_a")` → 0 (A not yet recorded) → condition passes.
3. R1's `recordSessionCall` executes and writes tool A's history entry.
4. R2 is allowed and tool B runs — even though tool A was (or is being) called in the same session.

The `maxCalls` condition was explicitly fixed to be atomic (`IncrementIfBelow` instead of a separate Peek+Increment, issue #204). `sequenceBlock` has no equivalent atomic primitive: it relies on `Peek` (non-incrementing read) and a separate `recordSessionCall` (IncrementAndGet), with no lock held across both.

**Trigger / repro:**

```yaml
capabilities:
  - target: tool:fetch_credentials
    actions: [call]

  - target: tool:exfiltrate
    actions: [call]
    conditions:
      - type: sequenceBlock
        afterTools: [fetch_credentials]
```

Send `fetch_credentials` and `exfiltrate` concurrently on the same session (e.g., two simultaneous HTTP requests with the same `Mcp-Session-Id`). With appropriate timing, `exfiltrate` passes the `sequenceBlock` check before `fetch_credentials` is recorded.

**Severity context:**

The race window is small (microseconds between `Peek` and the previous request's `recordSessionCall`) and requires a specific concurrent usage pattern. Real MCP sessions are typically sequential. However, a client that deliberately fires concurrent requests (or an attacker that can influence timing) can exploit it reliably under load.

**Suggested fix:**

The most targeted fix is to add a `CheckAndRecord` (or `PeekOrBlock`) operation to `callcounter.LimiterStore`/`callcounter.HistoryStore` that atomically checks whether any of a set of keys has a non-zero count and, if not, records the current key — all under one lock/Lua script. This mirrors the atomicity introduced for `maxCalls` in issue #204.

A simpler but less precise approach: document this as a known limitation scoped to concurrent-request sessions, and note that MCP's serial request model makes it unexploitable in practice.

**Session:** rev-20260614T0505Z-dbabd3

### LINKED ISSUE #356: [rev-20260614T0505Z-dbabd3] resourceSpecificity misparses backslash-escaped `[` outside a character class, producing wrong specificity score
**Location:** `pkg/enforcement/engine.go:880-913` (`resourceSpecificity`)

**Reference:** https://github.com/eunolabs/eunox/blob/e3cbf867db3c524b824c5c50b42c5fe03abc7f6e/pkg/enforcement/engine.go#L880-L913

**Severity:** Low

**Description:**

`resourceSpecificity` iterates over pattern runes with a switch statement that includes a `case '['` arm to handle character classes. When it encounters `[`, it increments `wildcardCount` and then runs an inner loop to skip past the closing `]`. The outer switch does NOT handle `'\\'` (backslash) specially — a backslash outside a class falls into `default: literalCount++`.

This creates a misparse when a pattern contains `\[` — a backslash-escaped literal `[` (not starting a class). `path.Match` correctly interprets `\[` as a literal `[` character, but `resourceSpecificity` sees:

1. `\` → `default: literalCount++` (correct)
2. `[` → `case '['`: `wildcardCount++`, inner loop runs, consuming runes until `]`

The `[` that should be a literal is instead treated as opening a character class, which:
- Inflates `wildcardCount` by 1 (makes the pattern appear less specific)
- Consumes subsequent characters as if they were class contents — misattributing literals as class internals and potentially skipping the character after the intended closing `]`

**Trigger / repro:**

Two constraints compete for `tool:file[1]` (a tool whose name is literally `file[1]`):

```yaml
capabilities:
  - target: "tool:file\\[1]"   # exact match via escaped '[' — should win
    actions: [call]
    conditions: [...strict policy...]

  - target: "tool:file*"        # broader glob — should lose
    actions: [call]
    conditions: [...permissive policy...]
```

`resourceSpecificity("file\\[1]", "file[1]")`:
- Correct score: 7 literals × 10 − 0 wildcards = **70** (the `\`, `[`, `1`, `]` are all non-wildcard runes in the pattern, plus `f`, `i`, `l`, `e`)
- Actual score: the `[` is treated as opening a class; `1` and `]` are consumed as class internals; score is **4×10 − 1 = 39**

`resourceSpecificity("file*", "file[1]")` scores 4×10 − 1 = 39 as well. The two constraints now tie, and the looser `file*` constraint may win on manifest order — the wrong policy is applied.

**Suggested fix:**

Add a `case '\\'` arm in the outer rune loop that increments `literalCount` twice (for the backslash and the escaped rune) and advances `i` by one extra step, so that the character following `\` is consumed without hitting the `case '['` arm:

```go
case '\\':
    literalCount++ // the backslash itself
    if i+1 < len(runes) {
        i++
        literalCount++ // the escaped rune
    }
case '[':
    // ... existing class-handling logic
```

This mirrors how `path.Match` itself processes backslash escapes and keeps `resourceSpecificity` consistent with the matching semantics it is meant to score.

**Session:** rev-20260614T0505Z-dbabd3

### LINKED ISSUE #357: [rev-20260614T0605Z-5ecf0b] Stop() reads r.cancel without r.mu, racing with Start() under the race detector
**Location:** `pkg/killswitch/redis.go:136-140` (`Stop`), `pkg/killswitch/redis.go:82-88` (`Start`)

**Reference:** https://github.com/eunolabs/eunox/blob/cb43edd90b9f32cd0190072470b9ae0ade429025/pkg/killswitch/redis.go#L136-L140

**Severity:** High

**Description:**

`Start()` writes both `r.lifecycleCtx` and `r.cancel` under `r.mu`:

```go
// redis.go:84-88
r.mu.Lock()
r.lifecycleCtx = subCtx
r.cancel = cancel
r.mu.Unlock()
```

The comment even documents why: _"Write lifecycleCtx/cancel under r.mu so that handlePubSubMessage, which reads lifecycleCtx under the same lock, has a consistent view."_

But `Stop()` reads `r.cancel` **without** holding `r.mu`:

```go
// redis.go:136-140
func (r *Redis) Stop() {
    if r.cancel != nil {  // ← unsynchronized read
        r.cancel()
    }
}
```

This is a data race: `Start()` writes `r.cancel` under a write lock; `Stop()` reads it without any lock. If `Stop()` is called from a different goroutine while `Start()` is still executing (a realistic pattern during test teardown, signal handling, or dependency injection with races), `go test -race` flags a concurrent write/read on `r.cancel`. The fix was applied to `handlePubSubMessage` for the identical field but was not applied to `Stop()`.

**Trigger / repro:**

```go
r := NewRedis(client)
var wg sync.WaitGroup
wg.Add(2)
go func() { defer wg.Done(); r.Start(ctx) }()
go func() { defer wg.Done(); r.Stop() }()
wg.Wait()
// go test -race detects concurrent write (Start, under mu) and read (Stop, no mu) on r.cancel
```

**Suggested fix:**

Read `r.cancel` under `r.mu.RLock()` in `Stop()`, consistent with how `Start()` writes it:

```go
func (r *Redis) Stop() {
    r.mu.RLock()
    cancel := r.cancel
    r.mu.RUnlock()
    if cancel != nil {
        cancel()
    }
}
```

**Session:** rev-20260614T0605Z-5ecf0b

### LINKED ISSUE #358: [rev-20260614T0605Z-5ecf0b] Kill switch denials forwarded in route-level audit mode — /control/kill has no effect when audit:true
**Location:** `cmd/eunox-mcp/http.go:809-824` (`handleHTTPToolsCall`), same pattern in `handleHTTPResourcesRead`, `handleHTTPResourcesSubscribe`, `handleHTTPPromptsGet`

**Reference:** https://github.com/eunolabs/eunox/blob/cb43edd90b9f32cd0190072470b9ae0ade429025/cmd/eunox-mcp/http.go#L809-L824

**Severity:** High

**Description:**

Every enforced-method handler uses the same pattern to decide whether to block or forward a PDP denial:

```go
if dec.Decision == capability.DecisionDeny {
    denial := dec.Denial
    observe := rt.audit || dec.AuditOnly   // ← the fatal OR
    if !observe {
        return denialResult(...)            // hard block
    }
    // falls through: request is forwarded to upstream
}
```

When a route is in audit mode (`rt.audit == true`), `observe` is `true` for **every denial**, including kill switch denials. `ManifestPDP.Decide()` (and `DecideResourceRead`, `DecidePromptGet`, `DecideSampling`) call `killCheck()` first and return a deny with code `KILL_SWITCH` / `KILL_SWITCH_ERROR` when the switch is active. But back in the handler, `rt.audit || false == true`, so that kill switch denial is logged as "would be denied" and the request is forwarded to the upstream anyway.

The `/control/kill` endpoint (`handleKill`) correctly calls `p.ks.ActivateGlobal()` or `p.ks.KillSession()`. The kill switch state is updated. But any route configured with `audit: true` will never observe the block — the kill switch becomes entirely inoperative for those routes.

**Trigger / repro:**

1. Deploy the proxy with a route that has `audit: true` (or pass `--audit` in stdio mode).
2. Issue `POST /control/kill {"all":true}`.
3. Send any tool call to the proxy.
4. Observe: the request is forwarded to the upstream and the audit log says "deny" (audit mode) — but the upstream executes the call.

**Suggested fix:**

Kill switch denials must be a hard block regardless of audit mode. The cleanest fix is to check the denial code before computing `observe`:

```go
if dec.Decision == capability.DecisionDeny {
    denial := dec.Denial
    // Kill switch is a hard block even in audit mode — it is an operator
    // emergency stop, not a policy decision to be observed-and-forwarded.
    isKillSwitch := denial.Code == capability.ErrCodeKillSwitch ||
                    denial.Code == capability.ErrCodeKillSwitchError
    observe := !isKillSwitch && (rt.audit || dec.AuditOnly)
    if !observe {
        return denialResult(...)
    }
    // audit forward
}
```

Or, add an `IsHardDeny bool` field to `EnforceResponse` that `killCheck` stamps and that the handler tests directly, keeping the policy coupling out of the handler.

**Session:** rev-20260614T0605Z-5ecf0b

### LINKED ISSUE #359: [rev-20260614T0605Z-5ecf0b] handleKill silently discards ActivateGlobal/KillSession errors and always returns {"ok":true}
**Location:** `cmd/eunox-mcp/http.go:1241-1254` (`handleKill`)

**Reference:** https://github.com/eunolabs/eunox/blob/cb43edd90b9f32cd0190072470b9ae0ade429025/cmd/eunox-mcp/http.go#L1241-L1254

**Severity:** Medium

**Description:**

`handleKill` discards errors from both kill-switch write operations and unconditionally returns HTTP 200 with `{"ok":true}`:

```go
if body.All {
    _ = p.ks.ActivateGlobal(r.Context())   // error discarded
    w.Header().Set("Content-Type", ctJSON)
    _, _ = fmt.Fprint(w, `{"ok":true,"killed":"all"}`)
    return
}
if body.SessionID == "" {
    http.Error(w, "sessionId or all required", http.StatusBadRequest)
    return
}
_ = p.ks.KillSession(r.Context(), body.SessionID)  // error discarded
w.Header().Set("Content-Type", ctJSON)
b, _ := json.Marshal(map[string]interface{}{"ok": true, "killed": body.SessionID})
_, _ = w.Write(b)
```

With the Redis backend, if the Redis server is temporarily unavailable, `ActivateGlobal` or `KillSession` returns a connection error. The error is silently dropped; the operator receives `{"ok":true,"killed":"all"}` and believes the emergency kill is in effect, while the proxy continues to serve all requests.

This violates the "fail closed" invariant. For an emergency stop (`/control/kill`), a write failure is critical operator-visible information. Returning success on a failed kill is strictly worse than returning an error: the operator takes no recovery action because they believe the system is safe.

**Trigger / repro:**

1. Deploy with the Redis kill-switch backend.
2. Shut down Redis (or let a network partition occur).
3. `POST /control/kill {"all":true}` → response: `200 {"ok":true,"killed":"all"}`.
4. The global kill switch is NOT active (Redis write failed). Requests continue to be served.

**Suggested fix:**

Check and propagate the error:

```go
if body.All {
    if err := p.ks.ActivateGlobal(r.Context()); err != nil {
        http.Error(w, "kill switch activation failed: "+err.Error(), http.StatusInternalServerError)
        return
    }
    w.Header().Set("Content-Type", ctJSON)
    _, _ = fmt.Fprint(w, `{"ok":true,"killed":"all"}`)
    return
}
...
if err := p.ks.KillSession(r.Context(), body.SessionID); err != nil {
    http.Error(w, "kill switch session kill failed: "+err.Error(), http.StatusInternalServerError)
    return
}
```

Note: the in-memory backend always returns nil, so this change is backward-compatible for single-instance deployments.

**Session:** rev-20260614T0605Z-5ecf0b

### LINKED ISSUE #360: [rev-20260614T0605Z-5ecf0b] KillAgent("") and KillSession("") silently succeed but never block; Redis backend also writes a polluting bare-prefix key
**Location:** `pkg/killswitch/memory.go:61-66` and `pkg/killswitch/redis.go:201-209` (`KillAgent`); same for `KillSession` at `memory.go:77-82` and `redis.go:225-234`

**Reference:** https://github.com/eunolabs/eunox/blob/cb43edd90b9f32cd0190072470b9ae0ade429025/pkg/killswitch/redis.go#L201-L209

**Severity:** Medium

**Description:**

Neither `KillAgent` nor `KillSession` validates that the supplied ID is non-empty. Both implementations accept `""` and return nil (success). But `ShouldBlock` unconditionally skips empty IDs:

```go
// redis.go / memory.go ShouldBlock
if agentID != "" && r.killedAgents[agentID] {
    return true, nil
}
```

So `KillAgent("")` is a silent no-op from the caller's perspective: it returns nil, but no traffic is ever blocked by it.

**Additional side-effects in the Redis backend:**

1. **Key pollution.** `KillAgent("")` writes key `killswitch:agent:` (the bare agent prefix) to Redis. On the next `refreshState`, `scanPrefix` matches it, strips the prefix, and inserts `""` into `newAgents`. After the refresh, `r.killedAgents[""] == true` and `Status()` returns `""` in `KilledAgents` — a phantom entry that no real kill corresponds to.

2. **Spurious full refresh.** `KillAgent("")` publishes `"agent:kill:"`. In `handlePubSubMessage`, `len("agent:kill:") > len("agent:kill:")` is false, so this payload falls through to the `default` case and triggers an unexpected full `refreshState` round-trip on every listening replica.

3. **Key collides with prefix scan.** If `deleteByPrefix(ctx, redisAgentPrefix)` (called from `Reset`) is run concurrently with another agent key being written, the bare `killswitch:agent:` key is only deleted if it was created before the scan's cursor reached it — a minor TOCTOU issue made possible by the existence of the malformed key.

**Trigger / repro:**

```go
ks := killswitch.NewInMemory()
_ = ks.KillAgent(ctx, "")       // returns nil
blocked, _ := ks.ShouldBlock(ctx, "", "") // returns false — kill had no effect
st, _ := ks.Status(ctx)
// st.KilledAgents == [""]  — confusing phantom entry

rks := killswitch.NewRedis(client)
_ = rks.KillAgent(ctx, "")     // writes "killswitch:agent:" to Redis
// next reconcile: rks.killedAgents[""] = true, Status shows phantom ""
```

**Suggested fix:**

Add an early validation guard in both backends:

```go
func (m *InMemory) KillAgent(_ context.Context, agentID string) error {
    if agentID == "" {
        return fmt.Errorf("killswitch: KillAgent: agentID must not be empty")
    }
    ...
}
```

The `Manager` interface doc for `KillAgent` should also note that an empty ID is an error. The same fix applies to `KillSession` and to both backends.

**Session:** rev-20260614T0605Z-5ecf0b

### LINKED ISSUE #361: [rev-20260614T0606Z-ce3350] File descriptor leak in auditSink.Close() when Sync() fails
**Location:** `cmd/eunox-mcp/audit.go:683-685`
**Reference:** https://github.com/eunolabs/eunox/blob/cb43edd90b9f32cd0190072470b9ae0ade429025/cmd/eunox-mcp/audit.go#L683-L685
**Severity:** Medium

**Description:**
In `auditSink.Close()`, when `s.f.Sync()` returns an error the function sets `closeErr` and immediately returns — without ever calling `s.f.Close()`:

```go
if s.f != nil {
        closeErr = err
        return        // s.f.Close() is never called
    }
    closeErr = s.f.Close()
}
```

Because `closeOnce` ensures `Close()` executes its body exactly once, there is no retry path. The file descriptor is leaked for the remainder of the process lifetime. On Linux an unsynced, unclosed file descriptor holds kernel page-cache dirty pages pinned until process exit, which can accumulate if the proxy is long-lived and the filesystem is flaky.

**Trigger / repro:**
Any filesystem condition that causes `f.Sync()` to fail (full disk, I/O error, network filesystem timeout, etc.) while `Close()` is executing will leak the audit-log file descriptor. The `closeOnce.Do` wrapper means the FD is permanently leaked — a subsequent call to `Close()` is a no-op.

**Suggested fix:**
Close the file even when `Sync` fails; prefer returning the `Sync` error but still attempt the close:

```go
if s.f != nil {
    syncErr := s.f.Sync()
    if err := s.f.Close(); syncErr != nil {
        closeErr = syncErr
    } else {
        closeErr = err
    }
}
```

**Session:** rev-20260614T0606Z-ce3350

### LINKED ISSUE #362: [rev-20260614T0606Z-ce3350] rotate() leaves s.logPath stale after failed reopen, causing permanent rotation failures
**Location:** `cmd/eunox-mcp/audit.go:888-914`
**Reference:** https://github.com/eunolabs/eunox/blob/cb43edd90b9f32cd0190072470b9ae0ade429025/cmd/eunox-mcp/audit.go#L888-L914
**Severity:** Medium

**Description:**
In the fallback path of `auditSink.rotate()`, after `os.Rename` succeeds but `os.OpenFile` for the new log file fails:

```go
rotated := s.rotatedPath()
if err := os.Rename(s.logPath, rotated); err != nil { ... }  // succeeds

f, err := os.OpenFile(s.logPath, ...)  // s.logPath no longer exists — it was just renamed!
if err != nil {
    // Falls back to writing to the renamed fd (still valid inode)
    s.written = 0
    return   // BUG: s.logPath is now stale
}
```

After this path executes:
- `s.f` is valid and writes to the **renamed file** (at the `rotated` path).
- `s.logPath` still holds the **original path** (which no longer exists on disk).
- When `s.written` grows back to `maxBytes`, `rotate()` is called again and `os.Rename(s.logPath, ...)` fails (source doesn't exist) every time.
- `s.written = 0` is reset on each failure, so the probe repeats every `maxBytes` bytes written — the drainer spins indefinitely, logging rename errors, and `pruneRotated()` is never called, so disk can fill.
- The actual audit log grows unboundedly at the wrong (rotated) path.

**Trigger / repro:**
1. Audit log reaches `maxBytes` and rotation starts.
2. `os.Rename` succeeds (old log now at `rotated` path).
3. `os.OpenFile(s.logPath, ...)` fails (e.g., permission error, full filesystem creating the new file).
4. From this point, every subsequent rotation attempt fails silently with `os.Rename(s.logPath, ...)`, and audit records keep accumulating in the renamed file at an unexpected path.

**Suggested fix:**
On the failed-reopen path, update `s.logPath` to `rotated` so the drainer knows the active file's real location, or use a separate field to track the actual open path. Additionally, ensure `pruneRotated()` is invoked even in degraded mode:

```go
if err != nil {
    fmt.Fprintf(os.Stderr, "[eunox-mcp] audit rotate error (reopen): %v; falling back to renamed file\n", err)
    s.logPath = rotated  // update to the file we're actually writing to
    s.written = 0
    return
}
```

**Session:** rev-20260614T0606Z-ce3350

### LINKED ISSUE #363: [rev-20260614T0606Z-ce3350] openAuditSink silently ignores f.Stat() error; rotation threshold starts at zero for existing large logs
**Location:** `cmd/eunox-mcp/audit.go:260-264`
**Reference:** https://github.com/eunolabs/eunox/blob/cb43edd90b9f32cd0190072470b9ae0ade429025/cmd/eunox-mcp/audit.go#L260-L264
**Severity:** Low

**Description:**
In `openAuditSink`, after opening the audit log file the code discards the error from `f.Stat()`:

```go
info, _ := f.Stat()
var written int64
if info != nil {
    written = info.Size()
}
```

If `f.Stat()` returns an error (e.g., a race with concurrent truncation, a network filesystem quirk, or `ENOMEM`), `written` is initialized to `0` regardless of the actual file size. The rotation threshold check in `writeRecord` is:

```go
    s.rotate()
}
```

With `written = 0` on a large existing log, rotation is delayed by exactly `maxBytes` extra bytes compared to intent — the log grows up to `2 × maxBytes` before the first rotation instead of triggering immediately. On a 100 MiB default threshold this means 200 MiB before first rotation after restart, silently violating the configured bound.

**Trigger / repro:**
1. Audit log is near `maxBytes` (e.g., 95 MiB of a 100 MiB limit).
2. Proxy restarts. `f.Stat()` returns an error.
3. `written = 0`; rotation is not triggered until another 100 MiB is written (file reaches 195 MiB).

**Suggested fix:**
Log or return the stat error explicitly; fall back to the safe conservative value of `s.maxBytes` (force immediate rotation on first write) if stat fails, or treat it as a hard error at startup:

```go
info, err := f.Stat()
if err != nil {
    // Force an immediate rotation on the next write rather than
    // silently starting the threshold from zero.
    written = rotateSizeBytes
    fmt.Fprintf(os.Stderr, "[eunox-mcp] WARNING: could not stat audit log %q: %v; rotation threshold reset\n", logPath, err)
} else {
    written = info.Size()
}
```

**Session:** rev-20260614T0606Z-ce3350

### LINKED ISSUE #364: [rev-20260614T0606Z-ce3350] Allow audit record has wrong audit_only flag when per-entry audit-mode deny is forwarded
**Location:** `cmd/eunox-mcp/http.go:856` (and matching lines in `handleHTTPResourcesRead`, `handleHTTPResourcesSubscribe`, `handleHTTPPromptsGet`)
**Reference:** https://github.com/eunolabs/eunox/blob/cb43edd90b9f32cd0190072470b9ae0ade429025/cmd/eunox-mcp/http.go#L856
**Severity:** Medium

**Description:**
When a **per-entry** audit-mode constraint fires (`dec.AuditOnly == true`, route-level `rt.audit == false`), the deny is observed and the call is forwarded. Two audit records are written: a deny (correctly marked `audit_only: true`) and an allow (incorrectly marked `audit_only: false`).

The deny record is written inside the `if dec.Decision == capability.DecisionDeny` block and correctly uses `rt.audit || dec.AuditOnly`:

```go
rt.sink.Record(ctx, sess.id, params.Name, "tools/call", "deny",
    denial.Code, denial.ConditionType, denial.Details, nil,
    rt.audit || dec.AuditOnly)   // ← correct
```

But the subsequent allow record (after the upstream call) loses the `dec.AuditOnly` flag:

```go
rt.sink.Record(ctx, sess.id, params.Name, "tools/call", "allow",
    "", "", details, oblNames,
    rt.audit)                    // ← BUG: should be rt.audit || dec.AuditOnly
```

`dec` is still in scope at that point; the flag is simply not carried forward. The engine only sets `AuditOnly` on deny responses (confirmed in `pkg/enforcement/engine.go`), so this manifests specifically when:
- Route is NOT in audit mode (`rt.audit = false`)
- A per-entry audit-mode constraint denies the request (`dec.Decision = Deny, dec.AuditOnly = true`)
- The call is forwarded and succeeds

The resulting allow record has `audit_only: false`, falsely implying the call received a genuine policy-allow rather than being forwarded under per-entry audit observation. This misleads SIEM queries and the `stats` subcommand that use `audit_only` to separate enforced decisions from observed ones.

The same bug exists in:
- `handleHTTPResourcesRead` (line ~911)
- `handleHTTPResourcesSubscribe` (line ~988)
- `handleHTTPPromptsGet` (line ~1044)

The stdio proxy handlers in `stdio.go` likely have the same pattern and should be audited.

**Trigger / repro:**
1. Configure a manifest with at least one entry using `enforcement: audit`.
2. That entry's conditions trigger a deny for an incoming call.
3. The proxy forwards the call (audit mode).
4. The "allow" record in the audit log will have `"audit_only": false` instead of `true`.

**Suggested fix:**
Capture `dec.AuditOnly` before it goes out of scope and pass it to the allow record:

```go
// Before the deny block:
decAuditOnly := dec.AuditOnly

// ...existing deny block unchanged...

// Allow record:
rt.sink.Record(ctx, sess.id, params.Name, "tools/call", "allow",
    "", "", details, oblNames,
    rt.audit || decAuditOnly)
```

**Session:** rev-20260614T0606Z-ce3350

### LINKED ISSUE #365: [rev-20260614T0606Z-ce3350] TOCTOU race in loadOrCreateAuditKeys allows two processes to generate divergent signing keys
**Location:** `cmd/eunox-mcp/audit.go:1022-1045`
**Reference:** https://github.com/eunolabs/eunox/blob/cb43edd90b9f32cd0190072470b9ae0ade429025/cmd/eunox-mcp/audit.go#L1022-L1045
**Severity:** Low

**Description:**
`loadOrCreateAuditKeys` uses a check-then-act pattern to create the HMAC key file when it does not exist:

```go
data, err := os.ReadFile(keyPath)          // (1) check: does the file exist?
if err == nil {
    return parseAuditKeys(data)
}
if !os.IsNotExist(err) {
    return nil, ...
}

key := make([]byte, 32)
rand.Read(key)                             // (2) generate a new key
hex.Encode(encoded, key)
os.WriteFile(keyPath, encoded, 0o600)      // (3) write — may overwrite a key another process just wrote
return [][]byte{key}, nil                  // (4) return the IN-MEMORY key, not the file's key
```

If two `eunox-mcp` processes start concurrently and both observe the key file missing at step (1):
- Both generate different random keys at step (2).
- One wins the `os.WriteFile` race at step (3); the other silently overwrites the winner's key.
- Each process returns its own in-memory key from step (4), even though only one was persisted.

Result: process A signs audit records with `key_A` while process B signs with `key_B`, but the key file only contains `key_B`. `audit-verify` fails for all of process A's records (unknown key ID after the race), and there is no warning.

This is a startup-only race with a narrow window, but multi-instance deployments with a shared filesystem key path (documented in `openAuditSink`'s REL-04 note) make it realistic.

**Trigger / repro:**
1. Deploy two `eunox-mcp` instances sharing the same `--audit-key-path` on a shared filesystem.
2. Start both simultaneously with no pre-existing key file.
3. Both see `ErrNotExist`, generate different keys, and race on `os.WriteFile`.
4. The losing process uses a key that is no longer in the file; its records fail `audit-verify`.

**Suggested fix:**
Use an atomic create (`os.O_CREATE|os.O_EXCL`) and re-read the file if the exclusive create fails (another process won the race):

```go
f, err := os.OpenFile(keyPath, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0o600)
if os.IsExist(err) {
    // Another process just created it; read and return that key.
    data, err := os.ReadFile(keyPath)
    if err != nil {
        return nil, err
    }
    return parseAuditKeys(data)
}
if err != nil {
    return nil, err
}
// Write the new key to the exclusively-created file.
_, _ = f.Write(encoded)
_ = f.Close()
return [][]byte{key}, nil
```

**Session:** rev-20260614T0606Z-ce3350

eunolabs:
## Bugs addressed (session 7gz87n)

This PR fixes the batch of open `automated-review` bugs (issues #353–#365). Each
is a focused, independent fix with a regression test where one can be written
deterministically.

- Closes #353 — `MatchValueGlob("/**", "")` returned `true`, letting an empty
  string bypass an absolute-path `allowedValues` pattern. Added an
  asymmetric-root pre-check (`pattern`/`value` must agree on a leading `/`) before
  segment splitting. `pkg/enforcement/glob.go`.
- Closes #354 — a `sequenceBlock` antecedent that ran under a per-entry
  `enforcement: audit` constraint with a failing condition was forwarded (the
  tool ran) but never recorded, so a later `sequenceBlock` failed **open**. Added
  exported `Engine.RecordSessionCall` and call it from the PDP after an audit-mode
  deny on all three entry points. `pkg/enforcement/engine.go`, `cmd/eunox-mcp/pdp.go`.
- Closes #355 — `sequenceBlock` TOCTOU under concurrent same-session requests.
  This is intrinsic to two independent requests racing (no single-request atomic
  primitive can order one request's check against another's write); MCP's serial
  per-session model makes it unexploitable in practice. Documented as a known
  limitation in the handler and the manifest guide, per the issue's accepted
  resolution. `pkg/enforcement/handlers.go`, `docs/capability-manifest-guide.md`.
- Closes #356 — `resourceSpecificity` misparsed a backslash-escaped `[` as a
  character class, inflating `wildcardCount` and letting a broad glob tie an exact
  match. Added a `case '\\'` arm that consumes the escaped rune as a literal.
  `pkg/enforcement/engine.go`.
- Closes #357 — `Redis.Stop()` read `r.cancel` without `r.mu`, racing with
  `Start()`. Now reads it under `r.mu.RLock()`. `pkg/killswitch/redis.go`.
- Closes #358 — kill switch denials were forwarded in audit mode, making
  `/control/kill` inoperative for `audit:true` routes. Kill switch denials now
  hard-block regardless of audit mode across all four enforced-method handlers
  plus the sampling path, in both transports. `cmd/eunox-mcp/{http,stdio,pdp}.go`.
- Closes #359 — `handleKill` discarded `ActivateGlobal`/`KillSession` errors and
  always returned `{"ok":true}`, so a failed emergency stop looked successful.
  Now propagates the error as `500`. `cmd/eunox-mcp/http.go`.
- Closes #360 — `KillAgent("")`/`KillSession("")` were silent no-ops (and the
  Redis backend wrote a polluting bare-prefix key). Both backends now reject an
  empty ID. `pkg/killswitch/{memory,redis,killswitch}.go`.
- Closes #361 — `auditSink.Close()` leaked the file descriptor when `Sync()`
  failed (it returned before `Close()`, and `closeOnce` made it one-shot). Now
  always closes, preferring the sync error. `cmd/eunox-mcp/audit.go`.
- Closes #362 — `rotate()` left `s.logPath` stale after a failed reopen, so every
  later rotation's rename failed and the drainer spun while the log grew unbounded
  at the wrong path. Now points `s.logPath` at the renamed file.
  `cmd/eunox-mcp/audit.go`.
- Closes #363 — `openAuditSink` discarded the `f.Stat()` error and started the
  rotation counter at zero, letting an existing large log grow to ~2× the bound.
  Now falls back to `rotateSizeBytes` (forces rotation on next write) and warns.
  `cmd/eunox-mcp/audit.go`.
- Closes #364 — the forwarded **allow** record after a per-entry audit-mode deny
  lost the `audit_only` flag, mislabeling an observed-and-forwarded call as a
  genuine policy-allow. Now carries `dec.AuditOnly` on all four handlers in both
  transports. `cmd/eunox-mcp/{http,stdio}.go`.
- Closes #365 — TOCTOU in `loadOrCreateAuditKeys` let two concurrent processes
  generate divergent signing keys. Replaced the check-then-write with a
  write-temp-then-`os.Link` handshake (first-writer-wins; the key path only ever
  becomes visible fully written), so the loser re-reads the winner's key.
  **Note:** the issue's suggested `O_EXCL`-then-write fix has its own window — a
  racer can read the file after `O_EXCL` create but before the write; the
  concurrency test below caught it, hence the link-based approach.
  `cmd/eunox-mcp/audit.go`.

## Testing

- `go build ./...`, `gofmt -l` (clean), `go vet ./...` (clean), `golangci-lint run ./...` (0 issues).
- New regression tests: glob root-vs-empty (#353); audit-mode antecedent arms
  sequenceBlock (#354); escaped-bracket specificity (#356); Redis Start/Stop race
  and empty-ID rejection, in-memory empty-ID rejection (#357, #360); kill switch
  hard-block in audit mode (#358); concurrent audit-key convergence (#365);
  forwarded allow record marked `audit_only` (#364).
- #355 is documentation-only (accepted resolution). #359, #361, #362, #363 are
  verified by reasoning + existing suites; their failure paths (Redis/Sync/Stat/
  reopen errors) cannot be injected cleanly in a unit test without flaky
  OS-fault simulation.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Pzz4kZz7kiCDFv6q1hQbEv)_


eunolabs:
✅ Ready to merge — all session 7gz87n bugs addressed (#353, #354, #355, #356, #357, #358, #359, #360, #361, #362, #363, #364, #365), self-review complete with no outstanding comments, CI green (Test, Lint, Build matrix, demos, OPA comparisons, gofmt, License, Enforcement compliance all passing).

Self-review surfaced and fixed two substantive issues before this point: (1) #359 was referenced but not initially implemented — `handleKill` now propagates kill-store errors; (2) the issue-suggested `O_EXCL` fix for #365 had its own empty-file race window (caught by the new concurrency test), so it was replaced with a write-temp-then-`os.Link` handshake, and a gosec G703 nolint annotation was added for the temp-file cleanup to satisfy CI's golangci-lint v2.12.2.

---

PR #373, opened 2026-06-14
Fix session rev-20260614T0705Z-c837c5 bugs

### LINKED ISSUE #370: [rev-20260614T0705Z-c837c5] checkMaxEntries lacks upper bound — ZRemRangeByRank overflows to no-op for math.MaxInt maxEntries
**Location:** `pkg/callcounter/callcounter.go:188-193` and `pkg/callcounter/redis.go:118`

**Reference:** https://github.com/eunolabs/eunox/blob/2efbcdc/pkg/callcounter/callcounter.go#L188-L193

**Severity:** Medium

**Description:**

`checkMaxEntries` only validates that `maxEntries >= 1` — it has no upper bound:

```go
// pkg/callcounter/callcounter.go:188-193
func checkMaxEntries(maxEntries int) error {
    if maxEntries < 1 {
        return fmt.Errorf("callcounter: maxEntries must be >= 1, got %d", maxEntries)
    }
    return nil
}
```

In `Redis.IncrementAndGet`, the trimming step computes:

```go
// pkg/callcounter/redis.go:118
pipe.ZRemRangeByRank(ctx, windowKey, 0, int64(-(maxEntries + 1)))
```

When `maxEntries = math.MaxInt` (allowed by `checkMaxEntries`):

1. `maxEntries + 1` wraps to `math.MinInt` via signed integer overflow (Go wraps without panic).
2. `-(math.MinInt)` wraps again to `math.MinInt` (two's complement: negating `MinInt` is undefined behaviour in C, but in Go it silently stays `MinInt`).
3. `int64(math.MinInt)` = `math.MinInt64`.
4. `ZREMRANGEBYRANK key 0 -9223372036854775808` — Redis interprets the negative rank from the end; since `stop < start` in absolute terms after rank resolution, this is a **no-op**.

The sorted set is never trimmed and grows without bound, consuming unbounded Redis memory proportional to the number of calls over the window lifetime. The TTL (`2 * windowSec`) is the only safety valve, but within that window the set is unbounded.

Compare with `checkWindowSec`, which explicitly caps at `MaxWindowSeconds` to prevent overflow downstream — `checkMaxEntries` is missing the analogous guard.

**Trigger / repro:**

Any caller that passes `maxEntries = math.MaxInt` to `IncrementAndGet` on the Redis backend. The `sequenceBlock` path uses `maxEntries = sequenceHistoryMaxEntries = 1`, so production traffic is not currently exposed. The risk is a future caller, a configuration path, or a test helper that does not know about the implicit constraint.

**Suggested fix:**

Add a symmetric upper bound to `checkMaxEntries`, analogous to `MaxWindowSeconds`:

```go
const MaxEntries = math.MaxInt32 // fits comfortably in int64 arithmetic

func checkMaxEntries(maxEntries int) error {
    if maxEntries < 1 {
        return fmt.Errorf("callcounter: maxEntries must be >= 1, got %d", maxEntries)
    }
    if maxEntries > MaxEntries {
        return fmt.Errorf("callcounter: maxEntries must be <= %d, got %d", MaxEntries, maxEntries)
    }
    return nil
}
```

Alternatively, add a compile-time assertion in `redis.go` that `maxEntries+1` cannot overflow before the `int64` cast, or use `int64(maxEntries)` before adding 1:

```go
pipe.ZRemRangeByRank(ctx, windowKey, 0, -(int64(maxEntries) + 1))
```

This keeps the arithmetic in `int64` where `math.MaxInt32 + 1` is well within range, and also makes `math.MaxInt64` the only boundary concern (at which point `checkMaxEntries` capping at `MaxEntries` is still the right fix).

**Session:** rev-20260614T0705Z-c837c5

### LINKED ISSUE #371: [rev-20260614T0705Z-c837c5] Cleanup Phase 1 holds exclusive lock during full map scan, unnecessarily blocking Peek readers
**Location:** `pkg/callcounter/memory.go:513-521`

**Reference:** https://github.com/eunolabs/eunox/blob/2efbcdc/pkg/callcounter/memory.go#L513-L521

**Severity:** Medium

**Description:**

Issue #237 introduced a two-phase `Cleanup` to reduce lock-hold time: Phase 1 collects keys to delete, Phase 2 deletes them in batches of 256 re-acquiring the lock per batch. However, Phase 1 still acquires the **exclusive** write lock (`m.mu.Lock()`) for the entire O(N) map scan:

```go
// pkg/callcounter/memory.go:513-521  (Phase 1)
m.mu.Lock()
now := m.now()
var toDelete []string
for key, e := range m.entries {
    if shouldDeleteEntry(e, now) {
    }
}
m.mu.Unlock()
```

Phase 1 only **reads** `m.entries` and `m.now()` — it never modifies the map. A read lock (`m.mu.RLock()`) is sufficient for this phase. Meanwhile, `Peek` was specifically changed to use `m.mu.RLock()` (per issue #239) to allow concurrent reads on the enforcement hot path. That intent is defeated: a `Peek` call that arrives during a cleanup pass is still blocked for the full O(N) Phase 1 duration.

In a gateway deployment with many concurrent clients and a large number of live session×tool pairs, the cleanup interval fires every few minutes and can block the enforcement hot path for non-trivial time.

**Trigger / repro:**

1. Create an `InMemory` store with a large number of entries (tens of thousands of unique session×tool keys).
2. Trigger a `Cleanup` pass.
3. Concurrently call `Peek` — it blocks until Phase 1 completes, even though Phase 1 is purely reading.

**Suggested fix:**

Change Phase 1 to use `RLock`:

```go
m.mu.RLock()
now := m.now()
var toDelete []string
for key, e := range m.entries {
    if shouldDeleteEntry(e, now) {
    }
}
m.mu.RUnlock()
```

Phase 2 (the batch deletion loop) already re-acquires `m.mu.Lock()` correctly. This change lets `Peek` and other concurrent readers proceed during the scan, and only blocks writers during the short per-batch deletion windows.

Note: Phase 2 must still re-check `shouldDeleteEntry` before each delete (or tolerate the TOCTOU where a key became non-stale between Phase 1 and Phase 2). The current implementation already does this implicitly because it only reads `toDelete` (a list of key strings) without assuming the entry is still stale — verify that the Phase 2 deletion is safe to run even if the entry has since been updated (i.e., re-check or use a no-op if the entry is gone).

**Session:** rev-20260614T0705Z-c837c5

### LINKED ISSUE #372: [rev-20260614T0705Z-c837c5] Silent type assertion for retryMicros in IncrementIfBelow produces zero retryAfter on unexpected Redis return type
**Location:** `pkg/callcounter/redis.go:248`

**Reference:** https://github.com/eunolabs/eunox/blob/2efbcdc/pkg/callcounter/redis.go#L248

**Severity:** Low

**Description:**

`IncrementIfBelow` extracts the `retryAfter` hint from the Lua script's return value using a silent type assertion:

```go
// pkg/callcounter/redis.go:248
retryMicros, _ := arr[2].(int64)
```

The Lua script (`incrIfBelowScript`) returns `retryMicros` as a Redis integer (the earliest timestamp in the sorted set, in microseconds). The `go-redis` client converts Redis integers to `int64`, but if a future Redis version, a client library update, or a custom Redis module changes the return encoding (e.g., returning a bulk string or a float), the type assertion fails silently and `retryMicros = 0`.

With `retryMicros = 0`, `retryAfter` becomes `0`, which triggers the fallback in `handleMaxCalls`:

```go
// pkg/enforcement/handlers.go:441-447
retrySec := int64(retryAfter / time.Second)
if retryAfter%time.Second != 0 {
    retrySec++
}
if retrySec < 1 {
    retrySec = int64(mc.WindowSeconds)  // full window fallback
}
```

The fallback masks the bug in production (the client sees the full window as retry-after instead of the actual remaining time), but the type mismatch goes completely undetected — no log entry, no error, no metric.

**Trigger / repro:**

This cannot be triggered by a well-configured stock Redis deployment with an unmodified `go-redis` client. It would surface if:

- A `go-redis` major version change alters how Lua integer returns are decoded.
- A Redis proxy or compatibility layer (Valkey, KeyDB, custom module) returns the integer as a different type.
- A test uses a mock that returns the wrong type.

**Suggested fix:**

Use a two-value type assertion and return an error on failure:

```go
retryMicros, ok := arr[2].(int64)
if !ok {
    return false, 0, fmt.Errorf("callcounter: redis IncrementIfBelow unexpected retryMicros type %T (value %v)", arr[2], arr[2])
}
```

`IncrementIfBelow` already returns `(bool, time.Duration, error)`, so the caller can handle the error fail-closed. This makes the type contract explicit, surfaces the problem immediately in tests or unusual deployments, and leaves no silent path where a wrong `retryAfter` is quietly swallowed.

**Session:** rev-20260614T0705Z-c837c5

eunolabs:
## Bugs addressed (session rev-20260614T0705Z-c837c5)

- Closes #370 — `checkMaxEntries` only enforced a lower bound, so `math.MaxInt` was accepted; in `Redis.IncrementAndGet` the trim computed `-(maxEntries+1)` in `int`, wrapping through `math.MinInt` and making `ZRemRangeByRank` a no-op (sorted set grows unbounded, re-opening #233). **Fix:** added a `MaxEntries` (`math.MaxInt32`) upper bound to `checkMaxEntries`, and compute the negative rank stop in `int64` (`-(int64(maxEntries)+1)`) as defence in depth.
- Closes #371 — `Cleanup` phase 1 held the exclusive write lock for the full O(N) map scan even though it only reads, blocking concurrent readers (notably `Peek`, which took `RLock` per #239) for the whole pass. **Fix:** phase 1 now takes `RLock`; phase 2 deletes still take the write lock per batch.
- Closes #372 — `IncrementIfBelow` decoded the Lua reply with silent type assertions, so a changed encoding (go-redis upgrade, Redis-compatible proxy, or a test mock returning a bulk string/float) defaulted each value to the `int64` zero. A zero `retryMicros` was then masked by the `handleMaxCalls` full-window fallback, leaving the mismatch undetected. **Fix:** extracted the decode into `parseIncrIfBelowReply` with checked assertions that return a structured error and fail closed.

## Testing

- `go build ./...`
- `go test -race -count=1 ./...` (all packages pass)
- `go vet ./...` clean; `gofmt` clean; `make check-license` clean
- New regression tests:
  - `TestRedis_IncrementAndGet_RejectsOversizedMaxEntries` / `TestInMemory_IncrementAndGet_RejectsOversizedMaxEntries` (#370, incl. the `math.MaxInt` case and `MaxEntries` boundary acceptance)
  - `TestInMemory_Cleanup_ConcurrentPeekRaceFree` (#371, race-free concurrent `Peek`/`Cleanup`)
  - `TestParseIncrIfBelowReply` (#372, table-driven wrong-type / wrong-shape cases that fail closed)

---
_Generated by [Claude Code](https://claude.ai/code/session_01FiqfZ69UMSHV5MvtTni1tm)_


pkg/callcounter/callcounter.go
```diff
@@ -162,6 +162,19 @@ const MaxWindowSeconds = min(
...
+// maxEntries only ever appears in int64-safe arithmetic on its own (Redis's
+// ZRemRangeByRank already casts to int64 before the -(maxEntries+1) computation),
+// so this bound is not about preventing duration overflow but about keeping the
+// rank arithmetic well-behaved: without an upper bound, checkMaxEntries would
+// accept math.MaxInt, and -(maxEntries+1) — evaluated in int — would wrap through
+// math.MinInt and turn ZRemRangeByRank's trim into a no-op, re-opening the
+// unbounded sorted-set growth issue #233 closes. math.MaxInt32 (~2.1e9) is far
+// larger than any real retention cap (sequenceBlock uses 1, maxCalls uses the
+// quota) yet leaves maxEntries+1 comfortably inside int on every platform, so the
+// trim arithmetic can never overflow regardless of int width.
```
eunolabs: Self-review (#370): this comment claims `maxEntries+1` "stays comfortably inside int on every platform", but that is false on a 32-bit platform, where `int` is 32-bit and `math.MaxInt32 == math.MaxInt`, so `MaxEntries+1` overflows `int`. The actual overflow protection now comes from the `int64` rank arithmetic in `redis.go` (`-(int64(maxEntries)+1)`), not from this cap. Reword the comment so it doesn't assert a platform-safety property it doesn't have, and explain that on 32-bit the cap equals `math.MaxInt` (imposes no further restriction) with the int64 trim being what guarantees no overflow on either width.

---

pkg/callcounter/redis_test.go
```diff
@@ -122,6 +123,29 @@ func TestRedis_IncrementAndGet_RejectsNonPositiveMaxEntries(t *testing.T) {
...
+// the upper bound, math.MaxInt would be accepted and the ZRemRangeByRank trim
+// computed -(maxEntries+1) in int, wrapping through math.MinInt and turning the
+// trim into a no-op — leaving the sorted set to grow unbounded (re-opening issue
+// #233). Fail closed instead, matching the windowSec upper-bound guard.
+func TestRedis_IncrementAndGet_RejectsOversizedMaxEntries(t *testing.T) {
+	mr := miniredis.RunT(t)
+	client := redis.NewClient(&redis.Options{Addr: mr.Addr()})
+	t.Cleanup(func() { _ = client.Close() })
+
+	counter := callcounter.NewRedis(client)
```
eunolabs, line 136: Self-review (#370): the oversized cases `{callcounter.MaxEntries + 1, math.MaxInt}` assume `math.MaxInt > MaxEntries`. That holds on 64-bit (CI), but on a 32-bit platform `math.MaxInt == MaxEntries`, so the `math.MaxInt` case would be *accepted* and this assertion would fail (and `MaxEntries+1` overflows to a negative there). Guard the oversized cases on `math.MaxInt > callcounter.MaxEntries` so the test is correct on both int widths; the boundary-acceptance check should always run. Same fix needed in the InMemory counterpart in `callcounter_test.go`.

---

eunolabs (commented):
Self-review round 1: two substantive portability issues in the #370 fix (incorrect 32-bit safety claim in the MaxEntries comment, and tests that assume math.MaxInt > MaxEntries which is false on 32-bit). Addressing now.

---

pkg/callcounter/callcounter.go
eunolabs: Fixed in ea38e4b: reworded the `MaxEntries` comment to drop the false "inside int on every platform" claim — it now states the int64 trim arithmetic in `redis.go` is the actual overflow guard, that on 32-bit the cap equals `math.MaxInt` (no further restriction), and that the cap's role is to reject absurd/misconfigured values and keep the bound explicit.

---

pkg/callcounter/redis_test.go
eunolabs: Fixed in ea38e4b: extracted a shared `oversizedMaxEntries()` helper that returns the oversized cases only when `math.MaxInt > callcounter.MaxEntries` (empty on 32-bit, where nothing is representable above the boundary). Both the Redis and InMemory tests now use it, and both still always assert that the boundary value `MaxEntries` is accepted.

---

eunolabs:
✅ Ready to merge — all session rev-20260614T0705Z-c837c5 bugs addressed (#370, #371, #372), self-review complete with no outstanding comments, tests passing.

---

PR #377, opened 2026-06-14
Fix session rev-20260614T0805Z-a1aa77 bugs

### LINKED ISSUE #374: [rev-20260614T0805Z-a1aa77] TokenCache.Start() has no guard against multiple calls; spawns redundant goroutines
**Location:** `pkg/capability/token_cache.go:108-110`
**Reference:** https://github.com/eunolabs/eunox/blob/8e7ab40579f92e2074c4741657b80217fceb59cb/pkg/capability/token_cache.go#L108-L110
**Severity:** Medium

**Description:**
`Start()` is:

```go
func (c *TokenCache) Start(ctx context.Context) {
    go c.cleanupLoop(ctx)
}
```

There is no guard — no `sync.Once`, no `started bool`, no check of `c.stopped` — so every call to `Start()` launches a fresh `cleanupLoop` goroutine. Three concrete problems follow:

1. **Redundant goroutines / spurious lock contention.** Each goroutine calls `purgeExpired()` on every tick, holding the write lock for an O(n) scan. Two goroutines started accidentally in the same test or initialization path double the lock-hold time on the enforcement hot path with no benefit.

2. **`Stop()` then `Start()` silently breaks the cache.** `Stop()` closes `c.stopCh` (guarded by `c.stopped`). Any goroutine started after `Stop()` enters `cleanupLoop`, immediately receives on the already-closed `stopCh`, and exits. From that point the cache runs forever without cleanup: expired entries pile up, the `insertOrder` list grows without bound, and eviction based on `MaxSize` still works but TTL-expired entries are never reaped.

3. **No way to wait for the goroutine to exit.** There is no `sync.WaitGroup`, so a caller who calls `Stop()` and then immediately shuts down the parent struct cannot be certain the goroutine has released the mutex and exited. In tests this surfaces as the goroutine observing freed or zero-valued cache state.

**Trigger / repro:**
```go
cache := capability.NewTokenCache(capability.TokenCacheConfig{})
cache.Start(ctx)
cache.Start(ctx) // second goroutine — no error, no panic
// Both goroutines now compete to purge; after Stop():
cache.Stop()
cache.Start(ctx) // goroutine exits immediately; cleanup is now permanently dead
```

**Suggested fix:**
Add a `cleanupOnce sync.Once` field (mirroring `callcounter.InMemory.cleanupOnce`) and a `sync.WaitGroup` for shutdown:

```go
func (c *TokenCache) Start(ctx context.Context) {
    c.startOnce.Do(func() {
        c.wg.Add(1)
        go func() {
            defer c.wg.Done()
            c.cleanupLoop(ctx)
        }()
    })
}

func (c *TokenCache) Stop() {
    c.mu.Lock()
    if c.stopped { c.mu.Unlock(); return }
    c.stopped = true
    c.mu.Unlock()
    close(c.stopCh)
    c.wg.Wait() // wait for the goroutine to exit
}
```

**Session:** rev-20260614T0805Z-a1aa77

### LINKED ISSUE #375: [rev-20260614T0805Z-a1aa77] verifyAuditLog corrupts chain state when json.Unmarshal fails on a malformed line
**Location:** `cmd/eunox-mcp/audit.go:817-851`
**Reference:** https://github.com/eunolabs/eunox/blob/8e7ab40579f92e2074c4741657b80217fceb59cb/cmd/eunox-mcp/audit.go#L817-L851
**Severity:** Medium

**Description:**
In `verifyAuditLog`, each scanned line is decoded with:

```go
var rec auditRecord
_ = json.Unmarshal(line, &rec) // best-effort: a malformed line fails HMAC below
```

The error is intentionally discarded. When `json.Unmarshal` fails on a truly malformed line (binary corruption, a mid-write truncation, a non-JSON marker), `rec` retains its zero value: `rec.HMAC = ""` and `rec.Seq = 0`.

Immediately after, the chain state is updated unconditionally:

```go
havePrev = true
prevHMAC = rec.HMAC   // = "" (wrong)
prevSeq  = rec.Seq    // = 0  (wrong)
```

The next valid record (say seq N+1, with `PrevHMAC` = HMAC of the last written record) is then checked against the stale `prevHMAC = ""` and `prevSeq = 0`:

```go
if rec.PrevHMAC != prevHMAC {   // "sha256:real..." != "" → false chain break
    res.chainBreaks++
}
if rec.Seq != prevSeq+1 {       // N+1 != 0+1 → false seq gap (unless N=0)
    res.chainBreaks++
}
```

A single disk-corrupted line produces **two false positives** at the malformed line's position and **two more** at the following valid line (four `chainBreaks` total). After the following valid record is processed, `prevHMAC`/`prevSeq` are restored to correct values and all subsequent records verify cleanly.

**Practical impact:**

- An operator running `eunox-mcp audit-verify` after a disk I/O error will see spurious chain break reports and conclude the log was tampered with, triggering unnecessary incident response.
- A genuine deletion or insertion that falls immediately after a corrupted line is masked: because `prevHMAC` was already wrong, the next record's chain link cannot be validated and the deletion goes undetected.

**Trigger / repro:**
1. Write a valid audit log with >= 3 records (seqs 1, 2, 3).
2. Corrupt line 2 so it is not valid JSON (e.g., overwrite its first byte).
3. Run `eunox-mcp audit-verify` on the log.
4. Observe 4 chain-break reports instead of 1 (or 0, since the corruption itself is detected by HMAC).

**Suggested fix:**
Only update `prevHMAC`/`prevSeq` from a successfully decoded record. The simplest approach is to check whether `json.Unmarshal` succeeded before updating chain state:

```go
var rec auditRecord
unmarshalOK := json.Unmarshal(line, &rec) == nil

if havePrev && unmarshalOK {
    if rec.PrevHMAC != prevHMAC {
        res.chainBreaks++
        fmt.Fprintf(out, "CHAIN BREAK at seq %d: ...\n", rec.Seq)
    }
    if rec.Seq != prevSeq+1 {
        res.chainBreaks++
        fmt.Fprintf(out, "SEQ GAP: ...\n", rec.Seq, prevSeq)
    }
}

if unmarshalOK {
    if !havePrev { res.firstSeq = rec.Seq }
    havePrev = true
    prevHMAC = rec.HMAC
    prevSeq  = rec.Seq
}
// HMAC verification still runs regardless (catches malformed + tampered lines)
```

This way a malformed line is counted as `invalid` (HMAC fails) without poisoning the chain state for subsequent records.

**Session:** rev-20260614T0805Z-a1aa77

### LINKED ISSUE #376: [rev-20260614T0805Z-a1aa77] jwksCache.doRefresh: waiting callers receive the leader's cancelled-context error even when their own context is still valid
**Location:** `cmd/eunox-mcp/pdp_jwt.go:243-262`
**Reference:** https://github.com/eunolabs/eunox/blob/8e7ab40579f92e2074c4741657b80217fceb59cb/cmd/eunox-mcp/pdp_jwt.go#L243-L262
**Severity:** Medium

**Description:**
`jwksCache` uses a hand-rolled singleflight mechanism to deduplicate concurrent JWKS refreshes. When multiple goroutines call `doRefresh` simultaneously, one becomes the **leader** and performs the HTTP fetch; the others become **waiters** and block on `done`:

```go
select {
case <-done:         // leader finished
case <-ctx.Done():   // waiter's own context
    return nil, ctx.Err()
}
c.sfMu.Lock()
jwks, err := c.sfJWKS, c.sfErr   // ← consume the leader's stored result
c.sfMu.Unlock()
return jwks, err
```

If the **leader's** context is cancelled or times out (e.g., a per-request 500 ms deadline on the incoming HTTP connection), `guardedFetch(ctx)` returns `ctx.Err()`, and the leader stores that error in `c.sfErr` then closes `done`. Every waiter unblocks on `<-done`, reads `c.sfErr`, and returns the leader's `ctx.Err()` — even if the waiter's own context still has minutes of budget remaining.

The comment on `guardedFetch` acknowledges the circuit-breaker angle ("a burst of client disconnects ... could open the breaker") but the fix there only avoids recording a breaker failure when the leader cancels. It does not prevent the error from propagating to waiters with valid contexts.

**Concrete scenario:**
1. The JWKS TTL expires. Ten concurrent requests all call `doRefresh`.
2. Request 1 (leader) has a 50 ms per-request timeout; requests 2-10 have 30 s budgets.
3. The JWKS endpoint is slow (say 200 ms). Request 1's context fires at 50 ms.
4. Leader stores `sfErr = context.DeadlineExceeded`, closes `done`.
5. Requests 2-10 unblock, read `sfErr`, and return 401 to their clients — nine legitimate clients are rejected due to one short-lived request timing out.

This is effectively a self-inflicted auth outage triggered by any single short-deadline caller racing the JWKS cache TTL.

**Trigger / repro:**
- Configure `--jwks-uri` with a slow JWKS endpoint (or add artificial latency in tests).
- Issue one request with a very short context deadline concurrently with many normal requests, all landing when the cache is stale.
- Observe 401s on the normal requests due to the short-deadline request's cancellation error propagating via `sfErr`.

**Suggested fix:**
Do not surface the leader's context error to waiters whose contexts are still live. The simplest fix is to have waiters re-attempt a fresh fetch when the shared result is a context error and their own context is still valid:

```go
jwks, err := c.sfJWKS, c.sfErr
c.sfMu.Unlock()
if err != nil && errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
    if ctx.Err() == nil {
        // Leader was cancelled but we still have budget: retry independently.
        return c.guardedFetch(ctx)
    }
}
return jwks, err
```

Alternatively, store only non-context errors in `sfErr` from `guardedFetch`, or adopt `golang.org/x/sync/singleflight` with context-aware `DoChan` semantics.

**Session:** rev-20260614T0805Z-a1aa77

eunolabs:
## Bugs addressed (session rev-20260614T0805Z-a1aa77)

- Closes #374 — `TokenCache.Start()` had no guard, so every call spawned a fresh `cleanupLoop` goroutine (redundant lock contention on the enforcement hot path), and `Start()` after `Stop()` launched a goroutine that immediately drained the closed `stopCh` and died, leaving the cache permanently un-reaped. `Start` now uses `sync.Once` (launch at most once; a no-op after `Stop`), and `Stop` waits on a `sync.WaitGroup` so the goroutine has exited before it returns.
- Closes #375 — `verifyAuditLog` discarded the `json.Unmarshal` error, so a corrupted/truncated line decoded to the zero `auditRecord` (`HMAC=""`, `Seq=0`) and was adopted as the chain's `prev` state — fabricating spurious chain breaks on the next valid record and masking a real deletion landing right after the corruption. Chain linkage is now maintained only across successfully decoded records; the malformed line is still caught by per-record HMAC verification (counted `invalid`).
- Closes #376 — A JWKS singleflight waiter inherited the leader's `ctx.Err()`: if the leader's per-request deadline fired mid-fetch, every joined waiter returned that error, so one short-deadline caller racing the JWKS TTL rejected every concurrent caller (a self-inflicted auth outage). Waiters now detect a context-error result and, when their own context is still live, fetch independently instead.

## Testing
- `gofmt -l` clean on every touched file; `go vet ./cmd/eunox-mcp/ ./pkg/capability/` clean.
- New regression tests, each reproducing the original bug:
  - `TestTokenCache_StartIsIdempotent` — repeated `Start` reaps via a single goroutine; `Stop` does not hang; `Start` after `Stop` is a no-op.
  - `TestAuditChain_MalformedLineDoesNotPoisonChain` — a trailing corrupted line yields 0 chain breaks (was 2) and 1 invalid record.
  - `TestJWKSCache_WaiterRetriesWhenLeaderContextCancelled` — leader cancelled mid-fetch; waiters with live contexts still get keys.

---
_Generated by [Claude Code](https://claude.ai/code/session_01D7mULuhM9Jacv5FD7J6LMN)_


eunolabs (commented):
Self-review (round 1).

**Finding (addressed in this PR):** `docs/threat-model-mcp.md` — the #375 fix changes `audit-verify` behavior on a documented subsystem. The threat model explicitly documents the `seq`/`prev_hmac` chain semantics (§3.4, line ~231) and even discusses "no spurious chain break" for the legacy-upgrade boundary (line ~251), and CONTRIBUTING requires a docs update in the same PR for behavior changes. The original commits touched only `cmd/eunox-mcp/audit.go` + tests, leaving the threat model silent on the new malformed-line handling.

Resolution: commit `9c92c7b` adds a §8 change-log entry (1.15) and a sentence to the operational guidance (line ~466) stating that a JSON-decode-failed line is counted `invalid` by its HMAC but excluded from chain linkage, so it no longer fabricates a chain break against the next intact record.

**Other checks (no action needed):**
- `cmd/eunox-mcp/pdp_jwt.go:262-279` (#376) — waiter retry triggers only on a context-class error with a still-live waiter context; the breaker still bounds a genuinely bad endpoint, and the cache-update mirrors the existing force-branch (lines 268-273). The only cost is up to N independent fetches on the rare leader-cancellation race, which is exactly the issue's recommended trade-off.
- `pkg/capability/token_cache.go:118-141` (#374) — `sync.Once` + `WaitGroup`; `Stop` releases `c.mu` before `wg.Wait()`, so the purge goroutine can finish and exit — no deadlock. Intended lifecycle is Start-at-setup / Stop-at-teardown (not concurrent).
- #375 legacy interaction — pre-chain records are valid JSON (`unmarshalOK=true`), so the documented "no spurious chain break on upgrade" behavior is preserved (full suite green).

Local: `go test -race -count=1 ./...` all pass; `gofmt -l` clean; `go vet` clean. Awaiting CI.

---

eunolabs:
✅ Ready to merge — all session rev-20260614T0805Z-a1aa77 bugs addressed (#374, #375, #376), self-review complete with no outstanding comments. All 42 CI checks pass (Test, Lint, gofmt, License Headers, builds on all platforms, govulncheck, Trivy, CodeQL, demos, OPA comparisons); `go test -race ./...` green locally.

---

PR #382, opened 2026-06-14
Fix session rev-20260614T0905Z-4231ad bugs

### LINKED ISSUE #378: [rev-20260614T0905Z-4231ad] verifyECDSA does not enforce algorithm-to-curve binding, accepting non-standard curves via co-signature
**Location:** `pkg/capability/proofs.go:136-160`
**Reference:** https://github.com/eunolabs/eunox/blob/8e7ab40579f92e2074c4741657b80217fceb59cb/pkg/capability/proofs.go#L136-L160
**Severity:** high

**Description:**
`verifyECDSA` derives the expected IEEE P1363 signature length from the *key's own curve parameters* rather than from the algorithm string it received. It never asserts that the curve in the key matches what the JWS algorithm mandates:

| Algorithm | Required curve (RFC 7518 §3.4) |
|-----------|--------------------------------|
| ES256     | P-256                          |
| ES384     | P-384                          |
| ES512     | P-521                          |

The only guard today is:
```go
keyByteLen := (ecKey.Curve.Params().BitSize + 7) / 8
if len(sigBytes) != 2*keyByteLen {
    return fmt.Errorf("invalid ECDSA signature length …")
}
```

Any ECDSA-compatible curve whose key component byte size matches that of the algorithm's intended curve bypasses this check. `secp256k1` (the Bitcoin/Ethereum curve, supported by several Go libraries) has 32-byte components—identical to P-256—so an `ES256` entry in a co-signature would accept a secp256k1 key and a secp256k1-signed payload with the wrong curve's security profile.

**Trigger / repro:**
1. A co-issuer registers a DID document with a secp256k1 public key (32-byte components).
2. Signs a token compact serialization with secp256k1 + SHA-256.
3. Presents an `IssuerSignature{Algorithm: "ES256", ...}` with that signature.
4. A resolver returns the secp256k1 key.
5. `verifyECDSA` length-checks 64 bytes against `2*32 = 64` ✓, then calls `ecdsa.Verify` with the secp256k1 key — verification succeeds.

The co-signature is accepted despite not using P-256 as ES256 requires.

**Suggested fix:**
After the type assertion to `*ecdsa.PublicKey`, add an explicit curve check before the length check:

```go
// verifyECDSA dispatches are: ES256→P-256, ES384→P-384, ES512→P-521.
// The allowed curve is passed in from verifyRawSignature so the binding
// is enforced at the entry point, not scattered across all call sites.
func verifyECDSA(message, sigBytes []byte, key gocrypto.PublicKey, hash gocrypto.Hash, wantCurve elliptic.Curve) error {
    ecKey, ok := key.(*ecdsa.PublicKey)
    if !ok {
        return errors.New("key type mismatch: expected *ecdsa.PublicKey")
    }
    if ecKey.Curve != wantCurve {
        return fmt.Errorf("curve mismatch: algorithm requires %s, key uses %s",
            wantCurve.Params().Name, ecKey.Curve.Params().Name)
    }
    // … rest unchanged
}
```

Update call sites in `verifyRawSignature`:
```go
case "ES256":
    return verifyECDSA(message, sigBytes, key, gocrypto.SHA256, elliptic.P256())
case "ES384":
    return verifyECDSA(message, sigBytes, key, gocrypto.SHA384, elliptic.P384())
case "ES512":
    return verifyECDSA(message, sigBytes, key, gocrypto.SHA512, elliptic.P521())
```

**Session:** rev-20260614T0905Z-4231ad

### LINKED ISSUE #379: [rev-20260614T0905Z-4231ad] IssuerSignature.IssuedAt is never validated in VerifyCoSignatures
**Location:** `pkg/capability/proofs.go:72-105`, `pkg/capability/token.go:77`
**Reference:** https://github.com/eunolabs/eunox/blob/8e7ab40579f92e2074c4741657b80217fceb59cb/pkg/capability/proofs.go#L72-L105
**Severity:** medium

**Description:**
`IssuerSignature` carries an `IssuedAt` field (Unix timestamp) that documents *when* a co-issuer signed the token. `verifyOneSig` (and `VerifyCoSignatures`) never read this field; any timestamp—including values far in the future or impossibly distant in the past—is silently accepted.

Two concrete problems arise from this gap:

1. **Future-dated co-signatures are accepted.** A co-issuer can produce a co-signature with `IssuedAt` set to years in the future. The verifier has no way to distinguish this from a valid signature, and the field provides false assurance that it carries a meaningful time constraint.

2. **Co-signatures pre-dating the token's own `iat` are accepted.** A co-issuer could sign with `IssuedAt` set before the primary JWT was even issued (e.g., zero, or an epoch from years ago). Defenders relying on audit logs to detect anomalies lose the ability to spot this inconsistency at verification time.

While the cryptographic binding (co-signature is over the full JWT compact serialization, which includes the primary issuer's signature) prevents a stale co-signature from being moved to a different token, it does not prevent a co-issuer from setting a misleading timestamp that erodes the auditability and integrity guarantees the field is meant to provide.

**Trigger / repro:**
Produce any valid co-signature and set `IssuedAt` to `0` (epoch), `math.MaxInt64` (far future), or a time predating the JWT's `iat`. `VerifyCoSignatures` will return `nil` — no error.

**Suggested fix:**
After decoding the signature bytes and before (or after) the key lookup, validate the co-signature's `IssuedAt` relative to the token's own `iat` and the current time:

```go
func verifyOneSig(ctx context.Context, message []byte, sig IssuerSignature, resolver PublicKeyResolver, tokenIAT int64, now time.Time) error {
    // …existing empty-field checks…
    if sig.IssuedAt != 0 {
        // Co-signature must not be from the future.
        if sig.IssuedAt > now.Unix()+clockLeeway {
            return fmt.Errorf("co-signature IssuedAt %d is in the future", sig.IssuedAt)
        }
        // Co-signature must not pre-date the token itself (with leeway).
        if tokenIAT > 0 && sig.IssuedAt < tokenIAT-clockLeeway {
            return fmt.Errorf("co-signature IssuedAt %d predates token iat %d", sig.IssuedAt, tokenIAT)
        }
    }
    // … rest unchanged
}
```

`VerifyCoSignatures` would receive the token's `iat` and a clock (or derive them from the parsed `TokenPayload`). A reasonable `clockLeeway` matches the 60-second leeway already used in `JWKSClient.VerifyToken`.

**Session:** rev-20260614T0905Z-4231ad

### LINKED ISSUE #380: [rev-20260614T0905Z-4231ad] Unbounded key iteration from PublicKeyResolver enables CPU-exhaustion DoS in verifyOneSig
**Location:** `pkg/capability/proofs.go:88-104`
**Reference:** https://github.com/eunolabs/eunox/blob/8e7ab40579f92e2074c4741657b80217fceb59cb/pkg/capability/proofs.go#L88-L104
**Severity:** medium

**Description:**
`verifyOneSig` calls `resolver.ResolvePublicKeys` and iterates over every returned key without any upper-bound check:

```go
keys, err := resolver.ResolvePublicKeys(ctx, sig.IssuerDID)
// … no len check beyond "empty" …
var lastErr error
for _, key := range keys {
    if verifyErr := verifyRawSignature(message, sig.Algorithm, sigBytes, key); verifyErr == nil {
        return nil
    } else {
        lastErr = verifyErr
    }
}
```

Each iteration performs a full asymmetric cryptographic operation (ECDSA verify, RSA verify, or Ed25519 verify). An attacker or a compromised DID resolver can return an arbitrarily large number of keys, forcing the verifier to burn CPU proportional to `len(keys)` per co-signature. With `N` co-signatures each resolving to `M` keys, the work is `O(N × M)` cryptographic operations before a decision is reached.

In the gateway deployment path where co-signature verification is invoked per inbound MCP request, this creates a practical DoS vector: a single malicious co-issuer DID that resolves to thousands of keys can degrade or stall the proxy for all tenants.

**Trigger / repro:**
1. Register a DID that resolves to 10,000 RSA-2048 public keys.
2. Present a capability token with a co-signature from that DID.
3. `VerifyCoSignatures` will attempt 10,000 RSA verifications before returning an error.

**Suggested fix:**
Apply a hard cap on the number of keys fetched from the resolver before the loop:

```go
const maxResolverKeys = 20 // co-issuer DID documents typically carry ≤ 5 keys

keys, err := resolver.ResolvePublicKeys(ctx, sig.IssuerDID)
if err != nil {
    return fmt.Errorf("resolve co-issuer keys: %w", err)
}
if len(keys) == 0 {
    return errors.New("no public keys found for co-issuer")
}
if len(keys) > maxResolverKeys {
    return fmt.Errorf("co-issuer DID returned %d keys, limit is %d", len(keys), maxResolverKeys)
}
```

The constant can be made configurable via `VerifyCoSignatures` options if needed, but a compile-time default that fails closed is preferable to no bound at all.

**Session:** rev-20260614T0905Z-4231ad

### LINKED ISSUE #381: [rev-20260614T0905Z-4231ad] verifyRSAPSS uses nil PSSOptions, accepting any PSS salt length contrary to RFC 7518
**Location:** `pkg/capability/proofs.go:179-194`
**Reference:** https://github.com/eunolabs/eunox/blob/8e7ab40579f92e2074c4741657b80217fceb59cb/pkg/capability/proofs.go#L179-L194
**Severity:** low

**Description:**
`verifyRSAPSS` calls `rsa.VerifyPSS` with `nil` for the options parameter:

```go
if err := rsa.VerifyPSS(rsaKey, hash, digest, sigBytes, nil); err != nil {
```

When `opts` is `nil`, Go's `(*PSSOptions).saltLength()` returns `PSSSaltLengthAuto` (0), which instructs the verifier to *auto-recover* the salt length from the signature encoding rather than requiring a specific value. This means the verifier will accept PSS signatures with any salt length, including `sLen = 0` (the minimum, which provides no randomisation benefit).

RFC 7518 §3.5 states:

> The size in octets of the salt value *s* SHOULD equal the size in octets of the hash function output.

The SHOULD is deliberately not a MUST, so this is a standards-compliance deviation rather than a broken signature check. However, for a security-critical policy proxy enforcing capability tokens, the permissive `Auto` mode weakens the verification contract:

- A signer using a non-conforming library that emits `sLen = 0` PSS signatures will pass here without warning.
- Signatures that would be rejected by a strict RFC 7518 verifier (e.g., `sLen` does not equal hash output length) are silently accepted.

**Trigger / repro:**
Sign a PS256 message with `rsa.SignPSS(..., &rsa.PSSOptions{SaltLength: 0})`.
`VerifyCoSignatures` will return `nil` — the zero-salt PSS signature is accepted.

**Suggested fix:**
Pass explicit options that enforce the hash-output-length salt:

```go
opts := &rsa.PSSOptions{SaltLength: rsa.PSSSaltLengthEqualsHash}
if err := rsa.VerifyPSS(rsaKey, hash, digest, sigBytes, opts); err != nil {
    return fmt.Errorf("RSA PSS verification failed: %w", err)
}
```

`rsa.PSSSaltLengthEqualsHash` (-1 in Go's stdlib) instructs both signing and verification to use `sLen == hLen`, matching the RFC 7518 recommendation and the behaviour of the standard `rsa.SignPSS` helper when no options are passed.

**Session:** rev-20260614T0905Z-4231ad

eunolabs:
## Bugs addressed (session rev-20260614T0905Z-4231ad)

All four issues are in `pkg/capability/proofs.go` (co-signature verification, `VerifyCoSignatures`).

- Closes #378 — `verifyECDSA` derived the expected signature length from the *key's own* curve rather than the curve the JWS algorithm mandates, so any curve with a matching component byte size (e.g. secp256k1 vs P-256) was accepted under ES256. Now the required curve (RFC 7518 §3.4: ES256→P-256, ES384→P-384, ES512→P-521) is passed from `verifyRawSignature` and the key's curve must match, with a nil-safe mismatch error.
- Closes #379 — `IssuerSignature.IssuedAt` was never read, letting a co-issuer future-date or back-date a signature undetected. A non-zero `IssuedAt` is now rejected if it is in the future (60s leeway, mirroring `JWKSClient.VerifyToken`) or predates the token's own `iat` (parsed best-effort from the JWT compact serialization). Added a `WithClock` option for deterministic testing.
- Closes #380 — `verifyOneSig` looped over every key the resolver returned, each a full asymmetric verification, so a malicious/compromised resolver could drive an O(keys) CPU-exhaustion DoS on the per-request path. The trial set is now capped at `maxResolverKeys` (20) and fails closed past it.
- Closes #381 — `verifyRSAPSS` passed `nil` PSSOptions (`PSSSaltLengthAuto`), accepting PSS signatures with any salt length, including 0. The salt is now pinned to `PSSSaltLengthEqualsHash` per RFC 7518 §3.5. The test signer (which used `nil` = max salt, non-conforming) was corrected to emit RFC-conforming signatures.

## Testing
- New table-driven tests in `pkg/capability/proofs_test.go` covering each fix: ECDSA curve mismatch across the full ES256/384/512 matrix; PSS zero-salt rejection; key-cap rejection (and inclusive at-cap acceptance); and IssuedAt future/predates rejection plus valid/zero acceptance.
- `make check-fmt`, `make check-license`, `go vet`, and `golangci-lint run` — all clean.
- Existing valid-signature tests continue to pass unchanged (zero `IssuedAt` leaves the new timestamp checks inert, preserving opaque-token behavior).

https://claude.ai/code/session_01BUB4bpLfDzKpot2tyy677Y

---
_Generated by [Claude Code](https://claude.ai/code/session_01BUB4bpLfDzKpot2tyy677Y)_


eunolabs:
✅ Ready to merge — all session rev-20260614T0905Z-4231ad bugs addressed (#378, #379, #380, #381), self-review complete with no outstanding comments, tests passing.

Self-review notes:
- Each fix maps directly to its issue and resolves it (ECDSA curve binding, IssuedAt validation, key-set cap, RSA-PSS salt pinning).
- Added allow/deny/edge tests per fix; full `go test -race ./...` passes locally and `pkg/capability` reports 97.7% coverage in CI.
- The earlier red **Test** check was the pre-existing flaky `pkg/killswitch` Redis pub/sub test (`connection refused`), unrelated to this diff (which only touches `pkg/capability` + docs). Re-running the failed jobs turned CI fully green.
- Conventions intact: SPDX headers, signed Conventional Commits, no emojis in source, docs updated, fail-closed invariant preserved.

---

Issue #319, opened 2026-06-14
bug: redactFields fails the entire tools/call response when any text content item is not JSON


eunolabs:
## Summary

`redactJSONText` in `cmd/eunox-mcp/pdp.go` is called for every `type=text` content item when `redactFields` is configured. If that item is not parseable JSON, the function returns an error and `applyRedactObligs` returns that error to the transport — failing the entire response:

```go
// pdp.go:868
func redactJSONText(text string, paths []string) (string, error) {
    var val interface{}
    if err := json.Unmarshal([]byte(text), &val); err != nil {
        return "", fmt.Errorf("redactFields: text content is not parseable JSON; cannot verify redaction of %v (fail closed)", paths)
    }
    ...
}
```

The transport (stdio.go:499, http.go:759) returns an error to the MCP client when `applyRedactObligs` fails.

## Impact

`redactFields` is fundamentally incompatible with any tool that ever emits plain-text content — which includes nearly all tools when they return error messages (`"Error: file not found"`, stack traces, etc.). A tool that normally returns JSON but returns a plain-text error under failure would cause `redactFields` to block the error message rather than forward it.

Operators who add `redactFields` to a constraint don't expect it to change the success/failure behaviour of the tool; they expect structured field redaction on success and pass-through on non-JSON output.

## Steps to reproduce

1. Configure a manifest with `directives: [{type: redactFields, fields: [password]}]` on any tool.
2. Exercise the tool in a way that returns a plain-text error string in a `type=text` content item.
3. Observe that the client receives a proxy error instead of the upstream error response.

## Expected behaviour

When a text content item is not valid JSON, `redactFields` should either:
- **Skip redaction** for that item (the named fields cannot exist in non-JSON text) and forward it unchanged, or
- **Strip the item** from the response (more conservative, but preserves the current intent of "no partial redaction").

The current behaviour — fail the whole response — is disproportionate: the tool's non-JSON text carries no structured field to redact, so there is no risk of leaking a named field. The fail-closed comment in the code conflates "cannot verify absence of a named field in free-form text" with "must fail the response", but free-form text has no named fields at all.

## Note

The code comment says:
> "Text that is not parseable JSON fails the whole response closed: path-based redaction cannot prove the named field is absent from free-form text, so forwarding it would be a fail-open leak."

This reasoning is overly conservative for plain-text items (as opposed to text that *should* be JSON but is malformed). A distinction between "this text is intentionally plain text" and "this text is malformed JSON" would allow the right policy.

PR #394, opened 2026-06-14
Fix session rev-20260614T100Z review bugs (#383–#389)

### LINKED ISSUE #383: [rev-20260614T1005Z-18c713] checkAuth rejects all JWT requests when --auth-token and --jwks-uri are both configured
**Location:** `cmd/eunox-mcp/http.go:425-438` (checkAuth), `cmd/eunox-mcp/http.go:542-568` (handleMCP)

**Reference:** https://github.com/eunolabs/eunox/blob/3170cc052ead7e4ffd08a6299656c637920d91c8/cmd/eunox-mcp/http.go#L425-L438

**Severity:** High

**Description:**

`handleMCP` calls `checkAuth` unconditionally before the JWT PDP check:

```go
func (p *HTTPProxy) handleMCP(w http.ResponseWriter, r *http.Request) {
    if !p.checkAuth(w, r) {   // static token check runs FIRST
        return
    }
    // ...
    if p.jwtPDP != nil {      // JWT check never reached
        ctx, err := p.jwtPDP.ValidateToken(r.Context(), authHeader)
```

When `p.authToken != ""`, `checkAuth` requires the `Authorization: Bearer` value to equal the static `authToken` via `hmac.Equal`. A client sending a valid IdP-issued JWT sends a token like `eyJhbGci...` which will never equal the short static secret, so `checkAuth` returns `false` and a 401 is sent before the JWT PDP ever runs.

Both mechanisms are wired independently in `serveHTTPGateway`:
```go
proxy := NewHTTPProxyGateway(HTTPGatewayOptions{
    AuthToken: cfg.Listen.AuthToken, // set from listen.authToken
    JWTPDP:    gwJWTPDP,             // set from --jwks-uri
    ...
})
```

There is no startup validation rejecting this combination, so it can be reached in production.

**Trigger / repro:**

Configure the gateway with both `listen.authToken` in the YAML and `--jwks-uri` on the CLI. Send any request with a valid IdP JWT (`Authorization: Bearer <jwt>`). The proxy returns 401 immediately from `checkAuth`; the JWT PDP is never consulted. JWT auth is completely non-functional in this configuration.

**Suggested fix:**

Pick one of:

1. **Document mutual exclusivity and add a startup error.** In `serveHTTPGateway`, after constructing `gwJWTPDP`, reject the config if both `cfg.Listen.AuthToken != ""` and `gwJWTPDP != nil`:
   ```go
   if cfg.Listen.AuthToken != "" && gwJWTPDP != nil {
       return fmt.Errorf("listen.authToken and --jwks-uri are mutually exclusive: JWT mode handles all authentication; remove listen.authToken when --jwks-uri is set")
   }
   ```

2. **Skip `checkAuth` when JWT PDP is active.** In `handleMCP`, only call `checkAuth` when `p.jwtPDP == nil`:
   ```go
   if p.jwtPDP == nil && !p.checkAuth(w, r) {
       return
   }
   ```
   This treats the static token and the JWT as alternative auth layers.

Option 1 is simpler and gives the operator a clear error rather than silently defeating security configuration.

**Session:** rev-20260614T1005Z-18c713

### LINKED ISSUE #384: [rev-20260614T1005Z-18c713] buildWWWAuthenticate embeds metaURL verbatim into quoted-string with no escaping
**Location:** `cmd/eunox-mcp/oauth_metadata.go:45-57`

**Reference:** https://github.com/eunolabs/eunox/blob/3170cc052ead7e4ffd08a6299656c637920d91c8/cmd/eunox-mcp/oauth_metadata.go#L45-L57

**Severity:** Medium

**Description:**

`buildWWWAuthenticate` constructs the `resource_metadata` parameter by embedding `metaURL` directly inside double-quotes with no escaping:

```go
if metaURL != "" {
    b.WriteString(`, resource_metadata="`)
    b.WriteString(metaURL)   // raw value, no escaping
    b.WriteString(`"`)
}
```

If `metaURL` contains a double-quote character (`"`), the closing `"` of the `resource_metadata` quoted-string is emitted prematurely, producing a malformed `WWW-Authenticate` header. For example, if `--oauth-resource` is set to `https://example.com/path"with-quote`, the output is:

```
Bearer realm="eunox", resource_metadata="https://example.com/path"with-quote"
```

This is a syntactically invalid auth-param per RFC 7235 §2.1 (which requires quoted-string values to use `\"-escaped` inner quotes) and would cause RFC-compliant OAuth clients to fail to parse the challenge and discover the authorization server.

Additionally, there is no validation that the operator-supplied `--oauth-resource` value is a well-formed URI, so other characters that are valid in a URI but not in an HTTP quoted-string (e.g. backslash) could also produce malformed output.

**Trigger / repro:**

Start the proxy with `--oauth-resource 'https://example.com/"bad'` and send a request without credentials. The `WWW-Authenticate` header will contain an unescaped inner `"` and be unparseable by strict OAuth clients.

**Suggested fix:**

1. **Validate the `--oauth-resource` URI at startup** to ensure it is a valid absolute URI with no characters illegal in a `quoted-string`. Reject the config if it fails.

2. **Escape the value** per RFC 7235's `quoted-string` rule before embedding it:
   ```go
   escaped := strings.ReplaceAll(metaURL, `\`, `\\`)
   escaped = strings.ReplaceAll(escaped, `"`, `\"`)
   b.WriteString(`, resource_metadata="`)
   b.WriteString(escaped)
   b.WriteString(`"`)
   ```

Option 1 is preferable: a URI should never contain a literal `"`, so the right fix is to reject the misconfiguration at startup rather than silently mangle the header at runtime.

**Session:** rev-20260614T1005Z-18c713

### LINKED ISSUE #385: [rev-20260614T1005Z-18c713] serveOAuthMetadata returns 405 for HEAD requests, violating HTTP/1.1
**Location:** `cmd/eunox-mcp/oauth_metadata.go:65-82`

**Reference:** https://github.com/eunolabs/eunox/blob/3170cc052ead7e4ffd08a6299656c637920d91c8/cmd/eunox-mcp/oauth_metadata.go#L65-L82

**Severity:** Low

**Description:**

`serveOAuthMetadata` rejects every non-GET method with 405 Method Not Allowed:

```go
func (p *HTTPProxy) serveOAuthMetadata(w http.ResponseWriter, r *http.Request) {
    if r.Method != http.MethodGet {
        http.Error(w, "Method Not Allowed", http.StatusMethodNotAllowed)
        return
    }
```

RFC 7231 §4.3.2 states: "The HEAD method is identical to GET except that the server MUST NOT send a message body in the response. … A server MUST respond to a HEAD request with the same header fields it would have sent in response to a GET request."

Go's `net/http` `ServeMux` does **not** automatically translate HEAD to GET for user-registered `HandlerFunc` routes. A `HEAD /.well-known/oauth-protected-resource` therefore receives a 405 rather than a header-only 200.

This breaks any client that uses HEAD for:
- Endpoint existence checks (RFC 9728 discovery)
- HTTP cache validation
- Automated health probes for the metadata path

The existing test `TestServeOAuthMetadata_MethodNotAllowed` tests POST, PUT, and DELETE but does not cover HEAD, so the gap was not caught.

**Trigger / repro:**

```
HEAD /.well-known/oauth-protected-resource HTTP/1.1
Host: proxy.example.com
```
→ `405 Method Not Allowed`  (expected: `200 OK` with headers, no body)

**Suggested fix:**

Extend the method check to treat HEAD like GET and suppress the body:

```go
func (p *HTTPProxy) serveOAuthMetadata(w http.ResponseWriter, r *http.Request) {
    if r.Method != http.MethodGet && r.Method != http.MethodHead {
        http.Error(w, "Method Not Allowed", http.StatusMethodNotAllowed)
        return
    }
    if p.oauthMeta == nil {
        http.NotFound(w, r)
        return
    }
    doc, err := json.Marshal(p.oauthMeta)
    if err != nil {
        http.Error(w, "internal error", http.StatusInternalServerError)
        return
    }
    w.Header().Set("Content-Type", "application/json")
    w.Header().Set("Cache-Control", "no-store")
    if r.Method == http.MethodHead {
        return  // headers sent, no body
    }
    _, _ = w.Write(doc)
}
```

Add a test case `HEAD` to `TestServeOAuthMetadata_MethodNotAllowed` (it should NOT be in that list) and add a new `TestServeOAuthMetadata_HEADReturns200NoBody`.

**Session:** rev-20260614T1005Z-18c713

### LINKED ISSUE #386: [rev-20260614T1005Z-18c713] RFC 9728 metadata document served without required `resource` field when --jwks-uri is set alone
**Location:** `cmd/eunox-mcp/main.go:585-594`

**Reference:** https://github.com/eunolabs/eunox/blob/3170cc052ead7e4ffd08a6299656c637920d91c8/cmd/eunox-mcp/main.go#L585-L594

**Severity:** Medium

**Description:**

The RFC 9728 metadata document is activated whenever `--jwks-uri` is set, even without `--oauth-resource`:

```go
var oauthMeta *oauthResourceMetadata
var oauthMetaURL string
if pf.jwksURI != "" || oauthResource != "" {
    oauthMeta = &oauthResourceMetadata{
        Resource:             oauthResource,      // empty string when --oauth-resource not set
        AuthorizationServers: oauthAuthzServers,
    }
    if oauthResource != "" {
        oauthMetaURL = strings.TrimRight(oauthResource, "/") + metadataBasePath
    }
}
```

Because `oauthMeta != nil`, `serveOAuthMetadata` serves the document. Since `Resource` is tagged `json:"resource,omitempty"`, an empty resource string is silently dropped from the JSON output. The resulting document may be:

```json
{"authorization_servers":["https://idp.example.com"]}
```
or even just:
```json
{}
```

RFC 9728 §3.2 states: "The `resource` parameter is REQUIRED." A metadata document missing this field is non-conforming. An RFC 9728-compliant OAuth client that validates the `resource` field — for example, by comparing it to the URL it used to discover the document — would reject the response and fail to discover the authorization server, causing every authentication attempt to fail.

**Trigger / repro:**

Start the proxy with `--jwks-uri https://idp.example.com/.well-known/jwks.json` but without `--oauth-resource`. Issue:

```
GET /.well-known/oauth-protected-resource
```

The response is a 200 with a JSON body that has no `"resource"` key.

**Suggested fix:**

Two complementary fixes:

1. **Require `--oauth-resource` when `--jwks-uri` is set.** In `serveHTTPGateway`, add:
   ```go
   if pf.jwksURI != "" && oauthResource == "" {
       return fmt.Errorf("--jwks-uri requires --oauth-resource: RFC 9728 mandates the 'resource' field in the protected-resource metadata document. Set --oauth-resource to the URI of this proxy.")
   }
   ```

2. **Guard `oauthMeta` construction more tightly** so the metadata endpoint is only activated when both the JWKS URI *and* the resource URI are configured (or resource alone):
   ```go
   if oauthResource != "" {
       oauthMeta = &oauthResourceMetadata{
           Resource:             oauthResource,
           AuthorizationServers: oauthAuthzServers,
       }
       oauthMetaURL = strings.TrimRight(oauthResource, "/") + metadataBasePath
   }
   ```

Option 1 is the stronger fix (fail closed). Option 2 avoids a hard error but means `--jwks-uri` without `--oauth-resource` simply does not expose the metadata endpoint (returning 404), which is still more RFC-compliant than serving a broken document.

**Session:** rev-20260614T1005Z-18c713

### LINKED ISSUE #387: Engineering review: 9 findings (2 high, 3 medium, 4 low)
Principal engineering review of the current codebase. Two correctness bugs with security impact, three medium issues (audit completeness + maintainability), four low/polish items. Filed as one issue for triage; break into sub-issues as needed.

---

### 1 · HIGH — `DecideSampling` silently skips conditions on `system:sampling/createMessage`

**File:** `cmd/eunox-mcp/pdp.go:528–552`

`SamplingAllowed` only checks for a matching constraint with `"allow"` action. `EvaluateConditions` is never called, so every condition declared on that entry — `timeWindow`, `maxCalls`, `ipRange`, anything — is silently skipped. A manifest like:

```yaml
- target: system:sampling/createMessage
  actions: [allow]
  conditions:
    - type: timeWindow
      notAfter: "2026-12-31T00:00:00Z"
    - type: maxCalls
      count: 50
      windowSeconds: 3600
```

grants unconditional sampling permission, contradicting the declared policy. This is a fail-open on server-initiated sampling — the only enforcement point on that path.

**Fix:** Replace `SamplingAllowed` in `DecideSampling` with a `findConstraint` + `EvaluateConditions` call, exactly as `Decide`, `DecideResourceRead`, and `DecidePromptGet` do. `SamplingAllowed` becomes an internal predicate only if needed; `JWTPDP.DecideSampling` should delegate to `inner.DecideSampling(ctx, sessionID)` when inner implements `SamplingAuthorizer`, which `ManifestPDP` now does correctly.

```go
func (p *ManifestPDP) DecideSampling(ctx context.Context, sessionID string) capability.EnforceResponse {
        return *deny
    }
    claims := jwtClaimsAsMap(ctx)
    matched := p.findConstraint(EnforceTarget{
        Type: capability.TargetTypeSystem,
        Name: "sampling/createMessage",
    }, claims)
    if matched == nil || !containsAction(matched.Actions, "allow") {
        return denyResponse("SAMPLING_DENIED", "",
            "server-initiated sampling requires an explicit system:sampling/createMessage opt-in in the manifest")
    }
        SessionID: sessionID,
        ToolName:  "sampling/createMessage",
        Arguments: map[string]interface{}{},
        Target:    &capability.EnforceRequestTarget{Type: string(capability.TargetTypeSystem), Name: "sampling/createMessage"},
        Claims:    claims,
    }
    if err != nil {
        return capability.EnforceResponse{Decision: capability.DecisionDeny,
            Denial: &capability.DenialInfo{Code: capability.ErrCodeEnforcementError, Message: err.Error()}}
    }
    return resp
}
```

Needs: a test in `cmd/eunox-mcp/` that declares a `system:sampling/createMessage` entry with a `timeWindow` condition and verifies the condition is evaluated (not silently passed).

---

### 2 · HIGH — `redactJSONText` forwards unredacted malformed JSON arrays

**File:** `cmd/eunox-mcp/pdp.go:961–983`

When a text content item fails JSON parsing AND does not start with `{`, it is returned unchanged. The code only fails closed for objects:

```go
if strings.HasPrefix(strings.TrimSpace(text), "{") {
    return "", fmt.Errorf("... fail closed ...")
}
return text, nil   // ← forwarded unredacted, including malformed arrays
```

A tool returning a truncated array of sensitive objects — `[{"ssn":"123","name":"Alice"}` (common under payload limits or streaming bugs) — is forwarded unredacted even when `redactFields: ["ssn"]` is configured. The manifest guide §5a and the inline comment both state the requirement: "MUST NOT return unredacted bytes when redact paths are non-empty and the response cannot be structurally verified." Arrays are equally dangerous as objects.

**Fix:** Treat a leading `[` the same as `{`:

```go
first := []rune(strings.TrimSpace(text))
if len(first) > 0 && (first[0] == '{' || first[0] == '[') {
    return "", fmt.Errorf("redactFields: text content looks like a JSON container but is not parseable; cannot verify redaction of %v (fail closed)", paths)
}
return text, nil
```

Needs: a test case in the redaction test file with a text item containing `[{"ssn":"secret"}` (truncated, parse fails) verifying the function errors rather than returning the text.

---

### 3 · MEDIUM — `handleHTTPList` writes no audit record on success

**File:** `cmd/eunox-mcp/http.go:932–949`

`tools/list`, `resources/list`, and `prompts/list` write a deny record on upstream error but no record on success. Every other enforced method — `tools/call`, `resources/read`, `resources/subscribe`, `prompts/get` — writes an allow record. Tool enumeration is a reconnaissance technique (prompt-injection attacks often begin with a list call); the absence of a record breaks SIEM correlation and leaves the audit tape structurally incomplete.

**Fix:** Add a success record before the final return in `handleHTTPList`:

```go
if rt.sink != nil {
    rt.sink.Record(ctx, sess.id, msg.Method, msg.Method, "allow", "", "", nil, nil, rt.audit)
}
upResp.ID = msg.ID
return upResp
```

Use the method name for both `identifier` and `method` (consistent with the deny path and with how unmapped methods are logged).

---

### 4 · MEDIUM — Condition-dispatch loop duplicated verbatim between `ValidateAction` and `EvaluateConditions`

**File:** `pkg/enforcement/engine.go:480–518` and `700–735`

39 lines of identical condition-dispatch code. `collectObligations` was already extracted (with a comment noting the duplication) but the loop was not. Any future change — adding tracing, changing AuditOnly propagation, adding a per-condition timeout — must be applied twice and can diverge silently.

**Fix:** Extract `runConditions`:

```go
// runConditions evaluates every condition on matched, returning a non-nil deny
// response on the first failure or unknown type, nil when all pass.
func (e *Engine) runConditions(ctx context.Context, req *capability.EnforceRequest, matched *capability.Constraint, requestID, now string) *capability.EnforceResponse {
    for _, cond := range matched.Conditions {
        condType := cond.ConditionType()
        e.mu.RLock()
        handler, exists := e.handlers[condType]
        e.mu.RUnlock()
        if !exists {
            r := capability.EnforceResponse{RequestID: requestID, Decision: capability.DecisionDeny,
                AuditOnly: matched.IsAuditOnly(), DecidedAt: now,
                Denial: &capability.DenialInfo{Code: capability.ErrCodeConditionFailed, ConditionType: condType,
                    Message: fmt.Sprintf("unknown condition type: %s", condType)}}
            return &r
        }
        if condErr := handler.Handle(ctx, cond, req); condErr != nil {
            r := capability.EnforceResponse{RequestID: requestID, Decision: capability.DecisionDeny,
                AuditOnly: matched.IsAuditOnly(), DecidedAt: now,
                Denial: &capability.DenialInfo{Code: condErr.Code, ConditionType: condErr.ConditionType,
                    Message: condErr.Message, Details: condErr.Details}}
            return &r
        }
    }
    return nil
}
```

No behavior change; existing tests verify the outputs are identical.

---

### 5 · MEDIUM — `evaluateJWTConditions` has no default deny in type switch

**File:** `cmd/eunox-mcp/pdp_jwt.go:1070–1158`

The switch handles `AllowedOperationsCondition` and `AllowedValuesCondition` only. Currently safe because `buildV2Constraint` only ever produces those two types. But there is no `default` case — an unrecognized condition silently passes (the loop continues to `return nil`). Adding a new condition type to `buildV2Constraint` without updating this switch would be a silent fail-open, inconsistent with the engine's "fail closed on unknown condition type" invariant.

**Fix:**

```go
default:
    resp := denyResponse(capability.ErrCodeConditionFailed, cond.ConditionType(),
        fmt.Sprintf("JWT condition type %q has no evaluator; deny fail-closed", cond.ConditionType()))
    return &resp
```

---

### 6 · LOW — Ten near-identical `asXxx` cast helpers (~110 lines of boilerplate)

**File:** `pkg/enforcement/handlers.go:1277–1385`

Each helper tries `*T` then `T`. If conditions are standardized to pointer storage in the manifest loader, the value-receiver assertions can be dropped and each helper collapses to one line. Alternatively, a generic helper `func asCondition[T any](cond capability.Condition) (*T, bool)` eliminates the pattern entirely.

---

### 7 · LOW — `JWTPDP.inner*FilterXxx` three-way triplication

**File:** `cmd/eunox-mcp/pdp_jwt.go:906–925`

Three 5-line helpers differing only in the `ListFilterer` method called. Inline the body at each call site (3 lines each) and delete the helpers. Also: the `lf != nil` guard after a successful interface assertion on a concrete-struct type is always true — remove it.

---

### 8 · LOW — Double audit record on per-entry audit-mode observed deny

**File:** `cmd/eunox-mcp/http.go:813–866`

When `dec.AuditOnly=true` (per-entry `enforcement: audit` constraint), the code writes a `"deny"+audit_only=true` record and then, after forwarding, an `"allow"+audit_only=true` record for the same request. SIEM consumers see contradictory allow+deny pairs on the same request ID, which can trip correlation rules. Consider a single record with decision `"observe"` or a deny record with a forwarded flag, or at minimum document the two-record behavior in the audit schema spec.

---

### 9 · LOW — `recordAuditModeAntecedent` silently discards counter error

**File:** `cmd/eunox-mcp/pdp.go:303–307`

```go
_ = engine.RecordSessionCall(ctx, req)
```

A Redis failure here is invisible — no stderr, no audit record. Audit mode not blocking is correct; silence is not. A Redis partition during audit-mode rollout would cause subsequent `sequenceBlock` checks to fail open with no indication in the log.

**Fix:**
```go
if err := engine.RecordSessionCall(ctx, req); err != nil {
    fmt.Fprintf(os.Stderr, "[eunox-mcp] WARNING: audit-mode antecedent record failed for tool %q: %v (sequenceBlock may fail open)\n", req.ToolName, err)
}
```

---

## Proposed implementation order

| # | Finding | File(s) | Effort |
|---|---------|---------|--------|
| 1 | `DecideSampling` condition bypass | `pdp.go` | S |
| 2 | Redact fail-open on JSON arrays | `pdp.go` | XS |
| 3 | List audit record on success | `http.go` | XS |
| 5 | JWT condition switch default deny | `pdp_jwt.go` | XS |
| 9 | Log counter error in audit mode | `pdp.go` | XS |
| 4 | Extract `runConditions` helper | `engine.go` | S |
| 6 | `asXxx` boilerplate | `handlers.go` | M |
| 7 | Inline `inner*FilterXxx` | `pdp_jwt.go` | XS |
| 8 | Double audit record design | `http.go` | L (schema change) |

### LINKED ISSUE #388: [rev-20260614T1004Z-554770] allowedValues exact-match fires before glob, allowing literal metacharacter strings to bypass pattern constraints
**Location:** `pkg/enforcement/handlers.go`, `handleAllowedValues` (~line 969–987)

**Reference:** https://github.com/eunolabs/eunox/blob/f2156b7b4c05e9be6b6154a1a7a245f8d9342f1b/pkg/enforcement/handlers.go

**Severity:** Low / Medium

**Description:**

`handleAllowedValues` checks exact equality (`reflect.DeepEqual`) *before* glob matching:

```go
for _, allowed := range av.Values {
    // Exact match (handles non-string types: bool, number, nil).
    if reflect.DeepEqual(allowed, argValue) {
        return nil           // ← fires for strings too
    }
            if MatchValueGlob(pattern, str) {
                return nil
            }
        }
    }
}
```

The comment says the exact match is for non-string types (bool, number, nil), but the `reflect.DeepEqual` path fires unconditionally — including for strings. The result:

An operator who writes:
```yaml
conditions:
  - type: allowedValues
    argument: count
    values: ["[0-9]"]
```

intends to allow only single-digit strings (`"0"`–`"9"`) via the `[0-9]` character-class glob. However, because the exact match fires first, the literal string `"[0-9]"` is *also* allowed — it is string-equal to the pattern itself. The glob would correctly reject `"[0-9]"` (`path.Match("[0-9]", "[0-9]")` is false), but the code never reaches it.

Affected patterns include any `allowedValues` entry that uses `path.Match` metacharacters: `[`, `]`, `*`, `?`. An attacker who knows (or guesses) the pattern string can pass the literal pattern text as the argument value and bypass the intended constraint.

**Trigger / repro:**

```yaml
- target: tool:query
  actions: [allow]
  conditions:
    - type: allowedValues
      argument: table_suffix
      values: ["[a-z]"]
```

A call with `table_suffix = "[a-z]"` is allowed (exact match). A call with `table_suffix = "A"` is denied (no match). The operator expected only lowercase single letters; they got the literal pattern string too.

**Suggested fix:**

Skip the exact match for string values; string allowance is handled entirely by `MatchValueGlob` (which covers the no-metacharacter case via `path.Match("foo","foo") == true`):

```go
for _, allowed := range av.Values {
            if MatchValueGlob(pattern, str) {
                return nil
            }
        }
        // non-string arg vs string pattern: no match (fall through)
        continue
    }
    // Exact match for non-string types (bool, number, nil).
    if reflect.DeepEqual(allowed, argValue) {
        return nil
    }
}
```

This is a breaking change for any manifest that currently relies on exact string matching where the value happens to equal a metacharacter-containing pattern (an extremely unlikely operator intent, but worth noting in the release notes).

Alternatively, document the exact-before-glob order explicitly in the manifest guide and add a capability manifest guide warning: "A values entry that contains glob metacharacters (`*`, `?`, `[...]`) also matches itself as a literal string."

**Test needed:**

In `pkg/enforcement/` (or `cmd/eunox-mcp/enforcement_gaps_test.go`), add a case: `values: ["[0-9]"]`, argument value `"[0-9]"` → should deny; argument value `"5"` → should allow.

**Session:** rev-20260614T1004Z-554770
**Reviewed file:** `pkg/capability/conditions.go` @ `f2156b7b4c05e9be6b6154a1a7a245f8d9342f1b`

### LINKED ISSUE #389: [rev-20260614T1004Z-554770] unsatisfiable argumentSchema (required field absent from properties with additionalProperties:false) not caught at load time
**Location:** `cmd/eunox-mcp/manifest.go` (`checkArgumentSchemaKeywords`, `validateLocalManifest`); `pkg/enforcement/schema.go` (`schemaValidateObject`)

**Reference:** https://github.com/eunolabs/eunox/blob/f2156b7b4c05e9be6b6154a1a7a245f8d9342f1b/cmd/eunox-mcp/manifest.go

**Severity:** Low

**Description:**

The load-time manifest validator (`checkArgumentSchemaKeywords`) accepts the following `argumentSchema` without error:

```yaml
argumentSchema:
  type: object
  required: [path]
  properties: {}
  additionalProperties: false
```

This schema is logically unsatisfiable:
- `required: [path]` demands that `path` be present in the argument object.
- `additionalProperties: false` forbids any key not in `properties`.
- `properties: {}` lists no keys.

Therefore `path` is simultaneously required and forbidden. No argument object can satisfy both constraints.

`checkArgumentSchemaKeywords` validates only that individual keywords belong to the supported set — it does not cross-validate consistency between `required`, `properties`, and `additionalProperties`. The misconfig is accepted silently at load time and surfaces only at runtime, when `schemaValidateObject` fires:

```go
// schemaValidateObject, pkg/enforcement/schema.go
for _, req := range s.Required {
    if _, ok := v[req]; !ok {
        return fmt.Errorf("%s: missing required field %q", p, req)  // triggers if path absent
    }
}
// ...
if s.AdditionalProperties != nil && !*s.AdditionalProperties {
    for name := range v {
        if _, ok := s.Properties[name]; !ok {
            extras = append(extras, name)
        }
    }
    if len(extras) > 0 {
        return fmt.Errorf("%s: additional property %q is not allowed", p, extras[0])  // triggers if path present
    }
}
```

Every call to a tool with this schema is denied. The denial message — `"additional property \"path\" is not allowed"` — is confusing: the operator sees their required field rejected as "additional" rather than receiving a clear "schema misconfiguration" error at startup.

**Trigger / repro:**

```yaml
capabilities:
  - target: tool:read_file
    actions: [allow]
    argumentSchema:
      type: object
      required: [path]
      properties: {}
      additionalProperties: false
```

Every call to `read_file` is denied with `additional property "path" is not allowed`, even with a correct `{"path": "/etc/motd"}` argument. `make validate` accepts the manifest without complaint.

**Suggested fix:**

In `validateLocalManifest` (or a dedicated `validateArgumentSchema` helper called from it), after decoding the schema, check that every name in `required` appears as a key in `properties` when `additionalProperties` is `false`:

```go
func validateArgumentSchemaConsistency(path string, s *capability.ArgumentSchema) error {
    if s == nil {
        return nil
    }
    if s.AdditionalProperties != nil && !*s.AdditionalProperties {
        for _, req := range s.Required {
            if _, ok := s.Properties[req]; !ok {
                return fmt.Errorf(
                    "%s: required field %q is not listed in properties, "+
                    "but additionalProperties is false — this schema is unsatisfiable; "+
                    "add %q to properties", path, req, req)
            }
        }
    }
    // recurse into subschemas...
    for name, sub := range s.Properties {
        if err := validateArgumentSchemaConsistency(path+".properties."+name, sub); err != nil {
            return err
        }
    }
    if s.Items != nil {
        if err := validateArgumentSchemaConsistency(path+".items", s.Items); err != nil {
            return err
        }
    }
    return nil
}
```

Call it from `validateLocalManifest` alongside the existing `CompileSchemaPatterns` call.

**Test needed:**

In `cmd/eunox-mcp/` (manifest validation tests), add a table-driven case with `required: [foo]`, `properties: {}`, `additionalProperties: false` and assert that `LoadManifest` returns an error containing "unsatisfiable".

**Session:** rev-20260614T1004Z-554770
**Reviewed file:** `pkg/capability/conditions.go` @ `f2156b7b4c05e9be6b6154a1a7a245f8d9342f1b`

eunolabs:
## Bugs addressed

This PR fixes every open review-filed bug across the current review session: the two `rev-20260614T1004Z-554770` findings, the four `rev-20260614T1005Z-18c713` findings, and all nine findings of the engineering-review meta-issue (#387).

### Correctness / security
- **Closes #389** — An `argumentSchema` with a `required` field absent from `properties` while `additionalProperties: false` is logically unsatisfiable (required-and-forbidden) and was accepted at load, denying every call at runtime with a misleading "additional property" error. `validateArgumentSchemaConsistency` now rejects it at load (recursing through `properties`/`items`).
- **Closes #388** — `allowedValues` checked `reflect.DeepEqual` before glob, so a string equal to a metacharacter pattern (e.g. `"[0-9]"`) was allowed even though the glob rejects it. String values now match by `MatchValueGlob` only (a metacharacter-free pattern still matches itself); exact equality is reserved for non-string values.
- **Closes #387** — engineering review, all nine findings:
  1. `ManifestPDP.DecideSampling` only checked the `allow` action and never ran `EvaluateConditions`, silently skipping every condition on `system:sampling/createMessage` (fail-open on the only sampling enforcement point). Now resolves the constraint and evaluates conditions, mirroring `DecideResourceRead`/`DecidePromptGet`.
  2. `redactJSONText` forwarded any non-object text that failed to parse, including a truncated array of secret-bearing objects (`[{"ssn":"x"}`). A leading `[` immediately followed by `{`/`[` (an object-bearing array — the only array shape that can carry a dot-path target) now fails closed, while scalar arrays and `[ERROR] …`/`[2026-…] …` log prefixes still forward (preserves #319).
  3. `*/list` wrote a deny record on upstream error but nothing on success. Both HTTP and stdio list handlers now write an allow record so the audit tape isn't blind to enumeration.
  4. Extracted the duplicated condition-dispatch loop into `Engine.runConditions` (no behavior change).
  5. `evaluateJWTConditions` had no `default` in its type switch — a future JWT condition type would be silently skipped. Added a fail-closed default.
  6. Collapsed the ten `asXxx` casts onto one generic `asCondition[T]`.
  7. Removed the always-true `lf != nil` guard in the three `JWTPDP.innerFilter*List` helpers.
  8. Documented the two-record (deny+allow, both `audit_only`) behavior of an observed-and-forwarded denial in the threat model, with SIEM-correlation guidance (no audit-schema change).
  9. `recordAuditModeAntecedent` discarded the `RecordSessionCall` error; it now warns on stderr (audit mode still must not block).
- **Closes #383** — `listen.authToken` + `--jwks-uri` together: the static token check ran first and rejected every IdP JWT. Rejected at startup as mutually exclusive.
- **Closes #386** — `--jwks-uri` without `--oauth-resource` served an RFC 9728 metadata document missing the REQUIRED `resource` field. Fixed via option 2 (the non-breaking fix): the metadata document is published **only** when a resource URI is configured, otherwise the endpoint returns 404 instead of a non-conforming document. JWT validation does **not** require advertising an OAuth discovery document, so `--jwks-uri` without `--oauth-resource` stays a valid configuration (it must, e.g. `demo/docker-compose.jwt.yml`).
- **Closes #384** — `buildWWWAuthenticate` embedded `metaURL` verbatim in a quoted-string. `--oauth-resource` is now validated at startup (well-formed absolute URI, no quoted-string-illegal chars) and the value is escaped in the header builder as defense-in-depth.
- **Closes #385** — `serveOAuthMetadata` returned 405 for HEAD. HEAD is now handled like GET (same headers, no body) per RFC 7231.

## Testing
- `go test -race -count=1 ./...` — all packages pass; `make check-fmt`, `make check-license`, `golangci-lint run` — clean; `pkg/` coverage 97.9% (gate 80%).
- New/updated tests for every fix: unsatisfiable schema (load-time, nested, satisfiable control); literal-pattern-string denial; sampling-condition evaluation (deny + allow); malformed object-bearing-array fail-closed **and** scalar-array/bracket-log-prefix forward-unchanged; HTTP and stdio `*/list` success record; JWT unknown-condition fail-closed; `validateOAuthResourceURI`; gateway authToken+jwks mutual-exclusivity; OAuth metadata 404-without-resource and HEAD-200-no-body; WWW-Authenticate escaping.
- Behavior-change docs updated in the same PR (manifest guide glob semantics + migration note; threat model redaction posture, two-record audit note, and `*/list` record).

🤖 Generated with [Claude Code](https://claude.ai/code)


eunolabs (commented):
Self-review (round 3) — one substantive finding in my own #387 finding-2 fix, now resolved in commit 4e99c3e:

**`cmd/eunox-mcp/pdp.go` `looksLikeUnparseableJSONContainer`** — my first cut flagged a leading `[` followed by *any* JSON value-start (digit, quote, `-`, `t/f/n`) as a fail-closed container. That over-reaches: a `[2026-06-14] ...` ISO-timestamped log prefix and a `["a","b"` scalar array would be failed closed, breaking any tool that emits a bracketed log line while `redactFields` is configured — a regression against the #319 plain-text-forwarding behaviour. Since a dot-path redaction target is an *object key*, only an **object-bearing array** (`[{` / `[[`) can actually leak one; scalar arrays carry no keys. Narrowed the heuristic to `[{`/`[[`, added `TestRedactFields_ScalarArrayAndBracketLogPrefix_ForwardedUnchanged` for the prose side, and updated the threat-model wording.

Earlier rounds also tightened doc comments (`DecideSampling`, `oauth-resource` build condition), updated threat-model §6.1 for the new `*/list` allow record, and added an end-to-end HTTP `*/list` success-record test to complement the stdio one.

Verification: `go test -race -count=1 ./...` green; `make check-fmt`, `make check-license`, `golangci-lint run` clean.

---

eunolabs:
✅ Ready to merge — all open review-session bugs addressed (#383, #384, #385, #386, #387, #388, #389), self-review complete with no outstanding comments, all 42 CI checks green (Test, Lint, builds, demo/JWT, demo/gateway, demo/stdio, OPA comparisons, Enforcement compliance, CodeQL, govulncheck, Trivy).

Self-review found and fixed two real defects in the diff itself:
- The finding-#2 redaction heuristic over-reached onto timestamped log prefixes (`[2026-…]`); narrowed to object-bearing arrays (`[{`/`[[`) only, with a forward-unchanged test for the prose side.
- The first cut of #386 hard-required `--oauth-resource` for `--jwks-uri`, which broke JWT-only deployments (caught by the `demo / JWT mode` job). Switched to the non-breaking fix: never publish a metadata document without a `resource` field (404 instead), without forcing `--oauth-resource`.

---

PR #396, opened 2026-06-14
Fix session rev-20260614T110x bugs

### LINKED ISSUE #390: [rev-20260614T1105Z-5df230] Stats().State diverges from State() after cooldown expires, giving contradictory answers
**Location:** `pkg/circuitbreaker/breaker.go:196-219`

**Reference:** https://github.com/eunolabs/eunox/blob/3170cc052ead7e4ffd08a6299656c637920d91c8/pkg/circuitbreaker/breaker.go#L196-L219

**Severity:** Medium

**Description:**

`State()` and `Stats().State` give contradictory answers about the same breaker at the same instant once the cooldown has elapsed but before `Allow()` is called to mutate the internal field.

`State()` (line 200) applies a cooldown check and returns `StateHalfOpen` when the cooldown has elapsed:

```go
    b.mu.Lock()
    defer b.mu.Unlock()
        return StateHalfOpen   // logical state
    }
    return b.state
}
```

`Stats()` (line 211) returns the raw field `b.state`, which is still `StateOpen` until `Allow()` mutates it:

```go
    b.mu.Lock()
    defer b.mu.Unlock()
    return Stats{
        State: b.state,   // raw internal state, not logical state
        ...
    }
}
```

Consequence: in the window between cooldown expiry and the first `Allow()` call,

```go
b.State()        // → StateHalfOpen
b.Stats().State  // → StateOpen      ← contradicts State()
```

Two concrete failure modes:
1. Monitoring / health-check code that reads `b.Stats().State` reports `open` when the breaker is actually ready to admit a probe. Operators see a "stuck-open" breaker that resolves itself mysteriously.
2. Code that branches on `stats.State == StateHalfOpen` to trigger recovery logic (e.g. "log that we're entering probe phase") will never fire, because `Stats().State` never reaches `StateHalfOpen` while `b.state` is `StateOpen`.

**Trigger / repro:**

```go
    FailureThreshold: 2, CooldownDuration: 5*time.Second, HalfOpenMaxProbes: 1,
}, circuitbreaker.WithClock(clock))
b.RecordFailure()
b.RecordFailure()
// advance clock past cooldown
fmt.Println(b.State())        // "half-open"
fmt.Println(b.Stats().State)  // "open"  ← bug
```

**Suggested fix:**

Apply the same cooldown logic inside `Stats()`:

```go
    b.mu.Lock()
    defer b.mu.Unlock()

    state := b.state
        state = StateHalfOpen
    }
    return Stats{
        State:            state,
        ConsecutiveFails: b.consecutiveFails,
        TotalFailures:    b.totalFailures,
        TotalSuccesses:   b.totalSuccesses,
        LastFailureTime:  b.lastFailureTime,
        LastTransitionAt: b.lastTransitionAt,
    }
}
```

This ensures `Stats().State` always matches `State()`, and adds one table-driven case to the stats test.

**Session:** rev-20260614T1105Z-5df230
**Reviewed file:** `pkg/circuitbreaker/breaker.go` @ `3170cc052ead7e4ffd08a6299656c637920d91c8`

### LINKED ISSUE #391: [rev-20260614T1105Z-5df230] singleflight in refreshKeys captures first caller's context, causing context cancellation to falsely trip the JWKS circuit breaker
**Location:** `pkg/capability/jwks.go:243-267`

**Reference:** https://github.com/eunolabs/eunox/blob/3170cc052ead7e4ffd08a6299656c637920d91c8/pkg/capability/jwks.go#L243-L267

**Severity:** Medium

**Description:**

`refreshKeys` deduplicates concurrent JWKS fetches via `singleflight.Do`. The closure it passes captures `ctx` — the context of the **first** goroutine to call `refreshKeys`:

```go
v, err, _ := c.sfGroup.Do("refresh", func() (interface{}, error) {
    fetch := func(fetchCtx context.Context) (*jose.JSONWebKeySet, error) {
        return c.fetchKeys(fetchCtx)   // fetchCtx == ctx (first caller's context)
    }
    if c.breaker != nil {
        jwks, ferr = circuitbreaker.Do(ctx, c.breaker, fetch)  // ctx from outer scope
        ...
    } else {
        jwks, ferr = fetch(ctx)
    }
    ...
})
```

If the first caller's context is canceled (short deadline, request abort) before the HTTP fetch to the JWKS endpoint completes, two bad things happen simultaneously:

1. **All concurrent callers get the wrong error.** Because `singleflight` shares the result, every goroutine waiting on `"refresh"` receives `context.Canceled` or `context.DeadlineExceeded` regardless of whether their own context is still valid. Callers with generous timeouts fail because of a short-timeout sibling.

2. **The circuit breaker records a spurious failure.** `circuitbreaker.Do` calls `b.RecordFailure()` on every non-nil error returned by `fn`. A context cancellation is not evidence that the upstream JWKS endpoint is unhealthy, but the breaker cannot distinguish caller cancellation from a real network failure. After `FailureThreshold` such cancellations the breaker opens, causing **all** subsequent JWKS fetches to return `ErrOpen` — even when the IdP is fully reachable.

In a gateway under load where callers frequently race to the cache expiry, a single slow request context can cascade into a fully open circuit breaker that blocks JWT verification for the entire cooldown window. Since JWT verification is required for every authenticated MCP request, this is a potential denial-of-service on the authentication path.

**Trigger / repro:**

1. Configure a `JWKSClient` with a breaker (`FailureThreshold: 2`, `CooldownDuration: 30s`).
2. Let the cache expire.
3. Issue two concurrent `VerifyToken` calls: the first with a 1ms deadline (will cancel before the JWKS fetch completes), the second with a 5s deadline.
4. Both calls fail. The breaker has recorded 1 failure.
5. Repeat once more → breaker opens.
6. Now even callers with valid, non-canceled contexts immediately receive `JWKS fetch blocked by circuit breaker`.

**Suggested fix:**

Decouple the HTTP fetch from any individual caller's context. Use a long-lived context (e.g. the server's root context, stored in `JWKSClient`) for the actual network call inside the singleflight closure, and pass `ctx` only for the `circuitbreaker.Do` pre-checks (cancellation before consuming a probe slot):

```go
// In JWKSClientConfig / JWKSClient, add:
//   ctx context.Context  // root context for background fetches

v, err, _ := c.sfGroup.Do("refresh", func() (interface{}, error) {
    // Use the long-lived context so a single caller's cancellation
    // does not abort the shared fetch or trip the circuit breaker.
    fetchCtx := c.rootCtx  // or context.Background() if no root is available

    fetch := func(fc context.Context) (*jose.JSONWebKeySet, error) {
        return c.fetchKeys(fc)
    }
    if c.breaker != nil {
        jwks, ferr = circuitbreaker.Do(fetchCtx, c.breaker, fetch)
        ...
    } else {
        jwks, ferr = fetch(fetchCtx)
    }
    ...
})
```

Alternatively, explicitly skip failure recording when the error is `context.Canceled` or `context.DeadlineExceeded` — though this would require a change to the `circuitbreaker` package contract and loses the legitimate "upstream is hanging" failure signal.

**Session:** rev-20260614T1105Z-5df230
**Reviewed file:** `pkg/capability/jwks.go` @ `3170cc052ead7e4ffd08a6299656c637920d91c8`

### LINKED ISSUE #392: [rev-20260614T1105Z-5df230] refreshKeys emits N "refreshed JWKS" log lines and acquires N write locks for a single singleflight fetch
**Location:** `pkg/capability/jwks.go:272-281`

**Reference:** https://github.com/eunolabs/eunox/blob/3170cc052ead7e4ffd08a6299656c637920d91c8/pkg/capability/jwks.go#L272-L281

**Severity:** Low

**Description:**

The cache-store and log call in `refreshKeys` are placed **outside** the `singleflight.Do` closure, in the code that runs for every caller after the group returns:

```go
v, err, _ := c.sfGroup.Do("refresh", func() (interface{}, error) {
    // ... single HTTP fetch runs once ...
    return jwks, nil
})
if err != nil {
    return nil, err
}

jwks := v.(*jose.JSONWebKeySet)

// Store the freshly fetched JWKS under a short write lock.
c.mu.Lock()           // ← runs N times, once per waiting goroutine
c.jwks = jwks
c.fetchedAt = c.now() // ← N slightly-different timestamps
c.mu.Unlock()

c.logger.Info("refreshed JWKS", slog.Int("keys", len(jwks.Keys)))  // ← N log lines
return jwks, nil
```

When N goroutines are blocked on the same `sfGroup.Do("refresh", ...)` call and it completes, `singleflight` unblocks all N at once. Each goroutine independently:
- Acquires `c.mu.Lock()` and writes `c.jwks` / `c.fetchedAt` (N redundant critical-section acquisitions, each with a slightly different `c.now()` timestamp)
- Emits `logger.Info("refreshed JWKS", ...)` — producing **N log lines** for a single actual HTTP fetch

Under load (e.g. 50 concurrent token verifications hitting a cold cache), this emits 50 "refreshed JWKS" messages for one network round-trip, making the log unreadable and breaking any SIEM rule or rate-limit alert that fires on JWKS refresh frequency.

**Trigger / repro:**

Start a gateway with a short `CacheTTL`. Let the cache expire. Simultaneously issue 50 `VerifyToken` calls. Observe 50 identical `"refreshed JWKS keys=3"` log lines, all within a millisecond.

**Suggested fix:**

Move the store and log inside the `sfGroup.Do` closure so they execute exactly once regardless of how many callers are waiting:

```go
v, err, _ := c.sfGroup.Do("refresh", func() (interface{}, error) {
    var jwks *jose.JSONWebKeySet
    var ferr error
    if c.breaker != nil {
        jwks, ferr = circuitbreaker.Do(ctx, c.breaker, fetch)
        if ferr != nil { ... }
    } else {
        jwks, ferr = fetch(ctx)
        if ferr != nil { return nil, ferr }
    }

    // Store and log inside the closure — runs exactly once per singleflight.
    c.mu.Lock()
    c.jwks = jwks
    c.fetchedAt = c.now()
    c.mu.Unlock()
    c.logger.Info("refreshed JWKS", slog.Int("keys", len(jwks.Keys)))
    return jwks, nil
})
if err != nil {
    return nil, err
}
return v.(*jose.JSONWebKeySet), nil
```

**Session:** rev-20260614T1105Z-5df230
**Reviewed file:** `pkg/capability/jwks.go` @ `3170cc052ead7e4ffd08a6299656c637920d91c8`

### LINKED ISSUE #393: [rev-20260614T1105Z-5df230] circuitbreaker.New accepts zero/negative Config values that produce degenerate, undocumented breaker behavior
**Location:** `pkg/circuitbreaker/breaker.go:94-104`

**Reference:** https://github.com/eunolabs/eunox/blob/3170cc052ead7e4ffd08a6299656c637920d91c8/pkg/circuitbreaker/breaker.go#L94-L104

**Severity:** Low

**Description:**

`New` applies no validation to `Config`. Three fields produce degenerate behavior when zero or negative, none of which is documented:

**`FailureThreshold = 0`**

In `RecordFailure()`, `consecutiveFails` is incremented before the threshold check:

```go
b.consecutiveFails++
// ...
case StateClosed:
    if b.consecutiveFails >= b.config.FailureThreshold {  // 1 >= 0 → true
        b.state = StateOpen
    }
```

The breaker opens on the very first failure, which may be intended, but callers constructing `Config{}` by zero-value or forgetting to set the field get a breaker that trips immediately with no warning.

**`CooldownDuration = 0` (or negative)**

In `Allow()`:

```go
case StateOpen:
    if b.now().Sub(b.lastFailureTime) >= b.config.CooldownDuration {  // X >= 0 is always true
        b.state = StateHalfOpen
```

Any call to `Allow()` after a failure immediately transitions to half-open, collapsing the Open state to a single request cycle. The breaker provides no back-pressure.

**`HalfOpenMaxProbes = 0`**

`requiredHalfOpenSuccesses()` clamps to 1 when `HalfOpenMaxProbes < 1`, so the breaker can still close after a single probe success. However, `Allow()` in the `StateHalfOpen` case checks `halfOpenProbes < 0` (always false), so no *additional* probes are admitted beyond the single one consumed in the Open→HalfOpen transition. The effective probe count is always 1, inconsistent with the documented meaning of "number of probe requests admitted per half-open window."

**Trigger / repro:**

```go
    FailureThreshold:  0,   // zero value — trips on first failure
    CooldownDuration:  0,   // zero value — no cooldown
    HalfOpenMaxProbes: 0,   // zero value — undocumented effective count of 1
})
b.RecordFailure()
fmt.Println(b.State())  // "open" — tripped on one failure
fmt.Println(b.Allow())  // "true"  — immediately half-open (cooldown = 0)
```

**Suggested fix:**

Add validation in `New()` and return an error or panic with a clear message:

```go
func New(cfg Config, opts ...Option) (*Breaker, error) {
    if cfg.FailureThreshold <= 0 {
        return nil, fmt.Errorf("circuitbreaker: FailureThreshold must be > 0, got %d", cfg.FailureThreshold)
    }
    if cfg.CooldownDuration <= 0 {
        return nil, fmt.Errorf("circuitbreaker: CooldownDuration must be > 0, got %v", cfg.CooldownDuration)
    }
    if cfg.HalfOpenMaxProbes <= 0 {
        return nil, fmt.Errorf("circuitbreaker: HalfOpenMaxProbes must be > 0, got %d", cfg.HalfOpenMaxProbes)
    }
    b := &Breaker{config: cfg, state: StateClosed, now: time.Now}
    for _, opt := range opts { opt(b) }
    return b, nil
}
```

Alternatively, a `Validate() error` method on `Config` lets callers decide how to handle the error. Either way, the current `DefaultConfig()` path is unaffected (all values are positive).

**Session:** rev-20260614T1105Z-5df230
**Reviewed file:** `pkg/circuitbreaker/breaker.go` @ `3170cc052ead7e4ffd08a6299656c637920d91c8`

### LINKED ISSUE #395: [rev-20260614T1104Z-cb1219] allowedValues: reflect.DeepEqual type mismatch between YAML integer manifest values and JSON float64 request arguments causes incorrect denials
## Location

`pkg/enforcement/handlers.go`, lines 969–986
Reference: https://github.com/eunolabs/eunox/blob/3170cc052ead7e4ffd08a6299656c637920d91c8/pkg/enforcement/handlers.go#L969-L986

## Severity

**Medium** — silent policy bypass in the allow direction is not possible (the call is incorrectly *denied*, not incorrectly allowed), but it causes valid calls to be rejected with no actionable error for the operator, and the manifest appears correctly configured.

## Description

`handleAllowedValues` uses `reflect.DeepEqual` for non-string exact matching:

```go
for _, allowed := range av.Values {
    if reflect.DeepEqual(allowed, argValue) {
        return nil
    }
            if MatchValueGlob(pattern, str) {
                return nil
            }
        }
    }
}
```

`reflect.DeepEqual` is type-sensitive: `reflect.DeepEqual(int(42), float64(42))` returns `false`.

There are two different deserialization paths that produce these mismatched types:

| Path | Decoder | Go type for integer `42` |
|---|---|---|
| Manifest YAML (`values: [42]`) | `gopkg.in/yaml.v3` | `int` |
| JSON-RPC request argument `{"count": 42}` | `encoding/json` | `float64` |

When both are `42` numerically, `reflect.DeepEqual` returns `false`, so the loop exhausts all `allowed` entries without matching, and the call is denied with a `CONDITION_FAILED` error even though the value is explicitly permitted in the manifest.

The string glob branch is unreachable for this case because neither `allowed` nor `argValue` is a string, so there is no fallback.

## Trigger / Repro

Manifest fragment:
```yaml
constraints:
  - type: allowedValues
    argument: count
    values:
      - 42
      - 100
```

MCP tool call JSON-RPC argument:
```json
{"count": 42}
```

Expected: `ALLOW`
Actual: `CONDITION_FAILED` (deny)

The same condition with `values: ["42"]` (string) works correctly because the string branch handles strings via `MatchValueGlob`. The bug is specific to bare integer (and potentially bare float) values in `values:`.

## Suggested Fix

Before the `reflect.DeepEqual` comparison, normalize numeric types to a common representation. The cleanest approach is to convert both sides to `float64` when both are numeric:

```go
func numericEqual(a, b any) bool {
    fa, aOK := toFloat64(a)
    fb, bOK := toFloat64(b)
    return aOK && bOK && fa == fb
}

func toFloat64(v any) (float64, bool) {
    switch n := v.(type) {
    case int:     return float64(n), true
    case int8:    return float64(n), true
    case int16:   return float64(n), true
    case int32:   return float64(n), true
    case int64:   return float64(n), true
    case uint:    return float64(n), true
    case uint8:   return float64(n), true
    case uint16:  return float64(n), true
    case uint32:  return float64(n), true
    case uint64:  return float64(n), true
    case float32: return float64(n), true
    case float64: return n, true
    }
    return 0, false
}
```

Then in the match loop:

```go
    return nil
}
```

Note: `float64` cannot represent all `int64` values exactly (integers > 2^53 lose precision), but this is not a practical concern for `allowedValues` in MCP tool arguments. If exact large-integer semantics are required, both sides can be formatted as `fmt.Sprintf("%v", v)` for comparison instead.

## Session

Session ID: `rev-20260614T1104Z-cb1219`
Reviewed file: `pkg/enforcement/handlers.go` @ `3170cc052ead7e4ffd08a6299656c637920d91c8`

eunolabs:
## Bugs addressed

This PR fixes every open automated-review bug filed today across the two review
sessions `rev-20260614T1104Z-cb1219` and `rev-20260614T1105Z-5df230`. One commit
per bug.

- Closes #395 — `allowedValues` denied numbers that were explicitly permitted: a
  manifest value decoded from YAML as `int(42)` never `reflect.DeepEqual`-matched
  a request argument decoded from JSON as `float64(42)`. Added a `numericEqual`
  bridge that compares any two numeric values by `float64` value (`bool`
  excluded, so `true` is never equal to `1`).
- Closes #390 — `Stats().State` contradicted `State()`: after the cooldown
  elapsed, `State()` reported `half-open` while `Stats()` still returned the raw
  `open` field until `Allow()` mutated it. `Stats()` now applies the same
  cooldown check, so the two never disagree.
- Closes #393 — `circuitbreaker.New` accepted a zero-value `Config` that tripped
  on the first failure, had no cooldown, and admitted no probes. `New` now
  replaces any non-positive field with its `DefaultConfig` value (fail-safe), and
  a new `Config.Validate` lets callers reject a degenerate config outright.
- Closes #392 — `refreshKeys` ran the cache store and the `refreshed JWKS` log
  line *after* `singleflight.Do` returned, so N waiting callers each took the
  write lock and emitted a log line for a single HTTP fetch. Moved the store and
  log inside the singleflight closure so they run exactly once.
- Closes #391 — the singleflight refresh closure captured the first caller's
  context, so that caller cancelling mid-fetch failed every other waiter and
  recorded a spurious circuit-breaker failure; enough cancellations opened the
  breaker and blocked all JWT verification for the cooldown (a DoS on the auth
  path even with a healthy IdP). The shared fetch now uses
  `context.WithoutCancel`, decoupling it from any single caller.

## Testing

- `go vet ./...` clean; `golangci-lint run` clean on the touched packages;
  `gofmt` clean; Apache-2.0 SPDX headers present.
- New table-driven / targeted tests added per fix:
  - `TestEngine_AllowedValues_NumericTypeMismatch` (int/float allow + deny,
    bool-vs-1 deny) — #395
  - `TestBreaker_StatsReportsLogicalStateAfterCooldown` — #390
  - `TestConfig_Validate`, `TestNewSanitizesDegenerateConfig` — #393
  - `TestJWKSClient_RefreshKeys_ConcurrentCallersLogOnce` (one hit, one log line
    for 20 racing callers) — #392
  - `TestJWKSClient_RefreshKeys_CallerCancellationDoesNotTripBreaker` — #391
- Docs: `docs/capability-manifest-guide.md` updated to note numeric value
  matching is type-agnostic (#395).

---
_Generated by [Claude Code](https://claude.ai/code/session_01GJUtemhhNiyLxkf32PSBec)_


eunolabs:
✅ Ready to merge — all open session bugs addressed (#395, #390, #393, #392, #391), self-review complete with no outstanding comments.

**Tests:** every package touched by this PR is green in CI — `pkg/capability`, `pkg/circuitbreaker`, `pkg/enforcement`, and `cmd/eunox-mcp` all pass (also `go test -race -count=1 ./...` passes locally, including the new regression tests). Lint, gofmt, license-headers, govulncheck, CodeQL, and all build/demo jobs are green.

The only red check is `Test`, failing solely on `pkg/killswitch`'s `TestRedis_MultiInstance_KillPropagatesAcrossInstances` with `dial tcp …: connect: connection refused` — an environmental CI issue (the runner has no Redis service for the multi-instance pub/sub test). This package is untouched by this PR and the test passes locally; re-running the failed job reproduced the same infra failure. It is unrelated to these changes.

---

Issue #401, opened 2026-06-14
[rev-20260614T1207Z-95b179] refreshKeys post-singleflight store and log execute once per waiter, not once per fetch


eunolabs:
**Location:** `pkg/capability/jwks.go:272-281` (`refreshKeys`, post-singleflight section)

**Reference:** https://github.com/eunolabs/eunox/blob/9782701af33996d6a91247e3e9c2045c955a9f36/pkg/capability/jwks.go#L272-L281

**Severity:** Low

**Description:**

After `singleflight.Do` returns, `refreshKeys` stores the result and emits a log line:

```go
v, err, _ := c.sfGroup.Do("refresh", func() (interface{}, error) { ... })
if err != nil {
    return nil, err
}

jwks := v.(*jose.JSONWebKeySet)

c.mu.Lock()
c.jwks = jwks           // ← store
c.fetchedAt = c.now()   // ← fetchedAt stamp
c.mu.Unlock()

c.logger.Info("refreshed JWKS", slog.Int("keys", len(jwks.Keys)))  // ← log
return jwks, nil
```

`singleflight.Do` deduplicates the fetch (one HTTP request) but returns the same shared result to **every** goroutine that was waiting. All of them then fall through to the code below the `Do` call, so for a burst of N concurrent callers that all triggered the same singleflight:

- The lock is acquired and released **N times** (N redundant writes of identical values to `c.jwks` and `c.fetchedAt`).
- `"refreshed JWKS"` is logged **N times** for a single actual network round-trip.

The redundant writes are harmless in terms of correctness (all N goroutines write the same `*jose.JSONWebKeySet` pointer), but the duplicate log lines are misleading: an operator observing the log will see `refreshed JWKS` fire N times and may (wrongly) conclude N separate fetches occurred. Under bursts of concurrent token verifications this produces significant log spam and makes the "refreshed JWKS" message useless as a signal.

**Trigger / repro:**

```go
// 20 goroutines all verify a token concurrently when the JWKS cache is cold.
// All 20 join the same singleflight, but "refreshed JWKS" is logged 20 times.
var wg sync.WaitGroup
for i := 0; i < 20; i++ {
    wg.Add(1)
    go func() {
        defer wg.Done()
        _, _ = client.VerifyToken(context.Background(), token)
    }()
}
wg.Wait()
// Observe 20 "refreshed JWKS" lines in the log for 1 HTTP request.
```

**Suggested fix:**

[...]

Issue #400, opened 2026-06-14
[rev-20260614T1207Z-95b179] refreshKeys singleflight captures first caller's context — cancellation fails all waiters and spuriously trips the circuit breaker


eunolabs:
**Location:** `pkg/capability/jwks.go:243-267` (`refreshKeys`, singleflight closure)

**Reference:** https://github.com/eunolabs/eunox/blob/9782701af33996d6a91247e3e9c2045c955a9f36/pkg/capability/jwks.go#L243-L267

**Severity:** High

**Description:**

`refreshKeys` uses `singleflight.Do` to deduplicate concurrent JWKS fetches. The singleflight closure captures `ctx` — the context of whichever caller first triggered the refresh:

```go
v, err, _ := c.sfGroup.Do("refresh", func() (interface{}, error) {
    fetch := func(fetchCtx context.Context) (*jose.JSONWebKeySet, error) {
        return c.fetchKeys(fetchCtx)
    }
    // ...
    if c.breaker != nil {
        jwks, ferr = circuitbreaker.Do(ctx, c.breaker, fetch)  // ← first caller's ctx
    } else {
        jwks, ferr = fetch(ctx)                                 // ← first caller's ctx
    }
    // ...
})
```

The code comment acknowledges the context is the first caller's but does not address its consequences:

> _"The context passed here is the first caller's — subsequent callers share the result even if their context differs."_

**Two concrete failure modes arise:**

**1. Spurious failures for all concurrent callers**

If the first caller's context has a short deadline (e.g. a 100 ms per-request timeout) and the JWKS server takes 150 ms to respond:
- The HTTP fetch is cancelled at 100 ms
- `singleflight.Do` completes with a context error
- Every goroutine waiting for the same "refresh" key receives the error, even those whose own contexts are still fully live

A single slow-context caller silently poisons the JWKS refresh for all concurrent token verifications, causing spurious authentication failures under load.

**2. Spurious circuit breaker failure recording**

When the breaker is wired (`cfg.Breaker != nil`), the fetch path goes through `circuitbreaker.Do(ctx, c.breaker, fetch)`. If the first caller's context cancels mid-fetch, `fetchKeys` returns a context error; `circuitbreaker.Do` sees `err != nil` and calls `RecordFailure()`. The circuit breaker counts a healthy JWKS endpoint as failed, potentially tripping it and blocking all subsequent JWKS fetches for the full cooldown period — even though the upstream was never at fault.

**Trigger / repro:**

```go
[...]

Issue #399, opened 2026-06-14
[rev-20260614T1205Z-600006] New() accepts degenerate Config values that produce silent misbehavior


eunolabs:
**Location:** `pkg/circuitbreaker/breaker.go:80-90` (`New()` constructor)

**Reference:** https://github.com/eunolabs/eunox/blob/9782701af33996d6a91247e3e9c2045c955a9f36/pkg/circuitbreaker/breaker.go#L80-L90

**Severity:** Low

**Description:**

`New()` performs no validation on the `Config` it receives. Several degenerate configurations compile and run without error but produce behavior that almost certainly does not match the caller's intent:

| Field | Degenerate value | Observed behavior | Likely intent |
|---|---|---|---|
| `FailureThreshold` | `0` | Opens on the **first** failure (`consecutiveFails(1) >= threshold(0)` is true) | "Disabled" or "requires 1 failure" |
| `FailureThreshold` | negative | Same as `0` | (same) |
| `CooldownDuration` | `0` | Breaker **never stays open**: the check `elapsed >= 0` is always satisfied, so the very next `Allow()` after any trip immediately transitions to `HalfOpen` | A meaningful recovery window |
| `CooldownDuration` | negative | Same as `0` (negative `time.Duration` is always less than any non-negative elapsed) | (same) |
| `HalfOpenMaxProbes` | `0` | Behaves identically to `1` because `Allow()` initialises `halfOpenProbes = 1` during the `Open→HalfOpen` transition regardless of the configured limit | "No probes" or "disabled half-open" |
| `HalfOpenMaxProbes` | negative | Same as `0` | (same) |

The `FailureThreshold = 0` and `CooldownDuration = 0` cases are the most dangerous because they effectively disable two of the three state-machine phases: a `CooldownDuration = 0` breaker opens but can never accumulate meaningful open time, cycling Open→HalfOpen on every `Allow()` call, and a `FailureThreshold = 0` breaker opens on the first error regardless of policy.

These are not obscure edge cases — they are the natural result of a zero-value `Config{}` struct, which Go callers often construct by field:

```go
// Caller forgets FailureThreshold; zero value silently trips on first error.
b := circuitbreaker.New(circuitbreaker.Config{
    CooldownDuration:  30 * time.Second,
    HalfOpenMaxProbes: 1,
})
```

**Trigger / repro:**

```go
b := circuitbreaker.New(circuitbreaker.Config{
    FailureThreshold:  0,           // ← zero value
[...]

Issue #397, opened 2026-06-14
[rev-20260614T1205Z-600006] Stats().State and State() report inconsistent values after cooldown elapses


eunolabs:
**Location:** `pkg/circuitbreaker/breaker.go:160-185` (`State()` and `Stats()` methods)

**Reference:** https://github.com/eunolabs/eunox/blob/9782701af33996d6a91247e3e9c2045c955a9f36/pkg/circuitbreaker/breaker.go#L160-L185

**Severity:** Medium

**Description:**

`State()` and `Stats().State` return different values for the same circuit breaker when the breaker is in `StateOpen` and the cooldown period has elapsed but `Allow()` has not yet been called.

`State()` has explicit "virtual state" logic — it reports `StateHalfOpen` without mutating internal state:

```go
func (b *Breaker) State() State {
    b.mu.Lock()
    defer b.mu.Unlock()
    if b.state == StateOpen && b.now().Sub(b.lastFailureTime) >= b.config.CooldownDuration {
        return StateHalfOpen   // virtual HalfOpen
    }
    return b.state
}
```

`Stats()` returns the raw internal field directly:

```go
func (b *Breaker) Stats() Stats {
    b.mu.Lock()
    defer b.mu.Unlock()
    return Stats{
        State: b.state,   // always StateOpen here — no virtual logic
        ...
    }
}
```

During the window between cooldown expiry and the first subsequent `Allow()` call:
- `b.State()` → `StateHalfOpen`
- `b.Stats().State` → `StateOpen`

These are contradictory. Any code (monitoring dashboards, health endpoints, operational tooling, tests) that reads `Stats().State` to decide what state the breaker is in will draw the wrong conclusion, while code using `State()` sees the correct logical state.

**Trigger / repro:**

```go
b := circuitbreaker.New(circuitbreaker.Config{
    FailureThreshold:  1,
    CooldownDuration:  5 * time.Second,
    HalfOpenMaxProbes: 1,
}, circuitbreaker.WithClock(fakeClock))

b.RecordFailure()                 // opens the breaker
advanceClock(6 * time.Second)     // cooldown elapsed, Allow() not yet called

fmt.Println(b.State())            // StateHalfOpen
fmt.Println(b.Stats().State)      // StateOpen  ← diverges
```

**Suggested fix:**

Apply the same virtual-state logic inside `Stats()`:

```go
func (b *Breaker) Stats() Stats {
    b.mu.Lock()
    defer b.mu.Unlock()

[...]

PR #402, opened 2026-06-14
Fix session rev-20260614T120x bugs

### LINKED ISSUE #398: [rev-20260614T1205Z-600006] RecordFailure() while StateOpen resets cooldown timer, can trap breaker permanently open
**Location:** `pkg/circuitbreaker/breaker.go:145-165` (`RecordFailure()`)

**Reference:** https://github.com/eunolabs/eunox/blob/9782701af33996d6a91247e3e9c2045c955a9f36/pkg/circuitbreaker/breaker.go#L145-L165

**Severity:** High

**Description:**

`RecordFailure()` unconditionally updates `lastFailureTime = b.now()` regardless of the breaker's current state:

```go
func (b *Breaker) RecordFailure() {
    b.mu.Lock()
    defer b.mu.Unlock()

    b.totalFailures++
    b.consecutiveFails++
    b.lastFailureTime = b.now()   // ← always updated, even when StateOpen

    switch b.state {
    case StateClosed:
        ...
    case StateHalfOpen:
        ...
    // No StateOpen case — state is unchanged, but lastFailureTime is reset.
    }
}
```

Both `Allow()` and `State()` use `lastFailureTime` as the epoch for the cooldown check:

```go
if b.now().Sub(b.lastFailureTime) >= b.config.CooldownDuration { ... }
```

**Consequence:** Any call to `RecordFailure()` while the breaker is open silently resets the cooldown timer. If errors continue to arrive and the caller records them directly on the `Breaker` (instead of routing through `Do()`, which enforces the admit-before-record contract), the cooldown never elapses and the breaker is permanently stuck open — all traffic is rejected forever, recovery requires an explicit `Reset()`.

While `Do()` itself is safe (it only calls `RecordFailure()` after a successful `Allow()`), the raw API is exposed publicly and callers can — and do — interact with the breaker outside `Do()`. For example:

- Callers that wrap `Do()` with additional error-handling layers and call `RecordFailure()` on errors from those outer layers.
- Callers that use the breaker for non-`Do()` bookkeeping (e.g., recording failures from asynchronous background checks).
- The `consecutiveFails` counter is also incremented, interfering with Closed-state thresholding when the breaker eventually resets.

**Trigger / repro:**

```go
cfg := circuitbreaker.Config{
    FailureThreshold:  2,
    CooldownDuration:  10 * time.Second,
    HalfOpenMaxProbes: 1,
}
b := circuitbreaker.New(cfg)

b.RecordFailure()
b.RecordFailure()
// Breaker is now open.

// Simulate a caller that records failures directly while breaker is open:
for i := 0; i < 100; i++ {
    time.Sleep(9 * time.Second)
    b.RecordFailure()  // resets lastFailureTime every 9s < 10s cooldown
}

// Allow() will never transition to HalfOpen; breaker is permanently open.
fmt.Println(b.Allow())  // false — indefinitely
```

**Suggested fix:**

Guard `lastFailureTime` (and both counter increments) on the current state in `RecordFailure()`. Failures that arrive outside an admitted probe slot carry no semantics for the cooldown clock:

```go
func (b *Breaker) RecordFailure() {
    b.mu.Lock()
    defer b.mu.Unlock()

    b.totalFailures++

    if b.state == StateOpen {
        // The caller violated the "admit before record" contract.
        // Silently ignore to preserve cooldown semantics.
        return
    }

    b.consecutiveFails++
    b.lastFailureTime = b.now()

    switch b.state {
    case StateClosed:
        if b.consecutiveFails >= b.config.FailureThreshold {
            b.state = StateOpen
            b.lastTransitionAt = b.now()
        }
    case StateHalfOpen:
        b.state = StateOpen
        b.halfOpenSuccess = 0
        b.lastTransitionAt = b.now()
    }
}
```

Note: `totalFailures` can still be incremented for observability, but the cooldown clock and state machine should not be affected.

**Session:** rev-20260614T1205Z-600006

eunolabs:
## Bugs addressed (sessions `rev-20260614T1205Z-600006` and `rev-20260614T1207Z-95b179`)

Five automated-review issues were filed today across these two sessions (#397–#401). All five were reviewed against `9782701`, but **four of them had already been fixed minutes earlier by PR #396** (merged as `7cca898`) and are now duplicates on `main`. They have been closed as duplicates with a pointer to the original fix; only the one genuinely-open bug is fixed here.

### Fixed in this PR
- Closes #398 — `RecordFailure()` unconditionally stamped `lastFailureTime` (the cooldown epoch used by both `Allow()` and `State()`) and incremented `consecutiveFails`, *regardless of state*. A failure reported while the breaker is already open never comes from an admitted probe — `Allow()` transitions Open→HalfOpen before handing out a probe slot, so genuine probe outcomes are recorded against half-open. A caller recording failures directly on the breaker (outside `Do`, which honors the admit-before-record contract) could, under a steady trickle of out-of-band failures, keep restarting the cooldown so it never elapsed, trapping the breaker open until an explicit `Reset`. `RecordFailure` now still counts `totalFailures` (observability) but returns early when open, leaving the cooldown clock, `consecutiveFails`, and the state machine untouched. The `Do` path is unaffected.

### Already fixed on `main` (closed as duplicates, no code in this PR)
- #397 — duplicate of #390 (`Stats().State` vs `State()` divergence after cooldown) — fixed in PR #396.
- #399 — duplicate of #393 (`New()` accepted a degenerate `Config`) — fixed in PR #396.
- #400 — duplicate of #391 (singleflight captured first caller's context) — fixed in PR #396.
- #401 — duplicate of #392 (JWKS store/log ran once per waiter) — fixed in PR #396.

## Testing
- New regression test `TestBreaker_RecordFailureWhileOpenDoesNotResetCooldown`: opens the breaker, then reports five stray failures 9s apart (cooldown 10s) while open; asserts the breaker reaches half-open and admits a probe (with the bug the cooldown never elapses), and that `TotalFailures` counts all seven failures while `ConsecutiveFails` stays at 2.
- `go vet ./pkg/circuitbreaker/` clean; `gofmt` clean; Apache-2.0 SPDX headers present.

https://claude.ai/code/session_017Fs8sxBAAFKvGGN9tsLGo1

---
_Generated by [Claude Code](https://claude.ai/code/session_017Fs8sxBAAFKvGGN9tsLGo1)_


eunolabs (commented):
Self-review, round 1 — one substantive finding.

**pkg/circuitbreaker/breaker.go:204-205 — `RecordFailure` godoc is now incomplete for a public API contract change.** The doc comment still reads only "If consecutive failures exceed the threshold, opens the circuit." After this PR, a failure reported while the breaker is already open is counted in `TotalFailures` but otherwise ignored (no cooldown reset, no `consecutiveFails` increment). That is a deliberate, observable change to the documented behavior of an exported method, and CONTRIBUTING requires behavior changes to keep their docs in sync. The godoc should state the open-state behavior so callers using the raw API understand it. Fixing now.

Everything else checks out: the `Do` path is unaffected (it records only after an admitted `Allow`, which leaves `StateOpen` before returning a probe slot), the regression test fails against the old code, and fmt/vet/license/tests are clean.

---

eunolabs:
Round 1 finding resolved in `9466616`: the `RecordFailure` godoc now documents the open-state contract (failures while open are counted in `TotalFailures` but otherwise ignored — no cooldown reset, no `consecutiveFails` increment).

---

eunolabs:
✅ Ready to merge — all session `rev-20260614T1205Z-600006` / `rev-20260614T1207Z-95b179` bugs addressed (#397, #398, #399, #400, #401): #398 fixed here, and #397/#399/#400/#401 closed as duplicates already resolved by PR #396. Self-review complete with no outstanding comments. All 35 CI checks green; `go test -race ./...` passing.

---

Issue #318, opened 2026-06-14
sec: list-filter functions fail open — return unfiltered bytes on re-marshal error


eunolabs:
## Summary

`filterToolsListResult`, `filterResourcesListResult`, and `filterPromptsListResult` in `cmd/eunox-mcp/pdp.go` all return the **original, unfiltered upstream bytes** when `json.Marshal` fails after filtering. This silently bypasses list filtering and exposes every capability to the client.

## Reproduction

The exit paths are at lines 622, 662, and 695:

```go
out, err := json.Marshal(list)
if err != nil {
    return resultBytes   // <-- unfiltered original bytes returned
}
```

If `json.Marshal` on the filtered struct ever fails (e.g. due to an unencodable value slipping through, a future field change, or a crafted upstream response that round-trips through `json.Unmarshal` into an unencodable Go value), the full upstream list—including entries that should be hidden by policy—is returned to the client without error.

The unmarshal-error path by contrast fails closed:

```go
if err := json.Unmarshal(resultBytes, &list); err != nil {
    empty, _ := json.Marshal(mcpToolsListResult{})
    return empty   // <-- correct: empty list on parse failure
}
```

The two error branches of the same function use opposite safety postures. This violates the fail-closed invariant.

## Fix

On re-marshal error, return the same empty struct that the unmarshal-error path returns — never the original bytes:

```go
out, err := json.Marshal(list)
if err != nil {
    empty, _ := json.Marshal(mcpToolsListResult{})
    return empty
}
```

Apply consistently to all three functions. The empty list may cause a degraded UX for the end user, but that is the correct fail-closed behavior.

## Affected files

- `cmd/eunox-mcp/pdp.go`: `filterResourcesListResult` (line 622), `filterToolsListResult` (line 662), `filterPromptsListResult` (line 695)

PR #412, opened 2026-06-14
fix: unify JWKS caches and close two JWT correctness gaps (#410)

### LINKED ISSUE #410: Engineering review: reconcile the two JWKS caches + two related correctness gaps + list-filter dedup
Findings from a principal-engineering review focused on simplicity and correctness. The codebase is in good shape (fail-closed posture is consistent, audit-chain logic is meticulous, critical-path coverage is broad); these are narrow, high-leverage items. No behavior change is proposed except where noted as a bug fix.

The root theme is that the JWT/JWKS layer evolved in two places in parallel, producing one structural duplication and two correctness gaps that fall out of it.

### A1 — Capability-token JWKS client doesn't force-refresh on a kid miss (Med, fail-closed-but-rejects-valid)
`pkg/capability/jwks.go` — `VerifyToken` kid-miss branch (~L134) calls `refreshKeys` (L228), but `refreshKeys` opens with a TTL double-check that returns the cached set whenever it's still within `cacheTTL`. So after an issuer rotates its signing key, a token carrying the new kid is rejected (`no matching key for kid`) until the TTL (default 5m) expires, even though a fresh fetch would resolve it. The sibling IdP cache in `cmd/eunox-mcp/pdp_jwt.go` has an explicit `forceRefresh` for exactly this case; this one is missing it. No test exercises mid-TTL rotation (only "missing required kid").

Fix: give the kid-miss branch a forced path that bypasses the TTL double-check (still singleflight-deduped). Subsumed by A3 if the caches are unified.

### A2 — JWT∩manifest intersection drops synthesized resource/prompt args (Med, fail-closed)
`cmd/eunox-mcp/pdp_jwt.go` — `JWTPDP.DecideResourceRead`/`DecidePromptGet` (L573–582) delegate to `Decide`, which after JWT checks always calls the generic `p.inner.Decide(...)` (L660, L704), never `inner.DecideResourceRead`/`DecidePromptGet`. Only those inner methods synthesize `{"uri": uri}` / `{"name": name}` (`pdp.go` L509, L671), so under `--jwks-uri` the inner manifest evaluates argument-value conditions on resource/prompt entries against an empty map → `MISSING_CONTEXT` deny.

Net effect: a `resource:`/`prompt:` entry carrying an `allowedValues`/`allowedOperations` condition works in manifest-only mode but always denies under JWT mode. It's fail-closed (denies valid requests, not a bypass) and narrow (`ipRange`/`timeWindow`/`maxCalls` are unaffected), but it silently contradicts the documented synthesis.

Fix: in `Decide`, dispatch the inner call by `target.Type` to `inner.DecideResourceRead`/`DecidePromptGet`/`Decide`. Add an `enforcement_gaps_test.go` case with a uri/name condition behind a JWT.

### A3 — Two JWKS caches; one hand-rolled (Med→Large, the highest-leverage cleanup)
`pkg/capability/jwks.go` (`JWKSClient`, uses `golang.org/x/sync/singleflight` + `circuitbreaker.Do` + `context.WithoutCancel`, ~110 lines) vs `cmd/eunox-mcp/pdp_jwt.go` (`jwksCache`, hand-rolls singleflight in `doRefresh` with a force-vs-non-force join protocol and manual breaker accounting in `guardedFetch`, ~230 lines, L161–400). Same concept implemented twice; the hand-rolled one is the trickiest concurrency in the repo and the most likely place for a future regression.

The hand-rolled cache's one genuine extra (`forceRefresh` on kid miss) is exactly what `pkg` is missing (A1). So extracting one shared cache (fetch + singleflight + breaker + TTL + force-refresh, naturally in `pkg/capability`) deletes the hairiest goroutine code and fixes A1 in the same stroke; the two verifiers keep only their distinct claim-validation logic.

Trade-off: this is surgery on verified, security-critical code with extensive tests — the highest-risk item. The conservative alternative is to fix A1 in place and leave the duplication.

### B1 — Six near-identical `*/list` filter functions (Med, mechanical)
`pdp.go` `filter{Tools,Resources,Prompts}ListResult` (L719–835) and `pdp_jwt.go` `filter{Tools,Resources,Prompts}ListByClaims` (L934–1019): six functions, ~165 lines, identical shape (empty-guard → unmarshal/fail-closed → filter by predicate → marshal/fail-closed), differing only in envelope type, name field, and keep-test (tools adds the `descriptionHash` check). The fail-closed rationale is copy-pasted six times.

Fix: one generic helper carries the scaffolding; each call site supplies only the predicate. Collapses ~165 lines to one helper + six one-liners. Behavior-preserving; existing tests pin it.

### Polish
- Orphaned `.`-only comment lines: `pdp.go` L52–53 and `pdp_jwt.go` L90–91 (reflow artifact; join onto the preceding line).
- Comment density: several functions carry comments 2–4× their length (`compositeCounterKey`, `resourceSpecificity`, `looksLikeUnparseableJSONContainer`, the JWKS caches). Largely justified for security code — note only; trim opportunistically when editing for another reason, not as standalone churn.

### Suggested order
A1 + A2 (small, self-contained correctness fixes) → B1 (clean mechanical dedup) → A3 only if investing in the structural cleanup (which subsumes A1).

eunolabs:
Closes #410.

The JWT/JWKS layer had evolved two parallel caches, producing one structural duplication and two correctness gaps. This PR addresses all five review findings (the full scope, including the optional A3 cleanup, which subsumes A1).

## A1 — capability-token JWKS client force-refreshes on a kid miss (correctness)
`VerifyToken`'s kid-miss branch called a TTL-respecting refresh, so after an issuer rotated its signing key, a token carrying the new kid was rejected (`no matching key for kid`) until the cache TTL (default 5m) elapsed — even though a fresh fetch would resolve it. It now forces a TTL-bypassing refresh (the same path the IdP validator already had). Subsumed by A3.

- Regression test: `TestJWKSClient_VerifyToken_MidTTLRotation` (rotation within a 1h TTL must verify immediately).

## A2 — JWT∩manifest intersection no longer drops synthesized resource/prompt args (correctness, fail-closed)
`JWTPDP.Decide` always delegated to `inner.Decide`, which forwards args verbatim; only `inner.DecideResourceRead`/`DecidePromptGet` synthesize `{"uri": uri}` / `{"name": name}`. So a `resource:`/`prompt:` entry carrying an `allowedValues` condition denied with `MISSING_CONTEXT` under `--jwks-uri` while working in manifest-only mode. `Decide` now dispatches the inner call by `target.Type` via a new `decideInner` helper.

- Regression test: `TestA2_JWTIntersection_ResourcePromptArgConditions` (matching value → allow; non-matching value → `CONDITION_FAILED`, **not** `MISSING_CONTEXT`).

## A3 — one shared JWKS cache (the structural cleanup)
Extracted `pkg/capability.JWKSCache` (fetch + singleflight + circuit breaker + TTL + force-refresh) and deleted the hand-rolled `jwksCache` in `cmd/eunox-mcp` (~230 lines, the trickiest concurrency in the repo). Both verifiers now keep only their distinct claim-validation logic.

The shared cache decouples the fetch from any caller's context (`context.WithoutCancel`) and routes through `circuitbreaker.Do`, which **structurally** avoids the manual force-vs-non-force join protocol and manual breaker accounting the hand-rolled cache needed (caller cancellation can no longer abort the shared fetch or trip the breaker). The cmd-side cache tests are removed; equivalent and new coverage lives in `pkg/capability` (`jwks_test.go`, new `jwkscache_test.go`).

## B1 — list-filter dedup (mechanical)
Collapsed the six near-identical `*/list` filter functions onto one generic `filterListResult` helper; each call site now supplies only the predicate. The fail-closed-to-empty scaffolding (#318) lives in one place. Behavior-preserving; existing tests pin it.

## Polish
Joined the orphaned `.`-only comment lines in `pdp.go` and `pdp_jwt.go`.

## Docs
`docs/architecture.md` updated to describe the single shared cache and the kid-miss force-refresh.

## Verification
- `go build ./...`, `go vet ./...`, `make lint` — clean
- `make test` (`go test -race -count=1 ./...`) — all packages pass
- `make check-fmt`, `make check-license` — pass
- `pkg/capability` coverage: 97.9% (gate: 80%)

Net **−785 lines**.

https://claude.ai/code/session_01HC7deV23yH4K5LDYVCqQjg

---
_Generated by [Claude Code](https://claude.ai/code/session_01HC7deV23yH4K5LDYVCqQjg)_


pkg/capability/jwkscache.go
```diff
@@ -0,0 +1,222 @@
...
+		c.mu.RLock()
+		if c.jwks != nil && c.now().Sub(c.fetchedAt) < c.cacheTTL {
+			keys := c.jwks
+			c.mu.RUnlock()
+			return keys, nil
+		}
+		c.mu.RUnlock()
+	}
+
+	v, err, _ := c.sfGroup.Do("refresh", func() (interface{}, error) {
```
eunolabs, line 151: Both `Refresh` and `ForceRefresh` share the constant singleflight key `"refresh"`, so a forced (kid-miss) caller can now **join an in-flight non-forced fetch** instead of guaranteeing its own round-trip — the isolation the deleted `doRefresh`/`sfForced` protocol enforced. This is the intended A3 simplification (and `WithoutCancel` + `circuitbreaker.Do` legitimately retire the rest of the hand-rolled protocol), so I'm not asking to revert it — just flagging the residual behavior so it's on the record:

- Because the singleflight closure always performs an HTTP `fetchKeys` (the TTL check is *outside* `Do`), a joining forced caller still gets fresh-from-endpoint keys in the overwhelming majority of cases, so the A1 fix holds.
- The one narrow window: TTL just lapsed, a non-forced fetch is in flight whose request reached the IdP *just before* a rotation completed (server still serving old keys), and a forced kid-miss caller joins it → it gets the stale set and returns `no matching key for kid`. It **fails closed and self-heals on the next attempt** (cache is then fresh-but-old, so `ForceRefresh` starts its own fetch). Acceptable, but worth a sentence in the `ForceRefresh` doc comment noting the join-may-return-stale-under-concurrency caveat.

Also note: the new force-refresh tests (`TestJWKSCache_ForceRefresh_BypassesTTL`, `TestJWKSClient_VerifyToken_MidTTLRotation`) are serial, so nothing exercises a concurrent forced-vs-non-forced refresh anymore (the deleted `TestJWKSCache_ForceRefresh_DoesNotJoinNonForcedInFlight` did). Fine given the behavior is intentionally changed, but the concurrency path is now untested.

---

cmd/eunox-mcp/pdp_jwt.go
```diff
@@ -701,12 +469,31 @@ func (p *JWTPDP) Decide(ctx context.Context, sessionID string, target EnforceTar
...
 
+// decideInner delegates the post-JWT decision to the inner PDP, dispatching by
+// target type so resources/read and prompts/get reach the inner methods that
+// synthesize the {"uri": uri} / {"name": name} argument maps (pdp.go). Routing
+// every type through inner.Decide (which is the tools/call path and forwards args
+// verbatim) would evaluate an inner resource/prompt entry's argument-value
+// conditions against an empty map, denying with MISSING_CONTEXT a request the
+// manifest actually permits — a JWT-mode-only fail-closed regression that
+// contradicts the documented synthesis (issue #410, A2).
+func (p *JWTPDP) decideInner(ctx context.Context, sessionID string, target EnforceTarget, args map[string]interface{}, sourceIP string) capability.EnforceResponse {
```
eunolabs, line 501: `decideInner` correctly fixes the inner-manifest synthesis (A2). Two follow-ups on the same flow:

1. **Symmetric gap on the JWT-claims side is still open.** `Decide` calls `evaluateJWTConditions(matched.Conditions, target.Name, args)` (line ~457) with `args == nil` for resource/prompt targets (both `DecideResourceRead`/`DecidePromptGet` pass `nil`). So a JWT *shorthand* condition on a resource/prompt claim — e.g. `resource:file:///data/*?uri=...` or `prompt:x?<key>=...` → `AllowedValuesCondition{Argument:"uri"/...}` — looks up an absent arg and denies with `MISSING_CONTEXT` *before* ever reaching `decideInner`. That's the mirror image of the very bug this PR fixes on the manifest side. Tool claims are unaffected (real args are passed). Low practical impact (resource/prompt JWT conditions are largely redundant with the target pattern), but worth deciding whether it's in scope here or a follow-up.

2. **`default` case is a re-introduction hazard.** If a future type-specific inner method with its own synthesized args is added (say a `DecideResourceSubscribe`), forgetting to extend this switch silently routes it through `inner.Decide` and re-introduces exactly the empty-arg `MISSING_CONTEXT` regression. A one-line comment on the `default` case pointing back at the synthesis requirement would guard against that.

---

cmd/eunox-mcp/pdp_jwt.go
```diff
@@ -425,16 +180,29 @@ type JWTPDPOptions struct {
...
+	// the architecture docs describe. Callers (and tests) may inject a breaker
+	// with a different config or clock. The library capability.NewJWKSCache leaves
+	// the breaker opt-in, so it is applied here rather than in the cache.
+	breaker := opts.Breaker
+	if breaker == nil {
+		breaker = circuitbreaker.New(circuitbreaker.DefaultConfig())
+	}
 	return &JWTPDP{
-		cache:    newJWKSCache(opts.JWKSURI, opts.CacheTTL, opts.Client, opts.Breaker),
+		cache: capability.NewJWKSCache(capability.JWKSCacheConfig{
```
eunolabs, line 202: Minor / latent: the cmd cache config doesn't set `Logger`, so the shared `JWKSCache` falls back to `slog.Default()` and now emits an Info `"refreshed JWKS"` line on every refresh — a log site that didn't exist in the deleted cmd cache. Safe today (the default slog handler writes to stderr and the binary never calls `slog.SetDefault`, so it can't corrupt the stdio JSON-RPC stream on stdout). But since stdout is the protocol channel in stdio mode, relying on the process-global default is a footgun: a later innocuous `slog.SetDefault(...os.Stdout...)` would inject log lines into the framing. Consider passing an explicit `Logger:` (stderr) here so the proxy's JWKS logging can never depend on the global default.

---

eunolabs (commented):
Self-review of the fix (correctness + cleanup angles). Overall the change is sound — the unification is the right depth, the A1/A2 fixes are correct, and coverage is good (`pkg/capability` 97.9%). No blocking issues found. Three inline notes above, all non-blocking:

1. **Force-refresh now shares one singleflight key with non-forced refresh** (`jwkscache.go`) — intended A3 simplification; residual narrow, fail-closed, self-healing window on concurrent rotation + a now-untested concurrency path. Suggest a doc-comment caveat.
2. **Symmetric JWT-claims-side condition gap remains** (`pdp_jwt.go` `decideInner`/`Decide`) — `evaluateJWTConditions` still sees `nil` args for resource/prompt, mirroring the bug fixed on the manifest side; plus a `default`-case re-introduction hazard.
3. **cmd cache relies on `slog.Default()`** — latent stdio-stream footgun; pass an explicit stderr logger.

Things I specifically checked and cleared (not issues):

- **`*new(L)` fail-closed marshal** in the new generic `filterListResult` is byte-identical to the old per-type `mcpXListResult{}` marshal (each envelope has one non-omitempty slice field) → `{"tools":null}` etc. Behavior-preserving.
- **Caller-cancellation / half-open-probe stranding** — preserved by construction: `context.WithoutCancel` + `circuitbreaker.Do` make caller cancellation invisible to the breaker, so the old manual `guardedFetch` accounting (and its deleted tests) are correctly obsolete.
- **`JWKSClient` has no default breaker** while `NewJWTPDP` installs one — pre-existing and intentional (library opt-in vs. shipped-binary default, per `architecture.md`), not a regression.
- **`WithoutCancel` + HTTP timeout** — the shipped client (`newJWKSHTTPClient`) sets a 10s `Timeout` and `NewJWKSCache` defaults a 10s client, so no unbounded hang; only an explicitly Timeout-less custom client (caller misuse) is exposed, same as the pre-existing `JWKSClient`.
- **Cross-file call sites** (route.go shared-cache assignment, `decideInner` inner-non-nil guards, all six `filterListResult` callers) — all type-correct and behavior-preserving.

Optional nit (not posted inline): `jwksAlgorithms` (jwks.go) and `jwksAlgorithmsIDP` (pdp_jwt.go) are byte-identical lists that "must stay in lockstep"; now that a shared `pkg/capability` surface exists, exporting one and deleting the copy would remove the drift risk — on-theme with this PR's consolidation, but pre-existing so reasonable to defer.

---

eunolabs:
Addressed the review notes in `58dbf1d`:

1. **Force-refresh singleflight caveat** — documented the join-may-return-stale-under-concurrency window on `JWKSCache.ForceRefresh`, and added `TestJWKSCache_ForceRefresh_JoinsInFlightNonForced` to cover the concurrency path the deleted hand-rolled test used to pin.
2. **Symmetric JWT-claims-side condition gap** — `Decide` now synthesizes `{"uri"}`/`{"name"}` (via `jwtConditionArgs`) before evaluating shorthand conditions, mirroring the inner-manifest synthesis, so a `resource:`/`prompt:` claim carrying an `allowedValues` condition is evaluated instead of denying with `MISSING_CONTEXT`. New test `TestJWT_ResourcePromptShorthandConditions`. Also added the requested guard comment on `decideInner`'s `default` case.
3. **slog logger** — the cmd cache now passes an explicit stderr `slog` logger instead of inheriting `slog.Default()`.

All green locally: `go build`, `go vet`, `make lint`, `make test -race` (all packages), `make check-fmt`, `make check-license`.

---

PR #413, opened 2026-06-14
Audit write-failure observability and kill-switch refresh logging

### LINKED ISSUE #361: [rev-20260614T0606Z-ce3350] File descriptor leak in auditSink.Close() when Sync() fails

eunolabs:
## Summary

Add observability for audit write failures (full disk, EIO, serialization errors) via a dedicated `WriteFailures()` counter and `eunox_audit_write_failures_total` metric, distinct from dropped records. Add edge-triggered logging for kill-switch background-refresh failures (reconcile ticks and pub/sub resyncs) to surface Redis outages without flooding the log. Make audit-sink `Close()` failures exit the proxy non-zero so buffered records are not silently lost.

## Why

**Audit write failures were invisible.** Records that reached the drainer but could not be written to disk (full disk, EIO, serialization error) were silently discarded with no metric or log breadcrumb. An operator scraping `eunox_audit_dropped_records_total` would see zero while audit coverage silently degraded — exactly the failure mode an audit tool exists to prevent. The two loss modes are distinct: dropped records are queue back-pressure on a healthy file; write failures mean the file itself is broken. They need separate observability.

**Kill-switch refresh failures were silent.** A failed reconcile-tick or pub/sub-triggered resync against Redis discarded the error silently, so the cache could serve stale state with no log signal. A sustained outage would emit nothing to the operator log, only surfacing via a stale `/healthz` check.

**Audit-sink close failures were swallowed.** A `Sync()` or `Close()` error at shutdown (e.g. a full disk that persists to process exit) was discarded, so buffered audit records could be lost with no signal to a supervisor.

Closes #361 (audit close error handling).

## Type of change

- Bug fix (audit write failures, kill-switch logging, close error handling)
- New feature (write-failure metric and health endpoint field)

## How it was tested

```
make test    # All tests pass; new tests added:
             # - TestAuditSink_WriteFailuresCounted (verifies write failures are counted, not dropped)
             # - TestRedis_ReconcileRefresh_LogsAndThrottles (verifies edge-triggered logging)
             # - Updated TestRedisKillSwitch_Integration and TestRedisKillSwitch_GlobalActivate
             #   to use require.Eventually instead of fixed sleep (more robust)
make lint    # go vet + golangci-lint; .golangci.yml simplified (removed stale suppressions)
```

## Security considerations

**Fail-closed preserved.** Write failures are counted and logged but do not change the enforcement decision — the tool call result is still applied and returned. The audit record is lost (unavoidable on a broken file), but the policy decision stands.

**Audit tape.** Write-failure counting is a new field (`writeFailures` atomic counter) and metric (`eunox_audit_write_failures_total`), not a change to record structure or signing. The counter is incremented before any write attempt, so a serialization error, rotation failure, or disk write error all surface the same way.

**Secrets.** No secrets are logged; the write-failure warning is one-shot and generic ("check disk space and I/O").

**Authority.** No change to what is permitted; this is pure observability.

## Checklist

- [x] Docs updated (`README.md`, `docs/threat-model-mcp.md`, `CHANGELOG.md`)
- [x] Health endpoint schema updated (`cmd/eunox-mcp/health.go` adds `auditWriteFailed` field)
- [x] PR description explains *why* (audit/kill-switch observability gaps)

https://claude.ai/code/session_012oBGE6eyMcHv4YGUe48FW7


cmd/eunox-mcp/audit.go
```diff
@@ -655,7 +689,18 @@ func (s *auditSink) writeRecord(rec *auditRecord) {
...
+		// The backing file is unwritable (full disk, EIO, ...). The record is
+		// lost and the chain head is intentionally not advanced, so the chain
+		// stays internally consistent — but the trail is now incomplete. Count
+		// every loss (eunox_audit_write_failures_total) and warn once; a flood of
+		// per-record stderr lines under a sustained outage is not actionable, and
+		// the count plus the non-zero Close/exit are the surfaces operators alert
+		// on. The loss cannot be self-recorded with an in-chain marker the way a
+		// back-pressure drop is, because the file the marker would go to is the
+		// thing that is failing.
+		if s.writeFailures.Add(1) == 1 {
```
eunolabs: **Low-severity / robustness.** This one-shot warning gate shares the `writeFailures` counter with the marshal-error increments (~lines 654/664) and the nil-file increments (~lines 673/680). If any of those bumps the counter to ≥1 *before* the first real write error, `s.writeFailures.Add(1) == 1` is false and this intended one-time WARNING never prints — the operator gets no live stderr alert, only the deferred `Close()` summary.

In practice the write-error path is hit first today (the `s.f == nil` branches are defensive — `rotate()` never nils `s.f` — and a marshal error on these records is essentially impossible), so this is latent, not live. But it's fragile: the nil-file path is itself silent, so if a future change ever makes it reachable it would both lose records silently *and* suppress this warning, quietly breaking the documented "warn once" contract.

Consider gating the live warning on a dedicated `sync.Once`/bool reserved for the write-error path, so the alert behaviour is independent of the other (silent) increment sources. The metric (`eunox_audit_write_failures_total`) correctly counts all sources regardless.

---

pkg/killswitch/redis.go
```diff
@@ -127,11 +133,37 @@ func (r *Redis) reconcileLoop(ctx context.Context) {
...
+// otherwise non-fatal: the switch fails open and serves the last-known local cache
+// until Redis recovers (ADR-0003).
+func (r *Redis) reconcileRefresh(ctx context.Context) {
+	err := r.refreshState(ctx)
+	if r.logger == nil {
+		return
+	}
+	r.mu.Lock()
+	wasLogged := r.reconcileErrLogged
+	r.reconcileErrLogged = err != nil
```
eunolabs: **Low-severity / observability.** The edge-trigger state (`reconcileErrLogged`) is updated in a separate critical section from the refresh result. `refreshState` runs *outside* this lock and authoritatively sets `lastRefreshErr`, but here we read/write `reconcileErrLogged` against this goroutine's own local `err` snapshot. Since `reconcileLoop` (the ticker) and `handlePubSubMessage` (the unknown-payload resync) can both call `reconcileRefresh` concurrently, their `refreshState` results and their `reconcileErrLogged` updates can interleave.

If Redis state changes between two concurrent refreshes, the log breadcrumb can drift from actual health: a duplicate failure WARNING for a single outage, or a false/missed "recovered" INFO (logged while `lastRefreshErr` is still non-nil). The kill-switch decision path is unaffected — authoritative health is `lastRefreshErr`/`HealthStatus` — so this is a log-accuracy issue only. Under a *steady* outage it still logs exactly once; the drift needs a flap during the narrow concurrent-resync window.

If we want exactly-once-per-edge under concurrency, derive the transition from `lastRefreshErr` inside `refreshState`'s own critical section (fold the edge decision + the result write under one lock) instead of tracking it separately here.

---

eunolabs (commented):
Self-review of the release-readiness fixes. CI is green/in-progress with no failures, and the changes do what they claim. Two **low-severity** observations below — both are latent/narrow (neither affects enforcement or the kill-switch decision path), so they're judgment calls rather than blockers:

1. **`audit.go`** — the one-shot write-failure stderr warning shares its counter with the (silent) nil-file and marshal-error increment paths, so it could be suppressed if one of those fires first. Latent today because those paths are defensive/near-impossible, but fragile.
2. **`pkg/killswitch/redis.go`** — the edge-triggered reconcile logging tracks its transition state separately from the authoritative refresh result, so under concurrent reconcile+pub-sub resync with a flapping Redis the *log breadcrumb* can drift (duplicate warning / false recovery). Log-accuracy only; health reporting via `lastRefreshErr`/`HealthStatus` stays correct.

Everything else verified clean: exit-code LIFO ordering flushes the audit queue before exit, the single-drainer concurrency model holds, the HMAC chain stays consistent on a failed write, the new tests are race-free and assert correctly, and there's no cross-file caller breakage. Happy to push fixes for either item if you'd like them tightened.

---

PR #421, opened 2026-06-14
Fix session rev-20260614T1500Z-248c33 bugs

### LINKED ISSUE #414: [rev-20260614T1500Z-248c33] audit key creation fails permanently on filesystems without hard-link support (CIFS/NFS/FAT32/some Docker volumes)
**Location:** `cmd/eunox-mcp/audit.go:1083-1117`
**Reference:** https://github.com/eunolabs/eunox/blob/2774a3fec35bd1bf28557da66bba03e8dd1228d4/cmd/eunox-mcp/audit.go#L1083-L1117
**Severity:** medium
**Session:** rev-20260614T1500Z-248c33

## Description

`loadOrCreateAuditKeys` uses `os.Link(tmpName, keyPath)` as an atomic write-and-publish handshake to prevent a concurrent-startup race where two processes generate divergent keys. The problem: `os.Link` (hard links) is not supported on many real-world filesystems:

- CIFS / Samba mounts (common in enterprise and CI environments)
- NFS with `no_hardlinks` option or certain NFS server configurations
- FAT32 / exFAT (USB sticks, cross-platform shared volumes)
- Some Docker bind-mount backends (notably the default on Docker Desktop for Mac/Windows pre-VirtioFS)
- Some container overlay filesystems in certain Kubernetes configurations

When `os.Link` fails with `EPERM` or `ENOSYS` on these systems, the error is **not** `os.IsExist(err)`, so the function falls through to:
```go
return nil, fmt.Errorf("publishing audit key file %q: %w", keyPath, err)
```
This propagates up through `openAuditSink` → `main`, causing the proxy to exit with an error on first run. Because `defer os.Remove(tmpName)` also runs, no partial state is left behind—but the proxy simply cannot start.

## Trigger / repro

1. Mount a CIFS share (e.g. `//server/share /mnt/audit -t cifs`) and point `--audit-key-path` at it
2. Start `eunox-mcp proxy` for the first time
3. The key file does not exist yet → `loadOrCreateAuditKeys` generates a key, writes the temp file, then calls `os.Link` which returns `EPERM` → proxy exits before serving any requests

Any filesystem that rejects `link(2)` with `EPERM`/`ENOSYS` triggers the failure. The temporary file at `dir/.audit-key-*.tmp` is cleaned up by the deferred `os.Remove`, so the failure is clean but final on each attempt.

## Suggested fix

When `os.Link` fails with a non-`EEXIST` error that indicates the filesystem doesn't support hard links (specifically `EPERM`, `ENOSYS`, or `EXDEV`), fall back to `os.Rename` for the atomic publish. `os.Rename` is supported on all filesystems and is atomic within a single directory on POSIX systems. The key race `os.Link` was protecting against (two processes both generating a key) can be addressed for the rename case by using `O_EXCL` semantics: attempt rename, then re-read the file if it already existed.

```go
if err := os.Link(tmpName, keyPath); err != nil {
    if os.IsExist(err) {
        // another process won the race — read its key
        ...
    }
    // Hard links not supported — fall back to atomic rename.
    // os.Rename on POSIX is atomic within the same filesystem,
    // but unlike Link it silently overwrites an existing target.
    // To preserve the race-safety guarantee, check for existence
    // under the rename path and re-read if we lost.
    if isNoHardlinkErr(err) {
        if renameErr := os.Rename(tmpName, keyPath); renameErr != nil {
            return nil, fmt.Errorf("publishing audit key file %q: %w", keyPath, renameErr)
        }
        // We may have overwritten a key another process just wrote.
        // Re-read and trust what's on disk so all processes converge.
        data, readErr := os.ReadFile(keyPath)
        ...
        return parseAuditKeys(data)
    }
    return nil, fmt.Errorf("publishing audit key file %q: %w", keyPath, err)
}
```

where `isNoHardlinkErr` checks `errors.Is(err, syscall.EPERM) || errors.Is(err, syscall.ENOSYS) || errors.Is(err, syscall.EXDEV)`.

Alternatively, unconditionally use `os.Rename` when the process holds exclusive creation via `os.CreateTemp` with a randomised name—the temp file name itself provides uniqueness sufficient to avoid clobbering another process's temp, and the re-read after rename detects if we lost the race.

### LINKED ISSUE #415: [rev-20260614T1500Z-248c33] audit key temp file not synced before os.Link — power failure during first-run key generation permanently bricks the key path
**Location:** `cmd/eunox-mcp/audit.go:1094-1102`
**Reference:** https://github.com/eunolabs/eunox/blob/2774a3fec35bd1bf28557da66bba03e8dd1228d4/cmd/eunox-mcp/audit.go#L1094-L1102
**Severity:** low
**Session:** rev-20260614T1500Z-248c33

## Description

`loadOrCreateAuditKeys` writes the hex-encoded key into a temp file, closes it, and then hard-links it into place with `os.Link`:

```go
if _, err := tmp.Write(encoded); err != nil {
    _ = tmp.Close()
    return nil, fmt.Errorf("writing audit key: %w", err)
}
if err := tmp.Close(); err != nil {
    return nil, fmt.Errorf("closing temp audit key file: %w", err)
}

if err := os.Link(tmpName, keyPath); err != nil {
```

`tmp.Close()` flushes Go's user-space buffer to the kernel but does **not** call `fsync(2)`. On Linux and most POSIX systems, the kernel can hold the data in the page cache without committing it to storage. If the system loses power (or the OS panics, or the storage device is hot-unplugged) between `Close()` and the kernel flushing the page cache, the inode at `keyPath` will exist on disk but its data blocks will contain stale/zero bytes.

On the next boot:
1. `os.ReadFile(keyPath)` **succeeds** (the inode exists)
2. `parseAuditKeys("")` returns `fmt.Errorf("contains no key lines")`
3. `openAuditSink` returns an error
4. The proxy refuses to start

Because the code explicitly treats a *present-but-invalid* key file as a hard error (not regenerated, to avoid silently invalidating an existing signed log), there is no automatic recovery. An operator must manually `rm keyPath` and restart.

## Trigger / repro

1. Deploy `eunox-mcp` on a system that lacks a battery-backed write cache (bare metal without UPS, embedded hardware, VM with pass-through NVMe in certain configurations)
2. Start the proxy for the first time (key file does not yet exist)
3. Pull power / trigger a kernel panic in the ~milliseconds between `tmp.Close()` and the page-cache flush
4. On reboot: `[eunox-mcp] audit key: audit key file "~/.eunox/audit.key": contains no key lines`; proxy will not start

The window is narrow, making this unlikely in practice, but the consequence (operator intervention required) is severe enough to warrant a fix.

## Suggested fix

Add `tmp.Sync()` after writing and before closing:

```go
if _, err := tmp.Write(encoded); err != nil {
    _ = tmp.Close()
    return nil, fmt.Errorf("writing audit key: %w", err)
}
if err := tmp.Sync(); err != nil {   // ← add this
    _ = tmp.Close()
    return nil, fmt.Errorf("syncing temp audit key file: %w", err)
}
if err := tmp.Close(); err != nil {
    return nil, fmt.Errorf("closing temp audit key file: %w", err)
}
```

This matches the existing pattern in `rotate()` which explicitly calls `s.f.Sync()` before closing the rotated log file for the same durability reason. The key file is written once at process startup and is far smaller than a log rotation, so the `fsync` overhead is negligible.

### LINKED ISSUE #416: [rev-20260614T1500Z-248c33] hardcoded 4 MiB scanner limit in verifyAuditLog and readLastAuditLine silently breaks audit-verify and chain resumption when large records are written (audit-only mode)
**Location:** `cmd/eunox-mcp/audit.go:809-810` (verifyAuditLog) and `cmd/eunox-mcp/audit.go:335` (readLastAuditLine)
**Reference:** https://github.com/eunolabs/eunox/blob/2774a3fec35bd1bf28557da66bba03e8dd1228d4/cmd/eunox-mcp/audit.go#L809-L810
**Severity:** medium
**Session:** rev-20260614T1500Z-248c33

## Description

Two functions impose a 4 MiB hard cap on the size of a single audit record:

```go
// verifyAuditLog (line 809-810)
scanner := bufio.NewScanner(r)
scanner.Buffer(make([]byte, 4<<20), 4<<20)

// readLastAuditLine (line 335)
const maxTail = 4 << 20
```

Audit records are written with `json.Marshal(rec)` and there is **no cap** on record size in `writeRecord`. In **audit-only (wiretap) mode**, the `record()` call logs `params.Arguments` verbatim in the `Details` field:

```go
Details: copyDetails(details),  // includes params.Arguments in audit-only allow path
```

A tool call whose arguments contain a large payload—for example, a `write_file` tool with a multi-megabyte document body, a `create_embedding` tool with a large text corpus, or any tool operating on binary-as-base64 data—can produce a single audit record exceeding 4 MiB. The `writeRecord` path has no guard against this and will successfully append the oversized record to the log.

### Impact 1: `audit-verify` fails permanently for any log containing an oversized record

`bufio.Scanner.Scan()` returns `false` and `scanner.Err()` returns `bufio.ErrTooLong` when a line exceeds `max` (4 MiB). `verifyAuditLog` propagates this as:

```go
if err := scanner.Err(); err != nil {
    return res, err  // verification stops; records after the oversized one are unverified
}
```

The oversized record itself is reported as a scanner error, not as an `INVALID` record. All records written *after* the oversized one are silently skipped. The chain integrity check (`chainBreaks`) never runs for those later records. The audit-verify command exits non-zero but the output gives no useful indication of *which* record is the problem.

### Impact 2: Chain resumption at startup picks the wrong prev_hmac / seq after a restart

`readLastAuditLine` reads the last 4 MiB of the log and finds the last `\n`-terminated line in that window:

```go
if i := bytes.LastIndexByte(trimmed, '\n'); i >= 0 {
    return string(bytes.TrimSpace(trimmed[i+1:]))
}
```

If the most-recent record exceeds 4 MiB, the 4 MiB tail begins in the *middle* of that record. `LastIndexByte` then finds a newline within the record's JSON payload (embedded in a string value), not the record boundary. The returned "last line" is a JSON fragment, `json.Unmarshal` fails, and `openAuditSink` proceeds with `s.seq = 0` and `s.prevHMAC = ""`. The next record is then stamped `seq=1` and `prev_hmac="sha256:genesis"`, breaking the chain at every process restart after an oversized record—causing spurious chain-break reports on all subsequent `audit-verify` runs.

## Trigger / repro

1. Deploy `eunox-mcp` in audit-only (wiretap) mode (`audit: true` in the manifest or `--audit-only`)
2. Issue a tool call with a large argument, e.g. a `write_file` with a ~5 MiB document body
3. The record is written successfully to the log
4. Run `eunox-mcp audit-verify`: the command exits with `scanner: token too long` and skips all subsequent records
5. Restart `eunox-mcp`: the chain head is resumed from a stale/wrong record; subsequent `audit-verify` reports spurious chain breaks

## Suggested fix

Two complementary changes:

**1. Cap large `Details` values at record-construction time in `record()`.**  
Before handing the record to the drainer channel, truncate or omit argument values in `Details` that exceed a configurable limit (e.g. 512 KiB per value, 1 MiB total Details). Log a synthetic marker field (`"truncated": true`) so the cap is visible in the record. This is the correct place to bound record size—before the channel send—rather than silently failing during verification.

**2. Make the `verifyAuditLog` scanner and `readLastAuditLine` tail configurable / self-healing.**  
If a record exceeds the buffer, `verifyAuditLog` should log the line number and byte offset of the oversize record, mark it as a skippable error, and continue scanning (resync on the next `\n`). For `readLastAuditLine`, if no complete JSON line is found in the tail window, it should seek backward further (doubling the window up to a hard maximum) or warn and return `""` so that the chain resumes safely from a known-good predecessor rather than a stale fragment.

The simpler near-term fix is **option 1** alone—bounding record size at write time makes the scanner limit a maintenance footnote rather than a correctness hazard.

eunolabs:
## Bugs addressed (session rev-20260614T1500Z-248c33)

- Closes #414 — Audit key creation aborted permanently on filesystems without hard-link support (CIFS/Samba, some NFS configs, FAT32/exFAT, some container bind-mounts) because `os.Link` returns `EPERM`/`ENOSYS`/`EXDEV` there. `loadOrCreateAuditKeys` now classifies those errnos via `isNoHardlinkErr` and falls back to an atomic `os.Rename`, then re-reads the key from disk (`readPublishedAuditKeys`) so concurrent starters converge on the persisted key. A genuine I/O error still propagates and the create-race (`EEXIST`) path is unchanged.

- Closes #415 — The first-run key temp file was only `Close()`d (page-cache flush) before publish, so a power loss could leave `keyPath` present but zero/stale, bricking startup (a present-but-invalid key is a hard error, never regenerated). Added `tmp.Sync()` after the write and before `Close()`, matching the durability pattern `rotate()` already uses for rotated logs.

- Closes #416 — In audit-only (wiretap) mode the allow path logs `params.Arguments` verbatim, so a multi-megabyte tool argument could produce a single record larger than the 4 MiB `audit-verify` scanner buffer and the chain-resume tail window — making `audit-verify` fail with "token too long" and skip all later records, and making chain resumption pick a stale JSON fragment after a restart. Records are now bounded at write time (`boundAuditDetails`, on the drainer goroutine before signing): any string value over 512 KiB is replaced with a self-describing placeholder, and if the whole `details` map still exceeds 1 MiB it is replaced with a single `_eunox_truncated` marker. Bounded records stay well under the scanner buffer, so both impacts are eliminated; the now-accurate invariant is documented at the scanner/tail call sites.

## Testing
- `go test -race -count=1 ./cmd/eunox-mcp/ ./pkg/...` — all pass.
- New tests: `boundAuditDetails` per-value/total-cap/nested behavior; an end-to-end guard writing a ~5 MiB argument and asserting the record stays under the buffer, `audit-verify` scans cleanly with the chain intact, and the chain resumes correctly across a restart; `isNoHardlinkErr` classification table; `readPublishedAuditKeys` valid/missing/corrupt paths; and a rename-fallback publish round-trip.
- `make check-fmt`, `go vet ./...`, and `golangci-lint run ./cmd/eunox-mcp/` are clean.

https://claude.ai/code/session_015awtGgEGS3EKoLvPXD9jBj

---
_Generated by [Claude Code](https://claude.ai/code/session_015awtGgEGS3EKoLvPXD9jBj)_


eunolabs (commented):
Self-review round 1 — two substantive findings on the #414 fix:

1. `cmd/eunox-mcp/audit.go` (rename-fallback comment, ~line 1271): the comment claims "whichever writer's rename landed last is what every process converges on, so no process keeps signing with a key that is not the one persisted." That overclaims. On a hard-link-less filesystem two concurrent first-starts both take the rename branch; a process that re-reads `keyPath` *before* a later writer's `os.Rename` lands will return — and keep signing with — a key that is then overwritten on disk. `os.Link`'s first-writer-wins guarantee genuinely closed this; rename only narrows it. The comment should state the residual window honestly rather than implying full convergence.

2. `cmd/eunox-mcp/audit_test.go` (`TestLoadOrCreateAuditKeys_RenameFallbackPublishes`): this test simulates the rename by hand and then calls `loadOrCreateAuditKeys`, which takes the *file-already-exists reload* path — so the real `isNoHardlinkErr` → `os.Rename` branch inside `loadOrCreateAuditKeys` is never exercised. The core of the #414 fix is untested. Add a test seam to inject an `os.Link` failure (EPERM) and assert the fallback actually creates a usable, reloadable key.

Will address both.

---

eunolabs:
Round 1 findings addressed in 13d95ff:

1. Rewrote the rename-fallback comment in `loadOrCreateAuditKeys` to state the downgrade honestly — it converges the *persisted* key on the last rename and lets a race-loser adopt the winner, but does not close the window where an earlier starter already re-read a key a later rename overwrote. It is explicitly framed as a deliberate, confined downgrade of `os.Link`'s first-writer-wins guarantee for hard-link-less paths (the os.Link path still gives the full guarantee elsewhere).
2. Added an `osLink` test seam and rewrote the fallback test (`TestLoadOrCreateAuditKeys_RenameFallbackOnNoHardlink`) to force `os.Link` to fail with EPERM, so the real `isNoHardlinkErr` → `os.Rename` branch now executes under test: it asserts a usable key is created, no temp file is left behind, and the key reloads identically on restart.

Tests, `golangci-lint`, and `gofmt` all clean.

---

eunolabs:
Self-review round 2 — one substantive finding, now addressed in ce45b39:

**Docs inconsistency (repo convention: behavior changes need a docs update).** The #416 fix truncates oversized argument values, which directly contradicted `docs/conformance.md` ("The audit log is not filtered, **not truncated**, and not rewritten"), and the threat-model §6.2 / its table described the observe-mode "full argument map" with no size bound. Updated:
- `docs/conformance.md` — reworded the guarantee: the record *set* is never filtered/dropped/rewritten, but an individual oversized value is size-capped (with a forward reference to §6.2).
- `docs/threat-model-mcp.md` §6.2 — added the "Argument size is bounded before write" paragraph (512 KiB per-value placeholder, 1 MiB total-map `_eunox_truncated` marker) and annotated the data-capture table row.
- `docs/threat-model-mcp.md` changelog — added a 1.16 entry covering all three fixes (#416 record bound, #415 key fsync, #414 hard-link rename fallback).

`README.md`'s high-level "full argument map" line is left as-is — it states what is captured, not a no-truncation guarantee, and defers detail to the threat model.

---

eunolabs:
✅ Ready to merge — all session rev-20260614T1500Z-248c33 bugs addressed (#414, #415, #416), self-review complete (4 rounds, no outstanding comments), CI fully green (Test, Lint, Build matrix, gosec/govulncheck/Trivy/CodeQL, demo + enforcement-compliance suites all passing), `mergeable_state: clean`.

---

PR #422, opened 2026-06-14
Fix session rev-20260614T1510Z-94931c bugs

### LINKED ISSUE #418: [rev-20260614T1510Z-94931c] verifyAuditLog never checks the genesis sentinel on the first record, missing leading-truncation tampering on non-rotated logs
**Location:** `cmd/eunox-mcp/audit.go:816-862`
**Reference:** https://github.com/eunolabs/eunox/blob/2774a3fec35bd1bf28557da66bba03e8dd1228d4/cmd/eunox-mcp/audit.go#L816-L862
**Severity:** medium

## Description

`verifyAuditLog` checks chain integrity by comparing each record's `prev_hmac` against the preceding record's `_hmac`. For the very first record in the file (`havePrev == false`), the code sets `firstSeq` and skips the chain check entirely:

```go
if !havePrev {
    res.firstSeq = rec.Seq
} else {
    if rec.PrevHMAC != prevHMAC {
        res.chainBreaks++
        ...
    }
    ...
}
havePrev = true
```

Every non-genesis record in the chain has its `prev_hmac` validated. But the first record's `prev_hmac` is **never** validated against anything.

For a fresh (non-rotated) log, the first record is always stamped `prev_hmac: "sha256:genesis"` by `writeRecord`:

```go
default:
    // Genuine first record of a fresh log: no prior record exists
    rec.PrevHMAC = auditGenesisPrev  // "sha256:genesis"
```

This invariant is verifiable: any file whose first record has `seq == 1` **must** carry `prev_hmac == "sha256:genesis"`. The verifier can and should check it.

## Trigger / repro

1. Start a fresh audit log. At least one record is written (seq=1, prev_hmac="sha256:genesis").
2. An attacker removes the first record from the log.
3. The second record becomes the new first; it has `prev_hmac` pointing to the deleted record's `_hmac` (not the genesis sentinel).
4. Run `eunox-mcp audit-verify`.

Result: no `CHAIN BREAK` is reported even though the leading record is missing. The missing genesis is undetected.

If `seq` of the new-first record is > 1, the `SEQ GAP` check fires:

```go
if rec.Seq != prevSeq+1 {
    res.chainBreaks++
    ...
}
```

But only between consecutive records; the comparison for the very first record is skipped because `havePrev` is false when it is processed. So even the sequence gap (0 → N instead of 0 → 1) is not reported.

## Suggested fix

When processing the first decodable record (`!havePrev`), check:

```go
if !havePrev {
    res.firstSeq = rec.Seq
    // For a log that starts at seq 1 (non-rotated, never-continued),
    // the genesis sentinel is verifiable.
    if rec.Seq == 1 && rec.PrevHMAC != auditGenesisPrev {
        res.chainBreaks++
        _, _ = fmt.Fprintf(out,
            "CHAIN BREAK at seq 1: prev_hmac is %q, expected genesis sentinel %q\n",
            rec.PrevHMAC, auditGenesisPrev)
    }
} else {
    ...
}
```

This closes the detectable leading-truncation gap for non-rotated logs (where seq starts at 1) while leaving rotated logs (where seq starts mid-chain) unchanged — a rotated log's first record legitimately has a non-genesis `prev_hmac` pointing to the last record of the previous file.

The existing code comment correctly notes that "trailing truncation ... and, on a rotated log, leading removal cannot be proven from the file alone." This fix closes only the provable case (non-rotated log, seq == 1), consistent with that statement.

**Session:** rev-20260614T1510Z-94931c

### LINKED ISSUE #419: [rev-20260614T1510Z-94931c] flushDropMarker advances lastDroppedMarked before the write succeeds, permanently losing drop evidence on disk-write failure
**Location:** `cmd/eunox-mcp/audit.go:569-593`
**Reference:** https://github.com/eunolabs/eunox/blob/2774a3fec35bd1bf28557da66bba03e8dd1228d4/cmd/eunox-mcp/audit.go#L569-L593
**Severity:** medium

## Description

`flushDropMarker` advances `s.lastDroppedMarked` **before** calling `writeRecord`, so that a write failure does not cause the same marker to be re-emitted on every subsequent record (the comment explains this explicitly). But the trade-off is asymmetric: the comment's goal of avoiding re-emission is achieved, but the cost is that a failed write **permanently** loses the drop evidence from the tamper-evident chain.

```go
func (s *auditSink) flushDropMarker() {
    cur := s.dropped.Load()
    if cur <= s.lastDroppedMarked {
        return
    }
    newDrops := cur - s.lastDroppedMarked
    marker := auditRecord{
        ...
        DenialCode: "AUDIT_RECORDS_DROPPED",
            "dropped":       newDrops,
            "dropped_total": cur,
        },
    }
    // Mark as accounted before the write attempt: if the write itself fails we do
    // not want to spin re-emitting the same marker on every subsequent record.
    s.lastDroppedMarked = cur  // <-- advanced HERE, before write
    s.writeRecord(&marker)     // <-- if this fails, the loss is gone forever
}
```

If `writeRecord` fails (disk full, `s.f == nil` due to a failed rotation, a kernel write error), `s.lastDroppedMarked` has already been set to `cur`. The drop window is now "accounted for" in memory but **not on disk**. On the next call to `flushDropMarker`, `cur <= s.lastDroppedMarked` and the function returns immediately — the gap is silently skipped.

## Impact

1. `audit-verify` sees no `AUDIT_RECORDS_DROPPED` marker in the log, even though records were dropped.
2. The chain sequence numbers continue uninterrupted, so `audit-verify` reports no chain break.
3. A forensic investigator relying on `audit-verify` to detect dropped records receives a **false negative**.
4. The drop marker is specifically designed to be tamper-evident ("it is a normal chained record, so deleting it breaks the chain like any other record"). Pre-marking defeats this: a disk-write failure is now indistinguishable from a clean log.

The failure mode can be triggered by:
- A full disk at the moment `flushDropMarker` calls `writeRecord`
- A failed rotation that sets `s.f = nil` (the rotate path's reopen failure branch already handles this)
- Any transient I/O error surfaced by `os.File.Write`

## Suggested fix

Track a separate "pending" vs "committed" drop count so that the marker is only considered accounted for once it has been durably written. One approach:

```go
func (s *auditSink) flushDropMarker() {
    cur := s.dropped.Load()
    if cur <= s.lastDroppedMarked {
        return
    }
    newDrops := cur - s.lastDroppedMarked
    marker := auditRecord{
        ...
        DenialCode: "AUDIT_RECORDS_DROPPED",
            "dropped":       newDrops,
            "dropped_total": cur,
        },
    }
    // Only mark as accounted AFTER writeRecord succeeds.
    // writeRecord advances s.seq and s.prevHMAC only on a successful write,
    // so a retry (after the drop marker grows by more drops) correctly
    // reflects the full accumulated count. To avoid a tight retry loop on
    // a persistently broken writer, the caller (drain) already limits how
    // often flushDropMarker is called — once per record.
    s.writeRecord(&marker)
    if s.seq == marker.Seq { // writeRecord advances seq only on success
        s.lastDroppedMarked = cur
    }
}
```

Alternatively, inspect whether `s.prevHMAC` changed (it only changes on a successful write) as a write-success signal, since `writeRecord` is already designed to leave the chain head unchanged on failure.

**Session:** rev-20260614T1510Z-94931c

### LINKED ISSUE #420: [rev-20260614T1510Z-94931c] copyDetails is documented as "deep clone" but passes []string values through uncopied, violating its race-free contract
**Location:** `cmd/eunox-mcp/audit.go:472-499` (`copyDetails` / `copyDetailValue`)
**Reference:** https://github.com/eunolabs/eunox/blob/2774a3fec35bd1bf28557da66bba03e8dd1228d4/cmd/eunox-mcp/audit.go#L472-L499
**Severity:** low

## Description

`copyDetails` is called in `record()` specifically to give the drainer goroutine an independent copy of the `Details` map, so that the calling goroutine and the drainer never share backing storage. The function's own docstring states:

> "Nested `map[string]interface{}` and `[]interface{}` (the only container shapes a JSON-decoded argument map produces) are cloned recursively by `copyDetailValue`. **Other values are copied as-is**: scalars (string, bool, float64, json.Number, ...) are immutable, and the only non-scalars the engine puts in denial details are `[]string` slices borrowed from the immutable, already-loaded manifest (allowedCIDRs, allowedOperations, ...), **which are never mutated after a decision**."

The correctness of the shallow copy for `[]string` rests entirely on the assertion that those slices are never mutated. That is true today because all `[]string` values in denial details come from fields like `ipr.CIDRs`, `ao.Operations`, `ae.Extensions`, `at.Tables`, etc. — all loaded once at manifest-parse time and never modified.

However `copyDetailValue` handles only two container types:

```go
func copyDetailValue(v interface{}) interface{} {
    switch t := v.(type) {
    case map[string]interface{}:
        return copyDetails(t)
    case []interface{}:
        out := make([]interface{}, len(t))
        for i, e := range t {
            out[i] = copyDetailValue(e)
        }
        return out
    default:
        return v  // []string, []int64, etc. returned unchanged — shared backing array
    }
}
```

A `[]string` falls through to the `default` arm and is returned as-is. The drainer goroutine and the caller then share the same slice backing array.

## Trigger / repro

Any code path that places a `[]string` in `Details` that is later mutated will produce a data race detectable by `go test -race`. Two concrete near-miss scenarios:

1. **Audit-mode allow path** (`record()` is called with the live `params.Arguments` map). If `Arguments` contained a `[]string` value (e.g. a tool that takes a `paths []string` argument), the drainer and the tool-call handler would share the backing array. The tool handler could then mutate the slice while the drainer marshals it.

2. **New condition handler** added by a contributor who constructs `Details["allowedValues"] = someSlice` where `someSlice` is populated from the current request (not a manifest constant) and later modified. The race would only surface under `-race`, not during regular testing.

## Suggested fix

Add a `[]string` case to `copyDetailValue`:

```go
func copyDetailValue(v interface{}) interface{} {
    switch t := v.(type) {
    case map[string]interface{}:
        return copyDetails(t)
    case []interface{}:
        out := make([]interface{}, len(t))
        for i, e := range t {
            out[i] = copyDetailValue(e)
        }
        return out
    case []string:
        // []string values (allowedCIDRs, allowedOperations, etc.) are immutable
        // manifest slices today, but the "deep clone" contract requires we never
        // share backing storage with the caller, regardless of the value's origin.
        return append([]string(nil), t...)
    default:
        return v
    }
}
```

This makes the guarantee unconditional — future contributors don't need to know whether a particular `[]string` value is "safe" to share — and adds negligible overhead for the small slices that currently appear in denial details.

**Session:** rev-20260614T1510Z-94931c

eunolabs:
## Bugs addressed (session rev-20260614T1510Z-94931c)

All three are in the audit subsystem (`cmd/eunox-mcp/audit.go`).

- Closes #420 — `copyDetailValue` fell through its `default` arm for `[]string`, returning the slice uncopied so the drainer goroutine shared the caller's backing array. This violated `copyDetails`' documented race-free deep-clone contract (which held only because every `[]string` in denial details happens to be an immutable manifest slice today). Added an explicit `[]string` case that copies element-wise, making the guarantee unconditional, and updated the docstrings.

- Closes #419 — `flushDropMarker` advanced `s.lastDroppedMarked` *before* attempting the write, so a failed `writeRecord` (full disk, EIO, a backing file lost to a failed rotation) permanently lost the `AUDIT_RECORDS_DROPPED` marker: the dropped window was accounted for in memory but never landed on disk, and the next flush returned early. `audit-verify` then saw no marker and no chain break — a false negative. Now the marked count advances only after `writeRecord` reports a durable write (`s.seq` advanced), so a later flush retries and the drop evidence survives a transient failure.

- Closes #418 — `verifyAuditLog` set `firstSeq` for the leading record and skipped the chain check entirely, never validating that record's `prev_hmac`. A fresh non-rotated log always stamps `seq 1` with the genesis sentinel, an invariant the verifier can prove: an attacker who deletes the true seq-1 record and rewrites the survivor to claim `seq 1` left a non-genesis `prev_hmac` that went unreported. Now checked when the first record has `seq 1`, exempting the empty `prev_hmac` used to resume onto a pre-chain legacy tail (#228) so a legacy continuation verified as a standalone slice is not a false positive (and a genesis→empty forgery is still caught by per-record HMAC).

## Testing

- New regression tests: `TestCopyDetailValue_ClonesStringSlice` (#420), `TestAuditDropMarker_RetriesOnWriteFailure` (#419), `TestAuditChain_DetectsRewrittenGenesis` + `TestAuditChain_GenesisSentinelHonored` (#418).
- `go test -race -count=1 ./...` — all packages pass, including the pre-existing `TestAuditChain_ResumesFromLegacyTail` which guards the #418 legacy-tail exemption.
- `go vet ./...`, `make check-fmt`, `make check-license` all clean.

https://claude.ai/code/session_01SoyXQ2drRshFhEdthgGowM

---
_Generated by [Claude Code](https://claude.ai/code/session_01SoyXQ2drRshFhEdthgGowM)_


eunolabs (commented):
Self-review round 1 — one substantive finding (docs convention).

**Missing docs update for the audit-verify behavior change.** CLAUDE.md requires that audit-behavior changes update the threat model in the same PR, and these three fixes change what the chain/`audit-verify` proves:

- `docs/threat-model-mcp.md:247` ("What the chain does and does not cover") states leading-record removal "on a rotated log" is unprovable, but does not mention that #418 now makes origin rewrite / leading removal **detectable on a non-rotated log** (first record `seq == 1` must carry the `sha256:genesis` sentinel). As written it understates the new coverage.
- `docs/conformance.md:306-313` enumerates what `audit-verify` can and cannot detect; same gap.
- The threat-model changelog table (latest row 1.16, `docs/threat-model-mcp.md:648`) has no row for #418 (genesis-sentinel check), #419 (drop-marker survives a write failure — previously a write failure could hide dropped records, a tamper-evidence false negative), or #420 (audit-detail `[]string` deep-clone race).

Code itself looks correct: #419 keys "durably written" off `s.seq` advancing, which only happens on a successful `writeRecord`; #418 correctly exempts the empty `prev_hmac` legacy-continuation marker (#228) so a standalone legacy slice isn't a false positive; #420's element-wise copy is correct. Fixing the docs now.

---

eunolabs:
Round 1 finding resolved in 2110c66: updated `docs/threat-model-mcp.md` §3.4 ("What the chain does and does not cover") to document the new non-rotated leading-removal detection, refreshed the `audit-verify` summary in `docs/conformance.md`, and added threat-model changelog row 1.17 covering #418/#419/#420. Re-reviewing the full diff now.

---

eunolabs:
✅ Ready to merge — all session rev-20260614T1510Z-94931c bugs addressed (#418, #419, #420), self-review complete with no outstanding comments. CI green on 2110c66 (Go CI + Demo Integration Tests), `go test -race ./...` passing locally, `go vet` / `make check-fmt` / `make check-license` clean.

---

PR #430, opened 2026-06-14
Fix session rev-20260614T1304Z-343add bugs

### LINKED ISSUE #403: [rev-20260614T1304Z-343add] handleAllowedTables returns MISSING_CONTEXT instead of CONDITION_FAILED when argument exists with wrong type
**Location:** `pkg/enforcement/handlers.go:719-738` (caller) and `pkg/enforcement/handlers.go:1282-1335` (`parseTableArgument`)

**Reference:** https://github.com/eunolabs/eunox/blob/38c4ee8273a340d955f3cf0fad6bd22d393b9331/pkg/enforcement/handlers.go#L719-L738

**Severity:** Medium

**Description:**

`parseTableArgument` returns `nil, nil` (not an error) for any argument type that is not `string`, `[]interface{}`, or `map[string]interface{}` — for example a number (`float64`) or boolean. The caller in `handleAllowedTables` then sees `len(tables) == 0` and returns a `MISSING_CONTEXT` denial with the message `"argument %q is missing or empty"`.

This is wrong on two counts:
1. The argument **is present** — it was found by `resolveArgument` (the `ok=true` branch) — so `MISSING_CONTEXT` is the wrong error code.
2. The message `"argument is missing or empty"` is factually incorrect; the argument carries a value of the wrong type.

The same bug class was fixed for `handleAllowedExtensions` and `handleRecipientDomain` in issue #324: both now have an explicit `default:` branch in their type switch that returns `CONDITION_FAILED` with a type-mismatch message. `handleAllowedTables` routes through `parseTableArgument`, which has no such branch, so the fix was never applied here. The comment in `handleAllowedExtensions` even says "mirroring `handleAllowedTables`", which is now ironic — it is `handleAllowedTables` that should mirror `handleAllowedExtensions`.

```go
// parseTableArgument — no default branch:
switch t := v.(type) {
case string:  ...
case []interface{}:  ...
case map[string]interface{}:  ...
}
return nil, nil  // <-- swallows type error; caller misreads as MISSING_CONTEXT
```

**Trigger / repro:**

Construct a capability with an `allowedTables` condition and call the tool with an integer or boolean as the named argument:

```go
    ToolName:  "run_query",
    Arguments: map[string]interface{}{"table": float64(42)}, // wrong type
}
```

The denial returned has `Code: "MISSING_CONTEXT"` and message `"argument \"table\" is missing or empty"`, even though the argument is present. An operator reading the audit log would spend time looking for a missing field rather than a type mismatch.

**Suggested fix:**

Add a `default` case to `parseTableArgument` that returns a descriptive error:

```go
default:
    return nil, fmt.Errorf("must be a string, object, or array of strings/objects; got %T", v)
```

The caller already checks `perr != nil` and converts to a `CONDITION_FAILED` denial, so no further changes are needed there. The fix is consistent with `handleAllowedExtensions` (lines 587–595) and `handleRecipientDomain` (lines 868–878).

**Session:** rev-20260614T1304Z-343add

### LINKED ISSUE #404: [rev-20260614T1304Z-343add] schemaValidateValue enum check lacks numericEqual bridge — int enum values never match JSON float64 arguments
**Location:** `pkg/enforcement/schema.go:110-117`

**Reference:** https://github.com/eunolabs/eunox/blob/38c4ee8273a340d955f3cf0fad6bd22d393b9331/pkg/enforcement/schema.go#L110-L117

**Severity:** Medium

**Description:**

`schemaValidateValue` checks the `enum` constraint using only `reflect.DeepEqual`:

```go
if len(schema.Enum) > 0 {
    for _, allowed := range schema.Enum {
        if reflect.DeepEqual(allowed, val) {
            return nil
        }
    }
    return fmt.Errorf("%s: value not in enum", jsonPath)
}
```

`reflect.DeepEqual` is type-sensitive: `int(1) != float64(1.0)`. JSON arguments always arrive as `float64` (Go's `encoding/json` decodes all JSON numbers to `float64` when the target is `interface{}`). If an `ArgumentSchema.Enum` is populated programmatically with `int` values — as happens in tests and in direct engine-embedding scenarios — every JSON request carrying an integer argument will be falsely denied even when the value appears in the allowlist.

`handleAllowedValues` in `handlers.go` faces the identical problem and solves it with `numericEqual`:

```go
// handlers.go:996
    return nil
}
```

The same bridge is absent from `schemaValidateValue`. The comment in `handlers.go` is explicit: *"numericEqual additionally bridges the type gap between manifest numbers (decoded from YAML as int) and request arguments (decoded from JSON as float64)"*. `argumentSchema.enum` suffers the same gap.

**Why production isn't broken today:** The manifest loading path (`LoadManifest`) converts YAML → `interface{}` via `yaml.Unmarshal` (which yields `int`), then immediately marshals to JSON and unmarshals back with `json.Unmarshal` into the Go struct. That JSON round-trip converts every integer to `float64`. So manifests loaded from files are unaffected. Direct API callers (unit tests, embedded use, gateway-test helpers) that construct `ArgumentSchema{Enum: []interface{}{1, 2, 3}}` with bare Go `int` literals do not go through this round-trip and will see silent false denials.

**Trigger / repro:**

```go
schema := &capability.ArgumentSchema{
    Properties: map[string]*capability.ArgumentSchema{
        "count": {Enum: []interface{}{1, 2, 3}}, // int, as written in Go test code
    },
}
err := enforcement.ValidateArgumentSchema(
    map[string]interface{}{"count": float64(1)}, // float64, as decoded from JSON
    schema,
)
// err != nil — "value not in enum" — false denial
```

**Suggested fix:**

Add the `numericEqual` bridge to the enum comparison, exactly as `handleAllowedValues` does:

```go
if len(schema.Enum) > 0 {
    for _, allowed := range schema.Enum {
        if reflect.DeepEqual(allowed, val) || numericEqual(allowed, val) {
            return nil
        }
    }
    return fmt.Errorf("%s: value not in enum", jsonPath)
}
```

`numericEqual` and `toFloat64` already exist in `schema.go`'s sibling file `handlers.go` in the same package; no new code is needed.

A companion test should use programmatic int enum values:

```go
{
    name: "enum int vs float64 (programmatic int, JSON float64)",
    args: map[string]interface{}{"count": float64(2)},
    schema: &capability.ArgumentSchema{
        Properties: map[string]*capability.ArgumentSchema{
            "count": {Enum: []interface{}{int(1), int(2), int(3)}},
        },
    },
    wantErr: false, // must pass after the fix
},
```

**Session:** rev-20260614T1304Z-343add

### LINKED ISSUE #405: [rev-20260614T1304Z-343add] recordSessionCall silently skips recording when req.ToolName is empty but req.Target.Name is set, causing sequenceBlock to fail open
**Location:** `pkg/enforcement/engine.go:340-364`

**Reference:** https://github.com/eunolabs/eunox/blob/38c4ee8273a340d955f3cf0fad6bd22d393b9331/pkg/enforcement/engine.go#L340-L364

**Severity:** Medium

**Description:**

`recordSessionCall` derives the bare tool name exclusively from `req.ToolName` via `splitEnginePrefix`. When `req.ToolName` is empty (or consists only of a bare prefix like `"tool:"`), the function returns `nil` immediately at the early guard:

```go
targetType, tool := splitEnginePrefix(req.ToolName)
if tool == "" {
    return nil   // <-- exits here; req.Target.Name is never consulted
}
// req.Target overrides targetType but never tool:
    targetType = req.Target.Type
}
_, err := e.counter.IncrementAndGet(ctx,
    sequenceHistoryKey(req.SessionID, targetType, tool), ...)
```

`req.Target.Name` is **never used as a fallback** for `tool`. This matters because direct callers of `engine.ValidateAction` or `engine.EvaluateConditions` can legally set `req.Target = &capability.EnforceRequestTarget{Type: "tool", Name: "read_credentials"}` while leaving `req.ToolName` empty or as a bare prefix. In that scenario the allowed call is **never recorded in session history**, and a later `sequenceBlock` condition that names `read_credentials` in `afterTools` will `Peek` a count of 0 — **silently passing (fail open)** even though the antecedent actually ran.

The security invariant from `recordSessionCall`'s doc-comment makes this critical: *"Swallowing the error would leave the session-history marker missing; a later sequenceBlock Peek on that empty key would then conclude the antecedent never ran and fail OPEN."* The same reasoning applies when the marker is never written because `ToolName` was empty.

**Why the production ManifestPDP is unaffected today:** Every ManifestPDP entry point (`Decide`, `DecideResourceRead`, `DecidePromptGet`) sets both `req.ToolName` (the bare name from `ParseTarget`) and `req.Target`. Direct engine consumers — test helpers, embedding code, future transport variants — may not set both fields, and the engine provides no compile-time or runtime warning.

**Trigger / repro:**

```go
e := enforcement.New(enforcement.WithCallCounter(myCounter))

// Antecedent call: ToolName intentionally empty, Target has the name
antecedentReq := &capability.EnforceRequest{
    SessionID: "sess-1",
    ToolName:  "",  // empty — simulates a direct caller using only Target
    Target:    &capability.EnforceRequestTarget{Type: "tool", Name: "read_credentials"},
    Arguments: map[string]interface{}{},
}
// ValidateAction calls recordSessionCall — which returns nil without recording
e.ValidateAction(ctx, antecedentReq, antecedentConstraint)

// Dependent call with sequenceBlock afterTools: ["read_credentials"]
// history.Peek returns 0 — sequenceBlock passes — fail open
```

**Suggested fix:**

After the early `if tool == ""` guard, fall back to `req.Target.Name` before giving up:

```go
targetType, tool := splitEnginePrefix(req.ToolName)
if tool == "" && req.Target != nil && req.Target.Name != "" {
    tool = req.Target.Name
    if req.Target.Type != "" {
        targetType = req.Target.Type
    }
}
if tool == "" {
    return nil
}
// existing Target.Type override for the case where ToolName was non-empty:
    targetType = req.Target.Type
}
```

`handleSequenceBlock` already uses the same `splitEnginePrefix(prior)` to build the Peek key from the `afterTools` manifest entry, so as long as the antecedent's bare name and type match what the manifest lists, the keys will align.

A test should construct a request with empty `ToolName` but non-empty `Target.Name` and confirm the history marker is written and later detected by a `sequenceBlock` Peek.

**Session:** rev-20260614T1304Z-343add

### LINKED ISSUE #406: [rev-20260614T1304Z-343add] matchGlobSegments has unbounded O(n^k) backtracking for allowedValues patterns with k non-adjacent ** segments
**Location:** `pkg/enforcement/glob.go:80-111`

**Reference:** https://github.com/eunolabs/eunox/blob/38c4ee8273a340d955f3cf0fad6bd22d393b9331/pkg/enforcement/glob.go#L80-L111

**Severity:** Medium

**Description:**

`matchGlobSegments` uses recursive backtracking to handle `**` segments. Each `**` group tries the remaining pattern against every suffix of the remaining value segments:

```go
for i := 0; i <= len(val); i++ {
    if matchGlobSegments(rest, val[i:]) {  // <-- recursive call
        return true
    }
}
```

For a pattern with `k` non-adjacent `**` groups matched against a value with `n` `/`-separated segments, the worst-case call tree is O(n^k). Adjacent `**` groups are collapsed (lowering effective `k`), but k=2 is still O(n²), k=3 is O(n³), and so on.

The code comment (`glob.go:72-78`) acknowledges this: *"Each '**' group backtracks over every suffix of the value, so a pattern with k non-adjacent '**' groups against an n-segment value is worst case O(n^k)."* It suggests capping the `**` count at load time but does not implement the cap.

**Why this matters:**

- `allowedValues` pattern strings come from the manifest (operator-controlled), but argument **values** come from the MCP host — attacker-influenced. An operator who writes `values: ["/reports/**/data/**/export"]` (k=2) against an attacker-supplied path with 1000 `/`-separated segments causes O(1 000 000) recursion depth and significant latency. For k=3, that is O(10⁹) — a practical DoS.
- A denial-of-service that only requires a crafted long argument value, no exploit of a protocol vulnerability, and no authentication bypass.
- The suggested mitigation (`"cap the '**' count at load time"`) is not implemented, leaving production deployments unprotected for any manifest that happens to include multi-`**` patterns.

**Trigger / repro:**

```yaml
# manifest
capabilities:
  - target: tool:export
    actions: [call]
    conditions:
      - type: allowedValues
        argument: path
        values: ["/a/**/b/**/c"]   # k=2 non-adjacent **
```

```python
# MCP host sends:
{"tool": "export", "arguments": {"path": "/a/" + "/x" * 1000 + "/c"}}
```

Each enforcement call for this request triggers O(1000²) = O(10⁶) recursive calls in `matchGlobSegments`. At k=3 (three non-adjacent `**` groups in `allowedValues`), a single 500-segment path causes ~O(10⁸) calls.

**Suggested fix:**

Two options, in order of preference:

1. **Reject at load time:** Validate `allowedValues` patterns at manifest load (in `validateLocalManifest` or `CompileSchemaPatterns`) and deny any pattern with more than a configurable cap of `**` segments (e.g., `maxDoubleStar = 2`). Surface a clear error: `"allowedValues pattern %q has %d '**' segments; maximum is %d"`. This matches the comment's own suggested mitigation.

2. **Replace backtracking with DP:** Convert `matchGlobSegments` to a dynamic-programming matcher that runs in O(n × m) time (n = value segments, m = pattern segments) by building a reachability table instead of recursing. This eliminates worst-case blow-up entirely but is a more substantial change.

Option 1 is the lower-risk fix and preserves the existing algorithm for the safe single-`**` case (which is O(n) and fine).

**Session:** rev-20260614T1304Z-343add

eunolabs:
## Bugs addressed (session rev-20260614T1304Z-343add)

- Closes #403 — `handleAllowedTables` reported a present-but-wrong-type argument (number/boolean) as `MISSING_CONTEXT` "argument is missing or empty". Added a `default` branch to `parseTableArgument` returning a type-mismatch error, so the caller now denies with `CONDITION_FAILED` and an accurate message, mirroring `handleAllowedExtensions`/`handleRecipientDomain`. Genuinely empty values (empty string/array) still fall through to `MISSING_CONTEXT`.
- Closes #404 — `schemaValidateValue`'s `enum` check used only `reflect.DeepEqual`, so an enum populated with bare Go ints never matched a JSON `float64` argument (silent false denial for programmatic/embedded callers). Added the `numericEqual` bridge, matching `handleAllowedValues`.
- Closes #405 — `recordSessionCall` derived the bare tool name only from `req.ToolName`; a direct caller setting `req.Target` but leaving `ToolName` empty never recorded the antecedent, so a later `sequenceBlock` Peek failed OPEN. Added a fallback to `req.Target.Name` before the empty-name guard.
- Closes #406 — `matchGlobSegments` used recursive backtracking, O(n^k) for an `allowedValues` pattern with k non-adjacent `**` groups against an attacker-influenced n-segment value (value-driven DoS). Replaced it with a linear O(n*m) bottom-up dynamic program; semantics (including adjacent-`**` collapse) are preserved.

## Testing
- New/extended tests: `TestEngine_AllowedTables_WrongArgumentType` (#403), `TestValidateArgumentSchema_EnumNumericBridge` (#404), `TestSequenceBlock_RecordsAntecedentFromTargetName` (#405), and `TestMatchValueGlob` multi-`**` cases plus `TestMatchValueGlob_NoBacktrackingBlowup` (#406, 4000-segment value returns instantly).

---


eunolabs:
### Self-review (round 1)

Reviewed the full diff as a critical reviewer. One substantive finding:

- **`docs/capability-manifest-guide.md` (allowedTables section, ~line 1085):** #403 changes the denial for a present-but-wrong-type scalar `allowedTables` argument from `MISSING_CONTEXT` to `CONDITION_FAILED`. The guide documented this for sibling conditions (`allowedOperations`, the object/array malformed cases) but not for the top-level scalar case. Per the "behavior changes need a docs update in the same PR" convention, added the missing sentence in commit c1e8b7f.

Verified, no further issues:
- **#404** — `toFloat64` deliberately excludes `bool`, so the `numericEqual` enum bridge cannot make `true`/`1` falsely equal; non-numeric enum entries (strings) still only match via `reflect.DeepEqual`. No false-allow.
- **#403** — the new `default` branch only fires for top-level scalar non-string/object/array values; empty string and empty array still fall through to `nil,nil` and are correctly reported as `MISSING_CONTEXT`. The array branch's inner switch still guards recursive calls.
- **#405** — fallback records under the real tool name that ran (`Target.Name`), keyed by `Target.Type`; aligns with `handleSequenceBlock`'s lookup.
- **#406** — the linear DP reproduces every existing `TestMatchValueGlob` case plus new multi-`**` cases; zero-segment matching subsumes the old adjacent-`**` collapse. `TestMatchValueGlob_NoBacktrackingBlowup` confirms a 4000-segment value returns instantly.

---

eunolabs:
✅ Ready to merge — all session rev-20260614T1304Z-343add bugs addressed (#403, #404, #405, #406), self-review complete with no outstanding comments, tests passing.

---

PR #431, opened 2026-06-14
Fix session rev-20260614T1305Z-9c72ec bugs

### LINKED ISSUE #411: [rev-20260614T1305Z-9c72ec] shouldDeleteEntry strict < boundary keeps boundary-exact entries one Cleanup cycle longer than intended, diverging from Redis TTL semantics
**Location:** `pkg/callcounter/memory.go:477-491` (`shouldDeleteEntry`)

**Reference:** https://github.com/eunolabs/eunox/blob/38c4ee8273a340d955f3cf0fad6bd22d393b9331/pkg/callcounter/memory.go#L477-L491

**Severity:** Low

**Description:**

`shouldDeleteEntry` guards Cleanup's Phase 2 deletion with a strict `Before` comparison:

```go
func shouldDeleteEntry(e *entry, now time.Time, staleAfter time.Duration) bool {
    e.mu.RLock()
    defer e.mu.RUnlock()
    if len(e.timestamps) == 0 {
        return true
    }
    latest := e.timestamps[len(e.timestamps)-1]
    return latest.Before(now.Add(-staleAfter))   // <-- strict <, not <=
}
```

`staleAfter` is `cleanupMarginFactor * windowSec` (i.e., `2 × windowSec`). An entry whose newest timestamp is at exactly `now - 2×windowSec` satisfies `latest == now.Add(-staleAfter)`, so `latest.Before(now.Add(-staleAfter))` is **false** — the entry is retained and not deleted until the next Cleanup sweep.

**Divergence from the Redis backend:**

The Redis backend sets each key's TTL to `2 × windowSec` (via `EXPIRE`) on every `IncrementAndGet` admission:

```go
// redis.go — Expire call in TxPipeline
pipe.Expire(ctx, key, time.Duration(windowSec)*2*time.Second)
```

Redis expires keys at or after the TTL boundary, meaning a key whose last touch was exactly `2 × windowSec` ago is eligible for eviction at that instant (Redis uses `>=` semantics for TTL expiry). `InMemory`'s `shouldDeleteEntry` uses strict `<`, so the same logical key is held one full Cleanup interval longer than the Redis backend would hold it.

**Impact:**

- For most workloads (calls uniformly spread over time) the difference is at most one Cleanup cycle (default: a few seconds to minutes) and affects only the internal storage map, not observable rate-limit decisions. Timestamps at exactly the boundary are already outside the active window (`windowSec`), so no active enforcement decision is affected.
- However, for the `sequenceBlock` history counter (`maxEntries=1`, `windowSec=86400`), this means a session's `sequenceHistoryKey` entry lingers in the InMemory map for one extra Cleanup cycle after it would be evicted by Redis, causing a minor inconsistency between the two backends in memory footprint and cleanup timing.
- In a high-throughput scenario where many keys reach exact boundary simultaneously (e.g. a test-controlled clock), the InMemory backend holds more keys alive than Redis, making backend-switching behavior slightly non-deterministic at the boundary.

**Trigger / repro:**

```go
store := callcounter.NewInMemory()
now := time.Now()
// Simulate an entry with latest timestamp exactly at now - 2*window
// Advance clock so now = recorded_time + 2*window
// Call Cleanup — shouldDeleteEntry returns false; entry survives
// Advance clock by 1ns more — now shouldDeleteEntry returns true
```

With `WithClock` / a controlled clock, this is straightforward to reproduce.

**Suggested fix:**

Change the strict `Before` comparison to include the boundary case (`<=` semantics):

```go
// was: return latest.Before(now.Add(-staleAfter))
return !latest.After(now.Add(-staleAfter))
```

`!t.After(u)` is `t <= u` for `time.Time`. This brings InMemory in line with Redis's `>=`-TTL eviction semantics and eliminates the off-by-one.

A table-driven test should cover the exact-boundary case: an entry whose latest timestamp is `now - staleAfter` exactly should be deleted by `shouldDeleteEntry` after the fix.

**Session:** rev-20260614T1305Z-9c72ec

eunolabs:
## Bugs addressed (session rev-20260614T1305Z-9c72ec)

- Closes #411 — `shouldDeleteEntry` (InMemory call counter Cleanup) guarded deletion with a strict `latest.Before(now-staleAfter)`, so an entry whose newest timestamp was *exactly* `staleAfter` (`2*windowSec`) old was retained one extra Cleanup cycle. The Redis backend sets the key TTL to the same `2*windowSec` and evicts at-or-after the boundary (`>=` semantics), so the two backends diverged at the boundary. Changed to `!latest.After(now-staleAfter)` (i.e. `<=`), aligning InMemory with Redis. No observable enforcement decision changes — timestamps at the boundary are already outside the active `windowSec`.

## Testing
- Extended `TestShouldDeleteEntry` with the exact-boundary case (newest timestamp exactly `2*window` old → now reclaimed) and a one-nanosecond-inside case (still kept), pinning the `<=` semantics.

---


eunolabs:
✅ Ready to merge — session rev-20260614T1305Z-9c72ec bug addressed (#411), self-review complete with no outstanding comments, tests passing.

---

PR #432, opened 2026-06-14
Fix session rev-20260614T1316Z-cce54c bugs

### LINKED ISSUE #407: [rev-20260614T1316Z-cce54c] refreshKeys TTL guard defeats force-refresh on key-miss, breaking key rotation
**Location:** `pkg/capability/jwks.go:128-143` (VerifyToken) and `pkg/capability/jwks.go:228-237` (refreshKeys)

**Reference:** https://github.com/eunolabs/eunox/blob/38c4ee8273a340d955f3cf0fad6bd22d393b9331/pkg/capability/jwks.go#L128-L143

**Severity:** High

**Description:**

`VerifyToken` implements the standard "re-fetch JWKS on key-miss" pattern for supporting key rotation. When the JWT's `kid` is not found in the cached key set, it calls `c.refreshKeys(ctx)` expecting to receive a freshly-fetched JWKS:

```go
// jwks.go:133-143
matchingKeys := c.findKeys(keys, kid)
if len(matchingKeys) == 0 {
    keys, err = c.refreshKeys(ctx)        // intended to force a network fetch
    ...
    matchingKeys = c.findKeys(keys, kid)  // still returns empty — same cached JWKS
```

However, `refreshKeys` has a TTL freshness guard at its very top that **short-circuits the fetch when the cache is still valid**:

```go
// jwks.go:231-237
c.mu.RLock()
if c.jwks != nil && c.now().Sub(c.fetchedAt) < c.cacheTTL {
    keys := c.jwks
    c.mu.RUnlock()
    return keys, nil  // returns same cached JWKS without any HTTP call
}
```

This means that during the cache TTL window (default 5 minutes), the "force-refresh on key-miss" call is a no-op. The identical JWKS is returned both times, and any newly-added key (`kid`) remains invisible to the verifier until the cache expires naturally.

**Evidence from the test suite:**

The existing test `TestJWKSClient_VerifyToken_RefreshFails` (`jwks_test.go:571`) uses `CacheTTL: time.Millisecond` specifically as a workaround, with the comment:

```go
// Short cache TTL so the second attempt re-fetches rather than using cached keys.
client := NewJWKSClient(JWKSClientConfig{
    ...
    CacheTTL: time.Millisecond,
})
```

This comment is a direct acknowledgement that without the tiny TTL the second call would **not** re-fetch — confirming the bug exists.

**Trigger / repro:**

1. Configure a `JWKSClient` with default `CacheTTL` (5 min).
2. Issue a token verification call to populate the JWKS cache.
3. Add a new signing key to the upstream JWKS endpoint.
4. Issue a JWT signed with the new key (new `kid`).
5. Within 5 minutes of step 2, call `VerifyToken` with the new JWT.
6. **Result:** `"no matching key for kid <new-kid>"` — the new key is never fetched.

Expected: The key-miss triggers an unconditional HTTP fetch and the new key is found.

**Suggested fix:**

Add a `forceRefreshKeys` method that skips the TTL double-check and always enters the `sfGroup.Do` path, then call it (instead of `refreshKeys`) from the key-miss branch in `VerifyToken`:

```go
// forceRefreshKeys fetches unconditionally (ignores TTL guard).
// Used when a kid is not found in the cached JWKS to support key rotation.
func (c *JWKSClient) forceRefreshKeys(ctx context.Context) (*jose.JSONWebKeySet, error) {
    v, err, _ := c.sfGroup.Do("refresh", func() (interface{}, error) {
        fetchCtx := context.WithoutCancel(ctx)
        // ... same fetch + store logic as in refreshKeys
    })
    if err != nil {
        return nil, err
    }
    return v.(*jose.JSONWebKeySet), nil
}
```

And in `VerifyToken`:
```go
if len(matchingKeys) == 0 {
    keys, err = c.forceRefreshKeys(ctx)  // was: c.refreshKeys(ctx)
    ...
}
```

Note: using `sfGroup.Do` still deduplicates concurrent force-refreshes, so this does not open a DoS vector.

**Session:** rev-20260614T1316Z-cce54c

### LINKED ISSUE #408: [rev-20260614T1316Z-cce54c] tok.Claims payload-decode errors misreported as "verify signature" failures in VerifyToken
**Location:** `pkg/capability/jwks.go:146-201`

**Reference:** https://github.com/eunolabs/eunox/blob/38c4ee8273a340d955f3cf0fad6bd22d393b9331/pkg/capability/jwks.go#L146-L201

**Severity:** Low

**Description:**

In `VerifyToken`, the loop over matching keys uses a single `lastErr` variable and the label `"verify signature:"` for all failures from `tok.Claims`:

```go
var lastErr error
for i := range matchingKeys {
    var claims jwt.Claims
    var payload TokenPayload

    if err := tok.Claims(&matchingKeys[i], &claims, &payload); err != nil {
        lastErr = err    // could be a JSON unmarshal error, not a signature error
        continue
    }
    ...
}
return nil, fmt.Errorf("verify signature: %w", lastErr)  // misleading category
```

`tok.Claims` in go-jose performs two distinct operations: (1) cryptographic signature verification and (2) JSON unmarshaling of the payload into BOTH `&claims` (`jwt.Claims`) and `&payload` (`TokenPayload`).

**Unmarshaling `TokenPayload` can fail independently of signature validity.** Concretely, `TokenPayload.Audience` is declared as a plain `string`:

```go
type TokenPayload struct {
    ...
    Audience string `json:"aud"`
    ...
}
```

The JWT specification allows `aud` to be either a JSON string or a JSON array of strings (RFC 7519 §4.1.3). `jwt.Claims.Audience` (the go-jose type) handles both via a custom unmarshaler. But `TokenPayload.Audience` (`string`) fails with a JSON type mismatch when `aud` is encoded as an array.

In that case `tok.Claims` returns a JSON unmarshal error, which accumulates in `lastErr` and is ultimately reported as:

```
verify signature: json: cannot unmarshal array into Go struct field TokenPayload.aud of type string
```

This error message attributes a payload-format problem to signature verification, confusing operators and automated alerting that keys on error messages.

**Trigger / repro:**

Send any JWT with `"aud": ["service-a", "service-b"]` (array form) to `VerifyToken`. The JWKS contains the correct signing key with the matching `kid`. The signature is valid. `VerifyToken` returns an error containing `"verify signature:"` even though the signature is fine — only the `aud` encoding is non-standard for `TokenPayload`.

**Suggested fix:**

Two complementary changes:

1. **Distinguish error categories** at the call site. After `tok.Claims` returns an error, inspect whether it is a signature-type error vs. a parsing error (go-jose exposes `jose.ErrCryptoFailure` and similar sentinel values) and wrap with appropriate context:

```go
if err := tok.Claims(&matchingKeys[i], &claims, &payload); err != nil {
    lastErr = fmt.Errorf("key %s: %w", matchingKeys[i].KeyID, err)
    continue
}
```

2. **Use separate unmarshal steps** so that signature verification and payload parsing are decoupled. Verify the signature using `tok.Verify(key)` first (returns only a crypto error), then unmarshal the raw payload into `claims` and `payload` separately — this makes error attribution unambiguous.

**Session:** rev-20260614T1316Z-cce54c

eunolabs:
## Bugs addressed (session rev-20260614T1316Z-cce54c)

- Closes #408 — In `VerifyToken`, `tok.Claims(key, &claims, &payload)` both verifies the signature and unmarshals the eunox-native `TokenPayload`, so a payload-format failure (e.g. an `aud` encoded as a JSON array, which RFC 7519 allows but `TokenPayload.Audience string` rejects) was reported as `verify signature: ...` even though the signature was valid. Split the two steps: verify the signature + standard claims via `tok.Claims(key, &claims)` (whose `aud` unmarshaler accepts string-or-array), then parse the payload separately via `tok.UnsafeClaimsWithoutVerification(&payload)`; a parse failure now reports `parse token payload: ...`. Signature-failure errors are also annotated with the key id.
- Closes #407 — Already resolved on `main`: `VerifyToken` calls `c.cache.ForceRefresh(ctx)` on a kid-miss, and `JWKSCache.ForceRefresh` → `refresh(force=true)` skips the TTL fast-path (the refactor from issue #410, which the older commit the report was filed against predates). Added a regression test (`TestJWKSClient_VerifyToken_KeyRotationWithinTTL`) that rotates in a new signing key while the cache is well within a 1-hour TTL and asserts a token signed by the new key verifies — locking in the force-refresh-on-key-miss behavior.

## Testing
- New tests: `TestJWKSClient_VerifyToken_ArrayAudienceIsPayloadParseError` (#408 — asserts the error contains `parse token payload` and not `verify signature`) and `TestJWKSClient_VerifyToken_KeyRotationWithinTTL` (#407).

---


eunolabs:
### Self-review (round 1)

Reviewed the full diff as a critical reviewer.

Finding (addressed): the comment in `TestJWKSClient_VerifyToken_RefreshFails` ("Short cache TTL so the second attempt re-fetches rather than using cached keys") was stale — the issue #410/#407 force-refresh makes a kid-miss re-fetch *regardless* of TTL, so the tiny TTL is no longer load-bearing. The #407 report even cited that comment as evidence of the bug. Corrected the comment and raised the TTL to 1h (test still passes), in commit 2a88161.

Verified, no further issues:
- **#408** — Happy path is unchanged: a valid string-`aud` token still verifies + parses identically. `UnsafeClaimsWithoutVerification` is safe here because `tok.Claims(key, &claims)` has already verified the signature immediately prior; it only re-parses the already-verified bytes. A payload-parse failure returns immediately (the bytes are key-independent, so retrying other matching keys would reproduce the same error); a wrong-key signature failure still `continue`s to the next matching key and ultimately reports `verify signature: ...` (now annotated with the key id).
- **#407** — confirmed already fixed on `main` via `JWKSCache.ForceRefresh` → `refresh(force=true)` (skips the TTL fast-path); the new `TestJWKSClient_VerifyToken_KeyRotationWithinTTL` locks it in with a 1h TTL and a deterministic (no goroutine/timing) rotation.

---

eunolabs:
✅ Ready to merge — all session rev-20260614T1316Z-cce54c bugs addressed (#407, #408), self-review complete with no outstanding comments, tests passing.

---

PR #433, opened 2026-06-14
Fix session rev-20260614T1710Z-c9e087 bugs

### LINKED ISSUE #424: [rev-20260614T1710Z-c9e087] verifyAuditLog silently classifies corrupted records as skipped (not invalid) when a filter is active
**Location:** `cmd/eunox-mcp/audit.go:985-1076`

**Reference:** https://github.com/eunolabs/eunox/blob/e8e44f1cde57fd84e784d3a9aca31c9b915caf1d/cmd/eunox-mcp/audit.go#L985-L1076

**Severity:** Medium

**Description:**

When `audit-verify` is run with a `--request-id` or `--since` filter, a line that fails `json.Unmarshal` (binary corruption, truncated write, injected non-JSON data) leaves `rec` at its zero value (`RequestID=""`, `Time=""`). The reporting filter then fires:

```go
// audit.go:1052-1060
if requestID != "" && rec.RequestID != requestID {
    res.skipped++
    continue   // ← jumps here for EVERY malformed line when a filter is set
}
if !since.IsZero() {
    t, err := time.Parse(time.RFC3339Nano, rec.Time)
    if err != nil || !t.After(since) {
        res.skipped++
        continue  // ← same for --since filter
    }
}

ok, err := verifier.VerifyRecord(line)  // ← never reached for malformed lines under a filter
```

`VerifyRecord` is never called. The corrupted line is counted as `total+1, skipped+1` but **not** `invalid+1`. As a result, `ok()` (`invalid == 0 && chainBreaks == 0`) returns `true`, producing a false "PASS" verdict even when the log contains corrupted records.

The code comment at lines 993–1000 explicitly claims this is handled:

> "Such a line is still caught below — VerifyRecord fails its HMAC and counts it invalid"

This is true **only when no filter is active**. Under any filter the `continue` fires first.

**Trigger / repro:**

1. Write an audit log with a valid run.
2. Corrupt one line in the middle (flip a byte, truncate it, etc.).
3. Run `eunox-mcp audit-verify --request-id <any-valid-id>`.
4. **Result:** `ok` — the corrupted line is silently skipped.
5. Run `eunox-mcp audit-verify` (no filter): the same line is correctly reported as `INVALID`.

Note: the chain state variables (`havePrev`, `prevHMAC`, `prevSeq`) are correctly NOT updated for a failed unmarshal (inside `if unmarshalOK {...}`), so the chain check itself is not poisoned. Only the per-record HMAC check is skipped.

**Suggested fix:**

Move the `VerifyRecord` call (and its `invalid` / `valid` accounting) to execute before the `requestID`/`since` reporting filter, or alternatively treat an unmarshal failure as an explicit invalid regardless of filter. For example:

```go
if !unmarshalOK {
    res.invalid++
    _, _ = fmt.Fprintf(out, "INVALID  (malformed JSON at byte offset ~%d)\n", res.total)
    continue
}
```

This keeps the reporting filter as a display filter only, matching the stated design: "It does NOT gate the chain verification above."

**Session:** rev-20260614T1710Z-c9e087

### LINKED ISSUE #425: [rev-20260614T1710Z-c9e087] JWKSCache caches an empty key set without error, causing silent JWT verification failure for the full TTL
**Location:** `pkg/capability/jwkscache.go:185-196` (refresh closure) and `pkg/capability/jwkscache.go:218-224` (fetchKeys after unmarshal)

**Reference:** https://github.com/eunolabs/eunox/blob/e8e44f1cde57fd84e784d3a9aca31c9b915caf1d/pkg/capability/jwkscache.go#L185-L196

**Severity:** Medium

**Description:**

`fetchKeys` parses the JWKS response body and returns it unconditionally, even when the parsed set has zero keys:

```go
// jwkscache.go:218-224
var jwks jose.JSONWebKeySet
if err := json.Unmarshal(body, &jwks); err != nil {
    return nil, fmt.Errorf("parse JWKS: %w", err)
}
return &jwks, nil   // ← returned even when len(jwks.Keys) == 0
```

The `refresh` singleflight closure then stores this empty set and logs at INFO level:

```go
c.jwks = jwks
c.fetchedAt = c.now()
// ...
c.logger.Info("refreshed JWKS", slog.Int("keys", len(jwks.Keys)))  // logs "keys=0"
```

From that point until the TTL expires (default 5 minutes), every call to `GetKeys` returns the empty set. `FindKeys` returns an empty slice. Every JWT is rejected — but no error is ever returned from the cache layer and no circuit-breaker failure is recorded (HTTP 200 with `{"keys":[]}` is treated as a success).

**When does an IdP serve `{"keys":[]}`?**

- Brief window during a key rotation when the IdP depublishes the old key before publishing the new one
- Misconfigured or degraded JWKS endpoint
- Network proxy that returns an empty body with status 200

In any of these cases, the empty set is cached for the full TTL (5 min by default), causing a prolonged authentication outage with no alertable signal beyond an INFO log entry.

**Trigger / repro:**

1. Configure `JWKSCache` with `CacheTTL: 5*time.Minute`.
2. Have the JWKS endpoint return `{"keys":[]}` (HTTP 200).
3. Call `GetKeys` — it succeeds, returns an empty JWKS.
4. All subsequent calls to `FindKeys` return nil; all JWT verification fails.
5. Endpoint returns valid keys again — but the cache still serves `{"keys":[]}` for up to 5 more minutes.

Note: the circuit breaker (if configured) is NOT tripped because the HTTP response was successful.

**Suggested fix:**

Return an error from `fetchKeys` when the parsed set is empty, so the caller fails closed and the circuit breaker can track the failure:

```go
if len(jwks.Keys) == 0 {
    return nil, fmt.Errorf("JWKS endpoint returned an empty key set")
}
return &jwks, nil
```

This causes `Refresh` to return an error (and the circuit breaker to record a failure) without caching the empty set. `GetKeys` will serve the previously-cached non-empty set (if any) until the circuit opens, then block further fetches until the upstream recovers — the standard fail-safe pattern the circuit breaker is there for.

**Session:** rev-20260614T1710Z-c9e087

### LINKED ISSUE #426: [rev-20260614T1710Z-c9e087] InMemory.StartCleanup silently exits if passed an already-canceled context, permanently disabling cleanup via sync.Once
**Location:** `pkg/callcounter/memory.go:423-444`

**Reference:** https://github.com/eunolabs/eunox/blob/e8e44f1cde57fd84e784d3a9aca31c9b915caf1d/pkg/callcounter/memory.go#L423-L444

**Severity:** Medium

**Description:**

`StartCleanup` launches a goroutine that exits when `ctx` is canceled. The goroutine is guarded by a `sync.Once` so it is started at most once. When the passed `ctx` is **already canceled at call time**, the goroutine exits on its very first `select` iteration (a closed `Done()` channel is always immediately selectable):

```go
func (m *InMemory) StartCleanup(ctx context.Context, interval time.Duration) bool {
    started := false
    m.cleanupOnce.Do(func() {
        started = true
        // ...
        go func() {
            t := time.NewTicker(interval)
            defer t.Stop()
            for {
                select {
                case <-t.C:
                    m.Cleanup()
                case <-ctx.Done():
                    return   // ← exits immediately when ctx is already canceled
                }
            }
        }()
    })
    return started
}
```

Because `cleanupOnce` has now fired, **no subsequent call to `StartCleanup` can launch a new goroutine**. The counter's entry map grows without bound until the process restarts, re-opening the unbounded-heap-growth sink that `WithMaxKeys` and the cleanup goroutine together are meant to prevent.

The documentation says "once its ctx is canceled it cannot be restarted on this counter — a counter whose cleanup must outlive a canceled context needs a fresh InMemory," but this guidance only addresses cancellation that occurs *after* a successful startup. There is no warning, no error return, and no observable signal when the goroutine dies instantly on the first select. `StartCleanup` returns `true` (indicating "I started cleanup") even though cleanup runs zero times.

**Trigger / repro:**

```go
cancel()  // context is already canceled

counter := callcounter.NewInMemory()
started := counter.StartCleanup(ctx, 5*time.Minute)
// started == true  (misleading — goroutine immediately exited)

// From this point, the entry map grows forever.
// A second call to StartCleanup with a live context is a no-op:
started2 := counter.StartCleanup(context.Background(), 5*time.Minute)
// started2 == false — cleanup is permanently disabled
```

This can happen non-obviously when a parent context is canceled between counter construction and `StartCleanup` (e.g., during a config reload or test teardown).

**Suggested fix:**

Check `ctx.Err()` before firing the `Once` and return `false` (not started) when the context is already done:

```go
func (m *InMemory) StartCleanup(ctx context.Context, interval time.Duration) bool {
    if ctx.Err() != nil {
        return false  // refuse to consume the Once with a dead context
    }
    started := false
    m.cleanupOnce.Do(func() {
        started = true
        // ... launch goroutine as before
    })
    return started
}
```

This keeps the Once available for a subsequent call with a live context, and returns a truthful `false` when cleanup was not actually started. There is a narrow TOCTOU window between the `ctx.Err()` check and the `Do` where cancellation could still race, but that is the same window the current code has once the goroutine is running — it's not meaningfully worse and is far less harmful than permanently consuming the Once.

**Session:** rev-20260614T1710Z-c9e087

eunolabs:
## Bugs addressed (session rev-20260614T1710Z-c9e087)

- Closes #424 — `verifyAuditLog` applied the `--request-id`/`--since` reporting filter before HMAC-verifying each record, so a malformed (non-JSON) line — whose zero-value `rec` matches no filter — was counted `skipped`, not `invalid`, yielding a false `PASS` over a corrupted log whenever a filter was active. Now a line that fails to decode is counted invalid and printed before the reporting filter (the filter stays display-only), so the verdict is filter-independent. Chain-state handling is unchanged (still maintained only across decoded records).
- Closes #425 — `fetchKeys` returned a parsed JWKS even when it had zero keys, so an HTTP 200 `{"keys":[]}` (depublish/republish rotation window, degraded endpoint, empty-body proxy) was cached as a success for the full TTL — every JWT rejected, no error, no circuit-breaker signal. Now an empty key set is returned as a fetch error: the caller fails closed, the breaker records it, and any previously-cached non-empty set keeps serving.
- Closes #426 — `InMemory.StartCleanup` fired its `sync.Once` even when passed an already-canceled context; the goroutine exited on its first `select` while the `Once` stayed consumed, permanently disabling cleanup (unbounded map growth) and still returning a misleading `true`. Now it checks `ctx.Err()` and returns `false` before touching the `Once`, leaving it available for a later call with a live context.

## Testing
- New regression tests: `TestAuditChain_MalformedLineInvalidUnderFilter` (#424), `TestJWKSCache_EmptyKeySetDoesNotClobberCachedKeys` (#425), `TestStartCleanup_CanceledContextDoesNotConsumeOnce` (#426). Updated `TestJWKSClientNoKeysInJWKS` to assert the new fail-at-fetch behavior for an empty key set (#425).

---


eunolabs:
### Self-review (round 1)

Reviewed the full diff as a critical reviewer. Two comment-accuracy findings, both addressed in commit 22020f1:

- **`cmd/eunox-mcp/audit.go` (~line 994):** the pre-existing comment said a malformed line is "still caught below — VerifyRecord fails its HMAC and counts it invalid". After #424 that line is now caught by the explicit `!unmarshalOK` block before the reporting filter, not by `VerifyRecord`. Reworded.
- **`pkg/callcounter/memory.go` (StartCleanup doc):** the function doc only addressed cancellation *after* startup. Added a sentence documenting that an already-canceled context is refused (returns false) without consuming the `Once` (#426).

Verified, no further issues:
- **#424** — fix counts a decode failure invalid before the filter; chain-state maintenance is unchanged (still only across decoded records). Test corrupts the *trailing* record so `chainBreaks==0`, isolating the invalid-count behavior; the no-match filter reproduces the old false PASS.
- **#425** — empty set surfaces as a fetch error so the breaker sees it and the cache isn't clobbered; `fetchKeys` is only reached via the refresh closure, which stores `c.jwks` only on success, so a cached non-empty set survives (covered by the new test). Updated the pre-existing `TestJWKSClientNoKeysInJWKS` whose assertion encoded the old buggy "no matching key" behavior.
- **#426** — `ctx.Err()` guard precedes the `Once`; return values (false → true → false across canceled-then-live calls) deterministically distinguish fixed from broken without relying on goroutine timing.

---

eunolabs:
✅ Ready to merge — all session rev-20260614T1710Z-c9e087 bugs addressed (#424, #425, #426), self-review complete with no outstanding comments, tests passing.

---

PR #434, opened 2026-06-14
Fix session rev-20260614T1810Z-0f21c9 bugs

### LINKED ISSUE #427: [rev-20260614T1810Z-0f21c9] verifyAuditLog reports all pre-signing legacy records as INVALID and generates spurious chain breaks between consecutive Seq=0 records
**Location:** `cmd/eunox-mcp/audit.go:985-1076` (`verifyAuditLog`) and `cmd/eunox-mcp/audit.go:875-908` (`VerifyRecord`)

**Reference:** https://github.com/eunolabs/eunox/blob/e8e44f1cde57fd84e784d3a9aca31c9b915caf1d/cmd/eunox-mcp/audit.go#L985-L1076

**Severity:** High

**Description:**

`verifyAuditLog` has no exemption for records written before HMAC signing was introduced (the "pre-chain" / legacy format). When `audit-verify` is run on a log that spans the pre-signing era, every legacy record produces two distinct failures:

**1. False INVALID (HMAC verification always fails for empty `_hmac`):**

`VerifyRecord` compares `storedHMAC` against a freshly computed MAC:

```go
for _, key := range s.keysToTry(rec.KeyID) {
    mac := hmac.New(sha256.New, key)
    mac.Write(body)
    want := "sha256:" + hex.EncodeToString(mac.Sum(nil))
    if hmac.Equal([]byte(storedHMAC), []byte(want)) {
        return true, nil
    }
}
return false, nil
```

Legacy records have no `_hmac` field; after JSON decode, `storedHMAC = ""`. `hmac.Equal([]byte(""), []byte("sha256:..."))` is always `false`, so `VerifyRecord` returns `(false, nil)` for every legacy record. Each is counted as `res.invalid++` and the result prints `INVALID seq=0 ...`.

**2. Spurious chain breaks between consecutive legacy records (Seq=0 gap):**

Legacy records carry no `seq` field; after JSON decode, `rec.Seq = 0`. The chain checker in `verifyAuditLog` fires for every pair of consecutive legacy records:

```go
// audit.go:1040-1043
if rec.Seq != prevSeq+1 {
    res.chainBreaks++
    _, _ = fmt.Fprintf(out, "SEQ GAP: record seq %d does not follow %d ...\n", rec.Seq, prevSeq)
}
```

For the first legacy record `prevSeq = 0`; for the second `rec.Seq = 0`, so `0 != 0+1 = 1` fires. Each subsequent legacy record triggers another `SEQ GAP` chain break. N consecutive legacy records generate N-1 spurious chain breaks.

**The inconsistency:** `openAuditSink` (line 326) explicitly exempts legacy records from its startup HMAC check:

```go
// A legacy pre-chain record (no _hmac) is exempt: it predates signing and
// is resumed as before.
if prev.HMAC != "" {
        ...
    }
}
```

The same exemption is absent from `verifyAuditLog`, making `audit-verify` unsafe to run on any log that spans the pre-signing-to-post-signing upgrade boundary.

**Trigger / repro:**

1. Start `eunox-mcp` before signing support was added; write N records (no `_hmac`, no `seq`).
2. Upgrade to a signing-capable binary; append M more records with valid HMAC chains.
3. Run `eunox-mcp audit-verify` on the combined log.
4. **Result:** N `INVALID` lines, N-1 `SEQ GAP` chain breaks, `ok() == false` — a false failure verdict on the entire log.
5. The M post-upgrade records are correctly valid; only the legacy records produce false errors.

**Suggested fix:**

In `verifyAuditLog`, detect legacy records (decoded with `HMAC == "" && Seq == 0`) and count them in a new `legacy` category (or existing `skipped`) rather than `invalid`. Skip the chain-seq check for sequences of Seq=0 records:

```go
// After unmarshal in verifyAuditLog:
if unmarshalOK && rec.HMAC == "" && rec.Seq == 0 {
    // Legacy pre-signing record: no HMAC to verify, no seq to chain-check.
    res.skipped++ // or a new res.legacy counter
    continue
}
```

For the chain break, similarly guard the seq-contiguity check:
```go
if rec.Seq != prevSeq+1 {
    // Only report a gap if both records are in the signed era (seq > 0).
    if rec.Seq > 0 && prevSeq > 0 {
        res.chainBreaks++
        ...
    }
}
```

Mirror the `openAuditSink` precedent: legacy records are a known upgrade artifact, not tampered records, and their presence should not poison the verdict for the post-upgrade signed chain.

**Session:** rev-20260614T1810Z-0f21c9

### LINKED ISSUE #428: [rev-20260614T1810Z-0f21c9] Breaker.RecordFailure increments consecutiveFails in StateHalfOpen, causing Stats().ConsecutiveFails to exceed FailureThreshold after a failed probe
**Location:** `pkg/circuitbreaker/breaker.go:212-248` (`RecordFailure`)

**Reference:** https://github.com/eunolabs/eunox/blob/e8e44f1cde57fd84e784d3a9aca31c9b915caf1d/pkg/circuitbreaker/breaker.go#L212-L248

**Severity:** Low

**Description:**

`RecordFailure` has an explicit guard that prevents failures reported while the breaker is in `StateOpen` from advancing the consecutive-failure counter or resetting the cooldown clock:

```go
// audit.go:228-230
if b.state == StateOpen {
    return
}

b.consecutiveFails++      // runs in StateHalfOpen AND StateClosed
b.lastFailureTime = b.now()
```

However, there is no equivalent guard for `StateHalfOpen`. When an admitted half-open probe fails, the code reaches `b.consecutiveFails++` before the state-switch, so the counter is incremented from whatever value it held when the breaker opened.

**Concrete scenario:**

```
1. FailureThreshold = 5
2. Closed: 5 failures → consecutiveFails = 5 → state = StateOpen
3. Cooldown elapses → Allow() → state = StateHalfOpen, halfOpenProbes = 1
4. Probe fails → RecordFailure():
     b.state = StateHalfOpen → guard NOT taken
     b.consecutiveFails++  → consecutiveFails = 6     ← inflated
     case StateHalfOpen: state = StateOpen
5. Stats().ConsecutiveFails = 6  (was 5 when the breaker first opened)
```

Each failed half-open probe adds 1 to `consecutiveFails`. After k failed probes, `Stats().ConsecutiveFails = FailureThreshold + k`, which has no useful meaning: it is neither the count that tripped the breaker nor a measure of the half-open probes.

**Existing test that confirms the intent — but only for Open-state failures:**

`TestBreaker_RecordFailureWhileOpenDoesNotResetCooldown` (line 569) asserts:
```go
if stats.ConsecutiveFails != 2 {
    t.Errorf("expected ConsecutiveFails=2 (stray open-state failures ignored), ...")
}
```

This shows the intent that stray out-of-band failures should not pollute `consecutiveFails`. No equivalent assertion exists for half-open probe failures, leaving the inflation uncaught.

**Impact:**

Behavioral correctness is unaffected: `consecutiveFails` is only used in the `StateClosed` switch arm to decide whether to trip the breaker, and is reset to 0 by the first successful probe (`RecordSuccess`). The bug is confined to observability: `Stats().ConsecutiveFails` reports a misleading value while the breaker is cycling through Open/HalfOpen, making it harder to understand the failure pattern from dashboards or logs.

**Trigger / repro:**

```go
cfg := circuitbreaker.Config{
    FailureThreshold:  3,
    CooldownDuration:  5 * time.Second,
    HalfOpenMaxProbes: 1,
}
b := circuitbreaker.New(cfg, circuitbreaker.WithClock(clock))

for i := 0; i < 3; i++ { b.RecordFailure() }
// consecutiveFails = 3, state = Open

advanceClock(6 * time.Second)
b.Allow()           // Open → HalfOpen, halfOpenProbes = 1
b.RecordFailure()   // HalfOpen → Open, consecutiveFails = 4  ← WRONG

stats := b.Stats()
// stats.ConsecutiveFails = 4, but only 3 failures preceded the open state
```

**Suggested fix:**

Reset `consecutiveFails` when transitioning `StateHalfOpen → StateOpen`, mirroring what `RecordSuccess` does when closing the breaker:

```go
case StateHalfOpen:
    b.state = StateOpen
    b.halfOpenSuccess = 0
    b.consecutiveFails = 0   // ← add this line
    b.lastTransitionAt = b.now()
```

Alternatively, reset it in `Allow()` when transitioning `Open → HalfOpen` (since any probe outcome should start the closed-state counter from zero). Either approach should be accompanied by a test:

```go
// After tripping the breaker and one failed half-open probe,
// ConsecutiveFails must not exceed FailureThreshold.
if got := b.Stats().ConsecutiveFails; got > cfg.FailureThreshold {
    t.Errorf("ConsecutiveFails inflated by half-open probe failure: got %d, want <= %d", got, cfg.FailureThreshold)
}
```

**Session:** rev-20260614T1810Z-0f21c9

### LINKED ISSUE #429: [rev-20260614T1810Z-0f21c9] InMemory and Redis backends disagree on window boundary for calls within the last microsecond of the window (nanosecond vs microsecond precision)
**Location:** `pkg/callcounter/redis.go:99-100, 235-236, 295-296` and `pkg/callcounter/memory.go:173-175, 296-298, 393-394`

**Reference:** https://github.com/eunolabs/eunox/blob/e8e44f1cde57fd84e784d3a9aca31c9b915caf1d/pkg/callcounter/redis.go#L99-L100

**Severity:** Low

**Description:**

The InMemory and Redis backends enforce the sliding window boundary at different precisions, causing them to disagree on whether a call made within the final sub-microsecond of the window is in-window or expired.

**InMemory** stores timestamps as `time.Time` (nanosecond precision) and computes the cutoff at nanosecond precision:

```go
// memory.go:173-175
now := m.now()
window := time.Duration(windowSec) * time.Second
cutoff := now.Add(-window)
// keeps ts if: ts.After(cutoff)  →  ts > now-window  (nanosecond strict)
```

**Redis** stores scores as `UnixMicro()` (microsecond precision) and computes the cutoff at microsecond precision:

```go
// redis.go:99-100
nowUnixMicro := now.UnixMicro()
cutoff := now.Add(-time.Duration(windowSec) * time.Second).UnixMicro()
// removes if: score <= cutoff  →  ts.UnixMicro() <= (now-window).UnixMicro()
```

**The discrepancy:** Consider a call made at `C = now - window + 500ns` (500 nanoseconds inside the window in wall time):

- **InMemory:** `C.After(now - window)` → `true` (500 ns inside) → **COUNTED**
- **Redis:** `C.UnixMicro()` truncates the 500 ns sub-microsecond part, yielding `(now-window).UnixMicro() = cutoff`. `score ≤ cutoff` → **REMOVED** (treated as expired)

The same discrepancy applies to `IncrementIfBelow` (both backends) and `Peek` (Redis `ZCount` uses exclusive `"("` prefix matching InMemory's `ts.After`, so Peek is actually consistent — but `IncrementAndGet` and `IncrementIfBelow` diverge).

**Why this matters:**

For `maxCalls`: a call that InMemory counts as in-window (consumed against the quota) is treated by Redis as expired (not counted). After a backend switch, the effective quota for boundary-straddling calls changes silently.

For `sequenceBlock`: a tool call at exactly `now - 24h + Xns` (X < 1μs) would be detected as in-history by InMemory but missed by Redis, causing `sequenceBlock` to fail open on the Redis backend.

In production this is unlikely to matter (clocks have millisecond-level jitter and windows are in seconds), but it creates subtle non-determinism in unit tests that use controlled clocks set to exact nanosecond boundaries, and makes the backends non-interchangeable at sub-microsecond resolution.

**Trigger / repro:**

```go
// Using fake clocks to place a call exactly at boundary+500ns
now := time.Unix(1_000_000, 0)
windowSec := 60

inMem := callcounter.NewInMemory(callcounter.WithTimeFunc(func() time.Time {
    return now.Add(-time.Duration(windowSec)*time.Second + 500) // 500ns inside window
}))
inMem.IncrementAndGet(ctx, "k", windowSec, 10) // record the call

// Advance clock to exactly `now`
inMem.WithTimeFunc(func() time.Time { return now })
count, _ := inMem.Peek(ctx, "k", windowSec)
// count = 1: InMemory sees the call as in-window

// Redis with same scenario: call at same boundary → score == cutoff → removed
// Redis Peek would return 0 for the same call
```

**Suggested fix:**

Align both backends to microsecond precision, or align Redis to nanosecond precision. The lower-churn fix is to store Redis scores at nanosecond precision (which is within `float64`'s exact-integer range up to ~9.0e15 ns ≈ year 2554, well beyond `MaxWindowSeconds`):

```go
// redis.go: use nanoseconds instead of microseconds for scores and cutoff
nowUnixNano := now.UnixNano()
cutoff := now.Add(-time.Duration(windowSec) * time.Second).UnixNano()
// Update ZADD score: float64(nowUnixNano)
// Update ZRemRangeByScore bound: strconv.FormatInt(cutoff, 10)
// Update Lua script ARGV[2]/ARGV[6] accordingly
```

Alternatively, truncate InMemory timestamps to microseconds at record time to match Redis. Either approach should be covered by a cross-backend parity test using a controlled clock.

**Session:** rev-20260614T1810Z-0f21c9

eunolabs:
## Bugs addressed (session rev-20260614T1810Z-0f21c9)

- Closes #427 — `verifyAuditLog` had no exemption for pre-signing legacy records (no `_hmac`, no `seq` → decode to `HMAC==""`, `Seq==0`), so every legacy record was counted `INVALID` (its empty HMAC can never match a recomputed MAC) and each consecutive pair tripped a spurious `SEQ GAP` (`0` does not follow `0`), producing a false failure verdict across the pre-signing→signing upgrade boundary. Now legacy records are exempted (counted in a new `legacy` bucket, not verified) and the seq-gap check fires only between two signed records (`seq > 0`). `prev_hmac` linkage is unchanged, so a forged legacy record spliced into a signed chain is still caught. Mirrors the exemption `openAuditSink` already grants on resume.
- Closes #428 — A failed half-open probe ran `consecutiveFails++` in `RecordFailure` without resetting, so `Stats().ConsecutiveFails` climbed to `FailureThreshold + k` after `k` failed probes — a meaningless value. Now the `HalfOpen → Open` transition resets `consecutiveFails` to 0, mirroring `RecordSuccess` on close. (Behavior was already correct — the counter only gates the closed-state trip — so this is an observability fix.)
- Closes #429 — InMemory compared the window boundary at nanosecond precision while Redis stores scores as `UnixMicro`, so a call landing in the final sub-microsecond of a window was counted in-window by InMemory but treated as expired by Redis. Aligned InMemory to Redis's microsecond precision by flooring recorded timestamps. Note: the issue's suggested "store Redis scores as `UnixNano`" is not viable — `UnixNano` (~1.75e18) exceeds float64's exact-integer range (2^53≈9e15), so sorted-set scores would round to ~256ns and still not match. Flooring uses subtraction rather than `time.Truncate` so the monotonic clock reading (relied on for jump-robust pruning) is preserved.

## Testing
- New regression tests: `TestAuditChain_LegacyPreSigningRecordsExempt` (#427, real legacy→signed resume via `openAuditSink`), `TestBreaker_HalfOpenProbeFailureDoesNotInflateConsecutiveFails` (#428), and `TestBackendParity_SubMicrosecondWindowBoundary` (#429, drives InMemory and a miniredis-backed Redis with one controlled clock and asserts identical boundary verdicts).

---


eunolabs:
### Self-review (round 1)

Reviewed the full diff as a critical reviewer. No substantive issues found; verifications below.

- **#427** — The legacy exemption requires *both* `HMAC==""` and `Seq==0`, so a tampered signed record (seq>0 with a stripped HMAC) is **not** treated as legacy: it still reaches `VerifyRecord` and is counted INVALID. The seq-gap guard (`rec.Seq>0 && prevSeq>0`) only suppresses gaps that involve a seq-0 record; `prev_hmac` linkage is untouched, so a forged legacy record spliced into the signed chain still trips a CHAIN BREAK (verified by reasoning through the legacy→signed and signed→legacy transitions). `ok()` is unaffected by the new `legacy` count, which is correct — legacy records are a known upgrade artifact, not a failure.
- **#428** — Resetting `consecutiveFails` to 0 on `HalfOpen→Open` is safe: the counter only gates the closed-state trip, and it is already 0 whenever the breaker eventually re-closes. `TotalFailures` still counts every reported failure.
- **#429** — `Peek` is left unfloored deliberately: with stored timestamps now microsecond-aligned, `ts.After(cutoff_nano)` produces the same verdict as Redis's `ZCount("(cutoff_micro")` for every case (proven for `M==cutoff_micro` and `M==cutoff_micro+1µs`), so flooring `Peek` would be a redundant no-op. The cross-backend test confirms parity empirically. The PR notes why the issue's "UnixNano scores" suggestion is not viable (exceeds float64's 2^53 exact-integer range).

Round 2 re-review of the diff produced no new comments. `go test -race ./...` green; `gofmt`/`go vet` clean.

---

eunolabs:
✅ Ready to merge — all session rev-20260614T1810Z-0f21c9 bugs addressed (#427, #428, #429), self-review complete with no outstanding comments, tests passing.

---

PR #437, opened 2026-06-14
Fix session rev-20260614T1910Z-f52fae bugs

### LINKED ISSUE #435: [rev-20260614T1910Z-f52fae] rotate() fallback sets s.logPath to rotated path, causing permanent path drift and broken retention after a failed reopen
**Location:** `cmd/eunox-mcp/audit.go:1101-1113` (`rotate()` fallback path)

**Reference:** https://github.com/eunolabs/eunox/blob/745e7754e3298914b3931fa503221e6ca7af6ba4/cmd/eunox-mcp/audit.go#L1101-L1113

**Severity:** Medium

**Description:**

The fix for issue #362 set `s.logPath = rotated` in the fallback path of `rotate()` (when `os.Rename` succeeds but `os.OpenFile` for the new file fails). This correctly prevents the rotation-loop bug where every subsequent rotation attempt would fail because `s.logPath` no longer existed on disk.

However, the fix introduces a permanent path drift with two secondary consequences that are not addressed:

**1. Retention (`pruneRotated`) becomes blind to pre-fallback rotated files**

`pruneRotated()` uses `filepath.Glob(s.logPath + ".*")` to find sibling rotated files. After `s.logPath` changes from the configured base path (e.g. `audit.jsonl`) to the rotated path (e.g. `audit.jsonl.20260614T191028.123456789Z`), the glob changes to `audit.jsonl.20260614T191028.123456789Z.*`. Any rotated file created before the fallback — e.g. `audit.jsonl.20260614T100000.000000000Z` — no longer matches the pattern and is never pruned. With `retain` configured, the disk quota for the audit directory silently breaks:

```
Before fallback (s.logPath = "audit.jsonl"):
  audit.jsonl.20260614T100000.000000000Z  ← found by glob, eligible for pruning
  audit.jsonl.20260614T110000.000000000Z  ← found by glob, eligible for pruning

After fallback (s.logPath = "audit.jsonl.20260614T191028.123456789Z"):
  audit.jsonl.20260614T100000.000000000Z  ← NOT found, accumulates forever
  audit.jsonl.20260614T110000.000000000Z  ← NOT found, accumulates forever
  audit.jsonl.20260614T191028.123456789Z.20260614T201028.123456789Z  ← found, pruned
```

**2. Chain continuity is broken on the next process restart**

`openAuditSink` always opens the configured path (`logPath` from config). After a fallback rotation, the active log is being written at `audit.jsonl.20260614T191028.123456789Z`, not the configured `audit.jsonl`. On the next restart:

- `openAuditSink` opens `audit.jsonl` (creates it fresh if absent)
- `readLastAuditLine("audit.jsonl")` finds an empty file → `seq = 0`, `prevHMAC = ""`
- The new chain starts at seq=1 with the genesis sentinel, disconnected from the records in the drifted path
- The audit-verify tool cannot reconstruct continuity across this gap

The records written to `audit.jsonl.20260614T191028.123456789Z` are orphaned: they are chained among themselves but are invisible to `readLastAuditLine` on the next startup, creating an undetectable gap in the tamper-evident chain.

**Additionally:** the active log path permanently changes from the configured base name to a timestamped name. Log shippers and syslog configurations that tail the configured path (`audit.jsonl`) stop receiving new records without any error or warning.

**Trigger / repro:**

1. Audit log reaches `maxBytes` and `rotate()` is called.
2. `os.Rename(s.logPath, rotated)` succeeds — e.g. the filesystem is nearly full so creates fail but renames succeed.
3. `os.OpenFile(s.logPath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o600)` fails.
4. Fallback: `s.logPath = rotated`, `s.written = 0`, return.
5. Future rotations succeed (they write new files relative to the rotated path), but pre-fallback rotated files are never pruned and the configured base path is abandoned.
6. On next process restart, chain continuity is silently broken.

**Suggested fix:**

Instead of drifting `s.logPath` permanently, track the active file path separately. One approach: keep `s.logPath` as the configured base path (for consistent glob patterns and restart resume), and introduce a `s.activePath` for the fd currently in use. `pruneRotated` and `readLastAuditLine` continue to operate on `s.logPath`; the fallback updates only `s.activePath`.

Alternatively, after a successful fallback, schedule a warn-and-retry at startup: log the drifted path prominently and instruct operators to run `audit-verify` and manually restore the base path. Either approach must be accompanied by a test that exercises the rename-succeeds/open-fails path and verifies that (a) old rotated files are still pruned and (b) a subsequent restart resumes the chain.

**Session:** rev-20260614T1910Z-f52fae
**Reviewed file:** `cmd/eunox-mcp/audit.go` @ `745e7754e3298914b3931fa503221e6ca7af6ba4`

### LINKED ISSUE #436: [rev-20260614T1910Z-f52fae] RecordSuccess() unconditionally resets consecutiveFails even in StateOpen, corrupting Stats().ConsecutiveFails
**Location:** `pkg/circuitbreaker/breaker.go:176-192` (`RecordSuccess()`)

**Reference:** https://github.com/eunolabs/eunox/blob/745e7754e3298914b3931fa503221e6ca7af6ba4/pkg/circuitbreaker/breaker.go#L176-L192

**Severity:** Low

**Description:**

`RecordSuccess()` unconditionally resets `consecutiveFails` to 0 on line 181, regardless of the breaker's current state:

```go
func (b *Breaker) RecordSuccess() {
    b.mu.Lock()
    defer b.mu.Unlock()

    b.totalSuccesses++
    b.consecutiveFails = 0   // ← runs even in StateOpen

    if b.state == StateHalfOpen {
        b.halfOpenSuccess++
        // ...
    }
}
```

`RecordFailure()` has an explicit guard that prevents out-of-contract calls from affecting state in `StateOpen`:

```go
func (b *Breaker) RecordFailure() {
    // ...
    b.totalFailures++

    if b.state == StateOpen {
        return   // ← guard: stray failures ignored in StateOpen
    }

    b.consecutiveFails++
    b.lastFailureTime = b.now()
    // ...
}
```

No equivalent guard exists in `RecordSuccess()`. A caller that invokes `RecordSuccess()` while the breaker is in `StateOpen` — violating the admit-before-record contract documented on `Breaker` — will:

1. Reset `consecutiveFails` to 0.
2. Leave the breaker in `StateOpen` (the `StateHalfOpen` if-block does not execute).

After this, `Stats().ConsecutiveFails` reports 0 while the breaker is open, making it appear that the breaker opened spontaneously with no preceding failures. This undermines the observability value of the stat and could mislead operators debugging a circuit-open event.

**Asymmetry with RecordFailure:**

The fix for issue #398 added the `StateOpen` guard to `RecordFailure()` precisely because out-of-contract calls must not corrupt the state machine. The same reasoning applies to `RecordSuccess()`, but the guard was not added symmetrically. The doc comment on `Breaker` documents the admit-before-record contract but does not say what happens when it is violated; `RecordFailure()` handles its violation defensively while `RecordSuccess()` does not.

**Behavioral correctness:** unaffected. `consecutiveFails` is only checked in the `StateClosed` switch arm of `RecordFailure()` to decide whether to trip the breaker, and the first `RecordSuccess()` call in `StateHalfOpen` resets it to 0 anyway before the breaker closes. The bug is confined to observability: `Stats().ConsecutiveFails` reports a misleading value during an open-state window where a spurious `RecordSuccess()` was called.

**Trigger / repro:**

```go
cfg := circuitbreaker.Config{
    FailureThreshold:  3,
    CooldownDuration:  30 * time.Second,
    HalfOpenMaxProbes: 1,
}
b := circuitbreaker.New(cfg)

for i := 0; i < 3; i++ {
    b.RecordFailure()
}
// State: Open, Stats().ConsecutiveFails = 3

// Caller violates the admit-before-record contract:
b.RecordSuccess()

fmt.Println(b.State())                   // "open"
fmt.Println(b.Stats().ConsecutiveFails)  // 0  ← wrong; breaker opened due to 3 failures
```

**Suggested fix:**

Add the same guard that `RecordFailure()` uses, incrementing `totalSuccesses` for observability but otherwise leaving the state machine untouched when in `StateOpen`:

```go
func (b *Breaker) RecordSuccess() {
    b.mu.Lock()
    defer b.mu.Unlock()

    b.totalSuccesses++

    // A success reported while the breaker is open did not come from an
    // admitted probe (Allow transitions Open→HalfOpen before returning true).
    // Ignore it to keep consecutiveFails meaningful in Stats.
    if b.state == StateOpen {
        return
    }

    b.consecutiveFails = 0

    if b.state == StateHalfOpen {
        b.halfOpenSuccess++
        if b.halfOpenSuccess >= b.requiredHalfOpenSuccesses() {
            b.state = StateClosed
            b.halfOpenProbes = 0
            b.halfOpenSuccess = 0
            b.lastTransitionAt = b.now()
        }
    }
}
```

The fix should be accompanied by a table-driven test mirroring `TestBreaker_RecordFailureWhileOpenDoesNotResetCooldown` that asserts `Stats().ConsecutiveFails` is not zeroed by a stray `RecordSuccess()` while open.

**Session:** rev-20260614T1910Z-f52fae
**Reviewed file:** `pkg/circuitbreaker/breaker.go` @ `745e7754e3298914b3931fa503221e6ca7af6ba4`

eunolabs:
## Bugs addressed (session rev-20260614T1910Z-f52fae)

- Closes #435 — The #362 rotation fix set `s.logPath = rotated` in the reopen-failure fallback. That stopped the rotation loop but drifted the configured base path permanently: `pruneRotated`'s `filepath.Glob(logPath + ".*")` then matched `<rotated>.*` and went blind to every pre-fallback rotated file (retention silently broke), and the active log abandoned the configured base name that log shippers and restart-resume expect. Introduced a separate `activePath` (the file the current fd writes to): `rotate` renames `activePath` (so the loop fix is preserved), while `logPath` stays the configured base for `rotatedPath`, the prune glob, and resume. The fallback drifts only `activePath`; a recovered rotation restores it to the base path.
- Closes #436 — `RecordSuccess` reset `consecutiveFails` to 0 unconditionally, even in `StateOpen`, so an out-of-contract success reported while the breaker was open made `Stats().ConsecutiveFails` read 0 — as if the breaker had opened with no preceding failures. Added the same open-state guard `RecordFailure` already has (issue #398), symmetric and observability-only (behavior was already correct).

## Testing
- New regression tests: `TestAuditRotate_ReopenFailureKeepsBasePathAndRecovers` (#435 — drives the reopen-failure fallback, asserts `logPath` stays the base while `activePath` drifts, then a recovering rotation returns the active file to the base path and retention prunes a pre-fallback rotated file) and `TestBreaker_RecordSuccessWhileOpenDoesNotResetConsecutiveFails` (#436).

---


eunolabs:
### Self-review (round 1)

Reviewed the full diff as a critical reviewer. One finding, addressed in commit 911c167:

- **`TestAuditRotate_RenameErrorKeepsFd`:** it forced the rename failure by pointing `logPath` at a missing path, but `rotate` now renames `activePath`, so the test still passed only because the rename *target* directory (derived from `logPath`) was missing — not the *source*, which is what its comment describes. Repointed `activePath` (the rename source) at the missing path so the test exercises its stated mechanism.

Verified, no further issues:
- **#435** — `rotate` renames `activePath` (preserving the #362 loop fix) while `logPath` stays the configured base, so `rotatedPath` always yields `<base>.<ts>`, `pruneRotated`'s glob keeps matching every sibling (the core bug), and a recovered rotation restores `activePath` to the base path. The fallback preserves the fd, so records keep flowing; across repeated failed rotations the active inode is renamed forward and only adopted as a rotated file once a reopen succeeds. The new two-phase test exercises fallback → recovery → prune.
- **#436** — the open-state guard is symmetric with `RecordFailure` (#398) and observability-only; `TotalSuccesses` still counts the reported success, and the breaker correctly stays open.

---

eunolabs:
✅ Ready to merge — all session rev-20260614T1910Z-f52fae bugs addressed (#435, #436), self-review complete with no outstanding comments, tests passing.

---

PR #442, opened 2026-06-14
Fix session sx37x1 bugs

### LINKED ISSUE #409: Code review findings: timeout precedence inversion, DecideSampling missing sourceIP, and minor cleanups
# Code Review Findings

Principal-engineer review of the full `eunox-mcp` codebase as of 2026-06-14. Three tiers: **(a) correctness/risk**, **(b) simplification**, **(c) polish**.

---

### A-1 · `resolveUpstreamTimeout`: config silently beats an explicit CLI `0` — Med/S

**File:** `cmd/eunox-mcp/http.go:74–83`

When an operator passes `--upstream-timeout=0` to disable upstream timeouts for a long-running tool, but the gateway config has `defaults.upstreamTimeoutMs: 30000`, the config wins and a 30-second timeout is applied. The test at `operational_hardening_test.go:36` documents this as expected (`{flag: 8000, cfg: 5000, want: 5000}`), but it inverts conventional CLI-over-config precedence and makes `--upstream-timeout=0` ineffective whenever a config is present.

```go
// Current — config beats explicit flag:
func resolveUpstreamTimeout(flagVal, cfgVal int) int {
    v := flagVal
    if cfgVal > 0 {
        v = cfgVal   // overwrites even an explicit 0
    }
    ...
}

// Suggested — explicit flag always wins:
func resolveUpstreamTimeout(flagVal, cfgVal int) int {
    if flagVal != upstreamTimeoutUnset {
        return flagVal
    }
    if cfgVal > 0 {
        return cfgVal
    }
    return defaultUpstreamTimeoutMs
}
```

Update the test case `"config overrides explicit flag"` expectation to `8000`. If the current behaviour is intentional policy (config is authoritative), add an explicit warning when the flag is overridden, and document it in the manifest guide.

---

### A-2 · `DecideSampling` missing `sourceIP` → `ipRange` permanently fails-closed on sampling — Med/S

**File:** `cmd/eunox-mcp/pdp.go:87–89` (interface), `557–607` (ManifestPDP impl)

`SamplingAuthorizer.DecideSampling(ctx, sessionID)` takes no `sourceIP`. `EnforceRequest.Context.SourceIP` is always `""` on the sampling path. `handleIPRange` returns `MISSING_CONTEXT` when `SourceIP == ""`, so any `ipRange` condition on `system:sampling/createMessage` **permanently and silently denies sampling**—no amount of CIDR configuration helps.

The comment at `pdp.go:579–580` explicitly lists `ipRange` as a condition that "MUST be evaluated" on sampling. It is evaluated, but it always denies, which means an operator who adds `ipRange` to a sampling opt-in to allow only certain networks will find sampling unconditionally blocked with no diagnostic other than `MISSING_CONTEXT`.

**Option A** — pass session client IP through:
```go
type SamplingAuthorizer interface {
    DecideSampling(ctx context.Context, sessionID, sourceIP string) capability.EnforceResponse
}
```
`httpSession` knows the client IP at creation; thread it through `decideSampling`.

**Option B** — document the limitation: remove `ipRange` from the comment listing supported conditions on sampling; add a code comment in `DecideSampling` noting `SourceIP` is intentionally absent because sampling has no meaningful "source IP" (upstream is a local subprocess or trusted remote).

Either option is acceptable; the current state (documentation says it works, code silently denies) is wrong.

---

### A-3 · `ManifestPDP.findConstraint` doesn't filter by action; `Engine.findMatchingCapability` does — Low/S

**File:** `cmd/eunox-mcp/pdp.go:347–375` vs `pkg/enforcement/engine.go:583–633`

`findConstraint` selects the best-matching constraint by target specificity + principal but ignores `actions`. `findMatchingCapability` calls `actionMatchScore` inside its loop and skips constraints whose actions don't match.

For a manifest with two equal-specificity entries for the same target but different actions (e.g. one `["read"]` and one `["call"]` on `tool:read_file`), `findConstraint` returns whichever appears first in the YAML, and the separate `containsAction` check then denies with a confusing message even though a correct-action constraint exists.

Unlikely with well-formed manifests, but the divergence is surprising. A comment on `findConstraint` noting this intentional difference is the minimum fix. Longer-term: export `actionMatchScore` from the engine and use it inside `findConstraint`.

---

### B-1 · `stripNamespacePrefix` duplicates `enforcement.StripEnginePrefix` — Low/S

**File:** `cmd/eunox-mcp/pdp.go:422–435`

`stripNamespacePrefix` is byte-for-byte equivalent to the exported `enforcement.StripEnginePrefix`. `matchResource` and `resSpecificity` in the same file already delegate to the engine.

```go
// Before (22 lines):
func stripNamespacePrefix(s string) string {
    if idx := strings.Index(s, ":"); idx >= 0 {
        prefix := s[:idx]
        switch prefix {
        case "tool", "resource", "prompt", "system":
            return s[idx+1:]
        }
    }
    return s
}

// After (3 lines):
func stripNamespacePrefix(s string) string {
    return enforcement.StripEnginePrefix(s)
}
```

Deleting the local copy means a future namespace addition to the engine is automatically reflected here, instead of silently diverging.

---

### B-2 · `Decide` / `DecideResourceRead` / `DecidePromptGet` / `DecideSampling` repeat the same 5-step skeleton — Low/M

**File:** `cmd/eunox-mcp/pdp.go:200–694`

All four methods execute: kill-check → `findConstraint` → nil-deny → `IsAuditOnly` defer → action-check → `ArgumentSchema` guard → build `EnforceRequest` → `EvaluateConditions` → `recordAuditModeAntecedent`. Differences are the target type, required action string, synthetic `Arguments` map, and `ArgumentSchema` applicability.

A `decideCore(...)` helper would shrink ~150 lines to ~30 and ensure all four paths stay in sync (the `ArgumentSchema` guard, for example, is subtly different across the four). However, this is security-critical code where duplication is sometimes preferable to abstraction; weigh carefully and require a full hardening-suite pass before merging.

---

### B-3 · Three filter functions are ~85% structurally identical — Low/S

**File:** `cmd/eunox-mcp/pdp.go:718–835`

`filterToolsListResult`, `filterResourcesListResult`, and `filterPromptsListResult` all unmarshal a typed list, filter entries via `findConstraint` + `containsAction`, and remarshal—failing closed on any error. They differ only in the Go types and required action string. A generic `filterListResult[L, E any]` would remove ~60 lines of near-duplicate logic.

---

### C-1 · `DecideResourceRead` / `DecidePromptGet` hardcode action strings — Low/S

**File:** `cmd/eunox-mcp/pdp.go:473, 635`

`Decide` calls `requiredActionFor(target.Type)` for indirection; the other two hardcode `"read"` and `"get"`. Inconsistent style and a maintenance trap.

```go
// pdp.go:473
- if !containsAction(matched.Actions, "read") {
+ if !containsAction(matched.Actions, requiredActionFor(capability.TargetTypeResource)) {

// pdp.go:635
- if !containsAction(matched.Actions, "get") {
+ if !containsAction(matched.Actions, requiredActionFor(capability.TargetTypePrompt)) {
```

### C-2 · `loadOrCreateAuditKey` (singular) lives in production, only called by tests — Low/S

**File:** `cmd/eunox-mcp/audit.go:1026–1033`

Every call site is in `audit_test.go` and `low_severity_fixes_test.go`. Move to a `_test.go` file or delete and update callers to `keys, _ := loadOrCreateAuditKeys(p); key := keys[0]`.

### C-3 · `resolveUpstreamTimeout` comment understates the precedence inversion — Low/S

**File:** `cmd/eunox-mcp/http.go:69–72`

Even if the config-beats-flag behaviour is intentional, the comment doesn't warn that `--upstream-timeout=0` is a no-op when `defaults.upstreamTimeoutMs` is set. Add an explicit note.

---

## Summary

| ID | Priority | Effort | What |
|----|----------|--------|------|
| A-1 | Med | S | `resolveUpstreamTimeout`: config overrides explicit `--upstream-timeout=0` |
| A-2 | Med | S | `DecideSampling` missing `sourceIP`; `ipRange` on sampling always denies |
| A-3 | Low | S | `findConstraint` vs engine action-filter divergence; add comment |
| B-1 | Low | S | `stripNamespacePrefix` duplicates `enforcement.StripEnginePrefix` |
| B-2 | Low | M | Four-way PDP method skeleton; consider `decideCore` helper |
| B-3 | Low | S | Three filter functions; consider generics unification |
| C-1 | Low | S | Hardcoded `"read"`/`"get"` instead of `requiredActionFor` |
| C-2 | Low | S | `loadOrCreateAuditKey` in prod code, only used in tests |
| C-3 | Low | S | Timeout comment doesn't warn about precedence inversion |

eunolabs:
## Bugs addressed (issue #409, session sx37x1)

This PR works through the principal-engineer code-review findings filed in #409. Each finding is its own commit referencing #409.

- **A-1 / C-3 — `--upstream-timeout` precedence inversion.** `resolveUpstreamTimeout` let any positive `defaults.upstreamTimeoutMs` override the flag, so `--upstream-timeout=0` (disable the deadline for a long-running tool) was a silent no-op whenever a config was present. Now follows conventional CLI-over-config precedence: an explicit (non-negative) flag wins — `0` disables, a positive value sets it — and a negative (unset) flag defers to config, then the built-in 30 s default. Flag help text, the `TestResolveUpstreamTimeout` table, and the `CHANGELOG` migration note are updated. (C-3, the under-stated comment, is resolved by the same change.)
- **A-2 — `DecideSampling` missing `sourceIP`; `ipRange` on sampling always denied.** `SamplingAuthorizer.DecideSampling` took no source IP, so the sampling-path `EnforceRequest` always had an empty `SourceIP` and any `ipRange` condition on a `system:sampling/createMessage` opt-in failed closed with `MISSING_CONTEXT` regardless of CIDRs — even though the comment/docs claimed it was enforced. Chose **Option A**: capture the originating client IP on the HTTP session at `initialize` and thread it through `decideSampling` → `DecideSampling` (ManifestPDP + JWTPDP) into the request context. stdio has no network client so it passes `""` (ipRange still fails closed there, as documented). Adds an allow/deny source-IP test and documents the sampling semantics in the manifest guide.
- **A-3 — `findConstraint` does not filter by action.** Documented the intentional divergence from `enforcement.findMatchingCapability` and the one edge it affects (two equal-specificity same-target entries differing only by action), per the issue's prescribed minimum fix.
- **B-1 — `stripNamespacePrefix` duplicated `enforcement.StripEnginePrefix`.** Replaced the byte-for-byte copy with delegation so the recognized namespace set stays a single source of truth.
- **C-1 — hardcoded `"read"`/`"get"`.** `DecideResourceRead` and `DecidePromptGet` now route through `requiredActionFor(target.Type)` like `Decide` does.
- **C-2 — `loadOrCreateAuditKey` in production, used only by tests.** Moved to `audit_test.go` so it ships only with the test binary.

### Findings intentionally not changed
- **B-2 (four-way PDP skeleton → `decideCore` helper).** Declined. The issue itself flags this as security-critical code where "duplication is sometimes preferable to abstraction; weigh carefully." Collapsing the four decision paths into one helper risks subtly changing the deliberately-divergent `argumentSchema`/fail-closed guards for marginal line savings — not worth the regression risk in the single enforcement point.
- **B-3 (three filter functions → generics).** Already done in the current tree: `filterToolsListResult`/`filterResourcesListResult`/`filterPromptsListResult` already delegate to a generic `filterListResult[E, L any]`. No change needed.

## Testing
- `make check-fmt`, `make check-license` — clean.
- `go vet ./...` and `golangci-lint run ./...` — 0 issues.
- New tests: `TestManifestPDP_DecideSampling_IPRange_SourceIPThreaded` (in/out-of-range source IP) and updated `TestResolveUpstreamTimeout` cases (explicit flag, including `0`, now beats config).

Closes #409

---
_Generated by [Claude Code](https://claude.ai/code/session_019k7dUDhFrg92tK2kUGCyNf)_


eunolabs:
✅ Ready to merge — all session sx37x1 bugs addressed (#409: findings A-1, A-2, A-3, B-1, C-1, C-2; C-3 folded into A-1; B-2 declined with rationale; B-3 already in tree). Self-review complete with no outstanding comments — it caught and fixed a real data race in the A-2 client-IP threading and a JWTPDP test gap. CI is fully green (44/44 checks) and tests pass.

---

PR #457, opened 2026-06-14
Fix session 9ut5v9 bugs

### LINKED ISSUE #439: [rev-20260614T2002Z-a02822] circuitbreaker: two b.now() calls in RecordFailure() cause lastFailureTime and lastTransitionAt to diverge
**Location:** `pkg/circuitbreaker/breaker.go:243-264`
**Reference:** https://github.com/eunolabs/eunox/blob/9150f24e4dc49ccf726d1e0d156da1e6f3d83f65/pkg/circuitbreaker/breaker.go#L243-L264
**Severity:** Medium

## Description

`RecordFailure()` makes **two separate `b.now()` calls** when recording the state transition to `StateOpen`:

```go
// Line 244
b.lastFailureTime = b.now()   // first call → time T_a

switch b.state {
case StateClosed:
    if b.consecutiveFails >= b.config.FailureThreshold {
        b.state = StateOpen
        b.lastTransitionAt = b.now()  // second call → time T_b ≥ T_a
    }
case StateHalfOpen:
    b.state = StateOpen
    b.halfOpenSuccess = 0
    b.consecutiveFails = 0
    b.lastTransitionAt = b.now()  // another separate call → time T_c ≥ T_a
}
```

This produces two distinct timestamps for a single logical event (the trip). The cooldown timer in `Allow()` and `State()` is anchored to `lastFailureTime` (T_a):

```go
if b.now().Sub(b.lastFailureTime) >= b.config.CooldownDuration {
```

But `Stats().LastTransitionAt` reports T_b (or T_c), which is fractionally later. These two values should be identical for the trip event.

## Trigger / Repro

**Production:** The discrepancy is nanoseconds on a real clock, making it practically invisible. The cooldown starts from T_a but `Stats().LastTransitionAt` shows T_b — a minor but misleading inconsistency in operational dashboards and metrics that compute "how long has the breaker been open".

**Tests with `WithClock`:** A custom clock that returns a new (advancing) value on each call will produce `lastFailureTime ≠ lastTransitionAt` for the same trip event. Any test that:
1. Uses `WithClock` with a step-advancing fake
2. Calls `RecordFailure()` to trip the breaker
3. Asserts `Stats().LastTransitionAt == Stats().LastFailureTime`

…will fail unexpectedly. Tests that advance the clock by exactly `CooldownDuration` from `lastFailureTime` to trigger the half-open transition will also behave differently than tests anchored to `lastTransitionAt`.

The same divergence affects the `StateHalfOpen → StateOpen` re-open path (lines ~255–264).

## Suggested Fix

Capture `b.now()` once and use the same value for both fields:

```go
b.consecutiveFails++
now := b.now()
b.lastFailureTime = now

switch b.state {
case StateClosed:
    if b.consecutiveFails >= b.config.FailureThreshold {
        b.state = StateOpen
        b.lastTransitionAt = now  // same instant as lastFailureTime
    }
case StateHalfOpen:
    b.state = StateOpen
    b.halfOpenSuccess = 0
    b.consecutiveFails = 0
    b.lastTransitionAt = now  // same instant as lastFailureTime
}
```

This ensures the cooldown anchor and the reported transition time are always the same value, and that tests with a step-advancing `WithClock` see consistent behavior.

**Session:** rev-20260614T2002Z-a02822

### LINKED ISSUE #440: [rev-20260614T2002Z-a02822] pdp_jwt: JWTPDP.Decide does redundant kill-switch round-trip in intersection mode while DecideSampling is optimized
**Location:** `cmd/eunox-mcp/pdp_jwt.go:413-491` and `cmd/eunox-mcp/pdp.go:202-204`
**Reference:** https://github.com/eunolabs/eunox/blob/9150f24e4dc49ccf726d1e0d156da1e6f3d83f65/cmd/eunox-mcp/pdp_jwt.go#L413-L491
**Severity:** Medium

## Description

`JWTPDP.Decide` (and the paths it implements — tool calls, resource reads, prompt gets) performs the kill-switch check **twice** in intersection mode (when both `--jwks-uri` and `--policy` are set):

1. **First check** — in `JWTPDP.Decide` (line 414):
   ```go
       return *deny
   }
   ```

2. **Second check** — delegated through `decideInner` → `ManifestPDP.Decide` (pdp.go line 202):
   ```go
       return *deny
   }
   ```

Both checks consult the **same kill-switch store** (`p.ks == mdp.ks` in the typical wiring). When the kill store is Redis, this doubles the Redis read operations per request on the enforcement hot path.

`DecideSampling` already has this optimization applied:
```go
// pdp_jwt.go:374-378
if !p.innerSharesKillSwitch() {
        return *deny
    }
}
```

The same guard is entirely absent from `Decide`, `DecideResourceRead`, and `DecidePromptGet`.

## Trigger / Repro

Any deployment running in intersection mode (`--jwks-uri` + `--policy`) with a Redis kill-switch backend. Under moderate traffic (e.g., 1 000 rps), this doubles kill-store reads from 1K to 2K rps with no security benefit (both checks consult identical state and return identical outcomes).

The discrepancy is observable by comparing Redis command logs or metrics between `sampling/createMessage` requests (one read) and `tools/call` requests (two reads) on the same route.

## Suggested Fix

Apply the same `innerSharesKillSwitch()` guard to `Decide`:

```go
func (p *JWTPDP) Decide(ctx context.Context, sessionID string, target EnforceTarget, args map[string]interface{}, sourceIP string) capability.EnforceResponse {
    // Skip the wrapper's kill check when the inner ManifestPDP performs the
    // identical check against the same manager — mirrors DecideSampling.
    if !p.innerSharesKillSwitch() {
            return *deny
        }
    }
    ...
```

`DecideResourceRead` and `DecidePromptGet` both delegate to `Decide`, so fixing `Decide` covers all three paths.

Note: the security outcome is unchanged — the inner ManifestPDP always performs the kill check when it shares the manager; the outer check is purely redundant overhead in that case.

**Session:** rev-20260614T2002Z-a02822

### LINKED ISSUE #441: [rev-20260614T2002Z-a02822] callcounter: InMemory.Peek uses non-floored cutoff, inconsistent with recorded timestamps and Redis backend
**Location:** `pkg/callcounter/memory.go:407`
**Reference:** https://github.com/eunolabs/eunox/blob/9150f24e4dc49ccf726d1e0d156da1e6f3d83f65/pkg/callcounter/memory.go#L407
**Severity:** Low

## Description

`InMemory.Peek` computes its window cutoff from a **raw, unfloored `m.now()`**:

```go
// memory.go:407
cutoff := m.now().Add(-time.Duration(windowSec) * time.Second)
```

By contrast, `IncrementAndGet` and `IncrementIfBelow` both floor the current time to microsecond precision before recording a timestamp **and** before computing the prune cutoff:

```go
// memory.go:173-175
now := floorToMicro(m.now())
window := time.Duration(windowSec) * time.Second
cutoff := now.Add(-window)
```

The `floorToMicro` comment explains why this matters:
> "Redis stores sorted-set scores as UnixMicro and so cannot distinguish sub-microsecond timing, which made the two backends disagree on whether a call landing in the final sub-microsecond of a window was in-window or expired."

Because `Peek` omits the floor, it has a different view of the window boundary than `IncrementAndGet`:

- A timestamp is stored as `floor(T_record)` — at most 999 ns earlier than `T_record`.
- `Peek`'s cutoff is `T_peek − W` (unrounded), where `T_peek ≥ T_record`.
- At the boundary (`T_record ≈ T_peek − W + ε` for tiny ε), `floor(T_record)` may fall exactly **at or before** the unfloored cutoff, causing `Peek` to report 0 while `IncrementAndGet` would have counted it.

The `sequenceBlock` condition relies on `Peek` to answer "has tool A already run in this session?" A sub-microsecond discrepancy at the window boundary could cause a sequenceBlock check to spuriously report that the antecedent tool has *not* run, even though `IncrementAndGet` just recorded it — a fail-open gate on the sequenceBlock condition.

Additionally, the unfloored cutoff makes `InMemory.Peek` semantically different from the Redis backend's Peek-equivalent, which uses `UnixMicro` as its cutoff. The two-backends-must-agree invariant (established for `IncrementAndGet` by issue #429) is violated for `Peek`.

## Trigger / Repro

This requires a call that arrives in the final sub-microsecond of a window. In practice this is extremely unlikely on production clocks, but:

1. A test using `WithTimeFunc` with a clock that returns sub-microsecond fractional values can reproduce it deterministically.
2. A Redis-vs-InMemory consistency test that calls `Peek` immediately after `IncrementAndGet` could observe a count-by-1 discrepancy at the window boundary.

The sequenceBlock fail-open path would manifest as: tool B is allowed to run even though tool A ran (because `Peek` returned 0 for A's session key just after A was admitted).

## Suggested Fix

Apply `floorToMicro` to the `Peek` cutoff, matching `IncrementAndGet`:

```go
func (m *InMemory) Peek(_ context.Context, key string, windowSec int) (int64, error) {
        return 0, err
    }

    m.mu.RLock()
    defer m.mu.RUnlock()

    e, ok := m.entries[storageKey(key, windowSec)]
    if !ok {
        return 0, nil
    }

    // Floor to microsecond precision to match recorded timestamps and the Redis
    // backend, which stores scores as UnixMicro (same fix as IncrementAndGet, #429).
    cutoff := floorToMicro(m.now()).Add(-time.Duration(windowSec) * time.Second)
    var count int64
        if ts.After(cutoff) {
            count++
        }
    }
    return count, nil
}
```

A regression test should verify that `Peek` and `IncrementAndGet` agree on the count immediately after a call is recorded at the microsecond boundary.

**Session:** rev-20260614T2002Z-a02822

### LINKED ISSUE #444: [rev-20260614T2004Z-6ecc46] collectObligations silently fails closed for value-typed RedactFieldsDirective
## Bug report — automated review session `rev-20260614T2004Z-6ecc46`

**Commit reviewed:** `9150f24e4dc49ccf726d1e0d156da1e6f3d83f65`

### Location

`pkg/enforcement/engine.go`, `collectObligations()` (~lines 774–795)

### Description

`collectObligations` type-switches on directive values using only the pointer branch:

```go
for _, dir := range matched.Directives {
    switch d := dir.(type) {
    case *capability.RedactFieldsDirective:   // pointer only
        obligations = append(obligations, capability.Obligation{
            Type:  capability.DirectiveTypeRedactFields,
            Paths: d.Fields,
        })
    default:
        return nil, &capability.EnforceResponse{
            Code:     "ENFORCEMENT_ERROR",
            ...
        }
    }
}
```

If a `Constraint` is constructed programmatically (e.g. in tests or external callers) with a value-typed directive—`capability.RedactFieldsDirective{Fields: [...]}` instead of `&capability.RedactFieldsDirective{Fields: [...]}`—the directive hits `default` and the engine returns `ENFORCEMENT_ERROR`, denying the request and suppressing any obligations that should have been collected.

### Asymmetry with the rest of the codebase

- `cmd/eunox-mcp/manifest.go` `validateLocalManifest()` (lines 233–244) handles **both** `*capability.RedactFieldsDirective` and `capability.RedactFieldsDirective` in its switch.
- `pkg/enforcement/engine.go` `asCondition[T]()` (the generic helper for condition handling) handles both `*T` and `T` value types.
- `collectObligations` is the only site that handles pointer only, creating an undocumented invariant that every `Directive` stored in a `Constraint` must be a pointer.

Normal manifest loading via `pkg/capability/directives.go` `newDirective()` always returns a pointer, so the happy path works. The failure mode surfaces when:
1. Tests build `Constraint` literals inline with `capability.RedactFieldsDirective{}`.
2. External code or future directive types are added and the author forgets to take the address.

### Impact

A valid, allow-listed request with a `redactFields` directive is denied with `ENFORCEMENT_ERROR` rather than allowed with the fields redacted. Audit logs record a spurious policy-engine failure. Fail-closed is the correct posture for true unknowns, but this is a false unknown: the directive is structurally identical to the pointer version.

### Suggested fix

Add the value branch, mirroring `validateLocalManifest`:

```go
case capability.RedactFieldsDirective:
    obligations = append(obligations, capability.Obligation{
        Type:  capability.DirectiveTypeRedactFields,
        Paths: d.Fields,
    })
```

Or centralise with a helper analogous to `asCondition[T]` that accepts both `D` and `*D`.

### Severity

Medium — only triggered by programmatically constructed constraints with value-typed directives; manifest-loaded paths are unaffected today. However, the silent ENFORCEMENT_ERROR deny is hard to diagnose and will bite the first author who adds a new directive type and forgets the `&`.

### LINKED ISSUE #445: [rev-20260614T2004Z-6ecc46] Audit log file grows without bound when rotate() rename fails repeatedly
## Bug report — automated review session `rev-20260614T2004Z-6ecc46`

**Commit reviewed:** `9150f24e4dc49ccf726d1e0d156da1e6f3d83f65`

### Location

`cmd/eunox-mcp/audit.go`, `rotate()` (~lines 1128–1182, specifically the rename-failure path around line 1146)

### Description

When `os.Rename` fails during log rotation, the code resets `s.written = 0` to avoid an immediate tight-loop retry:

```go
if err := os.Rename(s.activePath, rotated); err != nil {
    fmt.Fprintf(os.Stderr,
        "[eunox-mcp] audit rotate error (rename): %v; continuing with existing file\n", err)
    s.written = 0  // reset counter so we do not loop back into rotate immediately
    return
}
```

Resetting to **zero** means the file must accumulate another full `maxBytes` worth of records before `rotate()` is called again. Under persistent rename failures (e.g. a permission error on the audit directory, a cross-device rename, or a file-system full condition affecting only rename syscalls), this loop repeats indefinitely:

```
written reaches maxBytes → rename fails → written = 0
→ written reaches maxBytes again → rename fails → written = 0
→ ...
```

There is no cap on how large the active log file can grow. The only limit is free disk space.

### Impact

1. **Disk exhaustion**: on a busy proxy, audit records accumulate unboundedly in the active file if rotation is permanently broken.
2. **Audit availability**: once the disk is full, the drainer goroutine will fail to write new records, effectively disabling audit logging.
3. **HMAC chain integrity**: if disk is full and the process restarts, `readLastAuditLine()` may fail to read the tail of an oversize file, breaking the chain linkage.
4. **Silent degradation**: the error is only printed to stderr; there is no metric, no structured audit event, and no configurable error threshold at which the proxy should refuse to forward requests (fail-closed) rather than continue with a broken audit log.

### Suggested fix

Instead of resetting `s.written = 0`, set it to `s.maxBytes - s.maxBytes/10` (or any value just below the threshold). This still prevents an immediate tight-loop retry while bounding growth to ~10% of `maxBytes` per failed attempt:

```go
// Back off by one burst (10%) so we retry rotation soon but don't loop.
s.written = s.maxBytes - s.maxBytes/10
```

A complementary hardening would be a configurable `maxRotateErrors` counter: after N consecutive rotation failures, stop forwarding requests (fail-closed) or write a rate-limited structured error audit record.

### Severity

Medium — requires a persistent filesystem fault for rename to fail repeatedly; under normal operation rotation succeeds on the first attempt. However, the failure mode (silent disk exhaustion) is operationally dangerous and the fix is low-risk.

### LINKED ISSUE #446: [rev-20260614T2004Z-6ecc46] verifyAuditLog skips seq-gap check when prevSeq is zero, allowing genesis-record substitution
## Bug report — automated review session `rev-20260614T2004Z-6ecc46`

**Commit reviewed:** `9150f24e4dc49ccf726d1e0d156da1e6f3d83f65`

### Location

`cmd/eunox-mcp/audit.go`, `verifyAuditLog()` — the sequence-gap detection block

### Description

The seq-gap check reads (approximately):

```go
if rec.Seq != prevSeq+1 && rec.Seq > 0 && prevSeq > 0 {
    // report gap
}
prevSeq = rec.Seq
```

The guard `prevSeq > 0` means: **when the first record examined has `Seq == 0`, the gap check is skipped for the transition from that record to the next one.**

The legitimate genesis record has `Seq == 1`. The verifier correctly enforces this via a separate check (`if rec.Seq == 1 { ... genesis ... }`). But both checks are bypassable together:

1. An attacker who possesses the HMAC key (or can replay an older signed record) appends a **forged record with `Seq == 0`** at the beginning of a log segment.
2. Because `Seq == 0` is not `== 1`, the genesis check does not fire.
3. Because `prevSeq == 0` after processing this record, the next record's gap check (`prevSeq > 0`) is false — so a following record with any `Seq` value passes without a gap error.
4. The true seq-1 genesis record can be deleted; the seq-0 decoy suppresses the check that would have caught the deletion.

The result is that an attacker can strip the true genesis record from a log and replace it with a signed seq-0 stub without triggering a verification error, defeating one layer of the tamper-evidence guarantee.

### Precondition

Exploitation requires possession of a valid HMAC signing key or a previously signed seq-0 record (which should not exist in a correctly operated log). This is a defence-in-depth issue, not an immediately exploitable vulnerability against a correctly deployed system. However:

- The signing key is stored on disk alongside the log (`audit_keys.json`); compromise of the log directory also compromises the key.
- A seq-0 record signed with the correct key could be sourced from a prior session's chain linkage edge case.

### Suggested fix

Change the guard to reject `Seq == 0` explicitly and enforce that the first record in any verified segment has `Seq == 1` (genesis) or continues an explicitly provided chain anchor:

```go
if rec.Seq == 0 {
    errs = append(errs, fmt.Sprintf("record %d: Seq 0 is invalid", lineNum))
    continue
}
if prevSeq > 0 && rec.Seq != prevSeq+1 {
    errs = append(errs, fmt.Sprintf("seq gap: expected %d, got %d", prevSeq+1, rec.Seq))
}
```

Rejecting `Seq == 0` as malformed closes the substitution vector regardless of key possession.

### Severity

Low-Medium — requires HMAC key compromise (so the log is already partially compromised) and deliberate manipulation. The fix is a one-line guard change with no behaviour impact on correctly formed logs.

### LINKED ISSUE #447: [rev-20260614T2011Z-4a7b07] token_cache: Stop() violates "blocks until goroutine exits" contract for concurrent callers
**Location:** `pkg/capability/token_cache.go:132-142`
**Reference:** https://github.com/eunolabs/eunox/blob/9150f24e4dc49ccf726d1e0d156da1e6f3d83f65/pkg/capability/token_cache.go#L132-L142
**Severity:** Low

## Description

`TokenCache.Stop()` documents that it "blocks until [the cleanup goroutine] has exited, so a caller may safely tear down state the goroutine touches once Stop returns." That guarantee is violated for any caller that is not the first one to acquire the write lock.

```go
func (c *TokenCache) Stop() {
    c.mu.Lock()
    if c.stopped {
        c.mu.Unlock()
        return           // ← returns immediately, without wg.Wait()
    }
    c.stopped = true
    c.mu.Unlock()
    close(c.stopCh)
    c.wg.Wait()          // ← only the first caller ever reaches this
}
```

Execution sequence when two goroutines call `Stop()` concurrently:

1. **G1** acquires the lock, sets `stopped = true`, releases lock, calls `close(c.stopCh)`, then blocks on `c.wg.Wait()`.
2. **G2** acquires the lock, sees `stopped == true`, releases the lock, and **returns immediately** — before the cleanup goroutine has observed the shutdown signal or called `c.mu.Unlock()` after its final `purgeExpired()` pass.

G2's caller now believes it is safe to tear down state (maps, the context, etc.) that the still-running cleanup goroutine may concurrently hold under `c.mu.Lock()`. The cleanup goroutine itself holds a pointer to `c.entries` and `c.insertOrder` — a teardown that frees or nilifies those while the goroutine is in `purgeExpired()` would be a data race.

## Trigger / Repro

```go
var wg sync.WaitGroup
for range 2 {
    wg.Add(1)
    go func() {
        defer wg.Done()
        cache.Stop()
        // Both goroutines assert here that the cleanup goroutine has exited.
        // The second one may not have waited.
    }()
}
wg.Wait()
```

Run with `-race` — the data race detector may fire if a teardown action follows Stop() closely.

## Suggested Fix

Move `c.wg.Wait()` outside the early-return so every caller blocks until the goroutine exits, regardless of whether it was the one to close `stopCh`:

```go
func (c *TokenCache) Stop() {
    c.mu.Lock()
    if !c.stopped {
        c.stopped = true
        close(c.stopCh)
    }
    c.mu.Unlock()
    c.wg.Wait() // all concurrent callers block until the goroutine exits
}
```

`c.wg.Wait()` on a counter of 0 (if `Start()` was never called) returns immediately, so the no-Start path is unaffected. `c.wg.Wait()` is safe to call from multiple goroutines simultaneously — it blocks all of them until the counter reaches zero.

**Session:** rev-20260614T2011Z-4a7b07

### LINKED ISSUE #448: [rev-20260614T2011Z-4a7b07] audit: chain silently broken when process restarts after rotation-fallback state
**Location:** `cmd/eunox-mcp/audit.go:325` (chain resume) and `cmd/eunox-mcp/audit.go:1146-1164` (rotation fallback)
**Reference:** https://github.com/eunolabs/eunox/blob/9150f24e4dc49ccf726d1e0d156da1e6f3d83f65/cmd/eunox-mcp/audit.go#L325
**Severity:** Medium

## Description

When `rotate()` successfully renames the active log to a timestamped rotated path but then **fails to reopen `logPath`** (e.g., disk full at the `OpenFile` call), the sink falls back to continuing writes on the renamed file:

```go
// audit.go:1157-1163
fmt.Fprintf(os.Stderr, "[eunox-mcp] audit rotate error (reopen)...")
s.activePath = rotated   // ← now writing to the renamed file
s.written = 0
return
```

At this point `logPath` **does not exist on disk** (it was renamed away). If the process then restarts before a subsequent successful rotation can recreate it, `openAuditSink` does:

```go
// audit.go:325
if last := readLastAuditLine(logPath); last != "" {
```

`logPath` is absent (or newly created as an empty file by the `OpenFile` call a few lines above). `readLastAuditLine` returns `""`, so the chain is resumed with:

```go
s.seq      = 0
s.prevHMAC = ""          // → first new record gets PrevHMAC = auditGenesisPrev
```

The new log's first record carries `"prev_hmac":"sha256:genesis"` at `seq=1`, but the actual chain continuity was in the rotated fallback file (at a higher seq, with a real predecessor HMAC). There is **no visible chain break** between the two files and no warning on restart — `audit-verify` run against the new `logPath` alone will pass, and an operator has no signal that coverage was interrupted.

### Why this matters

The design goal of the tamper-evident chain is that deletion or reordering of records is detectable. The rotation-fallback + restart scenario silently resets the chain, making the period between the fallback and the restart undetectable as a gap. An adversary who can cause a `disk-full` during rotation (e.g., by filling a shared filesystem) can then trigger a process restart (SIGKILL, OOM, deploy) to silently orphan recent audit records.

## Trigger / Repro

1. Run `eunox-mcp` until the audit log approaches `rotate-size-bytes`.
2. Fill the filesystem so that after the rename, `os.OpenFile(logPath, ...)` fails.
3. eunox-mcp continues running (stderr shows the reopen error, writes go to the renamed file).
4. Kill and restart the process.
5. `audit-verify` against the new `logPath` reports all-valid with no chain breaks. The records written to the rotated fallback file during step 3 are not linked from the new log.

## Suggested Fix

On startup, when `readLastAuditLine(logPath)` returns empty AND the log was just created (size == 0), scan the directory for the most-recently-modified rotated sibling (`logPath.<timestamp>`) and attempt chain resumption from its tail. If found, resume from that file's seq/prevHMAC; log a warning so operators know the chain was resumed from a fallback file rather than the base path.

Alternatively, at the end of `rotate()` (both on success and on the reopen-fallback path), write the current `seq` and `prevHMAC` to a small sidecar file (e.g., `logPath.chain-state`) that `openAuditSink` reads preferentially. This is simpler than directory scanning and handles the restart-from-fallback case without ambiguity.

**Session:** rev-20260614T2011Z-4a7b07

### LINKED ISSUE #449: [rev-20260614T2011Z-4a7b07] audit: readLastAuditLine discards ReadAt byte count, corrupts last-line extraction on short read
**Location:** `cmd/eunox-mcp/audit.go:380-381`
**Reference:** https://github.com/eunolabs/eunox/blob/9150f24e4dc49ccf726d1e0d156da1e6f3d83f65/cmd/eunox-mcp/audit.go#L380-L381
**Severity:** Low

## Description

`readLastAuditLine` allocates a buffer exactly matching the expected number of bytes and then discards the `n` return value from `ReadAt`:

```go
buf := make([]byte, size-start)
if _, err := f.ReadAt(buf, start); err != nil && err != io.EOF {
    return ""
}
```

`f.ReadAt` may return `(n < len(buf), io.EOF)` when the file is shorter than `size` (e.g., the file was truncated between the `f.Stat()` call above and the `ReadAt`). In that case:

- `err == io.EOF` → the early-return condition `err != nil && err != io.EOF` is **false**, so execution continues.
- `buf[n:]` contains **zero bytes** (`\x00`), not the expected newlines or JSON.

The subsequent extraction logic does not trim null bytes:

```go
trimmed := bytes.TrimRight(buf, "\n")       // trims '\n' only; '\x00' is left
if i := bytes.LastIndexByte(trimmed, '\n'); i >= 0 {
    return string(bytes.TrimSpace(trimmed[i+1:]))  // may include '\x00' suffix
}
return string(bytes.TrimSpace(trimmed))
```

`bytes.TrimSpace` trims only Unicode space characters (space, tab, `\n`, `\r`, `\f`, `\v`); null bytes survive. If the valid content before position `n` does not end with `\n`, the returned "last line" includes a null-byte suffix. `json.Unmarshal` then fails on the corrupt string, causing `openAuditSink` to start a fresh chain (seq=0, genesis sentinel) instead of resuming from the previous record — a spurious chain break on restart.

### Sequence that triggers this

1. The audit file is, say, 1 000 000 bytes.
2. `f.Stat()` returns `size = 1 000 000`.
3. Another process truncates the file to 999 990 bytes (or the kernel flushes a partial write from a crash).
4. `ReadAt(buf[1 000 000-4194304 to end], start)` returns `(n=..., io.EOF)` where `n < len(buf)`.
5. `buf[n:]` is zero-filled, corrupting the last-line extraction.

In normal operation an audit log only grows (append-only), so this path requires unusual circumstances. However, it can also be triggered by a race between `Stat` and `ReadAt` on a slow NFS mount where the file's cached size lags.

## Suggested Fix

Slice `buf` to the actual number of bytes read before parsing:

```go
n, err := f.ReadAt(buf, start)
if err != nil && err != io.EOF {
    return ""
}
buf = buf[:n]   // discard any zero-filled tail from a short read
```

This ensures the extraction always operates on exactly the bytes the kernel returned, regardless of the gap between the Stat size and the actual file size at ReadAt time.

**Session:** rev-20260614T2011Z-4a7b07

### LINKED ISSUE #450: [rev-20260614T2010Z-45fd6b] audit: partial-write byte count not added to s.written, delaying log rotation
**Location:** `cmd/eunox-mcp/audit.go:811-832`
**Reference:** https://github.com/eunolabs/eunox/blob/d6588404977c810ce30e85e3a8b5d6fe6832a0a5/cmd/eunox-mcp/audit.go#L811-L832
**Severity:** Low-Medium

## Description

`writeRecord` uses a two-result write call and only updates `s.written` on the non-error path:

```go
// audit.go:808-832
n, err = s.f.Write(line)   // or s.writeLine(line)
if err != nil {
    s.writeFailures.Add(1)
    if !s.writeErrWarned { ... }
    return  // BUG: n bytes may have landed on disk, but s.written is not updated
}
s.written += int64(n)      // only reached when err == nil
```

The `io.Writer` contract allows a partial write: `(n > 0, err != nil)` is a valid outcome, and on Linux the kernel `write()` syscall can return a positive count alongside `ENOSPC` when the disk fills mid-write. When this occurs:

1. `n` bytes are physically written to the log file on disk.
2. The early-return skips `s.written += int64(n)`.
3. `s.written` is now permanently underestimated by `n` bytes.

The rotation threshold is checked as:

```go
    s.rotate()
}
```

Because `s.written` is underestimated, the next rotation is delayed by up to `n` bytes per failed write. Under repeated partial-write failures (e.g., filesystem filling slowly), the active log file can grow significantly beyond `maxBytes` before the next rotation is triggered.

## Why the existing comment does not cover this

The comment correctly notes that the chain head is intentionally not advanced on write failure — chain consistency is preserved. But `s.written` is a size accounting field for rotation scheduling, not a chain field. The partial bytes on disk count toward the file's actual size, so they should also advance `s.written` even when the chain head is not advanced.

## Trigger / Repro

1. Set `rotate-size-bytes` to a small value (e.g., 1 MiB).
2. Simulate a disk that fills slowly under write pressure (e.g., via a FUSE mount or `ulimit -f`).
3. The `os.File.Write` call returns `(n=4096, err=syscall.ENOSPC)` (partial flush before the syscall fails).
4. `s.written` is not updated; the file grows without triggering rotation at `maxBytes`.

## Suggested Fix

Update `s.written` unconditionally from `n`, then return on error:

```go
n, err = s.f.Write(line)
s.written += int64(n)   // count bytes that actually landed on disk, even on error
if err != nil {
    s.writeFailures.Add(1)
    if !s.writeErrWarned { ... }
    return
}
// chain head advanced below only when err == nil
s.seq = rec.Seq
s.prevHMAC = rec.HMAC
```

This keeps chain integrity (head not advanced on error) while keeping the rotation counter accurate.

**Session:** rev-20260614T2010Z-45fd6b

### LINKED ISSUE #451: [rev-20260614T2010Z-45fd6b] pdp_jwt: AllowedValues condition always denies non-string arguments, inconsistent with manifest engine
**Location:** `cmd/eunox-mcp/pdp_jwt.go:898-914` (`evaluateJWTConditions`)
**Reference:** https://github.com/eunolabs/eunox/blob/d6588404977c810ce30e85e3a8b5d6fe6832a0a5/cmd/eunox-mcp/pdp_jwt.go#L898-L914
**Severity:** Medium

## Description

`evaluateJWTConditions` handles `AllowedValuesCondition` with a strict string-only check:

```go
// pdp_jwt.go:898-914
case capability.AllowedValuesCondition:
    rawVal, argExists := args[c.Argument]
    if !argExists {
        resp := denyResponse(capability.ErrCodeMissingContext, ...)
        return &resp
    }
    val, isStr := rawVal.(string)
    if !isStr {
        resp := denyResponse(capability.ErrCodeMissingContext, capability.ConditionTypeAllowedValues,
            fmt.Sprintf("%q: argument %q is not a string", name, c.Argument))
        return &resp  // non-string arguments always denied here
    }
    if !matchesAllowedValues(val, c.Values) { ... deny ... }
```

The manifest engine's `handleAllowedValues` (`pkg/enforcement/handlers.go:987-996`) explicitly supports non-string argument types — booleans, numbers, and nil — via `reflect.DeepEqual` and `numericEqual`:

```go
// handlers.go:989-996
// Exact match for non-string allowed values (bool, number, nil), which
// carry no glob semantics. numericEqual additionally bridges the type gap
// between manifest numbers (decoded from YAML as int) and request
// arguments (decoded from JSON as float64) ...
    return nil
}
```

## Concrete broken scenario

An operator configures a JWT capability claim for a tool with a boolean argument:

```json
{
  "mcp.capabilities": ["tool:my_tool?enabled=true"]
}
```

`buildV2Constraint` parses this into an `AllowedValuesCondition{Argument: "enabled", Values: [true]}` (line 637-640, where `cp.value` is an `interface{}` holding `true`).

When the tool is called with `{"enabled": true}`, the JWT evaluator hits the `!isStr` branch and denies with `MISSING_CONTEXT "argument \"enabled\" is not a string"` — even though:
- The argument is present.
- Its value exactly matches the constraint.
- The manifest engine would allow it.

The JWTPDP is supposed to intersect with the manifest, restricting what the manifest permits. Instead, for boolean/numeric arguments it is more restrictive than any manifest policy could ever be (it always denies, making the tool permanently uncallable regardless of manifest).

## Additional concern: misleading error code

The error code used is `ErrCodeMissingContext` with message "is not a string". This is factually wrong: the argument is present; its type is the issue. The code suggests the argument is absent, which will mislead operators debugging why a call was denied.

## Suggested Fix

Mirror the manifest engine: after the string-type-assertion fails, attempt an exact match against the allowed values using `reflect.DeepEqual` (and optionally the same `numericEqual` bridge):

```go
val, isStr := rawVal.(string)
if !isStr {
    // Non-string argument: fall back to exact-value matching, mirroring
    // handleAllowedValues in the manifest engine (handlers.go:987-996).
    for _, allowed := range c.Values {
        if reflect.DeepEqual(allowed, rawVal) || numericEqual(allowed, rawVal) {
            goto nextCondition
        }
    }
    resp := denyResponse(capability.ErrCodeValueNotPermitted, capability.ConditionTypeAllowedValues,
        fmt.Sprintf("%q: argument %q value not in permitted set", name, c.Argument))
    return &resp
}
// String path: glob matching as before.
if !matchesAllowedValues(val, c.Values) { ... }
nextCondition:
```

Alternatively, extract a shared helper that both `evaluateJWTConditions` and `handleAllowedValues` call, so the two evaluators cannot diverge again.

**Session:** rev-20260614T2010Z-45fd6b

### LINKED ISSUE #452: [rev-20260614T2037Z-f0b3a9] verifyAuditLog: legacy-record exemption enables silent injection of fabricated audit entries
**Location:** `cmd/eunox-mcp/audit.go`, lines 1081–1092 (`verifyAuditLog`)

**Reference:** `eunolabs/eunox` @ `d658840`

**Severity:** HIGH

**Description:**

`verifyAuditLog` classifies any record where `HMAC == ""` **and** `Seq == 0` as a "legacy" record and increments `res.legacy` instead of `res.invalid`. Crucially, it then `continue`s without updating `lastHMAC` or `lastSeq`, so the chain-continuity check is also skipped for these records. `res.ok()` returns `true` when `res.invalid == 0 && res.chainBreaks == 0`; it never inspects `res.legacy`. This means `audit-verify` exits 0 even when legacy-classified records are present.

The problematic block (simplified):
```go
// lines 1081-1092
    res.legacy++
    continue   // skips HMAC check AND chain-state update
}
```

An attacker who can append bytes to the audit log file (e.g. via a compromised sidecar, log-shipping misconfiguration, or direct filesystem access short of the process) can inject records of the form:
```json
{"decision":"allow","target":"tool:dangerous_tool","hmac":"","seq":0}
```
These records survive `audit-verify --log <file>` undetected (`ok=true`, exit 0) and appear in the human-readable output only as a "(legacy)" count that operators typically ignore.

**Trigger / repro:**

1. Run the proxy, producing a normal signed log (`audit.jsonl`).
2. Append a fabricated line: `{"timestamp":"...","decision":"allow","target":"tool:exec","hmac":"","seq":0}`.
3. Run `eunox-mcp audit-verify --log audit.jsonl`.
4. Observe exit code 0 and output `"... 1 legacy ..."` — the injected record is not flagged.

**Suggested fix:**

The legacy exemption was added to handle pre-signing records produced by an older version of the proxy. It should be time-bounded or removed entirely. At minimum:

- Any record with `Seq == 0` and `HMAC == ""` that appears **after** a signed record (i.e. `lastSeq > 0`) should be classified as `invalid`, not `legacy`.
- If the exemption must be retained for genuine forward-compat reasons, `res.legacy` should cause `res.ok()` to return `false` (or at least emit a non-zero exit code with a prominent warning) so operators know the log contains unverifiable records.
- Document in `docs/threat-model-mcp.md` that legacy records are unverifiable and attacker-injectable.

**Session:** rev-20260614T2037Z-f0b3a9

### LINKED ISSUE #453: [rev-20260614T2037Z-f0b3a9] matchGlobSegments: attacker-controlled slash count causes unbounded DP table allocation (DoS)
**Location:** `pkg/enforcement/glob.go`, lines 83–107 (`matchGlobSegments`)

**Reference:** `eunolabs/eunox` @ `d658840`

**Severity:** MEDIUM

**Description:**

`MatchValueGlob` is called by the `allowedValues` condition handler to match an argument value against manifest-defined glob patterns. When a pattern contains `**`, `MatchValueGlob` delegates to `matchGlobSegments`, which allocates a `(m+1) × (n+1)` boolean DP table where:

- `m = len(strings.Split(pattern, "/"))` — pattern segment count, controlled by the manifest author (trusted)
- `n = len(strings.Split(value, "/"))` — **value segment count, controlled by the caller / attacker**

```go
func matchGlobSegments(pat, val []string) bool {
    m, n := len(pat), len(val)
    dp := make([][]bool, m+1)
    for i := range dp {
        dp[i] = make([]bool, n+1)  // n slices of length n+1
    }
    ...
}
```

A tool caller who can supply a string argument (e.g. a file path, URI, or resource identifier) to a tool that has an `allowedValues` condition with any `**`-containing pattern can craft a value with a large number of `/` characters. For example, with a pattern like `/reports/**` (`m = 2`) and a value like `strings.Repeat("/x", 500_000)` (`n = 500_000`), the allocation is `3 × 500_001` booleans — roughly 1.5 MB per call. With many concurrent requests, this can exhaust available memory.

Note that the whole-pattern wildcards `*` and `**` are short-circuited before reaching `matchGlobSegments`, so only patterns with embedded `**` (e.g. `/reports/**`) are vulnerable; `values: ["*"]` is safe.

**Trigger / repro:**

1. Define a manifest with an `allowedValues` condition containing a `**` pattern, e.g.:
   ```yaml
   conditions:
     - type: allowedValues
       argument: path
       values: ["/data/**"]
   ```
2. Send a `tools/call` request with `arguments.path` set to a string of 100 000 slash-separated segments.
3. Each enforcement call allocates `O(m × n)` memory; under concurrent load this causes OOM or severe GC pressure.

**Suggested fix:**

Add a segment-count cap before entering the DP loop. A reasonable limit (e.g. 1 000 segments) is far above any legitimate file path or URI depth and stops the attack with a single bounds check:

```go
const maxGlobSegments = 1000

func matchGlobSegments(pat, val []string) bool {
    if len(val) > maxGlobSegments {
        return false  // fail-closed: deny oversized values
    }
    ...
}
```

Alternatively, cap inside `MatchValueGlob` before calling `matchGlobSegments`, which keeps the guard in one place. The cap should be documented as a hard limit in `docs/capability-manifest-guide.md`.

**Session:** rev-20260614T2037Z-f0b3a9

### LINKED ISSUE #454: [rev-20260614T2037Z-f0b3a9] Constraint.MarshalJSON/UnmarshalJSON silently drop Principal field, enabling authorization bypass on JSON round-trip
**Location:** `pkg/capability/constraint.go`, lines 239–313 (`constraintJSON`, `MarshalJSON`, `UnmarshalJSON`)

**Reference:** `eunolabs/eunox` @ `d658840`

**Severity:** HIGH

**Description:**

`Constraint` uses a private `constraintJSON` intermediary struct for custom JSON marshaling. However, `constraintJSON` is missing the `Principal` field that exists on `Constraint`:

```go
// Constraint (the real struct)
type Constraint struct {
    Target     string              `json:"target"`
    Actions    []string            `json:"actions"`
    // ... other fields ...
    Principal  map[string][]string `json:"principal,omitempty"`  // <-- scopes the constraint to specific JWT identities
    Conditions []Condition         `json:"conditions,omitempty"`
    // ...
}

// constraintJSON (the intermediary — Principal is absent)
type constraintJSON struct {
    Target     string      `json:"target"`
    Actions    []string    `json:"actions"`
    // ... other fields ...
    // Principal is NOT here
    Conditions []conditionEnvelope `json:"conditions,omitempty"`
    // ...
}
```

`MarshalJSON` converts `Constraint` → `constraintJSON` and calls `json.Marshal`; `Principal` is never copied in.
`UnmarshalJSON` decodes into `constraintJSON` and copies fields back to `*Constraint`; `Principal` is never restored.

**Impact:**

Any code path that serializes a `Constraint` to JSON and reads it back will silently lose the principal restriction. The constraint then matches **all** callers, not only the intended JWT sub/iss/agent_id/task_id identities. Known affected paths:

1. `ValidateActionResponse.MatchedCapability *Constraint` — the matched constraint is returned to callers over JSON (e.g. `pdp.go` callers that log or forward it). If the response is cached and re-parsed, the principal scope is gone.
2. Any future manifest-caching layer, admin API, or config-reload path that marshals `Constraint` values to JSON (a natural extension) will silently strip principal restrictions, allowing privilege escalation without any manifest change.
3. Tests or tooling that round-trip constraints through JSON (e.g. `validate --live` diff output) will produce misleading results.

**Trigger / repro:**

```go
c := capability.Constraint{
    Target:    "tool:read_secrets",
    Actions:   []string{"call"},
    Principal: map[string][]string{"sub": {"alice"}},
}
data, _ := json.Marshal(c)
var c2 capability.Constraint
_ = json.Unmarshal(data, &c2)
// c2.Principal is nil — any sub can now match this constraint
fmt.Println(c2.Principal) // map[]
```

**Suggested fix:**

Add `Principal` to `constraintJSON` and propagate it in both directions:

```go
type constraintJSON struct {
    // ... existing fields ...
    Principal map[string][]string `json:"principal,omitempty"`
}

func (c Constraint) MarshalJSON() ([]byte, error) {
    return json.Marshal(constraintJSON{
        // ... existing fields ...
        Principal: c.Principal,
    })
}

func (c *Constraint) UnmarshalJSON(data []byte) error {
    var aux constraintJSON
    if err := json.Unmarshal(data, &aux); err != nil {
        return err
    }
    // ... existing field copies ...
    c.Principal = aux.Principal
    return nil
}
```

Add a round-trip test in `pkg/capability/` that verifies a principal-scoped `Constraint` survives `json.Marshal` → `json.Unmarshal` with the `Principal` field intact and that `PrincipalMatches` still returns the correct result after the round-trip.

**Session:** rev-20260614T2037Z-f0b3a9

eunolabs:
## Bugs addressed (session 9ut5v9)

This PR works through the open automated-review findings across five review sessions, one commit per bug.

### Security / audit integrity
- Closes #454 — `Constraint` JSON round-trip silently dropped the `Principal` field (constraint widened to all callers). Added `Principal` to `constraintJSON` and propagate it in `MarshalJSON`/`UnmarshalJSON`; round-trip test asserts `PrincipalMatches` still scopes.
- Closes #453 — `matchGlobSegments` allocated an O(m·n) DP table where n is the caller-controlled value segment count (DoS). Added a 1000-segment cap that fails closed for oversized values.
- Closes #452 — `verifyAuditLog` exempted any HMAC-less `seq 0` record as "legacy" and `ok()` ignored the legacy count, so a forged record appended to a signed log passed `audit-verify` (exit 0). The exemption is now valid only at the head of a log (before any signed record); a post-signing HMAC-less `seq 0` record is counted invalid.
- Closes #446 — A record with `seq 0` but a non-empty HMAC is structurally impossible (legacy records have no HMAC; signed records start at seq 1). It is now rejected as invalid, closing a forged-`seq-0`-decoy substitution of the genesis record even against a key holder.

### Audit durability / accounting
- Closes #450 — `writeRecord` counted bytes toward `s.written` only on the non-error path, so a partial write (`n>0, err!=nil`, e.g. ENOSPC) underestimated size and delayed rotation. Bytes that landed on disk now count unconditionally; the chain head still advances only on full success.
- Closes #449 — `readLastAuditLine` discarded `ReadAt`'s byte count, so a short read (file shrank between `Stat` and `ReadAt`) left a zero-filled tail that corrupted last-line extraction and spuriously reset the chain. Slice the buffer to `n`.
- Closes #448 — After a rotation reopen-fallback the base log can be renamed away; on restart the chain reset to genesis and silently orphaned records. When the base log is empty on startup, resume from the newest rotated sibling's tail.
- Closes #445 — On a rename failure `rotate()` reset `s.written` to 0, so the file grew a full `maxBytes` before retrying — unbounded under a persistent rename fault. Back off by one ~10% burst instead.

### Correctness / consistency
- Closes #451 — JWT PDP `AllowedValues` denied any non-string argument with the misleading `MISSING_CONTEXT` "is not a string". Mirror the manifest engine: exact-match non-string args (numeric int/float64 bridge), deny unmatched with `VALUE_NOT_PERMITTED`.
- Closes #447 — `TokenCache.Stop()` returned immediately for non-first concurrent callers, violating its "blocks until the goroutine exits" contract. `wg.Wait()` now runs for every caller.
- Closes #444 — `collectObligations` handled only `*RedactFieldsDirective`, denying a value-typed directive with `ENFORCEMENT_ERROR`. Added the value branch, mirroring `validateLocalManifest`.
- Closes #441 — `InMemory.Peek` used a non-floored cutoff, unlike `IncrementAndGet`/Redis. Floored to microseconds so the backends/paths can't disagree on the in-window boundary.
- Closes #440 — `JWTPDP.Decide` did a redundant kill-switch round-trip in intersection mode. Guarded with `innerSharesKillSwitch()` (mirroring `DecideSampling`); inner check stays authoritative so the security outcome is unchanged.
- Closes #439 — `RecordFailure()` called `b.now()` twice for one trip, diverging `lastFailureTime` and `lastTransitionAt`. Capture the instant once.

## Testing
- `make check-fmt`, `make check-license`, `make check-notice` — clean.
- `make lint` (`go vet` + `golangci-lint`) — 0 issues.
- New focused tests per fix (JSON round-trip, glob segment cap, concurrent `Stop`, value directive, Peek/Increment agreement, trip-timestamp equality, non-string AllowedValues, kill-switch single-check, partial-write accounting, short-read extraction, rotated-sibling resume, forged seq-0 / post-signing legacy rejection); updated the rename-failure and JWT non-string tests to the new behavior.

---
_Generated by [Claude Code](https://claude.ai/code/session_015ot9oDy76B8VHqqvoQT1d7)_


eunolabs:
✅ Ready to merge — all session 9ut5v9 bugs addressed (#454, #453, #452, #451, #450, #449, #448, #447, #446, #445, #444, #441, #440, #439), self-review complete with no outstanding comments, tests passing.

CI green: Go CI (build/lint/vet/race tests, coverage 84.2%, govulncheck, Trivy) and Demo Integration Tests both pass. One transient flake in the `pkg/killswitch` Redis pub/sub test (`connection refused` to miniredis) failed on the first attempt and passed on re-run; it is unrelated to this diff (no `pkg/killswitch` changes) and passes locally under `-race`.

---

PR #466, opened 2026-06-14
Release-readiness hardening: JWKS/upstream-error/IO fixes, supply-chain CI gating, provenance, fuzz, docs

eunolabs:
## What this is

Follow-up to the release-readiness audit. Addresses the actionable findings in one PR: engine/proxy security hardening, supply-chain CI gating + build provenance, fuzz targets, and doc corrections. No release-blocking defects were found in the audit; these are the Medium/Low hardening items.

All changes verified locally: `go test -race ./...` (9 pkgs, 0 fail, 0 races), `golangci-lint run` (0 issues), `go vet`, gofmt, `make check-license`/`check-notice`/`check-stale-syntax`, and YAML well-formedness of every workflow.

## Findings addressed

**Engine / proxy (`sec:`)**
- **M4 — JWKS unknown-`kid` amplification.** A stream of distinct unknown-`kid` tokens drove one upstream JWKS refetch each. Added a short-TTL (30s), size-bounded negative cache (`ForceRefreshForKID`) shared by both verifiers; suppressed lookups still fail closed. Tests cover suppression, TTL expiry, and known/empty-`kid` passthrough.
- **L1 — upstream-error info disclosure.** `upstreamErrInfo` now returns a host-safe reason and logs the full error (which for a remote upstream can carry its internal hostname/path or raw response body) only to operator stderr at a single choke point. Regression guard added.
- **L2 — unbounded upstream reads.** `io.LimitReader` on the remote-upstream JSON-RPC decode (4 MiB, matching the subprocess per-message cap), the error-body snippet, and the kill-command response.

**CI/CD (`ci:`)**
- **M1** — Trivy fs scan now has a blocking pass on fixable CRITICAL/HIGH (`ignore-unfixed`); govulncheck pinned to `v1.3.0` (was `@latest`); the published Linux image is scanned at release.
- **M2** — SLSA build-provenance attestations for release binaries/packages and for each pushed image (by digest, to GHCR); added `attestations: write`.
- **M3** — release + demo + both Dockerfiles pinned to Go `1.25.11` (was floating `1.25`).
- **L5** — `cmd/` coverage now gated at 80% (was computed but ungated); it's the network-facing surface (currently ~84%).

**Build (`build:`)**
- **L4 (partial)** — Windows image drops to the unprivileged `ContainerUser`; Go pinned in both Dockerfiles.
- **L9** — `make build` stamps a real version from `git describe` instead of `dev`.

**Tests (`test:`)**
- **L7** — fuzz targets for the JSON-RPC reader, manifest loader, target parser, and Constraint unmarshaller (seed corpora also run in the normal suite).

**Docs (`docs:`)**
- **S1** — README "Operating in production" expands the Redis kill-switch fail-open note (cold-start caveat + alert-on-health + run-Redis-HA guidance).
- **L3** — corrected the circuit-breaker description in `repo-guide.md`/`CLAUDE.md` to "guards JWKS fetches" (architecture.md was already accurate).

## Deliberately NOT changed (rationale)

- **Kill-switch fail-open behavior (S1):** ADR-0003 explicitly rejects fail-closed-on-partition (a Redis blip would take down the whole data plane). Handled via docs prominence, not a behavior change.
- **Wiring the breaker into upstream MCP calls (L3):** a behavior-changing feature (new denial mode, config, tests) — documented as JWKS-only instead, per the finding's sanctioned alternative.
- **`--log-level` / leveled logging (L6):** a feature touching ~125 call sites; deferred to its own design rather than bloating this PR.
- **stdio goroutine cap (L10/L11):** the stdio host is a single trusted local process; bounded already.
- **Base-image digest pinning + Actions SHA-pinning (part of L4):** requires resolving real digests/SHAs; deferred to avoid breaking builds blind.

## Note

CI/release YAML changes are validated by syntax + standard action usage; the provenance/image-scan steps are exercised only on a real PR/tag run.

https://claude.ai/code/session_011fsdQoR23wuA7i4JQvT1Pz

---
_Generated by [Claude Code](https://claude.ai/code/session_011fsdQoR23wuA7i4JQvT1Pz)_


cmd/eunox-mcp/http_remote.go
```diff
@@ -222,11 +235,11 @@ func doMCPHTTP(ctx context.Context, client *http.Client, endpoint string, msg rp
 		return rpcMsg{}, resp.Header, nil
 	}
 	if resp.StatusCode != http.StatusOK {
-		body, _ := io.ReadAll(resp.Body)
+		body, _ := io.ReadAll(io.LimitReader(resp.Body, maxUpstreamErrBodyBytes))
 		return rpcMsg{}, resp.Header, fmt.Errorf("upstream returned %d: %s", resp.StatusCode, strings.TrimSpace(string(body)))
 	}
 	var out rpcMsg
-	if err := json.NewDecoder(resp.Body).Decode(&out); err != nil {
+	if err := json.NewDecoder(io.LimitReader(resp.Body, maxUpstreamRespBytes)).Decode(&out); err != nil {
```
eunolabs, line 242: **Behavior change worth a conscious sign-off (and maybe a docs line).** This caps a *remote HTTP* upstream's JSON-RPC response at 4 MiB where it was previously unbounded. The intent is to match the subprocess transport's existing 4 MiB per-message limit (`jsonrpc.go`) so the two upstream transports behave identically — which is the right consistency argument. But it does mean a legitimate large `resources/read` from a remote HTTP upstream that exceeds 4 MiB now fails (`decoding upstream response: ...`) where it used to succeed.

Two things for the reviewer:
1. Confirm 4 MiB is the intended ceiling for *both* transports (I inherited it from the subprocess cap rather than picking a number).
2. If kept, consider a one-liner in the "Operating in production" table noting the max upstream message size, since it's now an enforced limit a user could hit.

---

cmd/eunox-mcp/http.go
```diff
@@ -782,12 +782,22 @@ func isInfraDenialCode(code string) bool {
...
+// The returned reason never embeds the underlying error text. For a remote HTTP
+// upstream that text can carry the upstream's internal hostname/path or its raw
+// response body; disclosing either to the MCP host leaks upstream internals to a
+// party that should only ever see the failure class. The full error is logged to
+// operator stderr at this single choke point — every forward site routes its
+// failures through here — so operators keep full diagnosability while the host
+// sees only the sanitized class. The timeout/canceled reasons carry no
+// upstream-supplied text and are inherently host-safe.
 func upstreamErrInfo(err error, upstreamTimeMs int) (code, reason string) {
+	fmt.Fprintf(os.Stderr, "[eunox-mcp] upstream error: %v\n", err)
```
eunolabs: Two side-effects of moving the log into this mapper that I want to flag explicitly:

1. **Logging from a function named `upstreamErrInfo`** is a little surprising — the name implies a pure mapping. I chose it as the single choke point so both transports log uniformly (http previously logged nothing here; stdio had its own line, which I removed). If you'd rather keep the mapper pure, the alternative is a tiny `logUpstreamErr(err)` helper called at each of the 13 forward sites — more churn, but no hidden side-effect. I went with the choke point for minimal churn; flagging the trade-off.
2. **It now logs once per upstream error including timeouts/cancels** (previously only the `default` branch carried detail). Under a sustained upstream outage that's one stderr line per failed request. That matches how the proxy already logs other per-request security events (e.g. redaction failures), so I think it's acceptable, but if you want it quieter, gate this line to the `default` branch only — the timeout/cancel reasons are already fully descriptive without the underlying `err`.

---

pkg/capability/jwkscache.go
```diff
@@ -132,6 +156,70 @@ func (c *JWKSCache) ForceRefresh(ctx context.Context) (*jose.JSONWebKeySet, erro
...
+func (c *JWKSCache) markKIDAbsent(kid string) {
+	c.negMu.Lock()
+	defer c.negMu.Unlock()
+	now := c.now()
+	for k, at := range c.negKIDs {
+		if now.Sub(at) >= negativeKIDTTL {
+			delete(c.negKIDs, k)
+		}
+	}
+	if _, ok := c.negKIDs[kid]; !ok && len(c.negKIDs) >= negativeKIDMaxLen {
```
eunolabs, line 217: Acknowledging the one tension in this defense: when the map is at `negativeKIDMaxLen`, *new* distinct kids stop being recorded — i.e. suppression lapses exactly under a high-cardinality flood (the case it's meant to blunt). The backstops are (a) the expiry sweep at the top of `markKIDAbsent` (entries older than 30s are dropped, so the cap only binds under a sustained >1024-distinct-kids/30s flood) and (b) the circuit breaker still bounding forced fetches. So the degradation is graceful (back to today's breaker-bounded behavior, never worse), but it's a real ceiling.

If we want the protection to hold under arbitrary cardinality without unbounded memory, the alternative is to skip the per-kid map entirely and instead rate-limit *forced refreshes* globally (e.g. a token bucket on `ForceRefresh`). Happy to swap to that if you'd prefer a cardinality-independent guarantee; I went with the per-kid cache because it's precise (only suppresses confirmed-absent kids) and the breaker already covers the tail. Your call.

---

.github/workflows/release.yml
```diff
@@ -164,6 +180,21 @@ jobs:
...
+      # filesystem scan in go-ci.yml covers the source tree's dependencies; this
+      # covers the assembled image (base layer + binary). ignore-unfixed keeps
+      # the gate actionable — a vuln with no available fix alerts but does not
+      # wedge the release. The distroless base keeps the surface tiny, so a
+      # fixable CRITICAL/HIGH here is a real, addressable signal.
+      - name: Scan published Linux image (gate on fixable CRITICAL/HIGH)
+        uses: aquasecurity/trivy-action@v0.36.0
+        with:
+          scan-type: image
+          image-ref: ghcr.io/eunolabs/eunox-mcp@${{ steps.build.outputs.digest }}
```
eunolabs, line 193: Heads-up on the CI/release changes in this PR (this step + the three `attest-build-provenance` steps + the blocking Trivy fs pass): none of them can run until a real PR/tag actually triggers the workflows, so they're validated here only by YAML syntax and standard action usage. Specifically worth a second pair of eyes before the next tag:

- `image-ref` uses `steps.build.outputs.digest`, which for a multi-arch build is the **index** digest. Trivy resolves it to the runner-arch manifest — fine, but it means the scan covers one arch's layers, not both. Acceptable since the base is identical distroless across arches.
- `ignore-unfixed: true` is deliberate so an unfixable base CVE can't wedge a release, but it does mean a *known-but-unfixed* HIGH won't block — it'll only show in the non-blocking SARIF report. That's the intended trade-off, calling it out.
- `attest-build-provenance@v2` `subject-path` globs (`dist/*.deb` etc.) will **error if they match nothing**. goreleaser's config produces all four extensions, so it should match, but if a future goreleaser change drops a format the release will fail at this step.

---

eunolabs (commented):
Self-review of this PR — flagging the spots where a human decision or a second look matters, rather than approving my own work.

**Tested locally (all green):** `go test -race ./...` (9 pkgs, 0 fail / 0 races), `golangci-lint` (0 issues), `go vet`, gofmt, `make check-license` / `check-notice` / `check-stale-syntax`, and workflow YAML well-formedness. Short active fuzz runs of `FuzzMsgReader` and `FuzzParseTarget` found no crashes.

**Decisions I'd like a reviewer to confirm (inline):**
1. `http_remote.go` — the new 4 MiB cap on remote-HTTP upstream responses is a real behavior change (aligns with the subprocess cap, but a >4 MiB `resources/read` from a remote upstream now fails). Confirm the ceiling + maybe document it.
2. `http.go` — `upstreamErrInfo` now logs as a side effect and logs on every error class; flag if you'd prefer a pure mapper + per-site helper, or to gate logging to the `default` branch.
3. `jwkscache.go` — the negative-cache cap means suppression lapses under a >1024-distinct-kid/30s flood (breaker is the backstop). Fine as graceful degradation; offered a global forced-refresh rate-limit as an alternative if you want a cardinality-independent guarantee.
4. `release.yml` — the provenance/image-scan/blocking-Trivy steps are only exercised on a real PR/tag; noted the index-digest, `ignore-unfixed`, and `subject-path`-must-match caveats.

**Intentionally not in this PR** (rationale in the description): kill-switch fail-open *behavior* (ADR-0003 forbids), breaker-into-upstream wiring, `--log-level`, stdio goroutine cap, and base-image-digest/Actions-SHA pinning.

---

cmd/eunox-mcp/http_remote.go
eunolabs: Resolved in d89f9d6. Kept the 4 MiB cap — it's the intended ceiling for both transports (the subprocess path already enforces it in `jsonrpc.go`, so this aligns remote HTTP to the same limit) — and documented it in the README "Operating in production" section: a per-message 4 MiB limit on either upstream, with the recommendation to chunk/paginate oversized resources upstream.

---

cmd/eunox-mcp/http.go
eunolabs: Resolved in d89f9d6. Scoped the stderr log to the `default`/UPSTREAM_ERROR branch only — the one whose host reason is stripped of detail — so timeouts/cancels no longer emit a per-request line under a sustained outage (their reason is self-describing and already in the audit trail). Left it at the choke point rather than threading a `logUpstreamErr` helper through 13 sites, and the doc comment now states the scoped logging as a deliberate side effect.

---

pkg/capability/jwkscache.go
eunolabs: Decision: keeping the per-kid cache as-is, no code change. Under a sustained >1024-distinct-kid/30s flood it degrades to exactly today's breaker-bounded behavior (never worse), the 30s sweep keeps the cap from binding outside such a flood, and the circuit breaker is the global backstop on forced fetches. Leaving this thread open since it's a judgement call — if you'd prefer a cardinality-independent guarantee I'll swap to a token bucket on `ForceRefresh`, but I'd rather not add that complexity without a concrete need.

---

.github/workflows/release.yml
eunolabs: No code change — this is an informational heads-up for the pre-tag review. The three behaviors (index-digest scan of one arch's layers, `ignore-unfixed` so an unfixable base CVE can't wedge a release, and `subject-path` erroring if it matches nothing) are all intentional. Leaving the thread open so it's visible when someone validates the workflows on the next real tag.

---

PR #468, opened 2026-06-14
Fix session rev-20260614T2310Z-879c44 bugs

### LINKED ISSUE #467: [rev-20260614T2310Z-879c44] circuitbreaker: Allow() calls b.now() twice in Open→HalfOpen transition, lastTransitionAt diverges from cooldown-check instant
**Location:** `pkg/circuitbreaker/breaker.go:153-158`
**Reference:** https://github.com/eunolabs/eunox/blob/e2dea12834a068d3129ba11bf116bed0c7672ba0/pkg/circuitbreaker/breaker.go#L153-L158
**Severity:** Low

**Description:**

The `Allow()` function calls `b.now()` **twice** inside the `StateOpen` branch:

```go
case StateOpen:
    if b.now().Sub(b.lastFailureTime) >= b.config.CooldownDuration {  // first call → T_a
        b.state = StateHalfOpen
        b.halfOpenProbes = 1
        b.halfOpenSuccess = 0
        b.lastTransitionAt = b.now()  // second call → T_b ≥ T_a
        return true
    }
```

The first call establishes the instant the cooldown-check fires. The second, separate `b.now()` call is stored as `lastTransitionAt` — the "when did we enter half-open" timestamp read by dashboards and `Stats()`.

`RecordFailure()` had the exact same defect and it was documented and fixed in #439:

> "Capture the instant once and reuse it: the trip is a single logical event, so the cooldown anchor (lastFailureTime) and the reported transition time (lastTransitionAt) must be identical. Two separate b.now() calls produced divergent timestamps for one trip — invisible on a real clock but a real inconsistency under a step-advancing test clock and in 'how long has the breaker been open' dashboards (#439)."

The #439 fix was applied to `RecordFailure` (which now uses `now := b.now()` captured once) but **not** to the analogous `Open→HalfOpen` path in `Allow()`.

**Trigger / repro:**

*Production wall clock:* `T_b − T_a` is nanoseconds; `lastTransitionAt` is fractionally late. Dashboards that compute "time in half-open = now − lastTransitionAt" underestimate by a few nanoseconds — invisible but logically wrong.

*Step-advancing `WithClock` test:* A custom clock that returns a new, incremented value on every call will see `T_a ≠ T_b`. Any test that:
1. Uses `WithClock` with a step-advancing fake,
2. Calls `Allow()` to trigger the Open→HalfOpen transition, and
3. Asserts `Stats().LastTransitionAt == <expected transition time>`

…will produce a `lastTransitionAt` that is one step ahead of the time used to decide the transition, making the assertion fail or the expected value wrong.

**Suggested fix:**

Capture `b.now()` once and reuse it, mirroring the #439 fix in `RecordFailure`:

```go
case StateOpen:
    now := b.now()
        b.state = StateHalfOpen
        b.halfOpenProbes = 1
        b.halfOpenSuccess = 0
        b.lastTransitionAt = now  // same instant as the cooldown check
        return true
    }
```

This makes `lastTransitionAt` for the Open→HalfOpen transition self-consistent: it records the same instant that was used to evaluate the cooldown, matching the documented invariant from #439.

**Session:** rev-20260614T2310Z-879c44

eunolabs:
## Bugs addressed (session rev-20260614T2310Z-879c44)

- Closes #467 — `Breaker.Allow()` read the clock twice on the `StateOpen` branch: one `b.now()` to check the cooldown and a second, later `b.now()` stored as `lastTransitionAt`. Under a step-advancing test clock (and, fractionally, on a real clock) the reported transition time diverged from the instant that actually decided the transition. Now the instant is captured once (`now := b.now()`) and reused for both the cooldown check and `lastTransitionAt`, mirroring the `RecordFailure` fix from #439.

## Testing

- Added `TestAllowOpenToHalfOpenTransitionTimestamp` in `pkg/circuitbreaker/breaker_triptime_test.go`: a step-advancing clock asserts the `Open->HalfOpen` path reads the clock exactly once and that `LastTransitionAt` equals the cooldown-check instant.
- `go test -race -count=1 ./pkg/circuitbreaker/` passes.

---


eunolabs:
✅ Ready to merge — session rev-20260614T2310Z-879c44 bug addressed (#467), self-review complete with no outstanding comments, tests passing (`go test -race ./pkg/circuitbreaker/`).

---

PR #469, opened 2026-06-14
Fix session rev-20260614T2210Z-3ae4a5 bugs

### LINKED ISSUE #462: [rev-20260614T2210Z-3ae4a5] newestRotatedSibling picks non-rotated files, silently seeding audit chain from wrong source
**Location:** `cmd/eunox-mcp/audit.go:381-388`

**Reference:** https://github.com/eunolabs/eunox/blob/ffde3bfd1e70b9b605f91c64a6e25a75fd197522/cmd/eunox-mcp/audit.go#L381-L388

**Severity:** medium

**Description:**
`newestRotatedSibling` uses `filepath.Glob(logPath + ".*")` and returns the lexically last match — but it applies **no `rotatedAuditRe` filter**, unlike `pruneRotated` (line 1292) which filters the same glob result against `rotatedAuditRe` (`^\.\d{8}T\d{6}\.\d{9}Z(\.\d+)?$`) before acting on it.

Any file in the same directory whose name sorts lexically after all genuine rotated siblings (e.g. `audit.jsonl.bak`, `audit.jsonl.lock`, `audit.jsonl.zzz`) is returned as the "newest rotated sibling". In ASCII order, lowercase letters sort after digits and 'Z', so `audit.jsonl.bak` always beats `audit.jsonl.20260614T...Z`. This function is called from `openAuditSink` (line 337) at startup when the base log is empty — exactly the reopen-fallback path — to resume the tamper-evident chain from the most recent rotated file.

If a spurious file is returned:

1. **Non-JSON last line**: `json.Unmarshal` fails, the chain silently restarts from the genesis sentinel (seq 1, `prev_hmac = sha256:genesis`), creating a detectable but noisy chain break when the real rotated records are still on disk.
2. **JSON-shaped last line** (e.g. a log-rotation backup of the audit file): `json.Unmarshal` succeeds; the chain is resumed with the wrong `seq` and `prevHMAC`, causing every subsequent record's `prev_hmac` link to be wrong. `audit-verify` then reports spurious chain breaks across the entire run, masking any real tampering.
3. **Attacker-crafted file** (requires local filesystem write access): a file named `audit.jsonl.zzz` with a hand-crafted JSON tail can seed an attacker-chosen `seq`/`prevHMAC` into the chain — the HMAC verification on the tail (`openAuditSink:354-357`) would warn but not abort, per the documented audit-failure tradeoff.

`pruneRotated` correctly filters with `rotatedAuditRe`, so the inconsistency is a straightforward omission in `newestRotatedSibling`.

**Trigger / repro:**
Place any file named `audit.jsonl.bak` (or any name starting with `.` followed by a lowercase letter) in the same directory as the audit log. Rotate the audit log (e.g. by exceeding `rotateSizeBytes`), then restart `eunox-mcp`. The chain is resumed from `audit.jsonl.bak` instead of `audit.jsonl.20260614T...Z`. Subsequent records carry the wrong `prev_hmac`; `audit-verify` reports chain breaks for all of them.

**Suggested fix:**
Filter the glob result in `newestRotatedSibling` the same way `pruneRotated` does:

```go
func newestRotatedSibling(logPath string) string {
    matches, err := filepath.Glob(logPath + ".*")
    if err != nil || len(matches) == 0 {
        return ""
    }
    sort.Strings(matches)
    // Walk backwards to find the lexically last genuine rotated sibling.
    for i := len(matches) - 1; i >= 0; i-- {
        if rotatedAuditRe.MatchString(matches[i][len(logPath):]) {
            return matches[i]
        }
    }
    return ""
}
```

This makes `newestRotatedSibling` and `pruneRotated` agree on what constitutes a rotated sibling, closing the inconsistency.

**Session:** rev-20260614T2210Z-3ae4a5

### LINKED ISSUE #463: [rev-20260614T2210Z-3ae4a5] StartCleanup TOCTOU race permanently consumes sync.Once with a goroutine that immediately exits
**Location:** `pkg/callcounter/memory.go:447-478`

**Reference:** https://github.com/eunolabs/eunox/blob/ffde3bfd1e70b9b605f91c64a6e25a75fd197522/pkg/callcounter/memory.go#L447-L478

**Severity:** low

**Description:**
`StartCleanup` guards against consuming `cleanupOnce` with an already-canceled context by checking `ctx.Err() != nil` before `cleanupOnce.Do`. The comment says "Returning false here keeps the Once available for a subsequent call with a live context (#426)."

But between the pre-check and the `Do`, there is a TOCTOU window in which the context can be canceled:

```
goroutine A                    goroutine B (cancels ctx)
─────────────────────────────────────────────────────
ctx.Err() → nil (passes)
                               cancel()   // ctx is now done
cleanupOnce.Do(func(){         
    started = true             
    go func() {               
        select {               
        case <-ctx.Done(): return  // exits immediately
        ...
        }
    }()
})
// cleanupOnce permanently consumed; goroutine exits immediately
```

After this race:
- The cleanup goroutine exits immediately on its first `select` since `ctx.Done()` is already closed.
- `cleanupOnce` is **permanently consumed**: no future call to `StartCleanup` (even with a fresh live context) can start cleanup — all subsequent calls return `false` without entering `Do`.
- The `InMemory` counter's entry map grows without bound between invocations until the process is restarted, because no background `Cleanup` goroutine is running.

The comment acknowledges "The check-then-Do TOCTOU window is the same one the running goroutine already has," but the consequences differ: the running goroutine exiting is a clean shutdown, whereas the Once being consumed with no live goroutine leaves cleanup permanently disabled — the subtle asymmetry the comment elides.

Under normal operation this window is sub-microsecond and extremely unlikely. It becomes relevant in tests that cancel a context before calling `StartCleanup`, or in any code path that creates a short-lived context and calls `StartCleanup` near its deadline.

**Trigger / repro:**
In a test or under tight scheduling:
```go
cancel() // cancel before StartCleanup

// This should return false WITHOUT consuming the Once...
started := counter.StartCleanup(ctx, 0)
// ...but with the race, the Once is consumed and cleanup is permanently dead.

ctx2, cancel2 := context.WithCancel(context.Background())
defer cancel2()
started2 := counter.StartCleanup(ctx2, 0) // returns false, no goroutine launched
```

The only reliable way to reproduce the exact race requires the cancel to occur in the nanosecond window between the `ctx.Err()` check and `cleanupOnce.Do`; a synthetic delay or scheduler preemption in that window makes it consistent.

**Suggested fix:**
Re-check `ctx.Err()` inside the `Do` closure, and if the context is already done, avoid launching the goroutine while still consuming the Once (so the Once is burned only on an actually-started goroutine, or report via a separate flag):

```go
m.cleanupOnce.Do(func() {
    if ctx.Err() != nil {
        // Context canceled in the TOCTOU window; don't launch the goroutine.
        // The Once is consumed here, so document this as a known limitation:
        // a counter whose first StartCleanup races a cancel needs a fresh InMemory.
        return
    }
    started = true
    // ... launch goroutine as before
})
```

Alternatively, replace `sync.Once` with a mutex+bool that can be retried on a dead goroutine, but this is heavier. The simpler fix at least eliminates the "live goroutine that immediately exits" scenario while keeping the already-acknowledged limitation that the Once is consumed only once.

**Session:** rev-20260614T2210Z-3ae4a5

### LINKED ISSUE #464: [rev-20260614T2210Z-3ae4a5] openAuditSink silently resets audit chain when last log line is non-JSON (no warning emitted)
**Location:** `cmd/eunox-mcp/audit.go:343-368`

**Reference:** https://github.com/eunolabs/eunox/blob/ffde3bfd1e70b9b605f91c64a6e25a75fd197522/cmd/eunox-mcp/audit.go#L343-L368

**Severity:** low

**Description:**
`openAuditSink` reads the last line of the audit log to resume the tamper-evident chain (`seq`, `prevHMAC`) across restarts. It correctly warns when the last line **parses as JSON but fails HMAC verification**:

```go
if prev.HMAC != "" {
        fmt.Fprintf(os.Stderr, "WARNING: audit log tail failed HMAC verification...")
    }
}
```

But when `json.Unmarshal` itself **fails** (truncated record from a partial write at shutdown, mid-write power loss, disk corruption, or the tail of an oversized record), the `if json.Unmarshal([]byte(last), &prev) == nil` guard is false and the whole block is silently skipped:

```go
if last != "" {
    var prev auditRecord
    if json.Unmarshal([]byte(last), &prev) == nil {
        // chain resumed here ... HMAC warning here ...
    }
    // ← no else: JSON parse failure is silently ignored
}
```

The result: `s.seq` stays at 0 and `s.prevHMAC` stays at `""`, so the first new record is stamped with `seq=1` and `prev_hmac=sha256:genesis` — as if this were a brand-new log. The real prior records are still on disk with a chain that ends at some `seq=N` with HMAC `"sha256:xyz"`. When `audit-verify` scans the combined file it sees a genuine chain break: the old chain ends at seq N, and then a new record appears with seq=1 and genesis `prev_hmac`. This is **undetectable from legitimate tampering** without an external high-water mark.

The symmetric case — HMAC verification failure — already emits a warning. The JSON parse failure path is the **more severe** case (it means the log is structurally corrupt, not merely tampered-with) yet it produces no diagnostic output at all. An operator restart after a hard kill or power loss would silently break chain continuity with no indication of why `audit-verify` later reports chain breaks.

**Trigger / repro:**
1. Start the proxy, write several records.
2. Kill the process mid-write (e.g. `kill -9`) so the last record in the log is truncated (no closing `}`).
3. Restart the proxy.
4. Run `eunox-mcp audit-verify`.
5. Observe chain breaks from the restart with no corresponding startup warning in stderr.

**Suggested fix:**
Add a warning in the `else` branch (or check the unmarshal error explicitly):

```go
if last != "" {
    var prev auditRecord
    if err := json.Unmarshal([]byte(last), &prev); err != nil {
        fmt.Fprintf(os.Stderr,
            "[eunox-mcp] WARNING: audit log tail failed to parse (%v); "+
                "the last record may be truncated or corrupt. "+
                "Restarting the chain from genesis — run 'eunox-mcp audit-verify' "+
                "and reconcile against your external sink.\n", err)
    } else {
        // existing HMAC verification and chain resume logic
        if prev.HMAC != "" {
            if ok, verifyErr := s.VerifyRecord([]byte(last)); verifyErr != nil || !ok {
                fmt.Fprintf(os.Stderr, "[eunox-mcp] WARNING: audit log tail failed HMAC verification...")
            }
        }
        s.seq = prev.Seq
        s.prevHMAC = prev.HMAC
        s.resumedLegacyTail = prev.HMAC == ""
    }
}
```

This mirrors the HMAC-verification warning already present and gives operators a clear signal when a crash or disk error has broken chain continuity, so they can reconcile promptly rather than discovering the break only after running `audit-verify`.

**Session:** rev-20260614T2210Z-3ae4a5

eunolabs:
## Bugs addressed (session rev-20260614T2210Z-3ae4a5)

- Closes #464 — `openAuditSink` silently restarted the tamper-evident chain from genesis when the last log line failed to parse as JSON (truncated/corrupt record from a partial write or power loss), with no diagnostic. The symmetric HMAC-verification failure already warned; the parse-failure path now warns too, so an operator can reconcile rather than discovering the break only via `audit-verify`.
- Closes #462 — `newestRotatedSibling` returned the lexical-max of `filepath.Glob(logPath + ".*")` with no filter, so a spurious sibling that sorts after every genuine rotated stamp (`audit.jsonl.bak`, `.lock`, `.zzz` — lowercase sorts after digits and `Z`) could be returned as the "newest rotated" tail and seed the resumed chain's `seq`/`prev_hmac` from the wrong source. It now filters by `rotatedAuditRe`, matching `pruneRotated`.
- Closes #463 — `InMemory.StartCleanup` pre-checked `ctx.Err()` before `cleanupOnce.Do`, but a cancel landing in the TOCTOU window left the `Once` consumed by a goroutine that exited on its first `select`, permanently disabling cleanup while reporting a misleading `true`. The context is now re-checked inside the `Once`: it does not launch a goroutine that would immediately exit and reports `false` (the `Once` consumption remains a documented limitation).

## Testing

- `TestOpenAuditSinkWarnsOnUnparseableTail` — asserts the stderr warning and that the chain restarts from genesis (seq 0, empty prev_hmac) on an unparseable tail.
- `TestNewestRotatedSiblingIgnoresNonRotatedFiles` + `TestOpenAuditSinkIgnoresSpuriousSiblingOnResume` — a spurious `.bak`/`.zzz` sibling is ignored; resume comes from the genuine rotated sibling, not an attacker-chosen `seq`.
- `TestStartCleanup_CancelInTOCTOUWindowDoesNotLaunchGoroutine` — a `flipCtx` whose `Err()` flips between the pre-check and the inner re-check verifies no goroutine launches and `StartCleanup` reports `false`.
- `go test -race -count=1 ./cmd/eunox-mcp/ ./pkg/callcounter/` and `go vet` pass; `go build ./...` clean.

---


eunolabs:
✅ Ready to merge — all session rev-20260614T2210Z-3ae4a5 bugs addressed (#464, #462, #463), self-review complete (one stale-comment consistency fix applied in round 2, no outstanding comments). `go test -race ./cmd/eunox-mcp/ ./pkg/callcounter/`, `go vet`, and `golangci-lint` all pass; `go build ./...` clean.

---

Issue #460, opened 2026-06-14
[rev-20260614T2010Z-59e144] InMemory.Peek computes its window cutoff from un-floored now, breaking the floorToMicro boundary invariant (#429)


eunolabs:
**Location:** `pkg/callcounter/memory.go:387-415` (specifically the cutoff at line 407)

**Reference:** https://github.com/eunolabs/eunox/blob/9150f24e4dc49ccf726d1e0d156da1e6f3d83f65/pkg/callcounter/memory.go#L407-L414

**Severity:** low

**Description:**
`#429` introduced `floorToMicro` so the InMemory backend stores timestamps at microsecond granularity and computes its window cutoff from that same floored `now`, matching the Redis backend (which stores `UnixMicro` scores) — the stated goal being that a call landing in the final sub-microsecond of a window is classified identically by both backends and by both the read and write paths.

`IncrementAndGet` (line 173) and `IncrementIfBelow` (line 309) honor this: `now := floorToMicro(m.now()); cutoff := now.Add(-window)`. But `Peek` does **not**:

```go
cutoff := m.now().Add(-time.Duration(windowSec) * time.Second)   // line 407, un-floored
...
for _, ts := range e.timestamps {
    if ts.After(cutoff) { count++ }   // ts values are floored to micro
}
```

`Peek`'s cutoff carries the sub-microsecond nanoseconds of `m.now()`, while the stored `ts` are floored to micro. Because `now_raw >= floorToMicro(now_raw)`, `Peek`'s cutoff is slightly later than the write path's cutoff for the same instant, so `Peek`'s effective window is up to ~1 µs **shorter**. A boundary timestamp that `IncrementAndGet` (or the Redis backend) still counts in-window can be seen as expired by `Peek` — the exact cross-path/cross-backend disagreement `#429` set out to eliminate.

This is the `sequenceBlock` read path (`handleSequenceBlock` → `history.Peek`), so the practical effect is a ~1 µs window at the 24h `sequenceHistoryWindowSec` boundary where an antecedent could read as expired and the block fails open. Real-world impact is negligible (sub-microsecond at a 24h edge), hence low severity — but it is a genuine violation of the backend-consistency invariant and the InMemory read/write paths disagreeing with each other.

**Trigger / repro:**
[...]

PR #470, opened 2026-06-14
Fix session rev-20260614T2010Z-59e144 bugs

### LINKED ISSUE #459: [rev-20260614T2010Z-59e144] Audit record envelope (attacker-controlled Target) is unbounded, can exceed the 4 MiB verify/resume window and break audit-verify + chain resume
**Location:** `cmd/eunox-mcp/audit.go:575-594` (`boundAuditDetails`), `cmd/eunox-mcp/audit.go:981-989` (`verifyAuditLog` scanner), `cmd/eunox-mcp/audit.go:359-388` (`readLastAuditLine`), `cmd/eunox-mcp/audit.go:631-655` (`deriveTargetFields`/`bareTargetName`)

**Reference:** https://github.com/eunolabs/eunox/blob/9150f24e4dc49ccf726d1e0d156da1e6f3d83f65/cmd/eunox-mcp/audit.go#L575-L594

**Severity:** medium

**Description:**
`#416`'s size guard (`boundAuditDetails`, applied in `writeRecord`) bounds only the `Details` map (per-value 512 KiB, total 1 MiB). It does **not** bound the record's envelope string fields. The most important of these is `Target`, which is derived verbatim from the raw MCP identifier:

- `stdio.go:472/492/523` pass `params.Name` (the tool name) as the identifier for `tools/call`.
- `stdio.go:568/584/597` pass `params.URI` (the resource URI) for `resources/read`.

`deriveTargetFields` → `bareTargetName` returns this identifier unchanged into `auditRecord.Target` (no length cap). Both `params.Name` and `params.URI` are attacker-controlled and are written even on the **deny** path (e.g. an unknown tool name).

The incoming request is capped at 4 MiB (`maxRequestBodyBytes` for HTTP, the `bufio.Scanner` 4 MiB buffer in `jsonrpc.go:310` for stdio). But the audit-verify scanner buffer (`audit.go:989`) and the chain-resume tail window (`readLastAuditLine`, `maxTail = 4<<20`) are **also exactly 4 MiB**. A minimal-envelope request whose tool name / URI fills nearly the whole 4 MiB body produces an audit record whose `Target` alone is ~4 MiB; the audit JSON envelope (`time` RFC3339Nano, `seq`, `request_id`, `prev_hmac`/`_hmac` ~71 chars each, `key_id`, etc. — ~400–500 bytes) is then added on top, pushing the serialized line **over 4 MiB**.

Consequences:
1. **`audit-verify` DoS / integrity loss:** a line longer than the 4 MiB buffer makes `scanner.Scan()` fail with `bufio.ErrTooLong`; `verifyAuditLog` returns that error and can no longer verify the log past that point. A single crafted request (even one that is denied) renders the audit trail unverifiable.
2. **Chain-resume corruption:** on restart, `readLastAuditLine` reads only the last 4 MiB. If the last record exceeds that, the window contains no record-boundary newline, so the returned "line" is a truncated fragment, `json.Unmarshal` fails, and the chain is silently **not** resumed (seq/prev_hmac reset to genesis) — a tamper-evident-chain discontinuity.

The comment in `readLastAuditLine` even states the assumption that "envelope fields (timestamps, ids, target) [are] realistically far smaller" — that assumption is violated by an attacker-controlled `Target`.

**Trigger / repro:**
Send a `tools/call` (or `resources/read`) request of ~4 MiB whose `params.name` (or `params.uri`) is a ~3.99 MiB string, with an otherwise minimal JSON-RPC envelope. The request is denied (unknown tool) but an audit record with a ~3.99 MiB `Target` is written; the serialized line exceeds 4 MiB. Then run `eunox-mcp audit-verify` (fails with a scan error) and/or restart the proxy (chain not resumed from the oversized tail).

**Suggested fix:**
Bound the envelope, not just `Details`. Either (a) cap the `Target`/identifier length when building the record (truncate with a visible marker, mirroring `boundDetailValue`), or (b) compute the full serialized record size in `writeRecord` and, if it exceeds a record cap strictly below the 4 MiB scan/tail window, replace oversized envelope fields with placeholders. Whichever is chosen, the per-record byte ceiling must be provably less than both the `verifyAuditLog` buffer and `readLastAuditLine`'s `maxTail`.

**Session:** rev-20260614T2010Z-59e144

### LINKED ISSUE #461: [rev-20260614T2010Z-59e144] Audit Details size cap is applied in the drainer, after record() deep-clones full arguments — hot-path O(n) clone and up to 4096 un-truncated multi-MiB records queued
**Location:** `cmd/eunox-mcp/audit.go:417-465` (`record()` → `copyDetails`), `cmd/eunox-mcp/audit.go:743-749` (`writeRecord` → `boundAuditDetails`), `cmd/eunox-mcp/audit.go:78` (`auditChannelSize = 4096`); audit-mode source of `details`: `cmd/eunox-mcp/stdio.go:514-523`

**Reference:** https://github.com/eunolabs/eunox/blob/9150f24e4dc49ccf726d1e0d156da1e6f3d83f65/cmd/eunox-mcp/audit.go#L451-L465

**Severity:** medium

**Description:**
`auditSink.Record`/`record` is documented as doing "only struct initialization and a non-blocking channel send", with all serialization, signing, and I/O deferred to the drainer to keep the policy hot path fast. But in audit/wiretap mode the allow path passes the **full, live tool-arguments map** as `details` (`stdio.go:517-518`: `details = params.Arguments`), and `record()` calls `copyDetails(details)` — a **recursive deep clone of the entire arguments structure** — synchronously on the request-serving goroutine before the channel send (`audit.go:462`).

Two problems follow from `boundAuditDetails` running only later, in the drainer (`writeRecord`, `audit.go:749`):

1. **Hot path is O(arguments size), not O(1).** A multi-MiB nested argument map (write_file body, base64 blob, large corpus — the very cases `boundAuditDetails` exists to cap) is deep-cloned in full on every audited request before enqueue. The documented "cheap struct-init + send" contract is violated precisely for large payloads.

2. **Bounded queue holds unbounded-per-record memory.** The cap that would shrink `Details` to 1 MiB is applied *after* the record is dequeued. So the 4096-slot channel can hold up to 4096 **un-truncated** clones, each up to ~the 4 MiB request limit — on the order of ~16 GiB of live heap under audit-mode load — before any truncation happens. Under a burst of large-argument requests in audit mode this is a memory-amplification / OOM vector, and it defeats the purpose of the 1 MiB cap as a memory bound (it only ever bounds what is *written to disk*, never what is *queued*).

**Trigger / repro:**
Run the proxy in audit (wiretap) mode and issue many concurrent `tools/call` requests each carrying a few-MiB `arguments` payload. Observe (a) per-request latency scaling with argument size on the request goroutine, and (b) RSS growing to ~N × (queued records) × payload size rather than being bounded by the 1 MiB Details cap.

**Suggested fix:**
Apply the size bound *before* enqueue / during the clone rather than in the drainer: have `copyDetails` (or a combined clone-and-bound pass) enforce `auditDetailValueCap`/`auditDetailsTotalCap` as it copies, so the cloned structure handed to the channel is already capped. This keeps `Record` cheap for large payloads and makes the channel's memory footprint bounded by `auditChannelSize × (1 MiB + envelope)`.

**Session:** rev-20260614T2010Z-59e144

eunolabs:
## Bugs addressed (session rev-20260614T2010Z-59e144)

- Closes #461 — The audit `Details` size cap was applied in the drainer (`writeRecord`), *after* `record()` had already deep-cloned the full tool-arguments map and enqueued it. In audit-only mode (where the allow path logs `params.Arguments` verbatim) the 4096-slot queue could therefore retain up to 4096 un-truncated multi-MiB clones (~16 GiB of live heap), since the 1 MiB cap only shrank `Details` after dequeue. Bounding is now applied in `record()` before enqueue (`boundAuditDetails(copyDetails(details))`), so each queued record's `Details` is already capped and the queue's footprint is bounded by `auditChannelSize x (1 MiB + envelope)`. The redundant drainer-side bound was removed.
- Closes #459 — The audit record envelope was unbounded. `Target` (derived verbatim from the attacker-controlled `params.Name` / `params.URI`, written even on the deny path) and an unrecognized raw `Method` could each be ~4 MiB, pushing the serialized audit line past the 4 MiB `audit-verify` scanner buffer and the 4 MiB chain-resume tail window — making the log unverifiable and silently breaking chain resume on restart. `deriveTargetFields` now bounds both fields to 8 KiB via `boundEnvelopeField` (visible truncation marker). Combined with the 1 MiB `Details` cap, the full record is provably far below the 4 MiB window.

## Already resolved (no change needed)

- #460 — `InMemory.Peek` computing its cutoff from an un-floored `now`. This was filed against an older blob (`9150f24`); current `main` already floors the cutoff at `pkg/callcounter/memory.go:413` (`floorToMicro(m.now())...`, landed in #457), and `TestInMemoryPeekFloorMatchesIncrement` already guards it. Closing #460 separately as a stale duplicate.

## Testing

- `TestRecordBoundsDetailsBeforeEnqueue` — inspects the queued record directly (drainer never started) and asserts the oversized value is already a placeholder at enqueue time.
- `TestDeriveTargetFieldsBoundsEnvelope` — oversized `Target` and raw `Method` are truncated with a marker; a legitimate short target is untouched.
- `TestRecordOversizedTargetStaysUnderVerifyWindow` — a ~4 MiB tool name (denied) produces an audit line < 1 MiB that still parses and verifies its chain HMAC.
- `go test -race -count=1 ./cmd/eunox-mcp/`, `go vet`, and `golangci-lint` all pass; `go build ./...` clean.

---


eunolabs:
✅ Ready to merge — session rev-20260614T2010Z-59e144 bugs addressed (#461, #459; #460 closed separately as already-fixed by #457). Self-review complete over 3 rounds (round 1: generic truncation marker; round 2: threat-model §6.2 docs update for the envelope bound + bound-before-enqueue), no outstanding comments. `go test -race ./cmd/eunox-mcp/`, `go vet`, and `golangci-lint` all pass; `go build ./...` clean.

---

PR #499, opened 2026-06-15
Fix session rev-20260615T0247Z-c77bdd bugs

### LINKED ISSUE #496: [rev-20260615T0247Z-c77bdd] circuitbreaker: State() and Stats() hold exclusive write lock in read-only methods
**Location:** `pkg/circuitbreaker/breaker.go:386-394, 399-415`
**Reference:** https://github.com/eunolabs/eunox/blob/08eaa5669e68f52a488f5c777db790e466c4f5a3/pkg/circuitbreaker/breaker.go#L386-L415
**Severity:** Medium

**Description:**

`State()` and `Stats()` both call `b.mu.Lock()` / `b.mu.Unlock()` (exclusive write lock) even though they only read fields — they never write to the `Breaker`. Using a write lock here serializes these methods with every other operation on the breaker, including concurrent callers of `State()` and `Stats()` themselves, `Allow()`, `AllowProbe()`, `RecordSuccess()`, and `RecordFailure()`. A read lock (`b.mu.RLock()`) is sufficient and would allow concurrent reads to proceed in parallel.

```go
// State()  — line 386
    b.mu.Lock()          // ← should be RLock
    defer b.mu.Unlock()  // ← should be RUnlock
    ...
}

// Stats()  — line 399
    b.mu.Lock()          // ← should be RLock
    defer b.mu.Unlock()  // ← should be RUnlock
    ...
}
```

**Trigger / repro:**

The circuit breaker is the production guard for JWKS endpoint fetches (the only wiring documented in the architecture). In JWT-authenticated deployments, every request that misses the token cache triggers a JWKS fetch, which calls the breaker. Any concurrent health check, metrics scrape, or `stats` subcommand that calls `State()` or `Stats()` while enforcement is in flight contends with those callers on the write lock. Under sustained load with many concurrent JWT clients, this unnecessary serialization degrades enforcement throughput.

Minimal reproduction:

```go
b := circuitbreaker.New(circuitbreaker.DefaultConfig())
go func() { _ = b.State() }()
go func() { _ = b.State() }() // serialized with above — should not be
```

The Go race detector will not catch this (no actual data race), but a mutex contention profiler will show write-lock contention on read-only calls.

**Suggested fix:**

```go
    b.mu.RLock()
    defer b.mu.RUnlock()
    ...
}

    b.mu.RLock()
    defer b.mu.RUnlock()
    ...
}
```

Neither method writes to any field, so `RLock` is correct. The `b.now()` function pointer is set once at construction and never written to again, so reading it under `RLock` is safe.

**Session:** rev-20260615T0247Z-c77bdd

### LINKED ISSUE #497: [rev-20260615T0247Z-c77bdd] TokenCache.Put() does not refresh the eviction position of an existing entry, contradicting "LRU-like" documentation
**Location:** `pkg/capability/token_cache.go:197-222`
**Reference:** https://github.com/eunolabs/eunox/blob/08eaa5669e68f52a488f5c777db790e466c4f5a3/pkg/capability/token_cache.go#L197-L222
**Severity:** Low

**Description:**

`TokenCache` is documented as an "LRU-like cache" (struct comment, line 56), but the `Put()` method does not move an existing entry to the back of the eviction list when it is refreshed. On the update path the existing list element is reused at its original position:

```go
if existing, exists := c.entries[key]; exists {
    // Update path: reuse the existing list element so the insertion-
    // order list stays consistent.
    entry.listElem = existing.listElem   // ← old position preserved
} else {
    // eviction + PushBack only for new entries
    elem := c.insertOrder.PushBack(key)
    entry.listElem = elem
}
c.entries[key] = entry
```

As a result, a token that is actively being refreshed (e.g., a long-running agent sending the same bearer token on every request) stays at the front of the eviction list if it was the oldest insertion. The next call to `Put()` for a genuinely-new token will evict the just-refreshed entry even though other, older-USED tokens remain in the cache.

**Trigger / repro:**

```
1. Fill cache to MaxSize entries, with token A inserted first.
2. Call Put(A, freshPayload) — refreshes A's TTL.
3. Call Put(X, payload) for a brand-new token X.
4. Eviction check: c.insertOrder.Front() is still A (its insertion
   position never changed), so A is evicted.
5. The next request bearing token A is a cache miss even though it
   was just refreshed in step 2.
```

This is observable in high-throughput JWT deployments where a gateway process handles many concurrent sessions with the same token: the token keeps cycling in and out of the cache, causing unnecessary JWKS re-verification and Redis revocation round-trips.

**Suggested fix:**

On the update path, move the entry to the back of `insertOrder` so that refreshing a token resets its eviction timer, matching the documented "LRU-like" semantics:

```go
if existing, exists := c.entries[key]; exists {
    // Move to back so a refreshed entry is not evicted before
    // entries that have not been used since their original insertion.
    c.insertOrder.MoveToBack(existing.listElem)
    entry.listElem = existing.listElem
} else {
        ...
    }
    elem := c.insertOrder.PushBack(key)
    entry.listElem = elem
}
c.entries[key] = entry
```

If insertion-order FIFO eviction is actually the intended policy (i.e., the "LRU-like" phrase in the docs is imprecise), the struct and `Put()` documentation should be corrected to say "insertion-order FIFO eviction" so callers can reason about the eviction behavior accurately.

**Session:** rev-20260615T0247Z-c77bdd

### LINKED ISSUE #498: [rev-20260615T0247Z-c77bdd] auditSink.Close() uses a stack-local error variable; concurrent callers always receive nil even when Sync/Close failed
**Location:** `cmd/eunox-mcp/audit.go:1010-1046`
**Reference:** https://github.com/eunolabs/eunox/blob/08eaa5669e68f52a488f5c777db790e466c4f5a3/cmd/eunox-mcp/audit.go#L1010-L1046
**Severity:** Low

**Description:**

`Close()` declares `closeErr` as a **stack-local variable** and captures it by reference inside the `closeOnce.Do` closure. Only the goroutine whose `Do` invocation actually runs the body ever receives the sync/close error — every other concurrent or subsequent caller returns the uninitialized zero value (`nil`):

```go
func (s *auditSink) Close() error {
    var closeErr error          // local to this stack frame
    s.closeOnce.Do(func() {
        ...
        syncErr := s.f.Sync()
        closeErr = s.f.Close() // ← written into the first caller's local only
        if syncErr != nil {
            closeErr = syncErr
        }
    })
    return closeErr             // second+ callers always return nil here
}
```

If two goroutines call `Close()` concurrently (e.g., a signal handler and main), the one that does not run the `Do` body returns `nil` regardless of whether `Sync` or `Close` reported a failure. A caller that checks the return value to decide whether to exit non-zero (the existing warning message at the end of `Do` hints at this intent) will silently miss the error.

**Trigger / repro:**

```go
var wg sync.WaitGroup
var errs [2]error
for i := range errs {
    wg.Add(1)
    go func(i int) {
        defer wg.Done()
        errs[i] = sink.Close()
    }(i)
}
wg.Wait()
// errs[1] is nil even if Sync returned EIO
```

Under normal single-caller usage this is harmless, but the `Close` API contract ("Close is idempotent") implies it is safe to call from multiple sites (e.g., a defer chain plus a signal handler). The subtle result is that a full-disk or EIO failure on the final fsync of the audit log goes unreported to whichever caller loses the `sync.Once` race.

**Suggested fix:**

Store the error in a struct field so all callers can read it:

```go
type auditSink struct {
    ...
    closeErr error // set by the first Close(); readable by all after closeOnce
    ...
}

func (s *auditSink) Close() error {
    s.closeOnce.Do(func() {
        ...
        syncErr := s.f.Sync()
        s.closeErr = s.f.Close()   // store in struct
        if syncErr != nil {
            s.closeErr = syncErr
        }
    })
    return s.closeErr              // all callers see the real result
}
```

`sync.Once.Do` provides a happens-before guarantee: the write to `s.closeErr` inside `Do` completes before any subsequent caller's `Do` returns, making the subsequent read of `s.closeErr` safe without additional synchronization.

**Session:** rev-20260615T0247Z-c77bdd

eunolabs:
## Bugs addressed (session rev-20260615T0247Z-c77bdd)

- Closes #498 — `auditSink.Close()` stored its result in a stack-local `closeErr`, so any caller that lost the `sync.Once` race got `nil` even when `Sync`/`Close` failed. The error is now stored on the `auditSink` struct, so all callers observe the real result (`sync.Once` provides the happens-before guarantee).
- Closes #497 — `TokenCache.Put()` did not refresh the eviction position of an existing entry, so a frequently re-validated token could be evicted before older, un-refreshed entries, contradicting the documented "LRU-like" semantics. The update path now `MoveToBack`s the entry.
- Closes #496 — `Breaker.State()` and `Breaker.Stats()` are read-only but took the exclusive write lock, serializing them with every other breaker operation. The mutex is now a `sync.RWMutex` and both methods take the read lock.

## Testing
- `go test -count=1 ./pkg/... ./cmd/...` — all pass.
- `go test -race -count=1 ./pkg/circuitbreaker/` — passes with race detector.
- Added `TestTokenCache_Put_RefreshMovesToBack` covering the new LRU refresh behavior.

---


eunolabs:
✅ Ready to merge — all session rev-20260615T0247Z-c77bdd bugs addressed (#498, #497, #496), self-review complete with no outstanding comments, tests passing (`go test ./pkg/... ./cmd/...`, race-clean circuitbreaker tests, fmt/vet clean).

---

PR #500, opened 2026-06-15
Fix session rev-20260615T0210Z-d6ac44 bugs

### LINKED ISSUE #493: [rev-20260615T0210Z-d6ac44] audit: filepath.Glob over unescaped logPath silently matches wrong rotated siblings
## Location

`cmd/eunox-mcp/audit.go`

- `newestRotatedSibling()` — line 393: `filepath.Glob(logPath + ".*")`
- `pruneRotated()` — line 1344: `filepath.Glob(s.logPath + ".*")`

## Reference

https://github.com/eunolabs/eunox/blob/affe5a66a16e39d9a6d0bd46391dac6ed24681be/cmd/eunox-mcp/audit.go#L393
https://github.com/eunolabs/eunox/blob/affe5a66a16e39d9a6d0bd46391dac6ed24681be/cmd/eunox-mcp/audit.go#L1344

## Severity

**Medium**

## Description

Both `newestRotatedSibling` and `pruneRotated` build a glob pattern by concatenating the raw `logPath` with `".*"` and passing it to `filepath.Glob`. `filepath.Glob` treats `[`, `]`, `*`, and `?` as metacharacters. If an operator configures a log path that contains any of those characters — valid in some filesystem paths and container volume mount names — the resulting pattern is interpreted as a glob expression rather than a literal prefix.

Three failure modes follow:

1. **Wrong chain resume source** (`newestRotatedSibling`): the glob matches a file that is not a genuine rotated sibling. `readLastAuditLine` loads its tail, and the audit chain restarts from a wrong `seq`/`prev_hmac` pair. All subsequent records fail chain verification.

2. **Wrong files pruned** (`pruneRotated`): the glob returns unrelated files in the same directory, which are then deleted to satisfy the `retain` limit. Data that was never part of the audit log is permanently lost.

3. **Silent no-op** (`filepath.ErrBadPattern`): on some path strings the resulting pattern is syntactically invalid (e.g. unbalanced `[`). Both functions treat a non-nil error from `filepath.Glob` as "no matches" — `newestRotatedSibling` returns `""` (no chain resume; audit chain breaks), and `pruneRotated` returns early (no pruning). Neither logs the error, so the operator has no indication the configuration is being silently ignored.

The `rotatedAuditRe` filter applied *after* the glob call (line 406 in `newestRotatedSibling`, implied in `pruneRotated`) does not close the issue: it can only reduce false positives when the metacharacter in `logPath` causes over-matching, and provides no protection against `ErrBadPattern` or under-matching.

## Trigger / Repro

Configure `audit.logPath` to a path containing a bracket, e.g.:

```yaml
audit:
  logPath: /var/log/eunox[1]/audit.log
```

`filepath.Glob("/var/log/eunox[1]/audit.log.*")` will be parsed as a character-class glob. On a filesystem where `/var/log/eunox1/audit.log.*` or `/var/log/eunox/audit.log.*` exists, the glob matches those paths instead of (or in addition to) the intended rotated siblings.

## Suggested Fix

Replace `filepath.Glob` with a manual directory scan using `os.ReadDir` plus a `strings.HasPrefix` filter. This avoids metacharacter interpretation entirely:

```go
func rotatedSiblings(logPath string) ([]string, error) {
    dir := filepath.Dir(logPath)
    base := filepath.Base(logPath) + "."
    entries, err := os.ReadDir(dir)
    if err != nil {
        return nil, err
    }
    var matches []string
    for _, e := range entries {
        if strings.HasPrefix(e.Name(), base) {
            matches = append(matches, filepath.Join(dir, e.Name()))
        }
    }
    return matches, nil
}
```

Apply the same `rotatedAuditRe` suffix filter on `e.Name()[len(base)-1:]` (keep the leading `.`) afterwards to retain the existing protection against non-sibling files.

## Session

- Session ID: `rev-20260615T0210Z-d6ac44`
- Reviewed commit: `affe5a66a16e39d9a6d0bd46391dac6ed24681be`
- Modules reviewed: `pkg/circuitbreaker`, `pkg/callcounter`, `cmd/eunox-mcp/audit.go`

### LINKED ISSUE #494: [rev-20260615T0210Z-d6ac44] enforcement: unreachable dead-code branch in handleSequenceBlock
## Location

`pkg/enforcement/handlers.go`

- Pre-check (makes the branch unreachable): lines 1109–1125
- Dead branch: line 1182–1184

## Reference

https://github.com/eunolabs/eunox/blob/affe5a66a16e39d9a6d0bd46391dac6ed24681be/pkg/enforcement/handlers.go#L1109-L1125
https://github.com/eunolabs/eunox/blob/affe5a66a16e39d9a6d0bd46391dac6ed24681be/pkg/enforcement/handlers.go#L1182-L1184

## Severity

**Low** (correctness / maintainability; no behaviour change)

## Description

`handleSequenceBlock` runs a pre-validation loop before its main antecedent-check loop. The pre-validation (lines 1109–1125) iterates every entry in `sb.AfterTools` and returns a `ConditionError` if any entry reduces to an empty bare name after `stripEnginePrefix` + `strings.TrimSpace`. If such an entry exists the function never reaches the main loop.

The main loop (starting after line 1125) then repeats the same split and trim:

```go
priorType, priorTool := splitEnginePrefix(prior)   // line 1179
priorTool = strings.TrimSpace(priorTool)            // line 1182
if priorTool == "" {                                // line 1183 — unreachable
    continue
}
```

Because `splitEnginePrefix` and `stripEnginePrefix` are identical in what they return for the bare name (`stripEnginePrefix` is literally `_, bare := splitEnginePrefix(s); return bare`), and the pre-check has already guaranteed no entry produces an empty bare name, the `priorTool == ""` branch at line 1183 can never be true.

The code is not wrong — it just contains a guard that can never fire. The risk is that future maintainers either:
- Over-rely on the guard and remove the pre-check (creating a fail-open path where an all-empty `afterTools` list silently continues without recording), or
- Misread the dead branch as evidence that the pre-check is optional and strip it during refactoring.

## Suggested Fix

Remove lines 1182–1184 (`priorTool = strings.TrimSpace(priorTool)` + `if priorTool == "" { continue }`) since the pre-check already enforces the invariant. Add a brief comment on the pre-check explaining that it is load-bearing for the main loop:

```go
// Pre-check: reject any entry that names nothing after prefix stripping.
// This guarantees priorTool is non-empty in the loop below (fail-closed:
// an all-empty list must never silently allow the call).
```

## Session

- Session ID: `rev-20260615T0210Z-d6ac44`
- Reviewed commit: `affe5a66a16e39d9a6d0bd46391dac6ed24681be`
- Modules reviewed: `pkg/circuitbreaker`, `pkg/callcounter`, `cmd/eunox-mcp/audit.go`

eunolabs:
## Bugs addressed (session rev-20260615T0210Z-d6ac44)

- Closes #494 — Removed the unreachable `if priorTool == "" { continue }` branch in `handleSequenceBlock`: the load-bearing pre-check already denies any `afterTools` entry whose bare name is empty after the identical strip+trim, so the guard could never fire. Annotated the pre-check as load-bearing for the loop. **Note:** the issue's suggested fix also removed `priorTool = strings.TrimSpace(priorTool)`, but that line is *not* dead — it normalizes manifest-padded entries (e.g. `"export "`) so they match the whitespace-free recorded target name. It is retained.
- Closes #493 — `newestRotatedSibling` and `pruneRotated` built a `filepath.Glob` pattern by concatenating the raw `logPath` with `".*"`, so a `logPath` containing glob metacharacters (`[`, `]`, `*`, `?`) was interpreted as a pattern — matching wrong files (corrupting chain resume / deleting unrelated files) or hitting `ErrBadPattern` (silent no-op). Both now use a new `rotatedSiblings` helper that scans the directory with `os.ReadDir` + a literal `strings.HasPrefix`, reconstructing each path as `logPath + <suffix>` so downstream prefix-stripping (`rotatedAuditRe`, `rotatedStampParts`) is unchanged.

## Testing
- `go test -count=1 ./pkg/... ./cmd/...` — all pass.
- Added `TestNewestRotatedSiblingHandlesGlobMetacharacters` covering a `logPath` with `[`, `]`, `?` plus a decoy directory the glob expansion would have matched.

---


eunolabs:
✅ Ready to merge — all session rev-20260615T0210Z-d6ac44 bugs addressed (#494, #493), self-review complete with no outstanding comments, tests passing (`go test ./pkg/... ./cmd/...`, fmt/vet clean).

---

PR #504, opened 2026-06-15
Fix session rev-20260615T0140Z-d1d341 bugs

### LINKED ISSUE #489: [rev-20260615T0140Z-d1d341] Redis killswitch ReviveAgent/ReviveSession accept empty IDs, triggering spurious full-refresh on all replicas
**Location:** `pkg/killswitch/redis.go`, `ReviveAgent` and `ReviveSession` methods (~lines 266–302)

**Reference:** SHA `2b748cb9`; contrast with `KillAgent` (~line 247) and `KillSession` (~line 279)

**Severity:** Low

**Description:**

`KillAgent` and `KillSession` in the Redis backend correctly reject an empty ID with an error:

```go
// KillAgent
if agentID == "" {
    return fmt.Errorf("killswitch: KillAgent: agentID must not be empty")
}
```

The corresponding `ReviveAgent` and `ReviveSession` have no such guard:

```go
// ReviveAgent — missing empty-ID check
func (r *Redis) ReviveAgent(ctx context.Context, agentID string) error {
    if err := r.client.Del(ctx, redisAgentPrefix+agentID).Err(); err != nil {
        return err
    }
    r.publish(ctx, "agent:revive:"+agentID)
    ...
}
```

Calling `ReviveAgent("")` with an empty ID:

1. **Deletes the bare prefix key** — attempts `DEL killswitch:agent:` which is a no-op (the key doesn't exist as a kill entry), but pollutes Redis command history and logs.
2. **Publishes a malformed pub/sub message** — sends `"agent:revive:"` (empty ID suffix) to the `killswitch:events` channel.
3. **Every subscribed replica's `handlePubSubMessage` triggers a full refresh** — the empty-suffix message does not match any of the fast-path handlers (`len(payload) > len("agent:revive:")` is false), so it falls into the `default` branch and schedules a full `refreshState` round-trip for each replica.
4. **Returns nil** — the caller receives no indication that the argument was invalid.

The same applies to `ReviveSession("")`.

**Impact:**

- In a multi-replica deployment, each `ReviveAgent("")` call causes every replica to make two Redis round-trips (GET global + SCAN agents + SCAN sessions). Under a misconfigured or buggy caller that invokes this repeatedly, this can generate a traffic spike proportional to the fleet size.
- Monitoring systems that alert on kill-switch reconcile activity may fire spuriously.
- The `InMemory` backend correctly silences the no-op (it just `delete(m.killedAgents, "")` and returns nil without any broadcast), so the behavior diverges across backends. Code that works fine with `InMemory` in tests can be surprising in production against Redis.

**Suggested fix:**

Add the same empty-ID guard to both `ReviveAgent` and `ReviveSession`:

```go
func (r *Redis) ReviveAgent(ctx context.Context, agentID string) error {
    if agentID == "" {
        return fmt.Errorf("killswitch: ReviveAgent: agentID must not be empty")
    }
    ...
}

func (r *Redis) ReviveSession(ctx context.Context, sessionID string) error {
    if sessionID == "" {
        return fmt.Errorf("killswitch: ReviveSession: sessionID must not be empty")
    }
    ...
}
```

The `InMemory` backend should receive the matching guard for consistency (no broadcast there, but the error keeps call-site contracts uniform).

**Session:** rev-20260615T0140Z-d1d341

eunolabs:
## Bugs addressed (session rev-20260615T0140Z-d1d341)

- Closes #489 — `Redis.ReviveAgent`/`ReviveSession` lacked the empty-ID guard that `KillAgent`/`KillSession` already have. Calling `ReviveAgent("")` would `DEL` the bare `killswitch:agent:` key and publish an empty-suffix `agent:revive:` event, which falls into the pub/sub `default` branch and forces a full `refreshState` round-trip on every subscribed replica (a fleet-sized traffic spike), while returning `nil` to the caller. Added the same guard to both Redis methods, and to the `InMemory` `ReviveAgent`/`ReviveSession` so the call-site contract is uniform across backends.

## Testing
- `go test -race -count=1 ./pkg/killswitch/` — all pass.
- Added `TestRedis_Revive_EmptyIDRejected` and `TestInMemory_ReviveEmptyIDRejected`.

---


eunolabs:
✅ Ready to merge — session rev-20260615T0140Z-d1d341 bug addressed (#489), self-review complete with no outstanding comments, tests passing (`go test -race ./pkg/killswitch/`, fmt/vet clean).

---

PR #509, opened 2026-06-15
Fix session rev-20260615T0121Z-12a936 bugs

### LINKED ISSUE #485: [rev-20260615T0121Z-12a936] circuitbreaker.Do records request-context cancellation as an upstream failure, falsely tripping the JWKS circuit
**Location:** `pkg/circuitbreaker/do.go:50-58`

**Reference:** https://github.com/eunolabs/eunox/blob/affe5a66a16e39d9a6d0bd46391dac6ed24681be/pkg/circuitbreaker/do.go#L50-L58

**Severity:** Medium

**Description:**

`Do` records any non-nil error from `fn` as a circuit failure:

```go
result, err := fn(ctx)
completed = true
if err != nil {
    b.RecordFailure()   // ← fires on context.Canceled and context.DeadlineExceeded too
    return zero, err
}
b.RecordSuccess()
return result, nil
```

`fn` is passed the caller's `ctx` directly. When a request's context is canceled or times out — for reasons unrelated to the upstream JWKS endpoint (e.g., the MCP host dropped the connection, or a per-request deadline fired before the fetch completed) — `fn(ctx)` returns `context.Canceled` or `context.DeadlineExceeded`. `Do` records this as a circuit failure.

In production, `circuitbreaker.Do` wraps JWKS endpoint fetches (see `NewJWTPDP` → `capability.NewJWKSCache`). If `FailureThreshold` (default 5) request-level context cancellations occur in succession — regardless of whether the JWKS endpoint itself is healthy — the circuit opens. During the 30 s cooldown (`CooldownDuration`) **no JWT can be validated** even if the JWKS endpoint is fully reachable, because `ValidateToken` → `cache.GetKeys` → `Do` returns `ErrOpen` and `ValidateToken` returns HTTP 401 for every incoming request.

This is distinct from issue #472 (stale probes from prior half-open windows): the bug here is that *closed-state failures* are over-counted, not that *half-open probes* race.

**Trigger / repro:**

1. Configure `--jwks-uri` with a JWKS endpoint that responds in ~2 s.
2. Set a per-request deadline of 1 s on the MCP host side (shorter than the JWKS fetch time).
3. Send 5 requests in rapid succession. Each request's context expires while `getKeys` is in-flight.
4. Each `Do` call records a failure; after 5 the circuit opens.
5. Subsequent requests see `ErrOpen` (HTTP 401) even though the JWKS endpoint is healthy.

**Suggested fix:**

Use a context that is *independent of the request context* for the JWKS fetch, so the circuit breaker only measures genuine JWKS endpoint health. The simplest approach is to derive a fresh background context with a dedicated JWKS fetch timeout inside `JWKSCache.GetKeys`:

```go
fetchCtx, cancel := context.WithTimeout(context.Background(), c.fetchTimeout)
defer cancel()
// pass fetchCtx (not the request ctx) to the HTTP round-trip and to Do
```

Alternatively, make `Do` accept an `errors.Is`-based exclusion set so callers can opt out of counting context errors as failures:

```go
if err != nil {
    if !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded) {
        b.RecordFailure()
    }
    return zero, err
}
```

Either approach breaks the false coupling between request-level deadlines and JWKS-endpoint health tracking.

**Session:** rev-20260615T0121Z-12a936

### LINKED ISSUE #486: [rev-20260615T0121Z-12a936] wrapRoutesWithJWT allocates and immediately discards a JWKSCache+Breaker per route
## Location

`cmd/eunox-mcp/route.go`, function `wrapRoutesWithJWT`, lines ~253–263

## Reference

```go
func wrapRoutesWithJWT(routes map[string]*upstreamRoute, opts JWTPDPOptions) *JWTPDP {
    validator := NewJWTPDP(opts)
    for _, rt := range routes {
        w := NewJWTPDP(JWTPDPOptions{   // allocates JWKSCache + circuit Breaker
            Issuer:     opts.Issuer,
            Audience:   opts.Audience,
            Inner:      rt.pdp,
            KillSwitch: opts.KillSwitch,
            // JWKSURI, CacheTTL, Client, Breaker intentionally omitted
        })
        w.cache = validator.cache  // immediately overwrites the allocated cache
        rt.pdp = w
    }
    return validator
}
```

## Severity

Low — no correctness impact at runtime (all routes end up sharing the correct cache), but the per-route `NewJWTPDP` call allocates a full `JWKSCache` (with its own HTTP client and circuit `Breaker`) that is thrown away on the very next line. In a gateway config with many routes this is proportional wasted allocation at startup.

## Description

`NewJWTPDP` constructs a `JWKSCache` initializing an `http.Client`, a `circuitbreaker.Breaker`, and related fields. Inside `wrapRoutesWithJWT` a fresh `JWTPDP` is created per route solely to set `Inner` and `KillSwitch` — the JWKS infrastructure it builds is never used because `w.cache = validator.cache` replaces it immediately afterward.

The intent is correct (all routes share a single JWKS cache so they all observe the same circuit state), but the implementation wastes the per-route allocation and makes the code read as if each route has an independent cache until the override line is noticed.

## Trigger

Any gateway config with `jwt:` block and two or more routes. N routes → N−1 wasted `JWKSCache`+`Breaker` allocations at startup.

## Suggested Fix

Construct per-route `JWTPDP` instances without triggering cache initialization, then assign the shared cache. The simplest approach is to add an unexported `newJWTPDPWithCache(opts, cache)` constructor (or accept a `cache` field in `JWTPDPOptions`) so the loop can build the per-route wrapper in a single step without the throwaway allocation:

```go
func wrapRoutesWithJWT(routes map[string]*upstreamRoute, opts JWTPDPOptions) *JWTPDP {
    validator := NewJWTPDP(opts)
    for _, rt := range routes {
        rt.pdp = newJWTPDPWithCache(JWTPDPOptions{
            Issuer:     opts.Issuer,
            Audience:   opts.Audience,
            Inner:      rt.pdp,
            KillSwitch: opts.KillSwitch,
        }, validator.cache)
    }
    return validator
}
```

## Session

rev-20260615T0121Z-12a936

### LINKED ISSUE #487: [rev-20260615T0121Z-12a936] manifestDigest silently returns empty string on marshal error, corrupting audit records
## Location

`cmd/eunox-mcp/route.go`, function `manifestDigest`, lines ~309–316

## Reference

```go
func manifestDigest(m *LocalManifest) string {
    b, err := json.Marshal(m)
    if err != nil {
        return ""   // silent failure — no error propagation
    }
    sum := sha256.Sum256(b)
    return "sha256:" + hex.EncodeToString(sum[:])
}
```

Called at route build time:

```go
r.policySHA256 = manifestDigest(manifest)
```

## Severity

Medium — silent data corruption in audit records. An empty `policy_sha256` makes it impossible for an auditor to verify which policy version was in effect, undermining the audit trail's forensic value. The invariant "audit records are structured and meaningful" is violated without any log or error surfacing the underlying marshal failure.

## Description

`manifestDigest` swallows `json.Marshal` errors and returns `""`. That empty string propagates to `upstreamRoute.policySHA256` and is then stamped into every audit record written by `routeSink` (via `s.r.policySHA256`) for the lifetime of the gateway process.

`json.Marshal` on a `*LocalManifest` should almost never fail in practice — the struct contains only plain Go types. However, the silent failure violates two invariants:

1. **Audit records must be structured and meaningful.** An empty `policy_sha256` is indistinguishable from "this field was never set" and cannot be detected as corrupt by downstream consumers.
2. **Fail-closed.** A marshal error is ambiguous: we don't know whether the digest is wrong, so we should refuse to start rather than running with a silent empty field.

Because this function is called during route construction (before any traffic flows), an error here almost certainly reflects a programming mistake or memory corruption — exactly the kind of condition that warrants a hard failure rather than a graceful degradation.

## Trigger

Any `json.Marshal` error on `*LocalManifest` during gateway startup. In practice this requires a type in the struct that json cannot marshal (e.g. a `chan`, `func`, or circular pointer added by a future refactor), but the silent failure makes it undetectable when it does occur.

## Suggested Fix

Return an error from `manifestDigest` and propagate it to `BuildRoutes`, causing startup to fail loudly:

```go
func manifestDigest(m *LocalManifest) (string, error) {
    b, err := json.Marshal(m)
    if err != nil {
        return "", fmt.Errorf("manifestDigest: marshal failed: %w", err)
    }
    sum := sha256.Sum256(b)
    return "sha256:" + hex.EncodeToString(sum[:]), nil
}
```

Call site in `BuildRoutes`:

```go
digest, err := manifestDigest(manifest)
if err != nil {
    return nil, err
}
r.policySHA256 = digest
```

## Session

rev-20260615T0121Z-12a936

eunolabs:
## Bugs addressed (session rev-20260615T0121Z-12a936)

- Closes #487 — `manifestDigest` swallowed `json.Marshal` errors and returned `""`, which then got stamped into every audit record's `policy_sha256` for the gateway's lifetime (silent audit corruption + fail-open). It now returns `(string, error)`; `loadUpstreamPDP` propagates the error so startup fails loudly, and `doctor` prints `<error: ...>` instead of a blank digest.
- Closes #486 — `wrapRoutesWithJWT` called `NewJWTPDP` per route only to overwrite the freshly-allocated `JWKSCache` (+ `http.Client` + circuit `Breaker`) on the next line, wasting N−1 allocations at startup. Added an unexported `newJWTPDPWithCache(opts, cache)` constructor that builds the per-route wrapper sharing the validator's cache without the throwaway allocation.
- Closes #485 — **Already resolved in the production path; pinned with a regression test.** The breaker's only production wiring (JWKS fetches) already runs on a context decoupled from the request via `context.WithoutCancel` in `JWKSCache.refresh` (issue #391/#466), so a request-level cancellation or deadline cannot trip the JWKS circuit. Added `TestJWKSCache_RequestDeadlineDoesNotTripBreaker`, which reproduces #485's exact scenario (a per-request deadline shorter than the fetch) at the `JWKSCache` layer and asserts the breaker stays closed with zero failures.

  I **deliberately did not** take the issue's secondary suggestion of changing `circuitbreaker.Do` to ignore context errors. `Do` intentionally records a deadline error as a failure (`TestDo_UpstreamHangs_DeadlineRecordedAsFailure`, `TestDo_UpstreamHangs_TripsBreakerAfterThreshold`): a hung upstream that exhausts its own per-call deadline *is* an unhealthy dependency that must trip the breaker, and `Do` cannot distinguish that from a request-level deadline. Decoupling the fetch context (the issue's *primary* suggestion) is the correct layer and is already in place. Skipping the record in `Do` would also leave a consumed half-open probe unaccounted for and could wedge the window.

## Testing
- `go test -count=1 ./pkg/... ./cmd/...` and `go test -race ./pkg/capability/` — all pass.
- Added `TestJWKSCache_RequestDeadlineDoesNotTripBreaker`.

---


eunolabs:
✅ Ready to merge — all session rev-20260615T0121Z-12a936 bugs addressed (#487, #486, #485), self-review complete with no outstanding comments, tests passing (`go test ./pkg/... ./cmd/...`, race-clean capability tests, fmt/vet clean). Note: #485 was already resolved in the production path by the existing `context.WithoutCancel` decoupling; this PR pins it with a regression test and documents why modifying `circuitbreaker.Do` would be incorrect.

---

Issue #477, opened 2026-06-15
[rev-20260615T0110Z-758862] verifyAuditLog: forged seq-0 decoy record successfully suppresses SEQ GAP diagnostic message


eunolabs:
**Location:** `cmd/eunox-mcp/audit.go:1137–1179` (the `unmarshalOK` chain-verification block inside `verifyAuditLog`)

**Reference:** https://github.com/eunolabs/eunox/blob/affe5a66a16e39d9a6d0bd46391dac6ed24681be/cmd/eunox-mcp/audit.go#L1137-L1179

**Severity:** Medium (forensic-clarity / security observability)

**Description:**

`verifyAuditLog` checks all decodable records for chain linkage unconditionally — before any of the structural-validity filters (forged seq-0, legacy, etc.) — and then rejects the forged seq-0 record with `continue`. The problem is that the chain-state update runs **before** the rejection:

```go
if unmarshalOK {
    // Chain verification (updates havePrev, prevHMAC, prevSeq, signedSeen)
    ...
    havePrev = true
    prevHMAC = rec.HMAC   // ← set to the forged record's HMAC
    prevSeq  = rec.Seq    // ← SET TO 0
    ...
}

// Only AFTER the chain state is updated does this fire:
if rec.Seq == 0 && rec.HMAC != "" {
    res.invalid++
    _, _ = fmt.Fprintf(out, "INVALID  record %d: seq 0 with a non-empty HMAC...\n", res.total)
    continue
}
```

The comment on that guard says the record is "used to suppress the seq-gap check for a following record" — and indeed that suppression succeeds. When a forged seq-0 record is processed, `prevSeq` is set to 0. The seq-gap check for the **immediately following** real record then tests `prevSeq > 0`, which evaluates to `false`, so the "SEQ GAP" diagnostic is never emitted for that record.

**Scenario:**

An attacker deletes real record seq 4 and inserts a forged seq-0 (HMAC != "") in its place:

| Record | prevHMAC after | prevSeq after | Diagnostics |
|--------|----------------|---------------|-------------|
| seq 3 (real) | H3 | 3 | — |
| forged seq-0 | H_forged | **0** | INVALID (correctly caught) |
| seq 5 (real) | H5 | 5 | CHAIN BREAK (prev_hmac mismatch) but **no SEQ GAP message** |

[...]

PR #510, opened 2026-06-15
Fix session rev-20260615T0122Z-07e355 bugs

### LINKED ISSUE #484: [rev-20260615T0122Z-07e355] verifyAuditLog updates chain state before forged-record detection, enabling tamper-evident chain bypass with signing-key access
**Location:** `cmd/eunox-mcp/audit.go`, `verifyAuditLog`, lines ~1137–1233

**Reference:** SHA `9bf298c4`; chain-verification logic in the `for scanner.Scan()` loop

**Severity:** High

**Description:**

`verifyAuditLog` maintains the chain head (`prevHMAC`, `prevSeq`, `signedSeen`) inside the `if unmarshalOK` block, which runs unconditionally for every decodable record. Two validity checks that classify certain records as invalid and `continue` past them come **after** the chain state update:

```go
if unmarshalOK {
    // ...check genesis / prev_hmac / seq gap...
    havePrev = true
    prevHMAC = rec.HMAC   // <-- chain state updated unconditionally
    prevSeq  = rec.Seq    // <-- (for forged seq-0 this sets prevSeq = 0)
    if rec.Seq > 0 { signedSeen = true }
}
// ...
// These checks come AFTER chain state was already updated:
if rec.Seq == 0 && rec.HMAC != "" {   // forged signed seq-0
    res.invalid++; continue
}
if rec.HMAC == "" && rec.Seq == 0 {   // unsigned seq-0 after signed record
    if signedSeen { res.invalid++; continue }
    ...
}
```

A decodable-but-invalid record (either type) therefore mutates `prevHMAC` and `prevSeq` to attacker-controlled values before the `continue` skips it.

**Impact — without signing key:**

Any actor who can append to the audit log can inject a forged seq-0 record with an arbitrary `_hmac` field (no key required for the injected value itself). `prevHMAC` is set to the injected value. The next legitimate record's `prev_hmac` no longer matches `prevHMAC`, producing a **false-positive chain break** that obscures where actual tampering occurred. This muddies forensic analysis even when the audit key is fully protected.

**Impact — with signing key:**

An attacker with signing-key access can conceal the deletion of an arbitrary audit record using only O(1) work, bypassing the chain guarantee that was supposed to require re-signing O(N) subsequent records:

1. Delete record seq=N.
2. Insert a forged seq-0 record (signed with the key) whose `_hmac` is some chosen value X. This record is detected and counted as `invalid`, but it has already set `prevHMAC = X` and `prevSeq = 0`.
3. Craft a replacement record at seq=N+1 with `prev_hmac = X` and a valid HMAC (signed with the key).
4. In `verifyAuditLog`:
   - The forged seq-0 is flagged `invalid` (1 count) — no chain break fires against it because it is the first decodable record in the window after seq=N-1 (which set `havePrev = true`, `prevHMAC = H(N-1)`). Wait — actually the forged seq-0 IS checked against `prevHMAC`... let me be precise.

   More precisely, to avoid a spurious chain break on the forged seq-0 itself, the attacker positions it so that `prevHMAC` at that point equals the forged seq-0's `prev_hmac` field — e.g., by making the forged seq-0's `prev_hmac` equal to `H(N-1)` (attacker knows this). The chain check for the forged seq-0 then passes (or is the first-record genesis check which only fires for seq == 1). After the forged seq-0, `prevHMAC = X` and `prevSeq = 0`.
   
   Then for the crafted seq=N+1 record:
   - `prev_hmac == X` → matches `prevHMAC` → **no chain break**
   - Seq gap check: `rec.Seq != prevSeq+1` → `N+1 != 1`, BUT the guard `prevSeq > 0` is **false** (prevSeq=0 after the forged seq-0) → **no seq gap fired**
   - HMAC: valid (attacker signed it) → `res.valid++`
   
   Net result: 1 invalid (the decoy), 0 chain breaks, deletion of seq=N completely hidden.

**Root cause:**

The two invalid-record detections should run before the chain state update, or the chain state update should be skipped for records classified invalid. The analogous path for malformed-JSON records correctly avoids this: `unmarshalOK` is false, so the `if unmarshalOK` block is entirely skipped, leaving chain state untouched.

**Suggested fix:**

Move the seq-0 validity checks inside the `if unmarshalOK` block, before the chain state update, and skip both the chain state update and the chain-check reporting for classified-invalid records:

```go
if unmarshalOK {
    // Classify before touching chain state.
    isForgedSignedSeq0  := rec.Seq == 0 && rec.HMAC != ""
    isForgedLegacySeq0  := rec.HMAC == "" && rec.Seq == 0 && signedSeen

    if !isForgedSignedSeq0 && !isForgedLegacySeq0 {
        // Only run chain checks and update chain state for legitimate records.
        if !havePrev { ... genesis check ... }
        else { ... prev_hmac and seq gap checks ... }
        havePrev = true
        prevHMAC = rec.HMAC
        prevSeq  = rec.Seq
        if rec.Seq > 0 { signedSeen = true }
    }
}

if !unmarshalOK {
    res.invalid++; ...; continue
}
if rec.Seq == 0 && rec.HMAC != "" {
    res.invalid++; ...; continue
}
    if signedSeen { res.invalid++; ...; continue }
    res.legacy++; continue
}
```

**Session:** rev-20260615T0122Z-07e355

eunolabs:
## Bugs addressed (session rev-20260615T0122Z-07e355)

- Closes #484 — `verifyAuditLog` updated the chain head (`prevHMAC`/`prevSeq`) for a decodable-but-invalid record before classifying it. The forged **signed** seq-0 decoy case was already fixed (#477: excluded via `!forgedSeq0`), but the symmetric forged **legacy** case — an HMAC-less `seq 0` record appearing *after* a signed record (`HMAC=="" && Seq==0 && signedSeen`) — still entered the chain block and overwrote `prevHMAC=""`, `prevSeq=0` *before* being rejected as invalid. That let an attacker who deletes an interior signed record splice in such a record to hide the deletion: the next crafted record could link with `prev_hmac=""` (no `CHAIN BREAK`) and the `SEQ GAP` check was suppressed because its `prevSeq > 0` guard was false (O(1) concealment even for a signing-key holder). 

  Fix: classify `forgedLegacySeq0` alongside `forgedSeq0` and exclude it from the chain-state update (`if unmarshalOK && !forgedSeq0 && !forgedLegacySeq0`), so `prevHMAC`/`prevSeq` stay anchored to the last legitimate record and the next real record still trips both diagnostics. A genuine head-legacy record (before any signed record, `signedSeen == false`) is **not** excluded, so the legacy→first-signed transition still verifies (#427).

## Testing
- Added `TestVerifyAuditLog_ForgedLegacySeq0DoesNotSuppressSeqGap`, which deletes an interior signed record, splices a stealthy HMAC-less seq-0 decoy (its `prev_hmac` linked to the surviving predecessor), and asserts the decoy is reported invalid **and** the `SEQ GAP: record seq 5 does not follow 3` diagnostic still fires. Verified the test **fails without the fix** (gap suppressed) and passes with it.
- `go test -count=1 ./pkg/... ./cmd/...` — all pass; existing `#446`/`#452`/`#477` seq-0 tests remain green.
- Updated `docs/threat-model-mcp.md` §3.4 changelog (v1.23).

---


eunolabs:
Self-review round 1 found one issue: the new threat-model changelog row (v1.23) was inserted before v1.22, breaking the table's ascending version order. Fixed in efe88d7 (moved 1.23 after 1.22). Re-reviewed the diff top-to-bottom: the core fix correctly excludes `forgedLegacySeq0` from the chain-state update (a `forgedLegacySeq0` record always has `signedSeen == true`, hence `havePrev == true`, so the `firstSeq`/genesis path is unaffected), and the regression test was verified to fail without the fix.

✅ Ready to merge — session rev-20260615T0122Z-07e355 bug addressed (#484), self-review complete with no outstanding comments, tests passing (`go test ./pkg/... ./cmd/...`, fmt/vet clean).

---

PR #511, opened 2026-06-15
Fix session rev-20260615T0310Z-bde0af bugs

### LINKED ISSUE #505: [rev-20260615T0310Z-bde0af] JWTPDP silently accepts tokens from any issuer when --jwt-issuer is unset
**Location:** `cmd/eunox-mcp/pdp_jwt.go`, `NewJWTPDP` (~line 213) and `ValidateToken` (~line 344)

**Reference:** https://github.com/eunolabs/eunox/blob/f65d1147ece58e5d1659a871485d66cb4cae0d1f/cmd/eunox-mcp/pdp_jwt.go#L344

**Severity:** Medium (security / fail-open risk)

**Description:**

`NewJWTPDP` warns on an empty `Audience` (when `AllowAnyAudience` is not set), surfacing the misconfiguration to the operator. No equivalent warning exists for an empty `Issuer`. Inside `ValidateToken`, the issuer check is guarded:

```go
if p.issuer != "" && stdClaims.Issuer != p.issuer {
    return nil, fmt.Errorf("token issuer %q does not match expected %q", ...)
}
```

When `--jwt-issuer` is omitted (or set to `""`), `p.issuer` is the empty string and the condition short-circuits — the issuer field in every incoming token is accepted without validation. An operator who configures a `jwks-uri` but omits `--jwt-issuer` believes they have a pinned-issuer deployment but is in fact accepting tokens from **any** issuer whose JWKS is reachable at the configured URI, or whose key happens to verify against the cached key set.

The fail-closed invariant says: "on any ambiguity … deny". An unconfigured issuer is ambiguous; the current code silently widens the trust boundary instead.

**Trigger / Repro:**

1. Start `eunox-mcp` with `--jwks-uri <uri>` and no `--jwt-issuer` flag.
2. Obtain a token signed by any key present in the JWKS response (including tokens issued by a different service sharing the same JWKS endpoint).
3. The proxy accepts the token without issuer validation; no warning appears in logs or startup output.

**Suggested Fix:**

Add a startup warning (matching the existing audience-empty warning) and consider defaulting to deny when issuer is empty:

```go
// In NewJWTPDP, alongside the audience check:
if opts.Issuer == "" {
    logger.Warn("jwt-issuer is not set; tokens from any issuer whose key is in the JWKS will be accepted — set --jwt-issuer to pin the expected issuer")
}
```

For a stricter change (fail-closed), reject the config entirely if `Issuer == ""` and no explicit `AllowAnyIssuer` flag is set, paralleling how audience is handled. At minimum, document the behavior explicitly in the `--help` output and the gateway config schema.

**Session:** rev-20260615T0310Z-bde0af
**Reviewed commit:** f65d1147ece58e5d1659a871485d66cb4cae0d1f
**Modules reviewed:** `cmd/eunox-mcp/pdp_jwt.go`, `pkg/circuitbreaker/breaker.go`, `pkg/callcounter/redis.go`, `pkg/capability/token_cache.go`, `pkg/capability/jwkscache.go`

### LINKED ISSUE #506: [rev-20260615T0310Z-bde0af] O(T×C) JWT capability claim re-parsing per list-filter call in JWTPDP
**Location:** `cmd/eunox-mcp/pdp_jwt.go`, `buildConstraintsFromClaims` called from `filterToolsListByClaims`, `filterResourcesListByClaims`, `filterPromptsListByClaims` (~lines 696–713, 803–825)

**Reference:** https://github.com/eunolabs/eunox/blob/f65d1147ece58e5d1659a871485d66cb4cae0d1f/cmd/eunox-mcp/pdp_jwt.go

**Severity:** Medium (performance)

**Description:**

`buildConstraintsFromClaims(caps []string, target EnforceTarget)` iterates over all C raw capability claim strings and parses each one via `parseV2Claim()` (which calls `strings.Cut`, `capability.ParseTarget`, and optionally `parseCondSuffix`) on every invocation. This function is called once **per list item** inside the three list-filter functions:

```go
// filterToolsListByClaims — simplified:
for _, tool := range tools {
    constraints := buildConstraintsFromClaims(claims.Capabilities, toolTarget(tool.Name))
    // ^ re-parses all C claims for every tool
    if evaluate(constraints) { permitted = append(permitted, tool) }
}
```

For a list response with T items and a JWT carrying C capability claims, the total cost is **O(T × C)** string-parse operations per list call. The raw `[]string` capability claims stored in `JWTClaims.Capabilities` are already validated by `ValidateToken` but never pre-parsed into `[]capability.Constraint`.

In a gateway serving a manifest with many tools/resources, a single `tools/list` or `resources/list` call parses the same C claim strings T times. For example, 50 tools × 20 capability claims = 1,000 redundant parse calls, all inside the list handler's critical path.

**Trigger / Repro:**

1. Issue a JWT with 20+ `mcp_capability_v2` claims.
2. Proxy an upstream that advertises 50+ tools.
3. Call `tools/list` through the proxy. `filterToolsListByClaims` is called once; inside it `buildConstraintsFromClaims` is called 50 times, re-parsing all 20 claims each time.
4. Profile shows string-parsing time scaling with T×C, not O(T+C).

**Suggested Fix:**

Parse claims once into a `[]capability.Constraint` slice after successful token validation and cache the result on `JWTClaims` (or in a thin wrapper struct returned by `ValidateToken`):

```go
type ParsedClaims struct {
    JWTClaims
    Constraints []capability.Constraint  // parsed once from Capabilities
}
```

Then list filters receive `ParsedClaims` and call `buildConstraintsFromClaims` zero times per list item — they filter against the pre-parsed slice directly, reducing cost to O(T + C).

Alternatively, change `buildConstraintsFromClaims` to accept a `target` and build a target-specific allow-set from the pre-parsed constraints in O(C), then do the T-item filter pass in O(T), for O(T + C) total.

**Session:** rev-20260615T0310Z-bde0af
**Reviewed commit:** f65d1147ece58e5d1659a871485d66cb4cae0d1f
**Modules reviewed:** `cmd/eunox-mcp/pdp_jwt.go`, `pkg/circuitbreaker/breaker.go`, `pkg/callcounter/redis.go`, `pkg/capability/token_cache.go`, `pkg/capability/jwkscache.go`

### LINKED ISSUE #507: [rev-20260615T0310Z-bde0af] TokenCache.purgeExpired holds write lock for full O(N) scan, blocking concurrent Get() calls
**Location:** `pkg/capability/token_cache.go`, `purgeExpired()`, lines 266–278

**Reference:** https://github.com/eunolabs/eunox/blob/f65d1147ece58e5d1659a871485d66cb4cae0d1f/pkg/capability/token_cache.go#L266-L278

**Severity:** Medium (performance / latency spike on token-validation hot path)

**Description:**

`purgeExpired()` acquires the exclusive write lock (`c.mu.Lock()`) and holds it for the entire O(N) scan-and-delete loop over all cache entries:

```go
func (c *TokenCache) purgeExpired() {
    now := c.cfg.Now()
    c.mu.Lock()          // exclusive write lock held for entire loop
    defer c.mu.Unlock()
    for key, entry := range c.entries {
            delete(c.entries, key)
            if entry.listElem != nil {
                c.insertOrder.Remove(entry.listElem)
            }
        }
    }
}
```

Any concurrent `Get()` call needs `c.mu.RLock()`, which is blocked for the full duration of the sweep. With the default configuration (`MaxSize = 4096`, `CleanupInterval = 60s`), every 60 seconds a cleanup run can block all token-validation reads for the time it takes to scan up to 4,096 entries — a periodic latency spike on the enforcement hot path.

By contrast, `pkg/callcounter/memory.go`'s `Cleanup()` already uses a two-phase approach: scan under `RLock` to identify stale keys (allowing concurrent reads to proceed), then acquire `Lock` only for the short batch-delete phase. The token cache should match this pattern.

**Trigger / Repro:**

1. Configure a `TokenCache` with `MaxSize = 4096` and `CleanupInterval = 60s` (defaults).
2. Fill the cache to capacity.
3. Every 60 seconds, `cleanupLoop` calls `purgeExpired()`. During that call, all concurrent `Get()` goroutines block on `c.mu.RLock()`.
4. Under load (many concurrent JWT validations), the 60-second cleanup tick causes a measurable latency spike in token verification.

**Suggested Fix:**

Apply the same two-phase pattern used in `pkg/callcounter/memory.go`:

```go
func (c *TokenCache) purgeExpired() {
    now := c.cfg.Now()

    // Phase 1: identify expired keys under read lock (allows concurrent Get).
    c.mu.RLock()
    var toDelete []string
    for key, entry := range c.entries {
        }
    }
    c.mu.RUnlock()

    if len(toDelete) == 0 {
        return
    }

    // Phase 2: delete under write lock (short critical section).
    c.mu.Lock()
    defer c.mu.Unlock()
    for _, key := range toDelete {
        if entry, ok := c.entries[key]; ok && now.After(entry.expiresAt) {
            delete(c.entries, key)
            if entry.listElem != nil {
                c.insertOrder.Remove(entry.listElem)
            }
        }
    }
}
```

The re-check `now.After(entry.expiresAt)` in phase 2 guards against a `Put()` that refreshed the entry between the two phases. This bounds the write-lock hold time to the number of expired entries rather than the total cache size, keeping `Get()` available during the scan phase.

**Session:** rev-20260615T0310Z-bde0af
**Reviewed commit:** f65d1147ece58e5d1659a871485d66cb4cae0d1f
**Modules reviewed:** `cmd/eunox-mcp/pdp_jwt.go`, `pkg/circuitbreaker/breaker.go`, `pkg/callcounter/redis.go`, `pkg/capability/token_cache.go`, `pkg/capability/jwkscache.go`

eunolabs:
## Bugs addressed (session rev-20260615T0310Z-bde0af)

- Closes #507 — `TokenCache.purgeExpired` held the exclusive write lock for the entire O(N) scan-and-delete, blocking all concurrent `Get()` (RLock) calls on the token-validation hot path on every cleanup tick. Rewritten as the two-phase pattern already used by `pkg/callcounter`'s `Cleanup`: scan under `RLock` to collect expired keys, then delete under `Lock` with a per-entry expiry re-check (so a `Put()` that refreshed an entry between phases — the #497 `MoveToBack` path — is not dropped).
- Closes #506 — The three `filter*ListByClaims` functions called `buildConstraintsFromClaims` once per list item, re-parsing all C capability claim strings for each of T items (O(T×C) parses per list call). Added `parseCapClaims` (parse once) + `anyCapCovers` (cheap namespace/bare-name match); the filters now parse the claim set once and compare per item, giving O(C) parses + O(T×C) comparisons. Behavior is identical (conditions are not consulted for `*/list`, matching the prior `len(...) > 0` check). `buildConstraintsFromClaims` is unchanged for the `Decide` path.
- Closes #505 — `NewJWTPDP` warned on an empty `Audience` but not an empty `Issuer`, so an operator who configured a JWKS URI but omitted `--jwt-issuer` silently accepted tokens from any issuer whose key verifies (the `ValidateToken` issuer check short-circuits when `p.issuer == ""`). Added a matching startup warning. (The stricter fail-closed-by-default option from the issue would require a new `--jwt-allow-any-issuer` opt-out flag + schema/docs and would break intentional no-pin deployments, so it's left as a separate design change; the warning surfaces the misconfiguration without a breaking change. Issuer *validation* behavior is already covered by `TestJWTPDP_ValidateToken_WrongIssuer`.)

## Testing
- `go test -count=1 ./pkg/... ./cmd/...` and `go test -race ./pkg/capability/` — all pass.

---


eunolabs:
✅ Ready to merge — all session rev-20260615T0310Z-bde0af bugs addressed (#507, #506, #505), self-review complete with no outstanding comments. The #506 refactor is behavior-preserving and covered by existing list-filter tests (`TestGap2/3/4_Filter*`, adversarial wildcard tests); `go test ./pkg/... ./cmd/...` and race-clean `pkg/capability` pass; fmt/vet clean.

---

PR #512, opened 2026-06-15
Fix session rev-20260615T0247Z-f0eff4 bugs

### LINKED ISSUE #501: [rev-20260615T0247Z-f0eff4] VerifyToken skips forced JWKS refresh on all-key signature failure for kid-less tokens
## Location

`pkg/capability/jwks.go`, `JWKSClient.VerifyToken`, lines ~109–203

## Reference

https://github.com/eunolabs/eunox/blob/08eaa5669e68f72a488f5c777db790e466c4f5a3/pkg/capability/jwks.go

## Severity

**Medium** — tokens signed with a new signing key and carrying no `kid` header are rejected for up to one full `cacheTTL` (default 5 minutes) after a JWKS key rotation.

## Description

`VerifyToken` has two code paths depending on whether `FindKeys(keys, kid)` returns any matches:

```go
matchingKeys := FindKeys(keys, kid)
if len(matchingKeys) == 0 {
    // kid-based path: forced refresh triggered here
    keys, err = c.cache.ForceRefreshForKID(ctx, kid)
    if err != nil {
        return nil, fmt.Errorf("force-refresh JWKS: %w", err)
    }
    matchingKeys = FindKeys(keys, kid)
    if len(matchingKeys) == 0 {
        return nil, fmt.Errorf("no JWKS key for kid %q", kid)
    }
}
// Both paths converge here
var lastErr error
for i := range matchingKeys {
    if err := tok.Claims(&matchingKeys[i], &claims); err != nil {
        lastErr = fmt.Errorf("key %q: %w", matchingKeys[i].KeyID, err)
        continue
    }
    return &payload, nil
}
// All keys exhausted — NO forced refresh before returning the error
```

For a token with no `kid` header, `FindKeys(keys, "")` returns **all** cached keys (non-empty slice). The `len(matchingKeys) == 0` branch — and thus `ForceRefreshForKID` — is never reached.

If the IdP rotates its signing key while the old public key is still in cache, every kid-less token signed with the new key will fail signature verification against every cached key and the function returns an error **without refreshing the JWKS cache**. The denial persists until the next natural TTL-based refresh.

Tokens that carry an explicit `kid` do not have this problem: on all-key failure the `len(matchingKeys) == 0` branch triggers a forced refresh (because the new `kid` will not be in cache).

## Trigger / Repro

1. Configure a JWKS-backed deployment with an IdP that issues tokens without `kid` headers (common with some legacy or self-hosted IdPs).
2. Rotate the IdP's signing key. New tokens are signed with the new key; the old public key is still in `JWKSCache` (within TTL).
3. Submit a valid new-key token. `VerifyToken` tries all cached keys (pre-rotation), all fail, returns error, no refresh.
4. All subsequent requests with new-key tokens receive `AUTHORIZATION_FAILED` for up to `cacheTTL` seconds.

## Suggested Fix

After the signature loop exhausts all matching keys and before returning the final error, attempt a single forced refresh and retry:

```go
var lastErr error
for i := range matchingKeys {
    if err := tok.Claims(&matchingKeys[i], &claims); err != nil {
        lastErr = fmt.Errorf("key %q: %w", matchingKeys[i].KeyID, err)
        continue
    }
    return &payload, nil
}
// All cached keys failed — the signing key may have rotated since last fetch.
// Attempt one forced refresh (rate-limited by ForceRefreshForKID's negative-KID
// cache so this cannot be amplified by a flood of bad tokens).
refreshed, err := c.cache.ForceRefreshForKID(ctx, kid)
if err == nil && len(refreshed) > len(keys) {
    // New keys appeared; retry against the expanded set.
    for i := range refreshed {
        if err := tok.Claims(&refreshed[i], &claims); err != nil {
            continue
        }
        return &payload, nil
    }
}
```

Alternatively, `ForceRefreshForKID` could be made to accept an empty `kid` and refresh unconditionally (subject to the same negative-KID rate limiting), keeping the fix at the cache layer rather than the verify loop.

## Session

- Session ID: `rev-20260615T0247Z-f0eff4`
- Reviewed commit: `08eaa5669e68f72a488f5c777db790e466c4f5a3`
- Modules reviewed: `pkg/circuitbreaker`, `pkg/callcounter`, `pkg/enforcement`, `pkg/capability/jwks.go`, `pkg/capability/token_cache.go`

### LINKED ISSUE #502: [rev-20260615T0247Z-f0eff4] TokenCache Put() does not advance token position in eviction queue on re-insertion
## Location

`pkg/capability/token_cache.go`, `Put()` method, lines ~197–222

## Reference

https://github.com/eunolabs/eunox/blob/08eaa5669e68f72a488f5c777db790e466c4f5a3/pkg/capability/token_cache.go

## Severity

**Low** — eviction fairness issue; no correctness or security impact, but frequently-refreshed tokens are evicted before dormant ones.

## Description

`Put()` handles the case where the cache key already exists by reusing the existing list element without moving it to the back of the insertion-order list:

```go
func (c *TokenCache) Put(key string, payload *TokenPayload, exp time.Time) {
    c.mu.Lock()
    defer c.mu.Unlock()

    entry := tokenEntry{payload: payload, exp: exp}
    if existing, exists := c.entries[key]; exists {
        entry.listElem = existing.listElem   // reuses old position — no MoveToBack
    } else {
        elem := c.insertOrder.PushBack(key)  // new entries go to back (most recent)
        entry.listElem = elem
    }
    c.entries[key] = entry
    // eviction loop below trims from Front (oldest position)
}
```

When the capacity limit is reached, `Put()` evicts entries from `c.insertOrder.Front()` (the oldest position). A token that is `Put()` repeatedly (e.g., fetched from the IdP on every request because it carries a short `exp`) retains its original front-of-list position from first insertion. A dormant token inserted later but never re-Put sits closer to the back. Under capacity pressure, the frequently-refreshed (and therefore most-recently-valid) token is evicted first, while the stale dormant token survives.

The cache is documented as "LRU-like" (by insertion order), but re-Put entries behave as FIFO-by-first-insert rather than LRU. The result is that the most actively used tokens are evicted soonest.

## Trigger / Repro

1. Create a `TokenCache` with `maxEntries = 2`.
2. `Put("tokenA", ...)` — tokenA is at position [A].
3. `Put("tokenB", ...)` — list is [A, B].
4. `Put("tokenA", ...)` (re-insert/refresh) — tokenA stays at front: list is [A, B].
5. `Put("tokenC", ...)` — cache is full; evicts `Front()` = A. List becomes [B, C].

Expected behavior (LRU): step 4 should move A to back → [B, A]; step 5 evicts B.

## Suggested Fix

Add `c.insertOrder.MoveToBack(existing.listElem)` in the re-insert branch:

```go
if existing, exists := c.entries[key]; exists {
    entry.listElem = existing.listElem
    c.insertOrder.MoveToBack(existing.listElem)  // update position to reflect recent use
} else {
    elem := c.insertOrder.PushBack(key)
    entry.listElem = elem
}
```

`list.MoveToBack` is O(1) and thread-safe under the held `c.mu` write lock.

## Session

- Session ID: `rev-20260615T0247Z-f0eff4`
- Reviewed commit: `08eaa5669e68f72a488f5c777db790e466c4f5a3`
- Modules reviewed: `pkg/circuitbreaker`, `pkg/callcounter`, `pkg/enforcement`, `pkg/capability/jwks.go`, `pkg/capability/token_cache.go`

### LINKED ISSUE #503: [rev-20260615T0247Z-f0eff4] Redis Lua rate-limit script uses ZRANGE WITHSCORES syntax requiring Redis 6.2+
## Location

`pkg/callcounter/redis.go`, `incrIfBelowScript` Lua script, lines ~181–200

## Reference

https://github.com/eunolabs/eunox/blob/08eaa5669e68f72a488f5c777db790e466c4f5a3/pkg/callcounter/redis.go

## Severity

**Low** — compatibility issue affecting the denial code path on Redis < 6.2; calls over the limit return a Lua error instead of a rate-limit response, but the fail-closed invariant is preserved (the call is still denied).

## Description

`incrIfBelowScript` contains this Lua snippet in the `count >= limit` denial branch:

```lua
local pivot = redis.call('ZRANGE', KEYS[1], idx, idx, 'WITHSCORES')
```

The four-argument `ZRANGE key min max [REV] [BYSCORE|BYLEX] [LIMIT offset count] [WITHSCORES]` syntax — where `WITHSCORES` is a trailing option rather than a separate `ZRANGEWITHSCORES` command — was introduced in **Redis 6.2**. On Redis 6.0 and earlier, this call fails with:

```
ERR wrong number of arguments for 'zrange' command
```

Because the `ZRANGE WITHSCORES` call is only reached when `count >= limit` (the denial branch), the error surfaces only when a rate limit is actually exceeded. The admission path (`count < limit`) exits the script before this line and is unaffected.

**Practical impact on Redis < 6.2:**
- Calls within the limit: admitted correctly (no regression).
- Calls over the limit: the Lua script returns an error to `IncrementIfBelow`. The error is treated as `CONDITION_FAILED` rather than `RATE_LIMITED`, so the call is still denied (fail-closed), but:
  - No `retryAfter` hint is computed or returned (the `pivot` variable is used to calculate when the oldest window entry expires).
  - The error code seen by the MCP host differs from the expected `RATE_LIMITED` code.
  - Audit records reflect `CONDITION_FAILED` rather than the more informative rate-limit outcome.

## Trigger / Repro

1. Deploy with Redis 6.0 or 6.1.
2. Configure a `maxCalls` condition with a limit of N.
3. Issue N+1 calls for the same tool within the time window.
4. The (N+1)-th call returns a Lua error from Redis instead of a clean rate-limit response.

## Suggested Fix

Replace the `ZRANGE ... WITHSCORES` call with the pre-6.2 `ZRANGEWITHSCORES` command, which has been available since Redis 1.2:

```lua
-- Before (Redis 6.2+ only):
local pivot = redis.call('ZRANGE', KEYS[1], idx, idx, 'WITHSCORES')

-- After (compatible with Redis 1.2+):
local pivot = redis.call('ZRANGEWITHSCORES', KEYS[1], idx, idx)
```

`ZRANGEWITHSCORES` returns results in the same `[member score member score ...]` flat-array format, so the rest of the Lua script needs no changes.

Alternatively, document Redis 6.2 as a hard minimum requirement in the deployment guide and add a version check to the `doctor` subcommand.

## Session

- Session ID: `rev-20260615T0247Z-f0eff4`
- Reviewed commit: `08eaa5669e68f72a488f5c777db790e466c4f5a3`
- Modules reviewed: `pkg/circuitbreaker`, `pkg/callcounter`, `pkg/enforcement`, `pkg/capability/jwks.go`, `pkg/capability/token_cache.go`

eunolabs:
## Bugs addressed (session rev-20260615T0247Z-f0eff4)

- Closes #501 — `JWKSClient.VerifyToken` only triggered a forced JWKS refresh on the `FindKeys(keys, kid) == 0` (kid-miss) path. A token with **no kid header** matches every cached key (`FindKeys(keys, "")` returns the whole set), so that path never ran for it: after the issuer rotated its signing key, a kid-less token signed by the new key failed against every cached key and was rejected for up to a full `cacheTTL`. Fix: refactored the per-key verify loop into a `tryKeys` helper (so the retry path can't diverge from the primary path — preserves the #408 terminal-vs-signature-error distinction exactly), and after all matching keys fail for a **kid-less** token, force one refresh and retry. Added `JWKSCache.ForceRefreshForVerify`, rate-limited via the negative-kid cache under an empty-string sentinel (a real kid is never empty) so a flood of bad-signature kid-less tokens triggers at most one upstream fetch per `negativeKIDTTL` — the same amplification protection `ForceRefreshForKID` gives the kid path.
- Closes #502 — **Duplicate of #497, already merged to `main`.** `TokenCache.Put` now calls `c.insertOrder.MoveToBack(existing.listElem)` on the update path (`pkg/capability/token_cache.go:210`), giving the documented LRU-like refresh behavior. Covered by `TestTokenCache_Put_RefreshMovesToBack`. No further change needed here.

## Note on #503 (not included)
#503 reports that the rate-limit Lua script's `ZRANGE ... WITHSCORES` requires Redis 6.2+ and suggests replacing it with `ZRANGEWITHSCORES`. This is **incorrect** and I did not make the change: the script uses the **index-based** form `ZRANGE key idx idx WITHSCORES`, which has existed since Redis 2.0 (6.2 only *added* the `BYSCORE`/`BYLEX`/`REV`/`LIMIT` options to `ZRANGE`). The suggested `ZRANGEWITHSCORES` is not a real Redis command and would break the script on every version. The current code is correct and is exercised by the denial-branch tests (`TestRedis_IncrementIfBelow_RetryAfterWhenCountExceedsLimit`, `...AdmitsUpToLimit`). I'll close #503 as invalid with this explanation.

## Testing
- Added `TestJWKSClient_KidlessToken_RotatedKey_ForcesRefresh` (verified it **fails without the fix**) and `TestJWKSClient_KidlessToken_BadSignature_FetchRateLimited` (amplification gate).
- `go test -race -count=1 ./pkg/capability/` and `go test ./pkg/... ./cmd/...` — all pass.

---


eunolabs:
✅ Ready to merge — session rev-20260615T0247Z-f0eff4 addressed: #501 fixed (kid-less rotation forced refresh, with amplification-bounded `ForceRefreshForVerify`); #502 already resolved on `main` by the merged #497 (`MoveToBack`); #503 is invalid and will be closed with an explanation (the `ZRANGE idx idx WITHSCORES` index form predates Redis 6.2 and the suggested `ZRANGEWITHSCORES` is not a real command). Self-review complete, `make lint` clean (0 issues), `go test -race ./pkg/capability/` and full `go test ./pkg/... ./cmd/...` pass. New #501 tests verified to fail without the fix.

---

PR #523, opened 2026-06-15
refactor: extract enforcedForward to deduplicate resource/prompt enforcement

eunolabs:
## Summary

Extracts the shared deny/observe/forward/record logic from the three argument-less enforced methods (`resources/read`, `resources/subscribe`, `prompts/get`) into a new `enforcedForward` helper in both `HTTPProxy` and `StdioProxy`, removing ~150 lines of duplicated enforcement code across the two transports while preserving exact behavior. It also folds in several smaller correctness / dedup / polish fixes.

## Why

The three argument-less enforced methods implemented identical PDP decision handling — deny-with-audit-observation, upstream-error handling, obligation recording, audit-sink calls — copy-pasted across both transports. That duplication made it easy for the transports to drift and added maintenance burden.

Alongside the dedup, this PR also:
- Consolidates the allowed-values matcher into `enforcement.MatchAllowedValue` so the JWT shorthand and the manifest engine share one implementation, removing the hand-synced numeric / allowedValues copies in `pdp_jwt.go` (issue #451).
- Replaces three near-identical sorted-key helpers with one generic `sortedKeys`.
- Fixes `ArgumentSchema` serialization to omit an absent `type` instead of emitting `"type":null` (behavior change — see below).
- Corrects drifted comments (the `validate_live` FM-5 exit-code note, a `fetchStdioTools` godoc) and an inaccurate `//nolint` rule id (`G703` → `G304`).
- Removes the dead `ManifestPDP.SamplingAllowed` and simplifies the sampling handlers to a clearer `observe` formulation. Sampling stays **binary** — `enforcement: audit` is rejected on `system:` targets at load, so there is no per-entry audit mode for it (no behavior change to sampling).

## Type of change

- [x] Refactor / chore — `enforcedForward`, `MatchAllowedValue`, `sortedKeys` dedup
- [x] Bug fix — `argumentSchema` `"type":null` serialization

## How it was tested

```
make test    # go test -race ./...
make lint    # go vet + golangci-lint (0 issues)
```

All existing enforcement/transport tests pass. New / updated tests: `TestArgumentSchema_MarshalJSON_OmitsNullType` (round-trips a typeless nested schema with no `type:null`), `TestSortedKeys`, and the `DecideSampling`-based sampling tests migrated off the deleted `SamplingAllowed`.

## Security considerations

**Fail-closed preserved:** the dedup moves identical code into shared helpers without changing decision logic. Deny paths, route-level audit observation, and kill-switch hard-blocks are unchanged.

**Audit tape:** no changes to record fields or signing. The `recordObligations` parameter preserves prior behavior (resources/subscribe records no obligation names).

**Authority:** no widening of permissions. The shared `MatchAllowedValue` is identical to the prior manifest-engine matcher; JWT shorthand now uses it instead of a hand-synced copy, reducing divergence risk.

## Checklist

- [x] `make check-license` / `make check-notice` pass
- [x] Docs updated (`docs/capability-manifest-guide.md` §2b sampling note; `validate_live` FM-5 exit-code comment)
- [x] Schema unchanged (no config/manifest field changes)
- [x] PR description explains *why*

https://claude.ai/code/session_018bERPupz6DpQ6TYYNPHuzg


cmd/eunox-mcp/enforcement_gaps_test.go
```diff
@@ -872,6 +872,43 @@ func TestGap5_StdioProxy_SamplingAllowed_ManifestHasSamplingEntry(t *testing.T)
 	}
 }
 
+func TestGap5_StdioProxy_Sampling_PerEntryAuditMode_ObservesAndForwards(t *testing.T) {
```
eunolabs: This test builds a `system:sampling/createMessage` constraint with `Enforcement: EnforcementAudit` directly — but that manifest can never be loaded. `LoadManifest` rejects `enforcement: audit` on any `system:` target (`manifest.go:144-147`, *"the sampling opt-in is binary"*), and `TestLoadManifest_Enforcement` has an explicit case asserting exactly that (`wantErr: "not supported on a system"`).

Since `Constraint.IsAuditOnly()` is solely `Enforcement == "audit"`, `DecideSampling` can never set `AuditOnly=true` from a real manifest. That makes the `dec.AuditOnly` branch this test exercises — and the matching #364 handler change in both `handleUpstreamRequest`/`handleHTTPUpstreamRequest` — **unreachable in production**. The test passes only because it bypasses the loader, so it gives false confidence rather than covering a real config.

Net: the codebase now asserts two contradictory things — "audit is binary / unsupported for `system:` targets" (validation) **and** "audit-mode sampling observes-and-forwards" (this fix). Please reconcile one way:
- if audit-mode sampling is intentionally unsupported, drop or clearly comment the `dec.AuditOnly` handling in both sampling handlers and the `IsAuditOnly()` branch in `DecideSampling` as deliberate defensive dead-code, and rewrite this test to load via a manifest so it documents the rejection; **or**
- if audit-mode sampling *should* work, relax the `system:`-target validation rule so the path is actually reachable.

---

pkg/enforcement/handlers.go
```diff
@@ -1016,6 +991,43 @@ func (e *Engine) handleAllowedValues(_ context.Context, cond capability.Conditio
...
+		}
+	}
+	return false
+}
+
+// NumericEqual reports whether a and b are both numeric and equal in value,
+// independent of their concrete Go type. It is exported so callers outside this
+// package (the JWT shorthand PDP) share one numeric comparison instead of keeping
+// a hand-synced copy.
+func NumericEqual(a, b any) bool { return numericEqual(a, b) }
```
eunolabs: `NumericEqual` is exported but has no callers anywhere in the tree — `git grep NumericEqual` finds only this definition and its own doc comment. `pdp_jwt.go` was switched to `enforcement.MatchAllowedValue`, which calls the package-private `numericEqual` internally, so it never touches this wrapper.

That makes this dead public API, and it makes the PR description slightly inaccurate — it says the JWT path now uses "the shared `enforcement.NumericEqual`", but it actually uses `MatchAllowedValue`. (`unused`/`deadcode` linters won't catch it because exported funcs are assumed to be API.) Suggest either dropping the export or having `pdp_jwt.go` call it directly if you want it to be the genuinely shared primitive.

---

cmd/eunox-mcp/http.go
```diff
@@ -1123,8 +1071,12 @@ func (p *HTTPProxy) handleHTTPUpstreamRequest(ctx context.Context, sess *httpSes
 	dec := decideSampling(ctx, rt.pdp, sess.id, sess.clientIP)
 
 	if dec.Decision != capability.DecisionAllow {
-		// Kill switch denials hard-block even in audit mode (issue #358).
-		if !rt.audit || isKillSwitchDenial(dec.Denial) {
+		// A per-entry `enforcement: audit` sampling opt-in (dec.AuditOnly) observes
+		// and forwards its own denial, exactly like route-level audit mode — matching
+		// every other enforced method (issue #364). A kill switch denial is a hard
+		// block even in audit mode (issue #358).
+		observe := !isKillSwitchDenial(dec.Denial) && (rt.audit || dec.AuditOnly)
```
eunolabs: The same per-entry audit-mode change was applied to the HTTP transport here, but only the stdio path got a new test (`TestGap5_StdioProxy_Sampling_PerEntryAuditMode_ObservesAndForwards`). Assuming this path survives the reachability question (see my note on that test), CLAUDE.md asks for new MCP-method behavior coverage in `enforcement_gaps_test.go` — an HTTP equivalent would guard against the two transports drifting, which is exactly what this refactor is meant to prevent.

---

eunolabs (commented):
## Review

Thanks — the core dedup is clean and, as far as I can verify, behavior-preserving. I checked out the head (`51ed39d`), built it, ran the affected packages (`pkg/enforcement`, `pkg/capability`, and the `TestGap5*` / `TestSortedKeys` / `TestArgumentSchema*` / `TestCrossNamespace*` cmd tests) — all pass, and all 43 CI checks are green.

### What the refactor gets right
- The `enforcedForward` extraction faithfully preserves the per-method asymmetries: the `auditID` vs `denialTarget` split keeps prompts logging `prompts/<name>` while denying on the bare name; `recordObligations=false` keeps `resources/subscribe` from logging obligation names (matching prior behavior); the stderr `kind` labels match the originals. Both transports now share the same tail, which is the right call.
- The `MatchAllowedValue` consolidation is semantically identical to the old string-glob / non-string-exact split (`matchesAllowedValues` + `matchesNonStringAllowedValue`): a string arg still only matches string globs, a non-string arg still only matches non-string exact/numeric values. JWT and engine genuinely share one matcher now (#451).
- `ArgumentSchema.MarshalJSON` shadow-field trick is correct (depth-0 `*SchemaType` shadows the embedded value field, `omitempty` elides an absent type) and is covered by a round-trip test.

### Main concerns (see inline)
1. **The #364 sampling fix is unreachable under current validation.** `enforcement: audit` on a `system:` target is rejected at load, so `dec.AuditOnly` is never true for sampling from a real manifest — the new test exercises an unloadable state. This is the one I'd want resolved (or explicitly documented as defensive dead-code) before merge, since it's a latent contradiction in the enforcement logic.
2. **`enforcement.NumericEqual` is exported but has zero callers** — dead public API, and the description's claim that the JWT path uses it is inaccurate (it uses `MatchAllowedValue`).
3. **The HTTP sampling change has no test** — only stdio got one.

### Metadata / scope (non-blocking)
- "Type of change: Refactor or chore (no behavior change)" and "the refactor does not change decision logic" aren't quite accurate. The #364 change *is* a decision-logic change (deny → observe-and-forward for a per-entry audit opt-in), and the `ArgumentSchema` null-type fix and the `validate_live` FM-5 exit-code note are behavior changes too. The "The refactor also:" list in the body does own these, so it's mostly the checkbox that's misleading — and the commits are nicely split per concern, which helps a lot.
- Leftover old-name references the "clean up test names" pass missed: `TestGap5_StdioProxy_SamplingAllowed_ManifestHasSamplingEntry` and `TestHTTPHandleUpstreamRequest_SamplingAllowedThroughJWTWrapper` still say `SamplingAllowed`.

Overall: the dedup itself is good and safe to land. I'd just want concern #1 reconciled (relax the validation, or trim/comment the unreachable `AuditOnly` sampling handling) so the codebase isn't simultaneously asserting "audit is unsupported on `system:` targets" and "audit-mode sampling observes and forwards."

---

eunolabs:
Thanks for the careful review — all addressed in `0dd268f`.

1. **Reachability contradiction (the blocking one).** Confirmed: `enforcement: audit` is rejected on `system:` targets at load (`validateLocalManifest`, covered by `TestLoadManifest_Enforcement`), so `DecideSampling` can never set `dec.AuditOnly` from a loaded manifest — the per-entry audit handling was unreachable and contradicted the "sampling opt-in is binary" invariant. My original premise was simply wrong. Reverted to honor the invariant: both sampling handlers now observe only on route-level `--audit` (`!isKillSwitchDenial && audit`), the `IsAuditOnly` defer is removed from `DecideSampling` (with a comment explaining why sampling has no per-entry audit), the dead test is deleted (the loader rejection is already covered), and the §2b docs note is corrected.

2. **Dead `NumericEqual`.** Removed. The JWT path uses `MatchAllowedValue`, which calls the package-private `numericEqual` — so there was no shared exported primitive to keep.

3. **HTTP sampling test.** Moot under the revert — the per-entry audit path no longer exists. Route-level audit sampling for HTTP stays covered (`TestHTTPHandleUpstreamRequest_SamplingThroughJWTWrapper` and the audit-mode tests).

Also renamed the two leftover `SamplingAllowed` test names.

On the metadata: after the revert the only remaining genuine behavior change is the `argumentSchema` `type:null` serialization fix; the sampling change is back to original semantics. Happy to tweak the "type of change" checkbox if you'd like.

---

eunolabs (commented):
## Re-review of `0dd268f`

All three concerns resolved cleanly — thanks for the quick turnaround. I checked out the new head, rebuilt, reran the affected packages (`pkg/enforcement`, `pkg/capability`, and the `TestGap5*` / `TestHTTPHandleUpstreamRequest*` / `TestLoadManifest_Enforcement` / `TestManifestPDP_DecideSampling` cmd tests) — all pass, and all 43 CI checks are green.

Verified on the follow-up:
- **Sampling reachability** — both handlers are back to `observe := !isKillSwitchDenial(dec.Denial) && <route>.audit`, i.e. the original pre-#364 semantics; the `IsAuditOnly` defer is gone from `DecideSampling` with a good comment explaining *why* sampling is binary (and noting the other `Decide*` methods correctly keep theirs). The unreachable test is removed and the loader rejection stays covered by `TestLoadManifest_Enforcement`. Docs corrected. No latent contradiction anymore.
- **`NumericEqual`** — removed; `MatchAllowedValue`'s doc now correctly references the package-private `numericEqual`.
- **Leftover names** — `TestGap5_StdioProxy_Sampling_ManifestHasSamplingEntry` / `TestHTTPHandleUpstreamRequest_SamplingThroughJWTWrapper` renamed.

The core dedup (`enforcedForward`, `MatchAllowedValue` consolidation, generic `sortedKeys`, `ArgumentSchema.MarshalJSON`) is unchanged from my first pass and remains correct/behavior-preserving.

One **non-blocking** loose end before merge: the PR **description body** is now stale relative to the code — it still lists "Fixes sampling's per-entry audit-mode handling (issue #364)", "Removes hand-synced numeric comparison code … in favor of the shared `enforcement.NumericEqual`", and "New test `TestGap5_StdioProxy_Sampling_PerEntryAuditMode_ObservesAndForwards` verifies the sampling audit-mode fix" — all three were reverted/removed. Worth trimming those bullets (and, as you offered, the "no behavior change" checkbox, since the `argumentSchema` `type:null` serialization fix remains a real behavior change) so the description matches what actually lands.

Code LGTM — would approve, but GitHub blocks approving one's own PR, so flagging it green here instead.

---

PR #533, opened 2026-06-15
Fix automated-review bug batch (sessions a74e64 / 02d8a4 / 63412f / 7ee6f4)

### LINKED ISSUE #513: [rev-20260615T0410Z-7ee6f4] timeWindow condition silently rejects valid RFC3339 sub-second timestamps
**Location:** `pkg/enforcement/handlers.go:195,217` and `cmd/eunox-mcp/manifest.go` (`validateTimeWindow`)

**Reference:** https://github.com/eunolabs/eunox/blob/f1e91d3a429b6f6100c6d493022ff99a8377be66/pkg/enforcement/handlers.go#L195-L217

**Severity:** Medium

**Description:**

`handleTimeWindow` parses `notBefore` and `notAfter` values with `time.Parse(time.RFC3339, ...)`:

```go
// pkg/enforcement/handlers.go:195
notBefore, err := time.Parse(time.RFC3339, tw.NotBefore)
// ...
// pkg/enforcement/handlers.go:217
notAfter, err := time.Parse(time.RFC3339, tw.NotAfter)
```

Go's `time.RFC3339` layout (`"2006-01-02T15:04:05Z07:00"`) does **not** include a fractional-seconds field, so `time.Parse` with that layout rejects any timestamp with sub-second precision (e.g. `"2026-01-01T00:00:00.5Z"`), returning a parse error. The handler treats a parse error as a condition failure and denies the call.

The manifest-load validator (`validateTimeWindow` in `cmd/eunox-mcp/manifest.go`) uses the same layout, so sub-second timestamps are also rejected at load time with an error message referencing `"expected RFC3339, e.g. 2026-04-01T00:00:00Z"` — misleading because RFC3339 (RFC 3339 §5.6) explicitly permits fractional seconds.

`time.RFC3339Nano` (`"2006-01-02T15:04:05.999999999Z07:00"`) accepts both forms and is the correct layout to use for maximum-fidelity RFC 3339 parsing in Go.

**Trigger / repro:**

Write a manifest with a `timeWindow` condition using a sub-second `notBefore` or `notAfter`:

```yaml
conditions:
  - type: timeWindow
    notBefore: "2026-01-01T00:00:00.000Z"
    notAfter:  "2026-12-31T23:59:59.999Z"
```

Loading this manifest produces an error (or if somehow loaded, every `tools/call` against the constrained tool is denied). Any clock-synchronized deployment that round-trips timestamps through a system that emits fractional seconds (e.g. a CI pipeline, a log aggregator) cannot use those timestamps verbatim in a manifest.

**Suggested fix:**

Replace `time.RFC3339` with `time.RFC3339Nano` in all three parse call sites:

```go
// pkg/enforcement/handlers.go:195
notBefore, err := time.Parse(time.RFC3339Nano, tw.NotBefore)

// pkg/enforcement/handlers.go:217
notAfter, err := time.Parse(time.RFC3339Nano, tw.NotAfter)
```

Apply the same fix to `validateTimeWindow` in `cmd/eunox-mcp/manifest.go` so load-time validation and runtime enforcement use the same layout. Update the example in the error message to include a fractional-second form, or note that both are accepted.

**Session:** rev-20260615T0410Z-7ee6f4

### LINKED ISSUE #514: [rev-20260615T0410Z-7ee6f4] audit-verify --since filter uses exclusive lower bound, dropping boundary record
**Location:** `cmd/eunox-mcp/audit.go:1385`

**Reference:** https://github.com/eunolabs/eunox/blob/f1e91d3a429b6f6100c6d493022ff99a8377be66/cmd/eunox-mcp/audit.go#L1385

**Severity:** Low

**Description:**

The `audit-verify --since` filter exits the reporting window when it sees a record whose timestamp is **not after** the `since` value:

```go
// audit.go:1385
if inReportWindow && !since.IsZero() {
    t, err := time.Parse(time.RFC3339Nano, rec.Time)
    if err != nil || !t.After(since) {   // <-- exclusive: t == since is dropped
        inReportWindow = false
    }
}
```

`!t.After(since)` is true when `t == since`, so a record whose timestamp matches the `--since` argument exactly is excluded from verification and counted as `skipped` rather than `valid`. The intended semantics of `--since` is an inclusive lower bound ("show everything from this point forward"), matching the intuition of every log-tailing tool.

A security auditor who specifies the exact timestamp of a known event (e.g. copied from a prior `audit-verify` run or from an alert) will silently miss that record. In an incident-response scenario this could lead to the conclusion that a specific action was not recorded.

**Trigger / repro:**

1. Run the proxy and capture the `time` field of one audit record, e.g. `"2026-06-01T12:00:00.000000000Z"`.
2. Run `eunox-mcp audit-verify --since "2026-06-01T12:00:00.000000000Z" --log audit.jsonl`.
3. The record at exactly that timestamp is **not** verified — it is counted as skipped.

**Suggested fix:**

Change the boundary check to use `t.Before(since)` (exclusive upper bound of the skip region), which is equivalent to `t >= since`:

```go
if err != nil || t.Before(since) {
    inReportWindow = false
}
```

This makes the lower bound inclusive: records at exactly `since` enter the reporting window and are verified.

**Session:** rev-20260615T0410Z-7ee6f4

### LINKED ISSUE #524: [rev-20260615T0610Z-63412f] killswitch Redis.Start() leaks goroutines and loses Stop() control when called more than once
**Location:** `pkg/killswitch/redis.go:57-110`

**Reference:** https://github.com/eunolabs/eunox/blob/9d0ba1b006f843d70f4d2b12761028581f270ec1/pkg/killswitch/redis.go#L57-L110

**Severity:** High

**Description:**

`Start()` has no idempotency guard. Every call creates a fresh `context.WithCancel` context, overwrites `r.lifecycleCtx` and `r.cancel`, and launches two new goroutines (`listenPubSub` + `reconcileLoop`). The previous cancel function is silently discarded:

```go
func (r *Redis) Start(ctx context.Context) {
    subCtx, cancel := context.WithCancel(ctx)
    r.mu.Lock()
    r.lifecycleCtx = subCtx
    r.cancel = cancel   // ← overwrites the prior cancel; prior goroutines are now un-cancellable
    r.mu.Unlock()
    ...
    go r.listenPubSub(subCtx, pubsub)
    go r.reconcileLoop(subCtx)
}
```

Consequences:
1. **Goroutine leak** – the first call's `listenPubSub` and `reconcileLoop` goroutines run forever; their subCtx can only be cancelled by the *parent* `ctx` (e.g. the process-level context), never by `Stop()`.
2. **Stop() only cancels the latest pair** – every pair spawned before the last `Start()` call is unreachable.
3. **Multiple reconcile loops** – duplicate reconcile tickers fire concurrently, each taking the write lock and writing to `r.killedAgents`/`r.killedSessions` independently, wasting resources and producing non-deterministic edge-triggered log messages from `reconcileRefresh`.

Note the contrast with `InMemory.StartCleanup`, which uses a `sync.Once` to prevent exactly this problem.

**Trigger / repro:**

```go
r := killswitch.NewRedis(client)
r.Start(ctx)   // goroutine pair 1 started, cancel₁ stored
r.Start(ctx)   // cancel₁ overwritten; goroutine pair 1 leaked; goroutine pair 2 started
r.Stop()       // only cancels goroutine pair 2; pair 1 runs until ctx is cancelled
```

Also triggered by any service that calls `Start` on re-initialization without calling `Stop` first.

**Suggested fix:**

Add a `sync.Once` (or a started-flag under the mutex) so that only the first `Start()` launches goroutines:

```go
type Redis struct {
    ...
    startOnce sync.Once
}

func (r *Redis) Start(ctx context.Context) {
    r.startOnce.Do(func() {
        subCtx, cancel := context.WithCancel(ctx)
        r.mu.Lock()
        r.lifecycleCtx = subCtx
        r.cancel = cancel
        r.mu.Unlock()
        // ... initial refresh, subscribe, reconcileLoop
    })
}
```

If re-startability is required, add an explicit `Restart(ctx)` that first calls `Stop()` (draining the old goroutines), resets `startOnce`, then calls `Start(ctx)`.

**Session:** rev-20260615T0610Z-63412f

### LINKED ISSUE #525: [rev-20260615T0610Z-63412f] circuitbreaker Stats() reports State=HalfOpen with stale ConsecutiveFails=FailureThreshold — internally contradictory snapshot
**Location:** `pkg/circuitbreaker/breaker.go:278-311` (`Stats`) and `allow()` Open→HalfOpen transition

**Reference:** https://github.com/eunolabs/eunox/blob/9d0ba1b006f843d70f4d2b12761028581f270ec1/pkg/circuitbreaker/breaker.go#L278-L311

**Severity:** Medium

**Description:**

`Stats()` and `State()` both project a *logical* HalfOpen state when `b.state == StateOpen` and the cooldown has elapsed — without mutating any internal field. This is documented. The problem is that `ConsecutiveFails` is **never cleared** during this logical transition, so the snapshot returned by `Stats()` is internally contradictory:

```
State:            HalfOpen    ← logical view (cooldown elapsed, b.state still Open)
ConsecutiveFails: 5           ← physical value, still == FailureThreshold from when breaker tripped
```

`ConsecutiveFails` is only reset to 0 when a probe actually reports an outcome — either in `recordSuccess` (on any success) or in `recordFailure`'s `StateHalfOpen` case (documented fix #428). Until then — which may be 30 seconds under the default cooldown — every call to `Stats()` during the logical-HalfOpen window returns this contradiction.

```go
// Stats() (line 278)
    b.mu.RLock()
    defer b.mu.RUnlock()

    state := b.state
        state = StateHalfOpen  // ← logical promotion
    }
    return Stats{
        State:            state,
        ConsecutiveFails: b.consecutiveFails, // ← NOT cleared here; still == FailureThreshold
        ...
    }
}
```

A monitoring system (dashboard, alerting rule, health endpoint) that reads `Stats()` will see `State=HalfOpen` alongside `ConsecutiveFails=N ≥ FailureThreshold` for the entire cooldown window. "HalfOpen with N consecutive failures" is ambiguous: it looks like the breaker is simultaneously in recovery AND still failing, which is not what is actually happening. An operator following this dashboard during an incident may mis-diagnose the situation.

**Trigger / repro:**

```go
    FailureThreshold:  3,
    CooldownDuration:  30 * time.Second,
    HalfOpenMaxProbes: 1,
})

// Trip the breaker
for i := 0; i < 3; i++ {
    b.Allow()
    b.RecordFailure()
}

// Advance clock past cooldown but before any probe
time.Sleep(31 * time.Second)

s := b.Stats()
// s.State == StateHalfOpen  (logical)
// s.ConsecutiveFails == 3   ← still the trip value, misleading
```

**Suggested fix:**

Clear `consecutiveFails` in `allow()` at the moment the physical Open→HalfOpen transition occurs:

```go
case StateOpen:
    now := b.now()
        b.state = StateHalfOpen
        b.halfOpenGen++
        b.halfOpenProbes = 1
        b.halfOpenSuccess = 0
        b.consecutiveFails = 0   // ← add this
        b.lastTransitionAt = now
        return true, b.halfOpenGen
    }
```

This makes the `Stats()` snapshot consistent from the moment the transition occurs: `State=HalfOpen` with `ConsecutiveFails=0`. The existing resets in `recordSuccess` (on close) and `recordFailure` (on re-open, #428) remain load-bearing and are not affected.

**Session:** rev-20260615T0610Z-63412f

### LINKED ISSUE #526: [rev-20260615T0610Z-63412f] callcounter compactTimestamps: len(ts)*4 multiplication overflows on 32-bit platforms causing spurious or skipped compaction
**Location:** `pkg/callcounter/memory.go` — `compactTimestamps` function

**Reference:** https://github.com/eunolabs/eunox/blob/9d0ba1b006f843d70f4d2b12761028581f270ec1/pkg/callcounter/memory.go#L193-L202

**Severity:** Low

**Description:**

`compactTimestamps` decides whether to copy a timestamp slice into a right-sized backing array using this condition:

```go
func compactTimestamps(ts []time.Time) []time.Time {
    if cap(ts) > compactMinCap && len(ts)*4 < cap(ts) {
        compact := make([]time.Time, len(ts))
        copy(compact, ts)
        return compact
    }
    return ts
}
```

`len(ts)*4` is computed in `int`. On a 32-bit platform (`int` is 32 bits, max ~2.1 × 10⁹), the multiplication overflows when `len(ts) > 536,870,912` (2²⁹). The result wraps to a negative value, which is always less than `cap(ts)` (non-negative), so the condition fires spuriously — triggering an unnecessary `make`+`copy` cycle — or, when it should fire, the wrapping could cause it to miss the trigger.

`MaxEntries` is defined as `math.MaxInt32 ≈ 2.1 × 10⁹` — well above the overflow threshold on 32-bit `int`. So the call to `checkMaxEntries` in `IncrementAndGet` allows slices up to `math.MaxInt32` on all platforms, but on 32-bit platforms the compaction guard misbehaves for slices larger than 2²⁹.

The fix is to invert the inequality to avoid the multiplication entirely:

```go
// Before (overflows on 32-bit when len > 2^29):
len(ts)*4 < cap(ts)

// After (equivalent, no overflow):
len(ts) < cap(ts)/4
```

`cap(ts)/4` is integer division, which correctly floors the threshold (e.g. `cap=65` → threshold 16, same as `len*4 < 65` → `len < 17`).

**Trigger / repro:**

On a 32-bit build, call `IncrementAndGet` with `maxEntries = math.MaxInt32` and allow the slice to grow past 2²⁹ live entries. `compactTimestamps` is called every call, so the overflow fires on the first invocation once `len(e.timestamps) > 536,870,912`. In practice, this requires ≈ 4 GB of heap just for the timestamp slice (536 M × 8 bytes per `time.Time`), so it is not reachable in any normal workload. The risk is latent rather than immediately exploitable.

**Suggested fix:**

```go
func compactTimestamps(ts []time.Time) []time.Time {
    if cap(ts) > compactMinCap && len(ts) < cap(ts)/4 {
        compact := make([]time.Time, len(ts))
        copy(compact, ts)
        return compact
    }
    return ts
}
```

This is semantically identical to the original intent, avoids the multiplication, and is safe on both 32-bit and 64-bit platforms.

**Session:** rev-20260615T0610Z-63412f

### LINKED ISSUE #527: [rev-20260615T0610Z-02d8a4] JWT op= scan-all bypasses non-SQL operation restriction via free-text string argument
**Location:** `cmd/eunox-mcp/pdp_jwt.go:991-1011` (`evaluateJWTConditions`, scan-all-args branch)

**Reference:** https://github.com/eunolabs/eunox/blob/9d0ba1b006f843d70f4d2b12761028581f270ec1/cmd/eunox-mcp/pdp_jwt.go#L991-L1011

**Severity:** Medium

**Description:**

When a JWT capability claim uses the `op=` shorthand without an explicit argument name (e.g. `"tool:send_message?op=publish"`), `buildV2Constraint` creates an `AllowedOperationsCondition{Argument: "", Operations: ["publish"]}`. `evaluateJWTConditions` then evaluates this in "scan-all-args" mode:

```go
var matchedOp string
for _, v := range args {
    s, ok := v.(string)
    if !ok { continue }
    parts := strings.Fields(s)
    if len(parts) == 0 { continue }
    word := strings.ToUpper(parts[0])
    if containsStringFold(c.Operations, word) {
        matchedOp = word
        continue
    }
    // A disallowed SQL statement in any argument is a hard denial.
    if isSQLVerb(word) {
        // deny
    }
}
if matchedOp == "" {
    // deny: no matching operation found
}
```

The logic requires that **some** string argument's first word matches the permitted operation, and **hard-denies** if any other argument's first word is a recognized SQL verb (`isSQLVerb`). For SQL tool calls this provides reasonable protection because the dangerous DML/DDL/admin verbs are in `sqlVerbs`.

**For non-SQL operations, the `isSQLVerb` guard provides no protection.** Consider:

| Scenario | JWT claim | args | Expected | Actual |
|----------|-----------|------|----------|--------|
| Bypass | `op=publish` | `{"action": "delete", "note": "publish report"}` | DENY (action is delete) | **ALLOW** (note starts with "publish") |
| Bypass | `op=read` | `{"method": "write", "content": "read the docs"}` | DENY (method is write) | **ALLOW** (content starts with "read") |

A call is allowed whenever ANY string argument happens to start with the granted verb — even if that argument is a content/description field, and the actual operation argument holds a disallowed verb. Because "delete", "write", "send", "update", etc. are not SQL verbs, `isSQLVerb` does not catch them, and the loop continues to the `matchedOp` success path.

The comment at line 44–49 acknowledges the model is "best-effort" and advises pairing with a least-privilege database role for SQL. It does **not** warn that non-SQL `op=` grants are consistently bypassable by any tool call that includes a free-text string argument starting with the permitted verb.

The manifest-form `allowedOperations` handler (`pkg/enforcement/handlers.go:479-486`) explicitly removed scan-all-args mode and its comment explains why: "the former scan-all-args mode allowed a request to carry the allowed verb in a benign argument while the actual operation field held something disallowed." The JWT evaluator retains the same vulnerability for non-SQL operations.

**Trigger / repro:**

1. Issue a JWT with claim `"tool:send_message?op=publish"` (non-SQL op, no argument name).
2. The proxy creates `AllowedOperationsCondition{Argument:"", Operations:["publish"]}`.
3. Send a `tools/call` for `send_message` with `{"action": "delete_all_users", "body": "publish: weekly digest"}`.
4. `evaluateJWTConditions` scans all string args: "body" starts with "publish" → `matchedOp="publish"` → allow.
5. The call succeeds despite `action="delete_all_users"` being the real operation.

**Suggested fix:**

Option A (safest): Remove scan-all mode from `evaluateJWTConditions` and require `Argument` to be non-empty, matching what `handleAllowedOperations` does. The `op=publish` JWT shorthand without an explicit argument name would become a CONDITION_FAILED, forcing callers to use the manifest form (Pattern C) which names the argument explicitly.

Option B (preserve current UX): Add a caller-configurable "operation verbs" set analogous to `sqlVerbs`, or require operators to also specify the argument name in JWT claims (`op=publish&arg=action`). Document the limitation clearly for non-SQL use.

**Session:** rev-20260615T0610Z-02d8a4

### LINKED ISSUE #528: [rev-20260615T0610Z-02d8a4] auditSink has no clock seam: rotatedPath and record Time use time.Now() directly, making rotation untestable and timestamps inconsistent under NTP jumps
**Location:** `cmd/eunox-mcp/audit.go:1537` (`rotatedPath`), `audit.go:586` (allow-record Time), `audit.go:907` (drop-marker Time)

**Reference:** https://github.com/eunolabs/eunox/blob/9d0ba1b006f843d70f4d2b12761028581f270ec1/cmd/eunox-mcp/audit.go#L1537

**Severity:** Low

**Description:**

`auditSink` has no injected clock field. All time-sensitive operations use `time.Now()` directly:

```go
// audit.go:1537 — rotation filename
func (s *auditSink) rotatedPath() string {
    base := s.logPath + "." + time.Now().UTC().Format("20060102T150405.000000000Z")
    return uniqueRotatedPath(base)
}

// audit.go:586 — allow/deny record Time field
Time: time.Now().UTC().Format(time.RFC3339Nano),

// audit.go:907 — drop-marker record Time field
Time: time.Now().UTC().Format(time.RFC3339Nano),
```

This has two consequences:

**1. Untestable rotation behavior.** `uniqueRotatedPath` was split out from `rotatedPath` specifically to be "directly testable without depending on host clock granularity" (comment at line 1542). But `rotatedPath` itself still calls `time.Now()`, so testing the full rotate-and-rename path requires waiting for real wall-clock time or accepting non-deterministic filenames. Any test that wants to assert a specific rotation output must couple to real time.

**2. Timestamp inconsistency during NTP corrections or monotonic clock resets.** An NTP step-correction between consecutive calls to `time.Now()` can produce a rotation filename whose embedded timestamp is earlier or later than the `Time` fields of the records written into that file. An operator running `audit-verify` and cross-referencing rotated file names to record timestamps (e.g. to find the file that contains a given event time) may be misled.

The audit chain integrity depends on `seq`, `prevHMAC`, and record `Time` being mutually consistent within a file. The rotation filename is not part of the chain, but its timestamp is the primary human-visible locator for when records were captured. A gap between filename timestamp and record timestamps increases incident-response friction.

Compare: `pkg/enforcement/Engine` accepts a pluggable `Clock` option; `pkg/callcounter` memory backend injects `now func() time.Time` in its config; `pkg/capability/TokenCache` accepts `Now func() time.Time`. The audit sink is the only time-sensitive component without a seam.

**Trigger / repro:**

Any environment with NTP step corrections or running audit rotation tests with controlled time. The discrepancy is silent — no error is logged when the rotation filename timestamp diverges from record timestamps.

**Suggested fix:**

Add a `now func() time.Time` field to `auditSink`, defaulting to `time.Now` at construction:

```go
type auditSink struct {
    ...
    now func() time.Time // defaults to time.Now; injectable for tests
}
```

Replace `time.Now()` calls in `rotatedPath`, the allow-record constructor, and `flushDropMarker` with `s.now()`. Expose the field through `newAuditSink`'s options or a functional option. Existing callers pass `nil` / accept the default. Test coverage for rotation with a fake clock can then be added without depending on real wall time.

**Session:** rev-20260615T0610Z-02d8a4

### LINKED ISSUE #529: [rev-20260615T0610Z-02d8a4] TokenCache.Len() includes expired-but-not-yet-purged entries, misleading capacity and diagnostic consumers
**Location:** `pkg/capability/token_cache.go:248-254` (`TokenCache.Len`)

**Reference:** https://github.com/eunolabs/eunox/blob/9d0ba1b006f843d70f4d2b12761028581f270ec1/pkg/capability/token_cache.go#L248-L254

**Severity:** Low

**Description:**

`Len()` returns the raw entry count from the internal map, including entries whose `expiresAt` is in the past but that have not yet been removed by the background `cleanupLoop`:

```go
func (c *TokenCache) Len() int {
    c.mu.RLock()
    defer c.mu.RUnlock()
    return len(c.entries)
}
```

The comment states "including entries that have expired but not yet been cleaned up", which documents the behavior. However, the gap between reported `Len()` and the number of *live* (Get-able) entries can be substantial:

- `CleanupInterval` defaults to 60 s; `MaxEntryTTL` defaults to 30 s.
- In the worst case, all 4096 (`MaxSize`) entries can expire between cleanup ticks, so `Len()` returns 4096 while `Get()` returns a cache miss for every key.
- During this window, new `Put()` calls see `c.insertOrder.Len() >= c.cfg.MaxSize` (true) and evict the oldest entries — entries that are already expired and would have been removed for free at the next purge. This means a burst of new tokens arriving within a cleanup interval can evict each other unnecessarily when the cache appears full of stale entries.

In addition, any external caller (e.g. a monitoring endpoint, a `doctor` subcommand check, or a test assertion) that uses `Len()` to infer cache occupancy or to trigger a manual `Stop()/Start()` cycle receives an inflated count.

**Trigger / repro:**

```go
cache := capability.NewTokenCache(capability.TokenCacheConfig{
    MaxEntryTTL:     100 * time.Millisecond,
    CleanupInterval: 10 * time.Second,
    MaxSize:         10,
})
cache.Start(context.Background())
defer cache.Stop()

// Fill the cache with short-lived tokens
for i := 0; i < 10; i++ {
    cache.Put(fmt.Sprintf("tok%d", i), &capability.TokenPayload{ExpiresAt: time.Now().Add(50 * time.Millisecond).Unix()})
}

time.Sleep(200 * time.Millisecond) // all entries are now expired, cleanup hasn't run yet
fmt.Println(cache.Len()) // prints 10 — but every Get() returns (nil, false)

// A new token evicts an already-expired entry unnecessarily
cache.Put("new-tok", &capability.TokenPayload{ExpiresAt: time.Now().Add(30 * time.Second).Unix()})
```

**Suggested fix:**

Option A (minimal — clarify the contract): Rename `Len()` to `EntryCount()` or add a `LiveLen()` method that counts non-expired entries under the read lock. Keep `Len()` as-is for backwards compatibility in tests.

Option B (eliminate the staleness): In `Put()`, before the size-eviction loop, run a single-pass inline purge of expired entries so the eviction decision is made against live-entry count only:

```go
// Remove already-expired entries from the front of insertOrder before evicting
    front := c.insertOrder.Front()
    if front == nil { break }
    key := front.Value.(string)
    if entry, ok := c.entries[key]; ok && c.cfg.Now().After(entry.expiresAt) {
        c.insertOrder.Remove(front)
        delete(c.entries, key)
        continue
    }
    break // front is live; evict it normally
}
```

This keeps eviction focused on expired-first, and `Len()` remains accurate immediately after a `Put()`.

**Session:** rev-20260615T0610Z-02d8a4

### LINKED ISSUE #530: pkg/callcounter/redis: non-zero-padded seq breaks lex ordering for same-nanosecond entries, causing ZRemRangeByRank to discard the newest member
## Location

`pkg/callcounter/redis.go` lines 101 (`IncrementAndGet`) and 239 (`IncrementIfBelow`)

## Reference

Commit `9d0ba1b006f843d70f4d2b12761028581f270ec1`

## Severity

Low / Medium (correctness invariant violation; limited practical impact with today's callers, but silently wrong for any future `maxEntries > 1` caller of `IncrementAndGet`)

## Description

The sorted-set member string is built as:

```go
member := fmt.Sprintf("%s-%d-%d", r.instanceID, now.UnixNano(), r.seq.Add(1))
```

The trailing `%d` for the monotonic sequence counter is **not zero-padded**. When two calls arrive at exactly the same nanosecond (same `now.UnixNano()` value — plausible on hypervisors with coarser timer resolution), the member strings differ only in the `seq` suffix:

```
abc123-1718467200000001000-9     ← seq 9
abc123-1718467200000001000-10    ← seq 10 (newer)
```

Both entries land in the same Redis sorted-set score bucket (same microsecond). Redis breaks ties within a score by **lexicographic member order**. Lexicographically `"…-10" < "…-9"` (ASCII `'1' < '9'`), so seq 10 is ranked **lower** than seq 9 even though it was added after.

In `IncrementAndGet`, the rank-cap step is:

```go
pipe.ZRemRangeByRank(ctx, windowKey, 0, -(int64(maxEntries)+1))
```

This keeps the **highest-ranked** (lexicographically largest, assumed newest) `maxEntries` members. When the lex invariant is violated at a digit boundary (9→10, 99→100, 999→1000, …), the just-added entry sits at the lowest rank and is **immediately evicted** while the older entry survives.

`IncrementIfBelow` constructs the same non-zero-padded member but its Lua script does not use `ZRemRangeByRank`; the lex ordering of members does not affect the count, limit check, or `retryAfter` calculation there. The bug is therefore specific to `IncrementAndGet`'s MULTI/EXEC pipeline.

## Trigger / Repro

```
1. Two goroutines call IncrementAndGet on the same key within the same nanosecond
   (time.Now() returns the same value for both).
2. The instance-level r.seq counter is at 9, 99, 999, … just before those two calls,
   so the first call gets seq N (e.g. 9) and the second gets seq N+1 (e.g. 10).
3. ZRemRangeByRank(0, -(maxEntries+1)) fires: seq 10's member has lower lex rank
   than seq 9, so it is chosen as the entry to discard.
4. After the pipeline, countCmd.Val() reflects the older entry (seq 9), not the
   new one (seq 10).
```

Nanosecond collision is more likely in virtualized environments where `CLOCK_REALTIME` is quantized to 100 ns or 1 µs increments (e.g., some AWS/GCP VMs and containers under `vDSO`).

## Current impact

The only production caller of `IncrementAndGet` is `recordSessionCall` (`pkg/enforcement/engine.go`), which uses `maxEntries = sequenceHistoryMaxEntries = 1`. With `maxEntries = 1` the rank-trim fires only when a second entry is added (which happens in the race scenario). When the newer entry is evicted, the older one persists — still signaling "tool was called in this session." `sequenceBlock` only needs presence, so functional correctness is maintained.

However:
- The member format's monotonicity invariant is silently broken.
- Any future `maxEntries > 1` caller of `IncrementAndGet` would see incorrect count semantics whenever same-nanosecond entries straddle a digit boundary.

## Suggested fix

Zero-pad `seq` to a fixed width so lexicographic order matches insertion order for same-nanosecond entries. Since `r.seq` is a `uint64`, 20 decimal digits cover the full range without overflow:

```go
member := fmt.Sprintf("%s-%020d-%020d", r.instanceID, now.UnixNano(), r.seq.Add(1))
```

Alternatively, use a zero-padded hex string (`%016x`) for a more compact representation:

```go
member := fmt.Sprintf("%s-%016x-%016x", r.instanceID, now.UnixNano(), r.seq.Add(1))
```

Both approaches restore the monotonicity invariant. The fix applies to both the `IncrementAndGet` (line 101) and `IncrementIfBelow` (line 239) member format for consistency, even though only `IncrementAndGet` is currently affected by lex ordering.

A regression test can verify this: add two entries with the same `now.UnixNano()` mock value at a digit boundary (seq 9 then 10) to a Redis instance, then call `IncrementAndGet` with `maxEntries = 1` and assert that `ZCard` returns 1 and `ZRANGEBYSCORE` retains the seq-10 member, not seq-9.

---

_Session: rev-20260615T0710Z-a74e64_

eunolabs:
This PR lands fixes for all nine open `automated-review` / `bug` issues across four review sessions in a single branch (one commit per bug, plus a few test/guard follow-ups referencing the same issue).

### Session rev-20260615T0710Z-a74e64
- Closes #530 — callcounter Redis sorted-set member used a non-zero-padded `seq`, so same-nanosecond entries straddling a digit boundary (9→10) broke lexicographic tie-ordering and `ZRemRangeByRank` could discard the newest member. Member format is now `%s-%020d-%020d` (instanceID, nano, seq) in both `IncrementAndGet` and `IncrementIfBelow`. Regression test added.

### Session rev-20260615T0610Z-02d8a4
- Closes #529 — `TokenCache.Put` now reclaims the contiguous front-run of already-expired entries before evicting a live one, so a burst of new tokens no longer displaces live entries while stale ones occupy the cache, and `Len()` is accurate after a full-cache `Put`. Regression test added.
- Closes #528 — added an injectable `now func() time.Time` clock seam to `auditSink` (defaults to `time.Now`), used for record timestamps (allow/deny + drop marker) and the rotation filename, so rotation/timestamping are testable with a fake clock. A `clock()` accessor falls back to `time.Now` for directly-constructed (verify-only/test) sinks. Fake-clock test added.
- Closes #527 — **security**: the JWT `op=` scan-all-args path is unsound for non-SQL operations (a benign free-text argument starting with the granted verb masked a disallowed operation argument). Non-SQL `op=` in the bare form now **fails closed** (`CONDITION_FAILED`); the SQL path (guarded by `isSQLVerb`) is preserved. Docs (`capability-manifest-guide.md`) updated; bypass + fail-closed regression tests added.

### Session rev-20260615T0610Z-63412f
- Closes #526 — `compactTimestamps` threshold `len(ts)*4 < cap(ts)` could overflow `int` on 32-bit platforms; rewritten as `len(ts) < cap(ts)/4` (equivalent, overflow-free).
- Closes #525 — circuit breaker now clears `consecutiveFails` at the physical Open→HalfOpen transition and projects `ConsecutiveFails=0` in the logical-promotion path of `Stats()`, so a half-open snapshot is no longer the contradictory "HalfOpen with ConsecutiveFails==FailureThreshold". Existing cooldown test updated to observe the "stray ignored" property while genuinely Open and assert the new contract once half-open.
- Closes #524 — `killswitch.Redis.Start()` is now idempotent (`sync.Once`), preventing a goroutine leak / lost `Stop()` control / duplicate reconcile loops when called more than once (mirrors `InMemory.StartCleanup`).

### Session rev-20260615T0410Z-7ee6f4
- Closes #514 — `audit-verify --since` is now an inclusive lower bound (`t.Before(since)` instead of `!t.After(since)`), so a record at exactly the `--since` timestamp is verified rather than silently skipped.
- Closes #513 — `timeWindow` `notBefore`/`notAfter` now parse with `time.RFC3339Nano` (runtime handler and load-time validator), accepting RFC 3339 sub-second timestamps that `time.RFC3339` rejected; error-message examples updated.

## Testing
- `gofmt -l` — clean
- `golangci-lint run ./...` — 0 issues
- `go test -race -count=1 ./...` — all packages pass (new regression tests for #530, #529, #528, #527; updated tests for #525, #527)

## Note on branching
All nine issues are in this repo and were fixed on the single development branch this session is scoped to (`claude/hopeful-thompson-r0ixf5`), so they ship as one PR grouped by session rather than one PR per session.

https://claude.ai/code/session_01ETDDyRNynutMatUj2minaL

---
_Generated by [Claude Code](https://claude.ai/code/session_01ETDDyRNynutMatUj2minaL)_


eunolabs (commented):
Self-review round 1 — one substantive finding:

**`cmd/eunox-mcp/audit.go` (`clock()` placement):** The new `clock()` method was inserted directly after `record()`'s doc-comment block, with no blank line separating them. As written, the `record is the gateway-aware variant…` comment now attaches to `func clock()`, and `func record()` is left with no doc comment at all (a godoc regression). Fix: move `clock()` above `record()`'s doc comment (or elsewhere) so `record()`'s documentation stays attached to `record()` and `clock()` has its own standalone comment.

Everything else reviewed (the other 8 fixes, the test updates, and the `Stats()` projection / `recordFailure` open-state guard interaction for #525) looks correct. Will fix the above and re-review.

---

eunolabs:
✅ Ready to merge — all nine batched bugs addressed across sessions `a74e64` (#530), `02d8a4` (#529, #528, #527), `63412f` (#526, #525, #524), and `7ee6f4` (#514, #513). Self-review complete: round-1 found one issue (a misplaced `clock()` doc comment in `audit.go`, now fixed in 8e58de7); round-2 found no further substantive issues. CI is fully green (all 42 checks: builds on 5 platforms, Test, Lint, gofmt, CodeQL, govulncheck, Trivy, demo/OPA/enforcement-compliance), and local `go test -race`, `go vet`, `gofmt`, and `golangci-lint` (0 issues) all pass.

---

PR #534, opened 2026-06-15
Fix automated-review bugs (session 961067): kid-less JWKS rotation + 32-bit retryAfter

### LINKED ISSUE #531: [rev-20260615T0810Z-961067] JWTPDP.ValidateToken missing ForceRefreshForVerify for kid-less key rotation
## Location

`cmd/eunox-mcp/pdp_jwt.go`, lines 282–411 (`ValidateToken`)

## Reference

Contrast with `pkg/capability/jwks.go` `JWKSClient.VerifyToken`, which was patched for this exact scenario (search for the `ForceRefreshForVerify` block added for issue #501).

## Severity

**High** — causes a service-disruption window every time the IdP rotates its signing key when clients use kid-less JWTs.

## Description

`JWKSClient.VerifyToken` (in `pkg/capability/jwks.go`) has a forced-refresh fallback for tokens that carry no `kid` header: after all cached keys fail signature verification it calls `c.cache.ForceRefreshForVerify(ctx)`, re-tries the fresh key set, and only then returns an error.

`JWTPDP.ValidateToken` has no equivalent path. Its key-selection branch for a missing `kid`:

```go
matchingKeys = FindKeys(keys, "")   // returns all cached keys when kid == ""
```

iterates all cached keys. If every cached key fails (because the IdP rotated and the new key is not yet in the cache), the function falls through directly to:

```go
return ctx, fmt.Errorf("JWT signature verification failed: %w", lastErr)
```

No forced refresh is attempted.

## How to trigger

1. Configure `--jwks-uri` with an IdP that issues tokens **without** a `kid` header (common with some legacy IdPs and certain OIDC configurations).
2. The IdP rotates its signing key.
3. Any JWT issued after the rotation arrives at `JWTPDP.ValidateToken` before the normal cache TTL expires (default 5 min).
4. `FindKeys(keys, "")` returns the old (now-stale) cached keys; all signature checks fail.
5. `ValidateToken` returns a verification error — the token is rejected even though it is valid.

All clients using kid-less JWTs are locked out for up to the full `CacheTTL` duration.

## Suggested fix

After the `for _, key := range matchingKeys` loop, add the same forced-refresh path that `JWKSClient.VerifyToken` uses:

```go
// kid-less token: if all cached keys fail, attempt a forced refresh once
// (mirrors pkg/capability/jwks.go JWKSClient.VerifyToken path for #501).
if kid == "" && lastErr != nil {
    if refreshed, ferr := j.cache.ForceRefreshForVerify(ctx); ferr == nil {
        for _, key := range FindKeys(refreshed, "") {
            payload, terminal, sigErr := tryVerify(key, rawToken)
            if terminal != nil {
                return ctx, terminal
            }
            if sigErr == nil {
                // success — validate claims and return
                ...
                return ctx, nil
            }
            lastErr = sigErr
        }
    }
}
return ctx, fmt.Errorf("JWT signature verification failed: %w", lastErr)
```

The exact shape must match the existing `tryVerify` / claims-validation sequence already present in `ValidateToken` so that the fast-path and refresh-path are consistent.

## Session

`rev-20260615T0810Z-961067` — automated review of HEAD `9d0ba1b`

### LINKED ISSUE #532: [rev-20260615T0810Z-961067] InMemory.IncrementIfBelow int(cur-limit) truncates on 32-bit, producing wrong retryAfter
## Location

`pkg/callcounter/memory.go`, line 366 (inside `IncrementIfBelow`)

## Severity

**Low** — incorrect `RetryAfter` estimate on 32-bit platforms only; no correctness or security impact because a truncated-negative value causes the fallback path (full window duration) to be returned instead.

## Description

`IncrementIfBelow` computes an index into the `valid` timestamp slice to estimate when the oldest excess call will expire:

```go
if idx := int(cur - limit); idx >= 0 && idx < len(valid) {
    retry = time.Until(valid[idx].Add(window))
}
```

`cur` and `limit` are both `int64`. Their difference is also `int64`, but the cast `int(cur - limit)` silently truncates to `int32` on 32-bit platforms (where `int` is 32 bits). If `cur - limit` exceeds `math.MaxInt32` (≈ 2.1 billion) the result is negative, `idx >= 0` is false, and `retry` stays at zero — the caller sees `RetryAfter == 0` and applies the full window as a conservative fallback instead of the tighter estimate.

In practice this requires `limit` values in the billions, so it only matters in configurations with astronomically large rate-limit caps. The safe fallback means no calls are incorrectly admitted or denied; only the advisory hint is wrong.

## How to trigger

```go
// 32-bit build (GOARCH=386 or arm)
store := callcounter.NewInMemory(...)
// configure limit = 1, then hammer to cur = 2^31 + 2 (requires artificial clock)
// idx = int(int64(2^31+2) - int64(1)) = int(2^31+1) overflows to -2^31+1 on 32-bit
// retry falls back to 0 instead of the correct timestamp offset
```

## Suggested fix

Keep the arithmetic in `int64` and guard the conversion:

```go
diff := cur - limit   // int64
if diff >= 0 && diff < int64(len(valid)) {
    retry = time.Until(valid[diff].Add(window))
}
```

This is a one-line arithmetic fix with no semantic change on 64-bit platforms.

## Session

`rev-20260615T0810Z-961067` — automated review of HEAD `9d0ba1b`

eunolabs:
Second batch from review session `rev-20260615T0810Z-961067` (filed after #533 was opened).

## Bugs addressed
- Closes #531 — **High**: `JWTPDP.ValidateToken` had no forced-refresh fallback for kid-less tokens. When the IdP rotates its signing key and clients use JWTs without a `kid` header, `FindKeys(keys, "")` returns every cached key so the kid-miss path never runs; once all cached keys fail signature verification the token was rejected for up to a full cache TTL. Extracted the per-key verification into a `tryKeys` helper and added a single rate-limited `ForceRefreshForVerify` retry for kid-less tokens, mirroring the existing `JWKSClient.VerifyToken` fix (#501). Regression test (`TestJWTPDP_ValidateToken_KidlessKeyRotation`) added.
- Closes #532 — **Low**: `InMemory.IncrementIfBelow` computed its `retryAfter` index as `int(cur-limit)`, which truncates to int32 on 32-bit platforms; a difference above `math.MaxInt32` wrapped negative and dropped the estimate to the full-window fallback. Index arithmetic is now kept in `int64`.

## Testing
- `gofmt -l` — clean
- `golangci-lint run ./...` — 0 issues
- `go test -race -count=1 ./...` — all packages pass (new regression test for #531)

## Note on branching
Landed on the session's designated development branch (`claude/hopeful-thompson-r0ixf5`), rebased onto the now-merged `main` from PR #533.

https://claude.ai/code/session_01ETDDyRNynutMatUj2minaL

---
_Generated by [Claude Code](https://claude.ai/code/session_01ETDDyRNynutMatUj2minaL)_


eunolabs:
✅ Ready to merge — both session `961067` bugs addressed: #531 (kid-less JWKS rotation forced-refresh in `JWTPDP.ValidateToken`) and #532 (int64 retryAfter index on 32-bit). Self-review found no substantive issues (the `tryKeys` refactor is behavior-preserving and mirrors the reviewed `JWKSClient.VerifyToken` #501 pattern). CI is fully green (all 42 checks), and local `go test -race`, `go vet`, `gofmt`, and `golangci-lint` (0 issues) all pass.

---

PR #544, opened 2026-06-15
Fix session rev-20260615T0910Z-6fd4dd bugs

### LINKED ISSUE #535: [rev-20260615T0910Z-6fd4dd] TokenCache.Get releases RLock before expiry check, allowing revoked token to be served
**Location:** `pkg/capability/token_cache.go:151-167`
**Reference:** https://github.com/eunolabs/eunox/blob/0682111917f6f47627b5d4e8d367fda4da61f50d/pkg/capability/token_cache.go#L151-L167
**Severity:** Medium (security)

**Description:**
`Get` releases the read lock before checking whether the entry has expired and before returning the payload:

```go
func (c *TokenCache) Get(tokenStr string) (*TokenPayload, bool) {
    key := tokenKey(tokenStr)
    now := c.cfg.Now()

    c.mu.RLock()
    entry, ok := c.entries[key]
    c.mu.RUnlock()      // ← lock released here

    if !ok {
        return nil, false
    }
    if now.After(entry.expiresAt) {   // ← expiry check after unlock
        return nil, false
    }
    return entry.payload, true        // ← payload returned after unlock
}
```

Between `c.mu.RUnlock()` and the `return entry.payload, true`, another goroutine can call `Invalidate(tokenStr)` which acquires the write lock, deletes the entry from the map, and removes it from `insertOrder`. After `Invalidate` returns, `Get` still holds a valid `entry` pointer (the GC keeps it alive) and sees a non-expired `expiresAt`, so it returns the payload of the now-revoked token.

The documentation for `Invalidate` says it is called "on explicit revocation events so that the cached entry does not outlive the propagation of the revocation." This TOCTOU race means a revocation can silently fail to take effect for an in-flight `Get` caller, defeating the explicit revocation path — the attack window exists between the read-lock release and the expiry check.

**Trigger / repro:**
1. Token is cached and valid.
2. Concurrent goroutine calls `Invalidate(token)`.
3. A simultaneous `Get(token)` read the entry pointer under the lock, released the lock, is preempted.
4. `Invalidate` completes and removes the entry.
5. `Get` resumes, finds the (now-orphaned but GC-alive) entry is not expired, returns the payload.

This is reproducible under any `GOMAXPROCS > 1` workload where revocation and lookup occur concurrently (e.g. Redis kill-switch signaling `Invalidate` while a request is mid-evaluation).

**Suggested fix:**
Hold the read lock across the expiry check so `Invalidate`'s write lock cannot interleave:

```go
func (c *TokenCache) Get(tokenStr string) (*TokenPayload, bool) {
    key := tokenKey(tokenStr)
    now := c.cfg.Now()

    c.mu.RLock()
    defer c.mu.RUnlock()

    entry, ok := c.entries[key]
    if !ok {
        return nil, false
    }
        return nil, false
    }
    return entry.payload, true
}
```

The read lock is held for a fixed O(1) map lookup and a single time comparison — no I/O or blocking — so contention cost is negligible.

**Session:** rev-20260615T0910Z-6fd4dd

### LINKED ISSUE #536: [rev-20260615T0910Z-6fd4dd] cnf.jkt DPoP sender-constraint parsed but never enforced — sender-constrained tokens accepted as plain bearer tokens
**Location:** `pkg/capability/token.go:68-73`
**Reference:** https://github.com/eunolabs/eunox/blob/0682111917f6f47627b5d4e8d367fda4da61f50d/pkg/capability/token.go#L68-L73
**Severity:** High (security)

**Description:**
The `Confirmation` struct carries the RFC 7800 `cnf` claim, with `JKT` as the JWK SHA-256 thumbprint of the sender's proof-of-possession key. The code's own comment acknowledges the gap explicitly:

```go
// LIMITATION: this field is PARSED but NOT ENFORCED. The proxy does not verify a
// presented DPoP/proof-of-possession against `cnf.jkt`, so a token marked
// sender-constrained is currently accepted as a plain bearer token with no
// possession check. Do not rely on `cnf.jkt` for sender constraint until a
// verification path (and the corresponding threat-model update) lands;
```

A token that carries `cnf.jkt` is structurally advertising that it is bound to a specific key and MUST NOT be accepted without a proof of possession from that key. Because the proxy never verifies the proof, any party that intercepts or exfiltrates the token — without access to the corresponding private key — can replay it freely against the proxy. The entire purpose of sender-constrained tokens is to defeat replay attacks; the current behavior silently downgrades every `cnf.jkt`-bearing token to an ordinary bearer token.

**Trigger / repro:**
1. Issue an eunox-native capability token with `cnf.jkt` set to a valid JWK thumbprint.
2. Copy that token to any other process or attacker machine.
3. Present the token to the proxy without a DPoP proof header.
4. The proxy accepts the token and authorizes the request, ignoring the sender constraint entirely.

**Suggested fix:**
Implement proof-of-possession verification on the token-validation path. At minimum, once `cnf.jkt` is present in a verified token:

1. Require a `DPoP` header on the request.
2. Verify the DPoP proof's JWK thumbprint matches `cnf.jkt` (using `crypto/ecdsa`, `crypto/rsa`, or `filippo.io/edwards25519` depending on the algorithm).
3. Verify the DPoP proof's `htm`/`htu` claims match the current request method and URI, and `iat` is within a short replay window.
4. Deny with a new `SENDER_CONSTRAINT_FAILED` error code if any check fails.

Until this is implemented, any manifest that relies on token sender-binding for security is operating under a false assumption. The threat model (`docs/threat-model-mcp.md`) should be updated to call this out as a current gap.

**Session:** rev-20260615T0910Z-6fd4dd

### LINKED ISSUE #537: [rev-20260615T0910Z-6fd4dd] Engine DecidedAt / input.context.timestamp uses RFC3339 (second precision) while audit log uses RFC3339Nano
**Location:** `pkg/enforcement/engine.go:485,713`
**Reference:** https://github.com/eunolabs/eunox/blob/0682111917f6f47627b5d4e8d367fda4da61f50d/pkg/enforcement/engine.go#L485
**Severity:** Low-Medium

**Description:**
Both `ValidateAction` and `EvaluateConditions` format the decision timestamp with second-precision RFC3339:

```go
// pkg/enforcement/engine.go:485 (ValidateAction)
now := e.clock.Now().UTC().Format(time.RFC3339)

// pkg/enforcement/engine.go:713 (EvaluateConditions)
now := e.clock.Now().UTC().Format(time.RFC3339)
```

The audit sink, however, stamps every record with nanosecond-precision RFC3339Nano:

```go
// cmd/eunox-mcp/audit.go
Time: s.clock().UTC().Format(time.RFC3339Nano),
```

This has two consequences:

1. **`EnforceResponse.DecidedAt` loses sub-second precision.** Any consumer that correlates enforcement responses with audit records by timestamp may fail to find the matching record when multiple decisions occur within the same wall-clock second.

2. **`input.context.timestamp` in Rego loses sub-second precision.** `now` is stored in the context via `ctx = context.WithValue(ctx, ctxTimestampKey{}, now)` and surfaced to `PolicyEvaluator` implementations as `input.context.timestamp`. Issue #321 added this field specifically so time-based Rego rules evaluate against the real request time. A Rego rule that gates on sub-second timing (e.g. a short burst window, rate-limiting by millisecond) silently receives truncated timestamps, producing incorrect decisions that are invisible at the Go level.

**Trigger / repro:**
1. Write a `policy` condition whose Rego rule compares `input.context.timestamp` to a sub-second boundary (e.g. `time.parse_ns("RFC3339Nano", input.context.timestamp) % 1000000000 < 100000000`).
2. The rule will always receive a timestamp truncated to the second (e.g. `"2026-06-15T09:10:00Z"` instead of `"2026-06-15T09:10:00.123456789Z"`), making the sub-second portion always evaluate to zero.
3. Audit records for the same decision will carry the nanosecond timestamp, diverging from `DecidedAt`.

**Suggested fix:**
Change both `Format` calls to use `time.RFC3339Nano`:

```go
now := e.clock.Now().UTC().Format(time.RFC3339Nano)
```

This is backward-compatible: RFC3339Nano is a strict superset of RFC3339 and is already accepted everywhere RFC3339 is (standard libraries parse either). The `wiretapAllow()` function in `pdp.go:167` uses `time.RFC3339` for the same field and should be updated for consistency.

**Session:** rev-20260615T0910Z-6fd4dd

### LINKED ISSUE #538: [rev-20260615T0910Z-6fd4dd] Audit-mode argumentSchema failure skips recordAuditModeAntecedent, leaving sequenceBlock open
**Location:** `cmd/eunox-mcp/pdp.go:250-264` (argumentSchema check in `Decide`) and `cmd/eunox-mcp/pdp.go:298` (`recordAuditModeAntecedent` call)
**Reference:** https://github.com/eunolabs/eunox/blob/0682111917f6f47627b5d4e8d367fda4da61f50d/cmd/eunox-mcp/pdp.go#L250-L300
**Severity:** Medium (security — sequenceBlock fail-open)

**Description:**
`ManifestPDP.Decide` validates `argumentSchema` before calling `EvaluateConditions`, and the `recordAuditModeAntecedent` call that records the antecedent in session history is placed only after `EvaluateConditions` returns:

```go
// step 4 — argumentSchema (early return, no recordAuditModeAntecedent)
if matched.ArgumentSchema != nil {
    if err := enforcement.ValidateArgumentSchema(args, matched.ArgumentSchema); err != nil {
            // AuditOnly: true (set by the deferred func)
            ...
        }
    }
}

// step 5
// ...
recordAuditModeAntecedent(ctx, p.engine, req, matched, resp)  // ← never reached on argSchema failure
return resp
```

When a constraint is in audit (`enforcement: audit`) mode and a call arrives with arguments that fail `argumentSchema` validation:

1. The early return at step 4 fires with `Decision=Deny, AuditOnly=true` (the defer sets `AuditOnly`).
2. The transport sees `AuditOnly=true` and **forwards the request to the upstream** — the tool runs.
3. `recordAuditModeAntecedent` is **never called**, so the session-history counter for this tool is never incremented.
4. A subsequent call guarded by `sequenceBlock: afterTools: [<this tool>]` calls `Peek` on an empty history key and concludes the antecedent never ran — **the block fails open**.

Issue #354 added `recordAuditModeAntecedent` precisely to fix this gap on the condition-evaluation path. The same gap exists on the `argumentSchema` path and was not closed at the same time.

**Trigger / repro:**
1. Define a manifest with tool `A` marked `enforcement: audit` and an `argumentSchema` that rejects the incoming request shape.
2. Define tool `B` with `sequenceBlock: afterTools: [A]`.
3. Send a call to `A` with args that fail the schema (e.g. a required field missing).
4. The call is logged as a deny (audit mode) and forwarded. Tool A runs.
5. Immediately send a call to `B`.
6. `B`'s `sequenceBlock` Peek sees count=0 for `A` and allows the call, despite `A` having just executed.

**Suggested fix:**
Move or duplicate the `recordAuditModeAntecedent` call to cover the `argumentSchema` early-return path. The cleanest fix is to call it unconditionally on any deny-with-AuditOnly response, regardless of which step produced the denial. For example:

```go
if matched.ArgumentSchema != nil {
    if err := enforcement.ValidateArgumentSchema(args, matched.ArgumentSchema); err != nil {
        resp = capability.EnforceResponse{
            AuditOnly: matched.IsAuditOnly(),
                Code:          capability.ErrCodeInvalidParams,
                ConditionType: "argumentSchema",
                Message:       err.Error(),
            },
        }
        if matched.IsAuditOnly() {
        }
        return resp
    }
}
```

Alternatively, restructure `Decide` so all deny paths flow through a common exit that calls `recordAuditModeAntecedent` when appropriate, instead of scattering early returns.

Note: The same issue may apply to the actions-list check (step 3), though that path is less common in practice since a well-formed manifest always includes the required action.

**Session:** rev-20260615T0910Z-6fd4dd

eunolabs:
## Bugs addressed (session rev-20260615T0910Z-6fd4dd)

- Closes #535 — `TokenCache.Get` released the RLock before the expiry check, leaving a TOCTOU window where a concurrent `Invalidate` (explicit revocation) could delete the entry yet `Get` still returned the revoked token's payload. The lock is now held across the lookup and the expiry check (the critical section is an O(1) map lookup + one time comparison, no I/O).
- Closes #536 — `cnf.jkt` (RFC 7800 sender constraint) was parsed but never enforced, so a verified native capability token advertising a sender binding was accepted as a plain bearer token and could be replayed by anyone who captured it. The proxy has no DPoP/proof-of-possession path, so `JWKSClient.VerifyToken` now **fails closed**: a verified token carrying a non-empty `cnf.jkt` is rejected as a terminal error. Updated the `Confirmation` doc and added a threat-model changelog entry (1.24).
- Closes #537 — `Engine.ValidateAction`/`EvaluateConditions` formatted `DecidedAt` and the Rego `input.context.timestamp` with second-precision RFC3339 while the audit sink uses RFC3339Nano, so the enforcement timestamp lost sub-second precision (broke audit-record correlation and any sub-second time-based Rego rule). Both now format with RFC3339Nano (a strict superset of RFC3339); `wiretapAllow` updated for consistency.
- Closes #538 — `ManifestPDP.Decide` called `recordAuditModeAntecedent` only after `EvaluateConditions`, so an audit-mode entry that denied-and-forwarded on the **actions** or **argumentSchema** early returns ran the tool without recording the antecedent in session history — a later `sequenceBlock` then Peeked an empty history and failed open. The `EnforceRequest` is now built up front and the antecedent recorded on every audit-downgradeable early-return deny, in `Decide` and the analogous `DecideResourceRead`/`DecidePromptGet` paths.

## Testing
- New regression tests: `TestJWKSClient_VerifyToken_SenderConstrainedRejected` (#536), strengthened `TestValidateAction_TimestampReachesRegoInput` to a sub-second clock (#537), `TestManifestPDP_SequenceBlock_AuditModeArgSchemaAntecedentArmsBlock` (#538).
- `make build`, `make check-fmt`, `make check-license`, `make lint` (go vet) — all clean.

---


eunolabs:
✅ Ready to merge — all session rev-20260615T0910Z-6fd4dd bugs addressed (#535, #536, #537, #538), self-review complete with no outstanding comments, tests passing (all 26 CI checks green; `go test -race ./...`, build, gofmt, license, lint all clean locally).

---

PR #549, opened 2026-06-15
Fix session rev-20260615T1010Z-958151 bugs

### LINKED ISSUE #540: [rev-20260615T1010Z-958151] circuit breaker halfOpenGen not incremented on HalfOpen→Closed allows stale probe failures to spuriously trip closed breaker
**Location:** `pkg/circuitbreaker/breaker.go:296–304` (`recordSuccess`, HalfOpen→Closed transition)
**Reference:** https://github.com/eunolabs/eunox/blob/0682111917f6f47627b5d4e8d367fda4da61f50d/pkg/circuitbreaker/breaker.go#L296-L304
**Severity:** Medium (correctness — spurious circuit-breaker trips under concurrent load)

**Description:**
`recordSuccess` transitions the breaker from HalfOpen to Closed when enough probe successes accumulate, but it does **not** increment `halfOpenGen`:

```go
// pkg/circuitbreaker/breaker.go ~296–304
if b.halfOpenSuccess >= b.requiredHalfOpenSuccesses() {
    b.state = StateClosed
    b.halfOpenProbes = 0
    b.halfOpenSuccess = 0
    b.lastTransitionAt = b.now()
    // halfOpenGen is NOT incremented here — stale probes still share the same gen
}
```

Compare with the two other places that invalidate the current half-open window:
- `Reset()` (line ~433): explicitly calls `b.halfOpenGen++`
- `allow()` Open→HalfOpen transition (line ~270): explicitly calls `b.halfOpenGen++`

`Do[T]` in `do.go` calls `AllowProbe()`, which returns a probe bound to the current `halfOpenGen`. After the probe's `fn` finishes, `probe.Failure()` calls `recordFailure(guard=true, gen=<captured gen>)`. Inside `recordFailure`:

```go
// guard path
if b.state == StateHalfOpen && gen != b.halfOpenGen {
    return  // ← drop stale probe
}
```

Because `halfOpenGen` was **not** incremented when the breaker closed, a slow probe that was admitted during the half-open window still has `gen == b.halfOpenGen` after the breaker has transitioned to Closed. `recordFailure` therefore does **not** drop it as stale. Instead it:
1. Sets `b.state = StateOpen` (re-opens a now-healthy closed breaker)
2. Increments `b.consecutiveFails`

This means any probe that was in-flight at the moment the breaker closed and subsequently fails (e.g. a JWKS fetch that times out after the endpoint recovered) will spuriously trip the breaker back open, causing a brief availability disruption for a healthy upstream.

**Trigger / repro:**
1. Configure a breaker with `HalfOpenMaxProbes ≥ 2`.
2. The breaker is in HalfOpen; two probes are admitted concurrently.
3. Probe A succeeds first and causes the breaker to close (`halfOpenSuccess >= required`).
4. Probe B, which was slow, now calls its underlying fn; the fn fails (network blip).
5. `probe.Failure()` is called with the original `gen`. Since `halfOpenGen` was not incremented at step 3, `gen == b.halfOpenGen` still holds.
6. `recordFailure` sees `state == StateClosed` (not HalfOpen), so the `guard` drop condition (`state == StateHalfOpen && gen != b.halfOpenGen`) evaluates false.
7. The failure is counted, `consecutiveFails` increments, and the breaker may reopen.

**Suggested fix:**
Add `b.halfOpenGen++` in `recordSuccess` immediately after setting `b.state = StateClosed`:

```go
if b.halfOpenSuccess >= b.requiredHalfOpenSuccesses() {
    b.state = StateClosed
    b.halfOpenProbes = 0
    b.halfOpenSuccess = 0
    b.halfOpenGen++             // ← invalidate probes still in-flight from the old window
    b.lastTransitionAt = b.now()
}
```

This makes the HalfOpen→Closed transition consistent with `Reset()` and `allow()`, and ensures any slow probe that completes after the breaker closes is correctly identified as stale and dropped by `recordFailure`'s guard check.

**Session:** rev-20260615T1010Z-958151

### LINKED ISSUE #541: [rev-20260615T1010Z-958151] TokenCache.Put evicts live entry instead of expired entry in middle of insertion-order list, wasting cache capacity
**Location:** `pkg/capability/token_cache.go:174–249` (`Put`, expired-entry reclaim + capacity eviction loops)
**Reference:** https://github.com/eunolabs/eunox/blob/0682111917f6f47627b5d4e8d367fda4da61f50d/pkg/capability/token_cache.go#L174-L249
**Severity:** Medium (correctness — live tokens evicted while expired tokens linger, degrading cache hit rate and causing unnecessary re-verification)

**Description:**
`Put` performs two loops when the cache is at or above capacity:

**Loop 1 — expired entry reclaim (lines ~186–200):**
```go
for {
    front := c.insertOrder.Front()
    oldest := front.Value.(string)
    e, ok := c.entries[oldest]
    if !ok || !now.After(e.expiresAt) {
        break   // ← stops at first live entry from the front
    }
    c.insertOrder.Remove(front)
    delete(c.entries, oldest)
}
```
This scans only a **contiguous prefix** of expired entries from the front of the list. It stops at the first live entry, even if expired entries exist further along (in the middle or rear of the list).

**Why expired entries exist in the middle:**
Token TTLs are `min(cfg.MaxEntryTTL, tokenRemaining)`. A token with a short `tokenRemaining` (e.g. 5 s) inserted after a long-lived token (e.g. 1 h) will expire and sit behind the long-lived token in the list. Loop 1 will never reach it.

**Loop 2 — LRU eviction (lines ~201–210):**
```go
    front := c.insertOrder.Front()
    oldest := front.Value.(string)
    c.insertOrder.Remove(front)
    delete(c.entries, oldest)   // ← evicts the front, which is now LIVE
}
```
Because Loop 1 stopped at the first live entry, the list front is now a valid token. Loop 2 evicts it to make room, while the expired entry in the middle continues to occupy a slot.

**Net effect:**
- A live token is discarded, requiring the next request bearing that token to re-verify it (JWT signature check + possible JWKS fetch through the circuit breaker).
- The expired entry remains in the map and list until the background `purgeExpired` sweep runs (background goroutine, periodic interval), wasting a slot.
- Under sustained load with many short-lived tokens, the working set of active tokens can be squeezed to well below `MaxSize` because expired entries hold the slots.

**Trigger / repro:**
1. Set `MaxSize = 3`.
2. Insert token A (`expiresAt = now + 1h`) → list: [A].
3. Insert token B (`expiresAt = now + 5s`) → list: [A, B].
4. Insert token C (`expiresAt = now + 1h`) → list: [A, B, C]. Cache at capacity.
5. Advance clock by 10s (B is now expired).
6. Insert token D. `Put` runs Loop 1: front is A (live), breaks immediately. Loop 2: list length is 3 >= MaxSize, evicts front (A). D is inserted. List: [B(expired), C, D].
7. A subsequent `Get(A)` misses; A must be re-verified. B's expired entry persists.

**Suggested fix:**
Before falling through to Loop 2, do a full scan of the list for any expired entry to evict first:

```go
// After the contiguous-prefix scan, do a full pass for any remaining expired entry.
    for el := c.insertOrder.Front(); el != nil; el = el.Next() {
        key := el.Value.(string)
        if e, ok := c.entries[key]; ok && now.After(e.expiresAt) {
            c.insertOrder.Remove(el)
            delete(c.entries, key)
            break
        }
    }
}
// Then the existing capacity-eviction loop as a fallback.
    front := c.insertOrder.Front()
    // ... evict front (now truly the oldest live entry)
}
```

Alternatively, maintain a min-heap sorted by `expiresAt` alongside `insertOrder` so expired entries are always O(log n) to find and remove; this keeps eviction fast even with large `MaxSize`.

**Session:** rev-20260615T1010Z-958151

### LINKED ISSUE #542: [rev-20260615T1010Z-958151] audit-verify misclassifies legitimate seq-0 record as forged decoy after uint64 sequence number wraps
**Location:** `cmd/eunox-mcp/audit.go` — `verifyAuditLog`, around the `forgedSeq0` check (approximately line 1258)
**Reference:** https://github.com/eunolabs/eunox/blob/0682111917f6f47627b5d4e8d367fda4da61f50d/cmd/eunox-mcp/audit.go
**Severity:** Low (correctness — false-positive tamper alert after extremely long-running deployment)

**Description:**
`verifyAuditLog` contains a guard that detects a forged sequence-0 record injected into an existing log to reset the HMAC chain:

```go
// (paraphrased from verifyAuditLog)
forgedSeq0 := rec.Seq == 0 && rec.HMAC != ""
if forgedSeq0 {
    // report as INVALID / tampered
}
```

The intent is correct: in a legitimate log, `Seq == 0` only ever appears as the very first record in a new log file, where `HMAC` is computed from the genesis `prevHMAC` and the record body. A record with `Seq == 0` appearing mid-stream alongside a non-empty `HMAC` can only be a forged reset attempt.

However, `s.seq` in `auditSink` is a `uint64` incremented unconditionally for every record written. After exactly **2^64 records**, `s.seq` wraps to `0`. The next legitimately signed record will have `Seq = 0` and a valid non-empty `HMAC` (correct in the chain). `verifyAuditLog` will classify this record as `forgedSeq0 = true` and report the log as tampered, even though no tampering occurred.

There is an analogous check `forgedLegacySeq0` for the legacy signed-body format that has the same property.

**Practical impact:**
At 1 record per millisecond continuously, overflow occurs after approximately 584 million years. This is not a near-term operational risk. However:
- The codebase has a stated goal of being a long-lived, auditable security component.
- The `audit-verify` command is the authoritative tamper-detection tool; a false positive from it in any foreseeable deployment is a correctness defect.
- A simpler invariant can be stated that avoids the edge case entirely.

**Trigger / repro:**
1. Write exactly 2^64 audit records to the same log (or manipulate `s.seq` in a test to start near `math.MaxUint64`).
2. The record immediately following the wrap will have `Seq = 0` and a valid HMAC.
3. `audit-verify` will emit a tamper-detection error for this record.

**Suggested fix:**
Replace the `Seq == 0 && HMAC != ""` heuristic with a position-in-file check. The forged-seq-0 attack is only viable when a record with `Seq == 0` appears **after at least one other record in the file** (i.e. it is not the first line). Track whether at least one record has been successfully processed:

```go
var seenAnyRecord bool
// ... in the per-record loop:
if rec.Seq == 0 && rec.HMAC != "" && seenAnyRecord {
    // forged seq-0 decoy
}
seenAnyRecord = true
```

This eliminates the false positive after wrap: the wrapped record will be the first record in a new log rotation (log files are rotated well before 2^64 records), so `seenAnyRecord` will be false when it is encountered. If log rotation is not guaranteed to happen before overflow — though in practice it always will — the check can additionally gate on `rec.Seq != math.MaxUint64+1_mod_whatever` but the positional check is sufficient and simpler.

**Session:** rev-20260615T1010Z-958151

eunolabs:
## Bugs addressed (session rev-20260615T1010Z-958151)

- Closes #540 — The circuit breaker's `recordSuccess` closed the breaker (HalfOpen→Closed) **without** incrementing `halfOpenGen`, unlike the Open→HalfOpen transition in `AllowProbe` and `Reset`. A probe still in flight from the just-closed window therefore kept a matching generation, so its late `Failure` passed `recordFailure`'s stale-probe guard (which runs in `StateClosed`, where the open-state guard does not apply) and could spuriously re-open a healthy breaker. The generation is now bumped on close, so any in-flight probe from that window is recognized as stale and dropped.
- Closes #541 — `TokenCache.Put`'s expired-entry reclaim only scanned the contiguous run of expired entries at the **front** of the insertion-order list. Since per-token TTLs are `min(MaxEntryTTL, tokenRemaining)`, a short-lived token inserted after a long-lived one expires while sitting *behind* a live entry, so the front scan never reached it and the capacity loop then evicted a **live** front entry while the expired middle entry kept its slot. Added a full-list sweep for expired entries, gated on still being at capacity after the cheap front reclaim so the O(1) fast path is preserved.
- Closes #542 — After the `uint64` sequence counter wraps (2^64 records), the next legitimately signed record carries `seq 0` with a valid HMAC, which `verifyAuditLog` flagged as a forged seq-0 decoy (false-positive tamper alert). Exempt **only** the genuine wrap case — a seq-0 record whose immediately preceding legitimate record had `seq == math.MaxUint64`. This is deliberately tighter than the originally-suggested "first record in file" positional check, which would not reliably fire (rotation is by size, independent of record count) **and** would re-open the genesis-substitution vector #446 closed by rejecting a seq-0+HMAC record even as the first line. Safe because the predecessor MaxUint64 record must itself verify (its HMAC needs the signing key).

## Testing
  - `TestBreaker_StaleProbeAfterCloseDoesNotReopen` (#540)
  - `TestTokenCache_PutEvictsExpiredMiddleEntryNotLiveFront` (#541)
  - `TestVerifyAuditLog_PostWrapSeq0_NotForged` + `TestVerifyAuditLog_Seq0AfterNonWrap_StillForged` (#542)

### Note on #542 scope
The issue's suggested `seenAnyRecord` positional fix was **not** applied as written: it would regress the #446 hardening and not reliably fix the wrap. The implemented wrap-anchored exemption resolves the false positive while preserving every existing tamper-detection guarantee (#446/#477/#484) — see the inline comment for the full rationale.

---


eunolabs:
✅ Ready to merge — all session rev-20260615T1010Z-958151 bugs addressed (#540, #541, #542), self-review complete with no outstanding comments, tests passing (all 26 CI checks green; each fix has a regression test verified to fail without it; `go test -race ./...`, build, gofmt, license, lint all clean).

---

PR #550, opened 2026-06-15
Fix session rev-20260615T1110Z-e294b5 bugs

### LINKED ISSUE #543: [rev-20260615T1110Z-e294b5] Kill switch bypassed in JWTPDP.Decide when JWT capability check denies before inner PDP is called
**Location:** `cmd/eunox-mcp/pdp_jwt.go:519–606` (`JWTPDP.Decide`)
**Reference:** https://github.com/eunolabs/eunox/blob/0682111917f6f47627b5d4e8d367fda4da61f50d/cmd/eunox-mcp/pdp_jwt.go#L519-L606
**Severity:** Medium (security / audit integrity — kill switch state not consulted; request is still denied, but denial code and metrics are wrong)

**Description:**

`JWTPDP.Decide` contains an optimisation that skips the kill-switch check when `innerSharesKillSwitch()` returns true (the inner `ManifestPDP` uses the same kill-switch manager):

```go
// pdp_jwt.go:527-531
if !p.innerSharesKillSwitch() {
        return *deny
    }
}
```

The comment states that this is safe because the delegated inner `ManifestPDP.Decide` performs the identical check. However, the optimisation assumes the inner PDP is **always** called — an assumption that is violated by two early-return paths that precede `decideInner`:

**Path 1 — target not in JWT capability claims (line 560-562):**
```go
constraints := buildConstraintsFromClaims(claims.Capabilities, target)
if len(constraints) == 0 {
    return denyResponse(capability.ErrCodeAuthorizationFailed, "jwtCapability",
        fmt.Sprintf("%s %q is not in the JWT capability claims", target.Type, target.Name))
}
```

**Path 2 — JWT shorthand conditions fail (line 595-597):**
```go
if lastDeny != nil {
    return *lastDeny
}
```

In both cases the kill-switch manager (`p.ks`, which is non-nil) is never consulted, yet it is not checked at the top either (because `innerSharesKillSwitch()` returned true). The same issue applies to `DecideResourceRead` and `DecidePromptGet`, which delegate through `Decide`.

**Trigger / repro:**

1. Configure a route with both `--jwks-uri` and `--policy` (intersection mode). The `ManifestPDP` and `JWTPDP` share the same kill-switch manager; `innerSharesKillSwitch()` returns true.
2. Operator issues a `/control/kill session=<session-id>` command, setting the kill flag in the store.
3. The agent (session-id) sends a `tools/call` for a tool that is **in the manifest** but **absent from the JWT capability claims** (or present but with conditions that the call arguments fail).
4. `JWTPDP.Decide` skips the kill-switch check (optimisation), finds the tool is not in the JWT allowlist (path 1) or the conditions fail (path 2), and returns `AUTHORIZATION_FAILED` / `CONDITION_FAILED` without ever consulting the kill-switch store.
5. The audit record carries the wrong denial code; kill-switch metrics do not fire; the operator cannot confirm from the log that the kill command was effective for this session.

**Effect:**

- The request **is** denied, so enforcement is not bypassed from the caller's perspective.
- The kill-switch check is completely absent for these paths, however:
  - Audit records show `AUTHORIZATION_FAILED` or `CONDITION_FAILED` rather than `KILL_SWITCH_ACTIVATED`.
  - Any kill-switch-specific alerting (e.g. SIEM rules matching `denial_code = KILL_SWITCH`) does not fire.
  - An operator reviewing the audit log cannot distinguish "the agent was legitimately denied by capability policy" from "the operator killed the session and it tried to call something the JWT denied first."

**Suggested fix:**

Remove the `innerSharesKillSwitch()` optimisation from `Decide()` and always perform the kill check unconditionally before any JWT evaluation:

```go
func (p *JWTPDP) Decide(ctx context.Context, sessionID string, target EnforceTarget, args map[string]interface{}, sourceIP string) capability.EnforceResponse {
    // Always check the kill switch before any JWT evaluation so a killed session
    // gets KILL_SWITCH_ACTIVATED regardless of whether the JWT would also deny.
        return *deny
    }
    // ... rest unchanged ...
}
```

The single redundant kill-store round-trip that this reintroduces (when the inner ManifestPDP would have checked the same store) is a negligible cost compared to the correctness benefit. The optimisation note in the comment can reference the now-correct invariant: "the kill check is always at the top of Decide, so the inner PDP's check is redundant but harmless."

`DecideSampling` is unaffected: when `innerSharesKillSwitch()` is true the inner PDP is always a `ManifestPDP` which implements `SamplingAuthorizer`, so `DecideSampling` always reaches the inner's kill check.

**Session:** rev-20260615T1110Z-e294b5

eunolabs:
## Bugs addressed (session rev-20260615T1110Z-e294b5)

- Closes #543 — `JWTPDP.Decide` skipped the wrapper's kill-switch check whenever the inner `ManifestPDP` shared the kill-switch manager (the #440 dedup), on the assumption that the inner's identical check is always reached. But `Decide` has early-return paths that **never** reach the inner — a target absent from the JWT capability claims, and JWT shorthand conditions failing — so on those paths the kill switch went entirely unconsulted. A killed session was still denied, but with `AUTHORIZATION_FAILED` / `CONDITION_FAILED` instead of `KILL_SWITCH`, kill-switch metrics/alerting did not fire, and the audit log could not show that the kill took effect. The kill check is now unconditional at the top of `Decide`. `DecideResourceRead` / `DecidePromptGet` delegate to `Decide`, so all three host-initiated paths are covered; `DecideSampling` is unaffected (its delegate always implements the inner kill check).

## Testing
- New `TestJWTPDP_Decide_KillSwitchOnEarlyReturnPaths` (#543) — verifies a killed session is denied with `KILL_SWITCH` on both early-return paths (target-not-in-claims and conditions-fail); confirmed to fail without the fix (it returned `AUTHORIZATION_FAILED` / `OPERATION_NOT_PERMITTED`).
- Updated `TestJWTPDP_Decide_KillSwitchCheckedOnce` → `TestJWTPDP_Decide_KillSwitchAlwaysCheckedInWrapper`: the #440 dedup is intentionally superseded by #543, so the allowed intersection path now consults the kill store twice (wrapper + inner). The existing `TestJWTPDP_Decide_KillSwitchStillEnforced` and the `DecideSampling` kill-switch tests are unchanged and still pass.

### Note
The trade-off is deliberate: the #440 optimization eliminated a redundant kill-store round-trip on the allowed path, but it is unsound because it assumed the inner PDP is always reached. Correctness and fail-closed denial semantics take priority over the micro-optimization; the redundant round-trip on the allowed path is the accepted cost.

---


eunolabs:
✅ Ready to merge — session rev-20260615T1110Z-e294b5 bug addressed (#543), self-review complete with no outstanding comments, tests passing (all functional CI checks green — Test, Lint, Builds, gofmt, License, Enforcement compliance, demo/OPA integration; both early-return paths have a regression test verified to fail without the fix; `go test -race ./...`, build, lint clean locally).

---

PR #551, opened 2026-06-15
Fix session rev-20260615T1211Z-30913c bugs

### LINKED ISSUE #545: [rev-20260615T1211Z-30913c] runInitHandshake does not capture upstreamServerVersion for local subprocess HTTP sessions
**Location:** `cmd/eunox-mcp/http.go:1547–1550` (`runInitHandshake`)

**Reference:** https://github.com/eunolabs/eunox/blob/0682111917f6f47627b5d4e8d367fda4da61f50d/cmd/eunox-mcp/http.go#L1547-L1550

**Severity:** Medium

**Description:**

`runInitHandshake` reads the MCP `initialize` response from the upstream subprocess and unmarshals it into `mcpInitResult`, but only saves `result.Capabilities` — it never extracts `result.ServerInfo["version"]` into `s.upstreamServerVersion`. As a result, every HTTP-transport session that uses a local subprocess upstream starts with `upstreamServerVersion == ""`.

The two other init paths both extract the version correctly:

- **Remote HTTP** (`http_remote.go:178–185`):
  ```go
  if json.Unmarshal(respMsg.Result, &result) == nil {
      s.upstreamCaps = result.Capabilities
      if sv, ok := result.ServerInfo["version"].(string); ok {
          s.upstreamServerVersion = sv   // present
      }
  }
  ```
- **Stdio proxy** (`stdio.go:337–343`):
  ```go
  if err := json.Unmarshal(msg.Result, &result); err == nil {
      p.upstreamCaps = result.Capabilities
      if sv, ok := result.ServerInfo["version"].(string); ok {
          p.upstreamServerVersion = sv   // present
      }
  }
  ```

The buggy path in `runInitHandshake` (`http.go:1547–1550`):
```go
var result mcpInitResult
if err := json.Unmarshal(msg.Result, &result); err == nil {
    s.upstreamCaps = result.Capabilities
    // s.upstreamServerVersion is never set
}
```

**Trigger / repro:**

1. Run `eunox-mcp` in HTTP-transport mode with a local subprocess upstream (not `--upstream-url`).
2. Set `serverVersion: "^1.2.0"` (or any constraint) in the capability manifest.
3. Start a session. The FM-4 drift check (`evaluateDrift` → `CheckManifestDrift`) always sees `serverVersion == ""`, so the pin never matches.
4. With `--strict-drift`, every session is rejected; without it, the version constraint is silently ignored.

`runHTTPDriftCheck` (`drift.go:584`) passes `sess.upstreamServerVersion` to `evaluateDrift`, which in turn passes it to `CheckManifestDrift`; the FM-4 branch at `drift.go:147–154` compares it against `manifest.ServerVersion`. When the value is always `""`, the check always fires.

**Suggested fix:**

```go
var result mcpInitResult
if err := json.Unmarshal(msg.Result, &result); err == nil {
    s.upstreamCaps = result.Capabilities
    if sv, ok := result.ServerInfo["version"].(string); ok {
        s.upstreamServerVersion = sv
    }
}
```

Mirrors the identical three-line block already present in `http_remote.go:182–184` and `stdio.go:340–342`.

**Session:** `rev-20260615T1211Z-30913c`

eunolabs:
## Bugs addressed (session rev-20260615T1211Z-30913c)

- Closes #545 — `runInitHandshake` (the local-subprocess HTTP init path) saved only `result.Capabilities` and never extracted `serverInfo.version` into `s.upstreamServerVersion`, unlike the remote-HTTP (`http_remote.go`) and stdio (`stdio.go`) init paths. Every HTTP-transport session with a local subprocess upstream therefore started with `upstreamServerVersion == ""`, so the FM-4 drift check (`CheckManifestDrift`) always compared against an empty version: a manifest `serverVersion` pin was silently ignored, or — under `--strict-drift` — every session was rejected. Mirror the three-line `serverInfo["version"]` extraction the other two paths already perform.

## Testing
- New `TestRunInitHandshake_CapturesServerVersion` (#545) drives the handshake with an in-memory init response and asserts `upstreamServerVersion` is captured; confirmed to fail without the fix (`upstreamServerVersion == ""`).

---


eunolabs:
✅ Ready to merge — session rev-20260615T1211Z-30913c bug addressed (#545), self-review complete with no outstanding comments, tests passing (all 26 CI checks green; regression test verified to fail without the fix; `go test -race ./...`, build, lint clean locally).

---

Issue #547, opened 2026-06-15
[rev-20260615T1211Z-b4e0ad] JWT v0.2 resource capability claims with ? in URI are misparsed as condition suffixes


eunolabs:
**Location:** `cmd/eunox-mcp/pdp_jwt.go` — `parseV2Claim`
**Reference:** https://github.com/eunolabs/eunox/blob/0682111917f6f47627b5d4e8d367fda4da61f50d/cmd/eunox-mcp/pdp_jwt.go
**Severity:** Medium — access-control bypass in one direction (intended grants silently become denials); affects any deployment where IdP-issued JWT claims reference HTTP/HTTPS resource URIs containing query strings

**Description:**

`parseV2Claim` uses `strings.Cut(claim, "?")` to split a v0.2 JWT capability claim into its `<namespace>:<name>` prefix and an optional `<key>=<value>` condition suffix:

```go
func parseV2Claim(claim string) (parsedV2Claim, error) {
    name, condSuffix, _ := strings.Cut(claim, "?")
    ...
}
```

`strings.Cut` splits on the **first** occurrence of `?`. For HTTP/HTTPS resource URIs — which are a common MCP resource naming convention — the `?` that introduces a query string is indistinguishable from the `?` the v0.2 format uses to introduce conditions.

Consider a claim intended to grant read access to all results of a fixed API query:

```
resource:https://api.example.com/search?q=widget
```

`strings.Cut` produces:
- `name = "resource:https://api.example.com/search"` (truncated URI)
- `condSuffix = "q=widget"` (treated as a condition `AllowedValues{Argument:"q", Values:["widget"]}`)

During `resources/read` enforcement, the condition args map is `{"uri": targetURI}`. The key `"q"` is absent from that map, so `evaluateJWTConditions` cannot satisfy the condition and returns a `MISSING_CONTEXT` denial. The JWT holder is denied access to a resource their claim explicitly grants.

The same truncation means the resource name stored for list-filtering is `"https://api.example.com/search"` (without the query string), so even the `resources/list` filter drops the resource if the upstream names it with the full URL.

**Trigger / repro:**

1. Configure an IdP to issue a JWT with `mcp.capabilities` containing `"resource:https://upstream/data?format=json"`.
2. Present the JWT to eunox-mcp and request `resources/read` for URI `https://upstream/data?format=json`.
3. Observe denial (`MISSING_CONTEXT` or `AUTHORIZATION_FAILED`) even though the claim matches.

**Suggested fix:**

[...]

PR #555, opened 2026-06-15
Fix session rev-20260615T1211Z-b4e0ad bugs

### LINKED ISSUE #546: [rev-20260615T1211Z-b4e0ad] filterListResult serializes filtered-empty list fields as null instead of []
**Location:** `cmd/eunox-mcp/pdp.go:722-750` (filterListResult), `cmd/eunox-mcp/drift.go:465-494` (mcpToolsListResult), `cmd/eunox-mcp/jsonrpc.go:130-157` (mcpResourcesListResult, mcpPromptsListResult)
**Reference:** https://github.com/eunolabs/eunox/blob/0682111917f6f47627b5d4e8d367fda4da61f50d/cmd/eunox-mcp/pdp.go
**Severity:** Medium — protocol conformance break; some MCP clients that type-check for an array will reject the response or crash

**Description:**

`filterListResult` declares its accumulator as `var kept []E`, which is a nil Go slice. When every entry in a `tools/list`, `resources/list`, or `prompts/list` result is filtered out by the manifest policy, `kept` is never appended to and remains nil. The assignment `*field = kept` stores nil back into the struct field. When that struct is later marshalled to JSON, `json.Marshal` encodes a nil slice as `null`, producing e.g. `{"tools":null}` instead of the MCP-spec-required `{"tools":[]}`.

The same defect exists on the error path inside `filterListResult`. The `failClosed` closure does:
```go
failClosed := func() json.RawMessage {
    empty, _ := json.Marshal(*new(L))
    return empty
}
```
`*new(mcpToolsListResult)` zero-initialises the struct, whose `Tools` field is `nil` (no `omitempty`), so `failClosed()` also returns `{"tools":null}`.

The code comment at the assignment site says "empty list" but the semantics are "null field":
```go
*field = kept  // nil here if nothing matched
```

**Trigger / repro:**

1. Start eunox-mcp with a manifest that permits zero tools (e.g. an empty `tools:` block or all entries denied).
2. Connect an MCP host and send `tools/list`.
3. The proxy forwards the upstream response, filters all entries, and returns `{"tools":null}`.
4. A client that does `if result.tools.length === 0` (JS) or dereferences the slice without a nil guard (Go) will panic or misreport.

Reproduce the `failClosed` path by providing a manifest with a `tools` entry whose name is the empty string (fails the name-lookup, triggers failClosed).

**Suggested fix:**

Change the accumulator initialisation in `filterListResult` from `var kept []E` to `kept := make([]E, 0)` so the slice is always non-nil, even when empty:

```go
kept := make([]E, 0)
for _, e := range *field {
    if keep(e) {
        kept = append(kept, e)
    }
}
*field = kept  // now encodes as [] when empty
```

For `failClosed`, explicitly initialise the slice field before marshalling, or use a helper that guarantees the zero value is an empty slice rather than nil.

**Session:** rev-20260615T1211Z-b4e0ad

### LINKED ISSUE #548: [rev-20260615T1211Z-b4e0ad] StartCleanup TOCTOU permanently disables cleanup when context is canceled between pre-check and sync.Once.Do
**Location:** `pkg/callcounter/memory.go` — `StartCleanup`
**Reference:** https://github.com/eunolabs/eunox/blob/0682111917f6f47627b5d4e8d367fda4da61f50d/pkg/callcounter/memory.go
**Severity:** Low — requires a context-cancellation race at startup; consequence is unbounded memory growth from a leaked InMemory counter

**Description:**

`StartCleanup` guards against duplicate goroutines with a `sync.Once`:

```go
func (m *InMemory) StartCleanup(ctx context.Context, interval time.Duration) bool {
    if ctx.Err() != nil {        // (A) pre-check passes
        return false
    }
    started := false
    m.cleanupOnce.Do(func() {   // (B) Once consumed here
        if ctx.Err() != nil {   // (C) context now canceled
            return              //     goroutine not started, but Once is spent
        }
        started = true
        go m.runCleanup(ctx, interval)
    })
    return started
}
```

If the context is canceled in the narrow window between check (A) and `Do` (B), the `Do` callback fires, hits check (C), returns without launching the goroutine, and `started` remains `false`. The caller correctly receives `false` and knows no goroutine is running.

However, `sync.Once.Do` is permanently consumed: **all future calls to `StartCleanup` on this `InMemory` instance, with any context**, will skip the `Do` body entirely and also return `false`. The cleanup goroutine can never be started for the lifetime of the counter.

In a long-running process (gateway mode with persistent sessions, or stdio mode with a long-lived agent), the `InMemory` callcounter then grows without bound as call-history entries accumulate and are never evicted. This is especially relevant for the `sequenceBlock` condition, which relies on the cleanup goroutine to expire 24-hour session history.

The function's public documentation says:

> "Returns false without consuming the Once if ctx is already done."

That contract is violated in this TOCTOU window. The internal comment in the code acknowledges the flaw: "The Once is still consumed here (the documented limitation)." The documentation and the implementation disagree.

**Trigger / repro:**

```go
m := callcounter.NewInMemory()
// Cancel context immediately before/during StartCleanup
cancel() // cancel just before, simulating the race

ok := m.StartCleanup(ctx, time.Minute)
// ok == false (correct, no goroutine)

// Now try with a valid context — should start cleanup but doesn't
ok2 := m.StartCleanup(context.Background(), time.Minute)
// ok2 == false (wrong — Once is already spent)
```

In production the race is narrow but real: a context derived from an HTTP request, a signal handler, or a timeout deadline can be canceled concurrently.

**Suggested fix:**

Remove the `sync.Once` pre-check entirely and instead return early *inside* the `Do` callback only. To preserve the "Once not consumed on invalid context" guarantee, use an atomic flag instead of `sync.Once`:

```go
func (m *InMemory) StartCleanup(ctx context.Context, interval time.Duration) bool {
    if ctx.Err() != nil {
        return false
    }
    if !m.cleanupStarted.CompareAndSwap(false, true) {
        return false  // already started
    }
    go m.runCleanup(ctx, interval)
    return true
}
```

`cleanupStarted` would be a `sync/atomic.Bool` on the struct. This removes the TOCTOU window: the CAS is atomic, and a failed CAS means another goroutine won the race (not that the context was canceled). The context-cancel case is handled by the pre-check; if the race is lost between pre-check and CAS, the goroutine starts with a canceled context and `runCleanup` exits immediately on first tick — benign.

**Session:** rev-20260615T1211Z-b4e0ad

eunolabs:
## Bugs addressed (session rev-20260615T1211Z-b4e0ad)

- Closes #546 — `filterListResult` accumulated kept entries in a nil-initialized slice (`var kept []E`), so when every `*/list` entry was filtered out the field marshaled to JSON `null` instead of the MCP-spec-required `[]`. The `failClosed` closure had the same defect via `*new(L)`. Initialize `kept` with `make([]E, 0)` and force the `failClosed` field to an empty slice, so the list always serializes as an array — never `null` (which crashes clients that type-check or index it), and never the original unfiltered bytes (#318).
- Closes #548 — `StartCleanup` guarded its goroutine with a `sync.Once` that, once fired, can never be reset. A context canceled in the TOCTOU window between the pre-check and the `Once` body consumed the `Once` without launching a goroutine, permanently disabling cleanup. Replaced the `Once` with an `atomic.Bool` and, on a post-CAS re-check that catches the raced cancel, **release** the flag and report `false` — so a later call with a live context recovers and starts cleanup instead of the entry map growing unbounded forever. (The realistic already-canceled-at-entry case was already handled by the #426 pre-check; this fixes the residual narrow race that #463 had documented as an accepted permanent-disable.)

## Not addressed here: #547 (needs a spec decision)
#547 (JWT v0.2 resource claims with `?` in the URI) is **deliberately not included**. The suggested Option A — skip condition parsing when the resource name is an absolute URI — **breaks an existing, tested feature**: `TestJWT_ResourcePromptShorthandConditions` (#410) relies on `resource:file:///data/*?uri=…` where `?uri=` is a legitimate `allowedValues` **condition** on a `://`-bearing resource. A URI query `?` and a condition `?` are syntactically identical for resource claims, so no heuristic can distinguish them. The only correct fix is the breaking grammar change (Option B: move the condition separator from `?` to `#`), which requires a coordinated spec update in `eunolabs/mcp-capability-manifest`, a token migration, and changes to every existing `?`-condition token/test. That is an architecturally significant decision I've left for maintainer sign-off — see my comment on #547.

## Testing
  - `TestFilterListResult_AllFilteredSerializesEmptyArray` + `TestFilterListResult_FailClosedSerializesEmptyArray` (#546)
  - `TestStartCleanup_CancelInTOCTOUWindowRecovers` (#548, updated from the old #463 test to assert graceful recovery)

---


eunolabs:
✅ Ready to merge — session rev-20260615T1211Z-b4e0ad bugs addressed (#546, #548), self-review complete with no outstanding comments, tests passing (Test with race detector + 80% coverage gate green, Lint green, all builds/gofmt/license green; each fix has a regression test verified to fail without it; `go test -race ./...` clean locally). #547 intentionally deferred to a maintainer decision (breaking grammar change) — see the comment on that issue.

---

PR #556, opened 2026-06-15
sec: cap JWKS key count and refuse redirects; fix numeric comparison

eunolabs:
## Summary

Three security and correctness improvements: (1) cap JWKS endpoint responses to 100 keys to prevent DoS via unbounded key trials on kid-less tokens; (2) configure the default HTTP client to refuse redirects, blocking SSRF pivots to internal addresses; (3) fix `numericEqual` to compare large integers exactly rather than collapsing them through lossy float64 conversion.

## Why

**JWKS key count cap:** A kid-less JWT token trials every key in the JWKS set until one verifies. An attacker controlling or on-path to the JWKS endpoint can force O(keys) asymmetric verifications per token by returning an oversized set. The 1 MiB response size cap is insufficient — a single MiB can hold thousands of keys. Capping at 100 keys (far above any legitimate issuer's rotation overlap) bounds the per-token cost while failing closed on oversized responses.

**Redirect refusal:** The JWKS endpoint is the root of trust for every token. A 30x redirect can pivot to an internal address (SSRF) or attacker-chosen key source. The default client now refuses all redirects; a legitimate endpoint answers 200 directly. This is the in-library defense for callers constructing a cache directly (the CLI validates scheme/host up front separately).

**Numeric comparison fix:** `numericEqual` is used for `allowedValues`/enum matching across manifest (int) and request (float64) types. The old implementation converted both to float64, which loses precision above 2^53 — two distinct int64 values could round to the same float and incorrectly match. The fix compares integers exactly when both operands are integral, preserving the manifest-int vs request-float64 bridge while distinguishing large integers.

## Additional changes in this PR (CI & docs)

Companion hardening on this branch, called out here so the diff isn't surprising:

- **CI — SHA-pinned every GitHub Action.** All 16 action references across the workflows are pinned to a full commit SHA with the version retained in a `# vN` comment. Each digest was verified against its tag with `git ls-remote`. The existing `github-actions` Dependabot ecosystem keeps the pins current.
- **Dependabot — added the `docker` ecosystem** for `deploy/docker`, so a base-image patch carrying a security fix surfaces as a routine PR instead of only being caught by the release-time Trivy gate.
- **Docs — added `docs/release-checklist.md`**, the release playbook `CONTRIBUTING.md` already links twice but which did not exist; plus clarifications in the manifest guide (independent version axes), the README (two of the eleven condition types are extension hooks), and the `--audit` flag help (audit-log sensitivity).

`CHANGELOG.md` is intentionally left empty for now (per request); GitHub Releases auto-generates per-release notes from Conventional Commits.

## How it was tested

- Added `TestJWKSCache_RejectsOversizedKeySet` to verify the 100-key cap is enforced and responses at the cap are accepted.
- Added `TestJWKSCache_DefaultClientRefusesRedirect` to confirm the default client does not follow 30x redirects.
- Added `TestNumericEqual` with 15 cases covering the int-vs-float bridge, large-integer distinction, mixed integer types, and non-numeric rejection (with argument symmetry).
- Full `go test -race ./...` passes; `go vet`, gofmt, golangci-lint, and the license/notice/stale-syntax gates are clean.

## Security considerations

- **Fail-closed preserved:** Oversized JWKS responses are rejected wholesale; the circuit breaker records the failure and any cached good set keeps serving. Refused redirects surface as fetch errors, failing closed.
- **Authority:** The numeric comparison fix prevents unintended matches in `allowedValues` constraints — two distinct large integers can no longer be conflated, tightening enforcement.
- **No new secrets or telemetry:** Changes are purely defensive (caps, client config, comparison logic).

## Checklist

- [x] `make test` passes — race detector clean, `pkg/` coverage stays >= 80%
- [x] `make lint` passes — `go vet` + golangci-lint (0 issues)
- [x] `make check-license` / `check-notice` / `check-stale-syntax` pass
- [x] Docs updated — manifest guide (version axes), README (condition hooks), release checklist added
- [x] PR description explains *why* (DoS prevention, SSRF defense, precision fix) and the companion CI/docs scope

https://claude.ai/code/session_01Jpy76YnDnQP93KfwrvNBjG


.github/dependabot.yml
```diff
@@ -20,3 +20,15 @@ updates:
     open-pull-requests-limit: 5
     labels:
       - dependencies
+
+  # Track the Docker base images (builder and distroless/nanoserver runtime).
+  # Without this, a base-image patch carrying a security fix is only caught by
+  # the release-time Trivy gate rather than surfaced as a routine PR.
+  - package-ecosystem: docker
+    directory: /deploy/docker
```
eunolabs, line 28: Dependabot's Docker updater keys off the Dockerfile filename. This repo's Dockerfiles are `Dockerfile.mcp` and `Dockerfile.mcp.windows` (per `release.yml`), not a bare `Dockerfile`. Dependabot's matcher does include the `Dockerfile*` glob, so `Dockerfile.mcp` *should* be picked up — but since the whole point of this entry is to surface base-image patches, please confirm a `docker` update PR actually appears (or that a base-image bump is detected) so this doesn't silently become a no-op.

---

eunolabs (commented):
## Review — JWKS hardening + numeric comparison fix

Reviewed the full diff and traced the touched paths (`pkg/capability/jwkscache.go`, `pkg/enforcement/handlers.go`, and the CLI JWKS wiring in `cmd/eunox-mcp/main.go`). Net: the three security/correctness changes are sound, well-commented, and well-tested. No blocking issues — a few notes below.

### Correctness — verified

**1. JWKS key-count cap (`maxJWKSKeys = 100`)** — Correct and fails closed. `fetchKeys` rejects an oversized set *before* it is cached; the breaker records the failure and any previously-cached good set keeps serving (consistent with the existing empty-set handling from #425). The ceiling sits well above any realistic rotation overlap.

**2. Redirect refusal** — Correct and fails closed. `CheckRedirect` returning `http.ErrUseLastResponse` surfaces the 30x as a non-200, which `fetchKeys` already rejects via `resp.StatusCode != http.StatusOK`. I confirmed this only changes behavior for *direct library callers*: the CLI supplies its own client (`newJWKSHTTPClient`) that re-validates the scheme on every redirect hop, so both paths fail closed. One consequence worth a CHANGELOG line — a library-direct caller whose IdP legitimately 30x-redirects its JWKS URL will now break. That's acceptable pre-1.0 and fail-closed, but it is a behavior change.

**3. `numericEqual` precision fix** — Correct. The exact int-vs-int path is tried first and only falls back to the float64 bridge when an operand isn't an integral, in-range value. `floatToInt64` guards the range *before* the `int64(f)` conversion (Go leaves out-of-range float→int conversions implementation-defined) and the `float64(i) != f` check rejects fractional values. `maxInt64Uint = uint64(1<<63 - 1)` is fine — it's evaluated as untyped-constant arithmetic, so no overflow. The table test covers the bridge, the >2^53 distinction, mixed signed/unsigned, non-numeric/bool rejection, fractional floats, and argument symmetry.

### Test coverage
Good. Each change ships a focused test: oversized/at-cap key sets, a refused redirect with an explicit target-hit assertion, and the 15-case numeric table including swapped-argument symmetry.

### Scope vs. description
The PR description only documents the three code changes, but the diff also (a) SHA-pins every GitHub Action with `# vN` comments, (b) adds a `docker` Dependabot ecosystem, and (c) adds `docs/release-checklist.md`. These are all reasonable hardening/docs, but the description should mention them so the diff isn't surprising. Good news: the `github-actions` Dependabot ecosystem is already present, so the new SHA pins will be kept current rather than going stale.

### Minor
- **Verify the pins.** SHA-pinning's entire security benefit depends on each digest actually matching its `# vN` tag — a transposed digit pins to an unintended (or hostile) ref. Worth a quick `git ls-remote --tags` spot-check on a couple of the actions.
- **`jwkscache.go` comment wording.** "Mirrors the `maxResolverKeys` cap" is slightly misleading since the value differs (100 here vs 20 on the co-issuer path). Consider "mirrors the *approach* of `maxResolverKeys`" to avoid implying the same ceiling.

Nice, tightly-scoped security work overall.

---

.github/dependabot.yml
eunolabs: Good catch to call out. The config is correct: Dependabot's Docker updater matches the `Dockerfile*` glob within the configured directory, and both `deploy/docker/Dockerfile.mcp` and `Dockerfile.mcp.windows` sit under `directory: /deploy/docker`, so both are in scope (a bare `Dockerfile` is not required).

The one thing I can't produce from here is the empirical "a PR actually appeared" confirmation — I can't trigger a Dependabot run from the agent. It will run on the weekly schedule configured here, or on demand via **Insights → Dependency graph → Dependabot → Recent update jobs → Check for updates**. After a run, the tagged base images (e.g. `golang:1.25.11-alpine`) should yield update PRs; if a manual check produces nothing, that would indicate a config problem to fix. Flagging this so it's tracked rather than assumed.

---

PR #557, opened 2026-06-15
Fix session rev-20260615T1229Z-dba2c3 bugs

### LINKED ISSUE #552: [rev-20260615T1229Z-dba2c3] serveStdioHost allows all calls instead of failing closed when no policy is configured and enforcement is not "audit"
**Location:** `cmd/eunox-mcp/main.go:750-759` (`serveStdioHost`)
**Reference:** https://github.com/eunolabs/eunox/blob/0682111917f6f47627b5d4e8d367fda4da61f50d/cmd/eunox-mcp/main.go#L750-L759
**Severity:** Medium — violates the fail-closed invariant; a `transport: stdio` config with no `policy:` and non-audit enforcement silently starts in allow-all mode instead of refusing to start

**Description:**

`BuildRoutes` (HTTP gateway) enforces SEC-05: a route with no `policy:` and `enforcement` not set to `"audit"` is refused at startup:

```go
// route.go:144-149
if len(u.Policy) == 0 && !r.audit {
    return nil, fmt.Errorf("upstream %q: no policy configured and enforcement is not %q — "+
        "every call on /mcp/%s would be allowed unenforced. Add 'policy:' to enforce, "+
        "or set 'enforcement: audit' to run this route in observe-only (wiretap) mode",
        u.Name, capability.EnforcementAudit, u.Name)
}
```

`serveStdioHost` (stdio host) has no equivalent guard. When `len(u.Policy) == 0` and `audit == false`, it prints a WARNING and continues:

```go
// main.go:756-759
case !audit:
    fmt.Fprintf(os.Stderr,
        "[eunox-mcp] WARNING: upstream %q has no policy configured — ALL calls will be ALLOWED without enforcement. "+
            "Add 'policy:', or set enforcement: audit to acknowledge.\n", u.Name)
```

`loadUpstreamPDP` then returns `alwaysAllowPDP{}` for a policyless upstream. The stdio host starts up and forwards every MCP call uncontrolled.

The warning message itself says "set enforcement: audit to acknowledge" — implying an explicit opt-in is required — yet the code allows the non-audit policyless case without any acknowledgment.

The CONTRIBUTING.md fail-closed invariant states: "On any ambiguity — missing manifest, unmapped method, malformed JWT, unset condition argument — deny, with a structured error code". An absent manifest is called out explicitly. The SEC-05 protection should apply to both transports.

As a consequence, the `noPolicyStartupRejection` helper (used by `validate` and `doctor` subcommands) also doesn't flag this for stdio, so `eunox-mcp validate` and `eunox-mcp doctor` give false assurance that the config is valid.

**Trigger / repro:**

```yaml
# eunox-gateway.yaml
schemaVersion: "0.1"
transport: stdio
upstreams:
  - name: my-server
    transport: stdio
    command: npx
    args: ["-y", "@modelcontextprotocol/server-filesystem", "/data"]
    # policy: intentionally omitted (e.g. typo, deleted file, incomplete rollout)
    # enforcement: defaults to "enforce"
```

```
$ eunox-mcp proxy --config eunox-gateway.yaml
[eunox-mcp] WARNING: upstream "my-server" has no policy configured — ALL calls will be ALLOWED without enforcement.
# Proxy starts up; all tools/call, resources/read, prompts/get pass through uncontrolled
```

The equivalent HTTP config (`transport: http` at top level) would fail at startup.

**Suggested fix:**

Add the same SEC-05 guard to `serveStdioHost` immediately after the `strictDrift` check, mirroring `BuildRoutes`:

```go
// main.go — after the configStrict check at line 734
if len(u.Policy) == 0 && !audit {
    return fmt.Errorf("upstream %q: no policy configured and enforcement is not %q — "+
        "every call would be allowed unenforced. Add 'policy:' to enforce, "+
        "or set 'enforcement: audit' to run in observe-only (wiretap) mode",
        u.Name, capability.EnforcementAudit)
}
```

Also update `noPolicyStartupRejection` in `gateway_config.go` to drop the `hostTransportHTTP` condition from the SEC-05 case, so it covers both transports:

```go
// Before:
case cfg.hostTransport() == hostTransportHTTP && !cfg.auditModeFor(u):
// After:
case !cfg.auditModeFor(u):
```

**Session:** rev-20260615T1229Z-dba2c3

### LINKED ISSUE #553: [rev-20260615T1229Z-dba2c3] Breaker.Stats() reports stale LastTransitionAt during the logical half-open window before Allow() is called
**Location:** `pkg/circuitbreaker/breaker.go:406-430` (`Stats`)
**Reference:** https://github.com/eunolabs/eunox/blob/0682111917f6f47627b5d4e8d367fda4da61f50d/pkg/circuitbreaker/breaker.go#L406-L430
**Severity:** Low — observability/dashboarding only; no impact on the circuit breaker's admission or state-transition correctness

**Description:**

`Stats()` and `State()` logically promote the state from `StateOpen` to `StateHalfOpen` when the cooldown has elapsed, without actually mutating the breaker (the real physical transition happens only in `allow()` when the first probe is admitted). This logical promotion was introduced to make `Stats().State` never contradict `State()`.

Issue #525's fix correctly projects `ConsecutiveFails` to 0 for the logical HalfOpen snapshot:

```go
    state = StateHalfOpen
    consecutiveFails = 0  // corrected for logical state
}
return Stats{
    State:            state,
    ConsecutiveFails: consecutiveFails,
    ...
    LastTransitionAt: b.lastTransitionAt,  // NOT corrected — still the Open time
}
```

`LastTransitionAt` is NOT projected. During the logical HalfOpen window (cooldown elapsed, but `allow()` has not yet been called), `b.lastTransitionAt` still holds the time the breaker went Open (the last physical `Closed→Open` or `HalfOpen→Open` transition). The actual `lastTransitionAt` update to the HalfOpen time happens inside `allow()` at the physical Open→HalfOpen transition.

A dashboard computing "time the breaker has been half-open" as `now - stats.LastTransitionAt` would receive a value that includes the entire Open-state cooldown period — i.e., the breaker appears to have been half-open since it first opened, rather than since the cooldown elapsed.

The inconsistency: `ConsecutiveFails` is corrected for the logical state but `LastTransitionAt` is not. A snapshot with `{State: HalfOpen, ConsecutiveFails: 0, LastTransitionAt: <Open time>}` is internally contradictory: it shows HalfOpen without failures but a transition time that predates when HalfOpen became logically true.

**Trigger / repro:**

```go
    FailureThreshold:  1,
    CooldownDuration:  100 * time.Millisecond,
    HalfOpenMaxProbes: 1,
})
b.Allow()
b.RecordFailure()               // trips the breaker; LastTransitionAt = T_open
time.Sleep(150 * time.Millisecond)  // cooldown elapses

// At this point allow() has not been called yet.
s := b.Stats()
// s.State == StateHalfOpen  (logical promotion)
// s.LastTransitionAt == T_open  (Open time, NOT the cooldown-elapsed time!)
// A dashboard: "HalfOpen since T_open" -- but the breaker was Open until T_open+100ms
```

**Suggested fix:**

Project `LastTransitionAt` the same way `ConsecutiveFails` is projected. Capture the "would-have-transitioned-at" time as the instant when the cooldown elapsed:

```go
if b.state == StateOpen {
    now := b.now()
        state = StateHalfOpen
        consecutiveFails = 0
        // Project lastTransitionAt to when the cooldown actually elapsed
        // (lastFailureTime + CooldownDuration), not the future (now), so
        // the reported HalfOpen time is accurate rather than always "right now".
        lastTransitionAt = b.lastFailureTime.Add(b.config.CooldownDuration)
    }
}
return Stats{
    ...
    LastTransitionAt: lastTransitionAt,
}
```

This keeps `Stats()` read-only (no mutation) and makes the logical HalfOpen snapshot fully self-consistent: `State=HalfOpen`, `ConsecutiveFails=0`, `LastTransitionAt=<cooldown-elapsed time>`.

**Session:** rev-20260615T1229Z-dba2c3

### LINKED ISSUE #554: [rev-20260615T1229Z-dba2c3] TokenCache.Get() captures clock before acquiring RLock, allowing expired entries to be served under RLock contention
**Location:** `pkg/capability/token_cache.go:151-167` (`Get`)
**Reference:** https://github.com/eunolabs/eunox/blob/0682111917f6f47627b5d4e8d367fda4da61f50d/pkg/capability/token_cache.go#L151-L167
**Severity:** Low — the window is bounded by RLock contention duration; in practice only relevant under sustained high concurrency with tokens close to their TTL boundary

**Description:**

`Get()` calls `c.cfg.Now()` BEFORE acquiring `c.mu.RLock()`:

```go
func (c *TokenCache) Get(tokenStr string) (*TokenPayload, bool) {
    key := tokenKey(tokenStr)
    now := c.cfg.Now()    // (A) clock captured here

    c.mu.RLock()          // (B) may block if write lock is held
    entry, ok := c.entries[key]
    c.mu.RUnlock()

    ...
    if now.After(entry.expiresAt) {   // (C) expiry checked with stale `now`
        return nil, false
    }
    return entry.payload, true
}
```

Under sustained `RLock` contention (many concurrent `Get()` calls, or a `Put()`/`purgeExpired()` write lock in progress), there can be a measurable delay between (A) and (B). During this delay the captured `now` becomes stale.

If an entry's `expiresAt` falls between (A) and when the check at (C) actually executes, `now.After(entry.expiresAt)` is `false` — the stale `now` predates the actual expiry — and `Get()` returns a payload for a token that has already expired in real time. The entry is expired when the caller uses the payload.

The security contract documented in `TokenCache`'s doc comment says:
> "a revoked token whose JWTID has been added to the revocation store will continue to be accepted for at most MaxEntryTTL after the cache entry was populated"

The `MaxEntryTTL` bound can be violated by the lock-wait duration (however small). For JWKS-validated tokens where the `exp` claim is matched to `expiresAt`, this means a structurally expired JWT can still be served.

**Trigger / repro:**

Under a synthetic tight-TTL workload (e.g., cache entries with TTL=1ms) and concurrent write-lock activity, `Get()` can return payloads for tokens whose `expiresAt` is now in the past, because `now` was sampled before a write-lock cycle that took longer than the TTL.

Easier to see with a stepped test clock:

```go
tc := NewTokenCache(TokenCacheConfig{MaxEntryTTL: 10*time.Millisecond, ...})
tc.Start(ctx)
payload := &TokenPayload{ExpiresAt: time.Now().Add(5 * time.Millisecond).Unix()}
tc.Put(rawToken, payload)

// Simulate: Now() returns T, then clock advances past expiry before RLock acquired
// (e.g. inject a time function that returns T on first call and T+6ms on second)
p, ok := tc.Get(rawToken)
// ok==true, but p is from an expired entry
```

**Suggested fix:**

Move `now := c.cfg.Now()` to after the `RUnlock` so that the freshest available time is used for the expiry check. Calling `Now()` under the RLock (between `RLock` and `RUnlock`) would also work but holds the lock slightly longer for the clock call:

```go
func (c *TokenCache) Get(tokenStr string) (*TokenPayload, bool) {
    key := tokenKey(tokenStr)

    c.mu.RLock()
    entry, ok := c.entries[key]
    c.mu.RUnlock()

    if !ok {
        return nil, false
    }
    // Sample now after releasing the lock: the time between RUnlock and this
    // call is negligible (nanoseconds), whereas the time between a pre-lock
    // Now() and the RLock acquisition can be significant under contention.
    if c.cfg.Now().After(entry.expiresAt) {
        return nil, false
    }
    return entry.payload, true
}
```

**Session:** rev-20260615T1229Z-dba2c3

eunolabs:
## Bugs addressed (session rev-20260615T1229Z-dba2c3)

- Closes #552 — `serveStdioHost` only printed a WARNING and started **allow-all** when an upstream had no `policy:` and was not in audit mode, while the HTTP gateway (`BuildRoutes`) fails closed at startup (SEC-05). A `transport: stdio` config with a missing/typo'd policy therefore forwarded every call unenforced — a fail-closed-invariant violation. Added the same fail-closed guard to `serveStdioHost`, removed the now-dead allow-all warning branch, and dropped the http-only condition from `noPolicyStartupRejection` so `validate`/`doctor` flag the stdio case too.
- Closes #553 — `Stats()` logically promotes Open→HalfOpen once the cooldown elapses (without mutating), and #525 projected `ConsecutiveFails` to 0 for that snapshot, but `LastTransitionAt` was left at the physical Open time. The result was an internally contradictory `{State: HalfOpen, ConsecutiveFails: 0, LastTransitionAt: <open time>}`, and a "half-open since" dashboard that wrongly included the whole cooldown. Project `LastTransitionAt` to `lastFailureTime + CooldownDuration` (the instant the cooldown elapsed), matching the `ConsecutiveFails` projection.
- Closes #554 — `TokenCache.Get` sampled `c.cfg.Now()` at function entry, **before** acquiring the RLock. Under RLock contention the captured time went stale, so an entry that expired during the lock wait still passed the expiry check and was served past its TTL. Sample the clock **under** the lock, immediately before the check.

  > Note: the issue suggested sampling *after* `RUnlock`, but that would reopen the #535 revocation TOCTOU (which the merged fix closed by holding the lock across the expiry check). Sampling under the lock satisfies both: fresh clock **and** no `Invalidate` interleave.

## Testing
  - `TestValidateConfigRoutes_StdioHostPolicylessNonAuditFails` (#552)
  - `TestBreaker_Stats_LastTransitionAtProjectedInLogicalHalfOpen` (#553)
  - `TestTokenCache_Get_SamplesClockUnderLock` (#554, concurrency test under `-race`)

---


eunolabs:
✅ Ready to merge — session rev-20260615T1229Z-dba2c3 bugs addressed (#552, #553, #554), self-review complete with no outstanding comments, tests passing (Test with race detector + 80% coverage gate green, Lint green; each fix has a regression test verified to fail without it, including the #554 concurrency test under `-race`; `go test -race ./...`, build, gofmt, license, lint all clean locally).

---

PR #569, opened 2026-06-15
Resolve all code-review findings from #539 (A-1 through C-2)

### LINKED ISSUE #539: Code review: correctness & simplification findings (A-1 through C-2)
## Summary

Engineering review of the eunox-mcp codebase. The architecture and security design are solid — fail-closed throughout, audit chain integrity is correct, atomic `IncrementIfBelow` is the right fix for the sliding-window lockout, and the SSE write-timeout workaround via `ResponseController.SetWriteDeadline(time.Time{})` is correct. The findings below are ordered by impact.

---

### A-1 · `evaluateJWTConditions`: MISSING_CONTEXT emitted for a present-but-non-string named op-argument

**Location:** `cmd/eunox-mcp/pdp_jwt.go:994–1014`  
**Priority:** Medium · Effort: S

When `AllowedOperationsCondition.Argument != ""` and the argument exists but is not a `string` (e.g. a number or bool), the inner type assertion fails silently, `op` stays `""`, and the denial code is `MISSING_CONTEXT` with message "argument missing or empty". The manifest engine's `handleAllowedOperations` correctly distinguishes the two cases — `MISSING_CONTEXT` for absent, `CONDITION_FAILED` for non-string. An audit record that says "argument missing" when it's actually present at the wrong type misleads triage.

```go
// Current — silent fall-through on type mismatch:
var op string
if v, ok := args[c.Argument]; ok {
    if s, ok := v.(string); ok {
        if parts := strings.Fields(s); len(parts) > 0 {
            op = strings.ToUpper(parts[0])
        }
    }
}
if op == "" {
    resp := denyResponse(capability.ErrCodeMissingContext, ...,
        fmt.Sprintf("%q: argument %q is missing or empty", name, c.Argument))
    return &resp
}

// Proposed:
var op string
if rawVal, ok := args[c.Argument]; ok {
    s, isStr := rawVal.(string)
    if !isStr {
        resp := denyResponse(capability.ErrCodeConditionFailed,
            capability.ConditionTypeAllowedOperations,
            fmt.Sprintf("%q: argument %q must be a string, got %T", name, c.Argument, rawVal))
        return &resp
    }
    if parts := strings.Fields(s); len(parts) > 0 {
        op = strings.ToUpper(parts[0])
    }
}
if op == "" {
    resp := denyResponse(capability.ErrCodeMissingContext, ...) // unchanged
    return &resp
}
```

---

### A-2 · `evaluateJWTConditions` bypasses `resolveArgument` for both AllowedValues and AllowedOperations

**Location:** `cmd/eunox-mcp/pdp_jwt.go:996–1007, 1083`  
**Priority:** Low · Effort: S

Both `AllowedOperationsCondition` (named-arg path) and `AllowedValuesCondition` use `args[c.Argument]` directly. The manifest engine uses `resolveArgument(args, c.Argument)` which supports `$.nested.key` path references. JWT shorthand (`?key=value`) cannot produce path-style keys today, so this is unreachable — but it's a divergence that will silently deny with MISSING_CONTEXT if `$.path` keys are ever added to JWT claim conditions.

```go
// Current:
rawVal, argExists := args[c.Argument]

// Proposed (in both AllowedValues and AllowedOperations named-arg cases):
rawVal, argExists := resolveArgument(args, c.Argument)
```

---

### A-3 · `SupportedPrincipalClaims` comment incorrectly states custom claims are not captured

**Location:** `pkg/capability/constraint.go:149–157`  
**Priority:** Low · Effort: S

```go
// Current (inaccurate):
// Matching on arbitrary custom claims is not yet supported — the token's custom claims are not
// captured — so the manifest loader rejects any other principal key rather than
// letting it silently never match.
```

`jwtClaimsAsMap` in `cmd/eunox-mcp/pdp.go:851–878` does capture all `Extra` claims for Rego policy conditions (`input.claims.<name>`). The restriction is only that `principal` map keys are limited to these four dimensions for constraint selection. Operators reading this comment may incorrectly believe custom claims are unavailable for Rego policies.

```go
// Proposed:
// Matching arbitrary custom claims via the `principal` map is not yet supported;
// use a `policy` condition (input.claims.<name>) to inspect non-standard identity claims.
// The manifest loader rejects any unrecognized key so a typo never silently fails to match.
```

---

### A-4 · `wiretapAllow()` bypasses the injectable clock

**Location:** `cmd/eunox-mcp/pdp.go:163–168`  
**Priority:** Low · Effort: S

```go
func wiretapAllow() capability.EnforceResponse {
        DecidedAt: time.Now().UTC().Format(time.RFC3339), // ← hard-coded
    }
}
```

Every other decision path uses the `Engine.clock.Now()` injectable clock. Tests that freeze time to validate audit record timestamps get a real-wall-clock `DecidedAt` for wiretap-mode allows. Accept a `now time.Time` parameter (callers already have it), or thread a clock into `alwaysAllowPDP`.

---

### B-1 · `ManifestPDP.Decide/DecideResourceRead/DecidePromptGet/DecideSampling` duplicate ~40 lines of scaffolding

**Location:** `cmd/eunox-mcp/pdp.go:206–300, 455–537, 552–605, 611–693`  
**Priority:** Medium · Effort: M

All four methods share: kill-check → `findConstraint` → audit-mode defer → action check → optional schema guard → `EvaluateConditions` → `recordAuditModeAntecedent`. The only variation is how the `capability.EnforceRequest` is synthesized. When enforcement logic changes, all four must be updated manually. Issue #387 (where `DecideSampling` initially skipped condition evaluation) is a concrete example of this drift risk.

A shared helper:

```go
func (p *ManifestPDP) decideTarget(
    ctx context.Context,
    sessionID string,
    target EnforceTarget,
    req *capability.EnforceRequest,
    validateSchema bool,
) (resp capability.EnforceResponse) {
        return *deny
    }
    matched := p.findConstraint(target, jwtClaimsAsMap(ctx))
    if matched == nil {
            Denial:   &capability.DenialInfo{Code: capability.ErrCodeAuthorizationFailed, ...},
        }
    }
    if matched.IsAuditOnly() {
    }
    if !containsAction(matched.Actions, requiredActionFor(target.Type)) {
        return ... // CAPABILITY_DENIED
    }
    if validateSchema && matched.ArgumentSchema != nil {
        if err := enforcement.ValidateArgumentSchema(req.Arguments, matched.ArgumentSchema); err != nil {
            return ... // INVALID_PARAMS
        }
    } else if !validateSchema && matched.ArgumentSchema != nil {
        return ... // ENFORCEMENT_ERROR (fail-closed for stray schema)
    }
    if err != nil {
        return ... // ENFORCEMENT_ERROR
    }
    return resp
}
```

Each `Decide*` method then becomes ~10 lines: synthesize the `EnforceRequest`, call `decideTarget`.

---

### B-2 · `loadOrCreateAuditKeys` is 100+ lines doing three separate jobs

**Location:** `cmd/eunox-mcp/audit.go:1601–1702`  
**Priority:** Low · Effort: M

The function: (1) reads and parses the key file, (2) generates a fresh key when absent, (3) handles the temp-write → link/rename race. Jobs 2–3 are independently testable but embedded. Extracting `generateAndPersistAuditKey(keyPath string) ([]byte, error)` for lines 1619–1701 makes `loadOrCreateAuditKeys` a readable 20-line dispatch and lets the generation path be unit-tested without mocking the read path.

---

### C-1 · `httpWriteTimeout == defaultUpstreamTimeoutMs` is unexplained

**Location:** `cmd/eunox-mcp/http.go:52–66`  
**Priority:** Low · Effort: S

Both are 30 s. The SSE path correctly overrides this via `SetWriteDeadline(time.Time{})` (noted in the `Serve` comment), so the equality is intentional. A brief comment or making `httpWriteTimeout = 35 * time.Second` (slightly larger than the upstream default) would prevent future confusion about whether they should diverge.

---

### C-2 · `filterListResult`'s `failClosed` closure uses the non-obvious `*new(L)` idiom

**Location:** `cmd/eunox-mcp/pdp.go:728–730`  
**Priority:** Low · Effort: S

```go
failClosed := func() json.RawMessage {
    empty, _ := json.Marshal(*new(L))
    return empty
}
```

`*new(L)` is correct (zero value of type parameter `L`) but unfamiliar to readers not yet fluent in generic Go. A one-line comment clarifies.

---

## Priority summary

| # | Finding | Priority | Effort |
|---|---------|----------|--------|
| A-1 | JWT non-string op-arg emits wrong error code | **Med** | S |
| B-1 | `ManifestPDP.Decide*` duplication | **Med** | M |
| A-2 | `resolveArgument` bypass in JWT conditions | Low | S |
| A-3 | Wrong `SupportedPrincipalClaims` comment | Low | S |
| A-4 | `wiretapAllow()` uninjectable clock | Low | S |
| B-2 | `loadOrCreateAuditKeys` decomposition | Low | M |
| C-1 | Timeout equality comment | Low | S |
| C-2 | `*new(L)` comment | Low | S |

eunolabs:
Closes #539.

Addresses every finding from the engineering review in #539. Each commit maps to one finding (or the doc/comment cluster), so the PR can be reviewed finding-by-finding.

## Correctness & Risk

- **A-1 — JWT `allowedOperations` non-string named arg emitted the wrong error code.** The named-argument path collapsed a present-but-non-string argument into the same `MISSING_CONTEXT` denial used for an absent one, so an audit record read "argument missing" for an argument that was actually present at the wrong type. Now mirrors the manifest engine's `handleAllowedOperations` three-way split: absent → `MISSING_CONTEXT`, present-but-non-string → `CONDITION_FAILED`, whitespace-only → `MISSING_CONTEXT`. (+ tests, + CHANGELOG)
- **A-2 — `resolveArgument` bypass in JWT conditions.** Both JWT named-arg paths used a bare `args[c.Argument]` lookup while the engine uses `resolveArgument` (which also understands `$.nested.key`). Exported it as `enforcement.ResolveArgument` (matching the package's other shared helpers) and used it on both paths. No behavior change for current inputs; closes the divergence before it can bite.
- **A-3 — inaccurate `SupportedPrincipalClaims` comment.** It claimed custom claims "are not captured"; they are, and are available to `policy`/Rego conditions via `input.claims`. Corrected to point operators at the `policy` condition.
- **A-4 — `wiretapAllow()` used a hard-coded `time.Now()`.** Gave `alwaysAllowPDP` an optional `enforcement.Clock`; a nil clock falls back to the wall clock so production wiring (`alwaysAllowPDP{}`) is unchanged, while tests can freeze time to validate wiretap audit timestamps. (+ test)

## Simplification & Refactoring

- **B-1 — `ManifestPDP.Decide*` duplication.** The three target-style methods (tools/call, resources/read, prompts/get) now share `decideTarget(... validateSchema bool)`, and all four paths (including `DecideSampling`) share `evaluateAndRecord` for the condition-evaluation tail — directly closing the drift that let `DecideSampling` once skip condition evaluation (#387). `DecideSampling` keeps its own head because its gate is the binary `SAMPLING_DENIED` opt-in, not the target allowlist. **Behavior-preserving**: error codes and operator-facing messages are byte-identical (`targetOperationPhrase` reproduces the per-type wording); all existing PDP tests pass unmodified. Net −108 lines.
- **B-2 — `loadOrCreateAuditKeys` decomposition.** Extracted the generate-and-persist path into `generateAndPersistAuditKey`, leaving `loadOrCreateAuditKeys` a short read-or-create dispatch. (+ direct unit test)

## Polish

- **C-1 —** documented why `httpWriteTimeout == defaultUpstreamTimeoutMs` and why the SSE per-response deadline override means they needn't diverge.
- **C-2 —** added a comment clarifying the `*new(L)` generic zero-value idiom in `filterListResult`.

## Verification

- `go build ./...`, `go vet`, `gofmt`, `make check-fmt`, `make check-license` — clean
- `go test -race ./...` — all packages pass
- `golangci-lint` — 0 issues

https://claude.ai/code/session_01NhGXWFyxZwi7d6Xb4EkLyM

---
_Generated by [Claude Code](https://claude.ai/code/session_01NhGXWFyxZwi7d6Xb4EkLyM)_


eunolabs:
1. Inline nit — [cmd/eunox-mcp/pdp_jwt_allowedvalues_test.go](https://github.com/eunolabs/eunox-internal/blob/claude/determined-goodall-7202aw/cmd/eunox-mcp/pdp_jwt_allowedvalues_test.go) (the new non-string case, args: map[string]interface{}{"sql": 42}):

Nit (non-blocking): 42 is a Go int, but arguments arriving over the wire are JSON-decoded, so a real non-string value would be float64 (or bool/map). The CONDITION_FAILED branch is exercised either way, so this isn't wrong — but using a JSON-realistic value (e.g. float64(42) or a nested object) would make the test mirror production and also cover what %T actually renders in the audit message.

2. Optional style note — decideTarget(..., validateSchema bool): the trailing positional boolean is the classic "what does true mean at the call site" smell. It's well-mitigated here (both call sites carry an explanatory comment), so this is purely optional — a named type (schemaMode) would make the call sites self-documenting without the comment.

3. Very minor — targetOperationPhrase default branch returns string(t). Only tool/resource/prompt reach it today (sampling has its own path), so the "system" fallthrough is unreachable and harmless; flagging only so a future caller doesn't assume it's exhaustive.

eunolabs:
Thanks — addressed in 23c36a7:

1. **Test value** — switched `{"sql": 42}` to `float64(42)` so the case uses the value type a JSON-decoded argument actually has in production (and the type `%T` renders into the audit message).
2. **Positional bool** — `decideTarget`'s trailing `bool` is now a named `schemaMode` type with `validateSchema` / `rejectSchema` constants, so the call sites read for themselves.
3. **`targetOperationPhrase` default** — left as-is intentionally. Only tool/resource/prompt reach it (sampling has its own path), and the function doc already states the default returns the bare target type, so the `system` case is a documented, harmless fallthrough rather than an assumed-exhaustive switch. Happy to convert it to an explicit per-type map + `default: panic`/fallback if you'd prefer it to fail loud for a future caller.

---

PR #577, opened 2026-06-15
sec/ci: per-job release token scoping + safe default session cap

eunolabs:
Two hardening fixes from an external security review. Both are peripheral to the enforcement core (which reviewed clean); these close a supply-chain token-scope gap and an unsafe availability default.

## F-1 — Least-privilege release token (`ci:`)
`.github/workflows/release.yml` declared one workflow-level `permissions` block
(`contents`/`packages`/`id-token`/`attestations: write`) inherited by **every**
job — including `validate`, which runs the tag's own code and dependency graph
under `go build`/`go test`. A compromised build/test-time dependency would
execute holding the Sigstore OIDC + GHCR push scopes.

- Workflow default → `contents: read`.
- `validate` → `contents: read` (no publish scopes).
- `goreleaser` → `contents` + `id-token` + `attestations: write` (no `packages` — it does not push images).
- `docker-{linux,mac,windows}` → `packages` + `id-token` + `attestations: write`, `contents: read`.

No functional change to what the pipeline produces; each job retains exactly the scopes its steps use (verified against the cosign keyless-sign and `attest-build-provenance` steps).

## F-2 — Cap concurrent HTTP sessions by default (`sec:`)
Each HTTP session owns an upstream subprocess/connection, and `--max-sessions`
defaulted to `0` (unlimited). A client that keeps opening sessions could spawn
upstreams without bound — a resource-exhaustion / fork-bomb DoS, reachable
without authentication on a listener bound off-loopback.

- `--max-sessions` now defaults to a generous backstop (`512`) instead of unlimited (mirrors the existing `defaultMaxCallCounterKeys` backstop philosophy).
- `0` still disables the cap; a positive `listen.maxSessions` in the config still overrides the flag.
- README defaults table updated. Demo/CI gateway (binds `0.0.0.0`, opens a couple of sessions) is unaffected by the backstop.

## Testing
- `go build ./...` clean.
- `go test -race -count=1 ./cmd/eunox-mcp/` passes.
- `release.yml` parses; per-job permissions confirmed via YAML load.

Per request: no issue-number references added in code, and the changelog is untouched.

https://claude.ai/code/session_01BjKjPsHWzmywguMvqe9Pq7

---
_Generated by [Claude Code](https://claude.ai/code/session_01BjKjPsHWzmywguMvqe9Pq7)_


cmd/eunox-mcp/main.go
```diff
@@ -227,7 +239,7 @@ Flags:
...
 	shutdownTimeout := fs.Int("shutdown-timeout", 5000, "Milliseconds to wait for graceful upstream shutdown before SIGKILL.")
 	upstreamTimeout := fs.Int("upstream-timeout", upstreamTimeoutUnset, "Milliseconds to wait for the upstream to respond. An explicit value takes\nprecedence over the config's defaults.upstreamTimeoutMs: 0 disables the timeout,\na positive value sets it. Negative (the default) defers to the config, or to the\nbuilt-in default (30000) when the config does not set one either.")
-	maxSessions := fs.Int("max-sessions", 0, "Cap on concurrent client sessions (transport: http). A new session beyond the cap\nis refused with 503 rather than spawning an unbounded number of upstreams.\n0 = unlimited. Overridden by the config's listen.maxSessions.")
+	maxSessions := fs.Int("max-sessions", defaultMaxSessions, "Cap on concurrent client sessions (transport: http). A new session beyond the cap\nis refused with 503 rather than spawning an unbounded number of upstreams.\nDefaults to a safe backstop (512); pass 0 to disable the cap (unlimited).\nA positive listen.maxSessions in the config overrides this flag.")
```
eunolabs, line 242: With the flag default now `512`, `listen.maxSessions: 0` in a config file no longer means "unlimited". The merge in `cmdProxy` only applies the config value when it's **positive**:

```go
maxSessions := pf.maxSessions          // now 512 by default
if cfg.Listen.MaxSessions > 0 {        // 0 does NOT override
    maxSessions = cfg.Listen.MaxSessions
}
```

So a config that sets `maxSessions: 0` to request unlimited silently collapses to the `512` backstop (`0 > 0` is false → the flag's 512 wins). The PR description's claim that "0 still disables the cap" holds only for the CLI flag (`--max-sessions 0`), not for the config field.

This also contradicts the JSON Schema, which the PR didn't update — `schemas/eunox-gateway-config.schema.json:64` still says:

> "0 (default) means unlimited."

That's now wrong on both counts: 0 is no longer the effective default, and a config `0` no longer yields unlimited. Per CONTRIBUTING/CLAUDE conventions ("schema changes update `schemas/` plus a roundtrip test", "behavior changes need a docs update in the same PR"), this should be reconciled.

Two ways to fix:
- **Preserve config-expressed unlimited:** make `Listen.MaxSessions` an `*int` so "unset" is distinguishable from an explicit `0`, then apply config whenever it's non-nil (and let `0` mean unlimited as documented). This also fixes the existing `cfg.Listen.MaxSessions != 0` "is-it-set" check in `gateway_config.go:279`, which already conflates unset with 0.
- **Or accept the new semantics** (unlimited only via `--max-sessions 0`) and update the schema description + the README row accordingly, so a config author isn't misled into expecting `maxSessions: 0` to disable the cap.

Adding a precedence test (config `0` + default flag → expected effective cap) to `operational_hardening_test.go` would lock whichever choice you make.

---

README.md
```diff
@@ -308,7 +308,7 @@ layer over the config file with a matching CLI flag.
 
 | Concern | Control | Default | Set it with |
 | ------- | ------- | ------- | ----------- |
-| **Runaway sessions** — each session owns an upstream subprocess/connection; an unbounded client could spawn many. | Cap concurrent sessions; a new session past the cap gets `503`. | unlimited | `listen.maxSessions` / `--max-sessions` |
+| **Runaway sessions** — each session owns an upstream subprocess/connection; an unbounded client could spawn many. | Cap concurrent sessions; a new session past the cap gets `503`. | **512** (backstop; pass `0` to disable) | `listen.maxSessions` / `--max-sessions` |
```
eunolabs: Minor wording: "pass `0` to disable" is true for `--max-sessions 0` but not for `listen.maxSessions: 0` (see the note on `main.go` — config `0` no longer reaches the unlimited path). Worth disambiguating here so a config-only operator isn't surprised, and to keep this table consistent with `schemas/eunox-gateway-config.schema.json`, which still documents `0` as unlimited.

---

eunolabs (commented):
## Review: sec/ci hardening (F-1 token scoping, F-2 session cap)

Two well-scoped, sensible hardening fixes. F-1 is correct as written; F-2 is the right idea but has one semantic gap and a doc-sync miss. Details below.

### F-1 — Per-job release token scoping ✅
I traced every job's steps in `release.yml` against its new `permissions:` block, and they match exactly:

- **`validate`** — checkout + `go mod verify`/`build`/`test` only → `contents: read`. Correct; this is the key win, since this is the job that runs the tag's own code and dependency graph, and it no longer holds any publish scope.
- **`goreleaser`** — creates the GitHub Release (`contents: write`), cosign keyless-signs `checksums.txt` (`id-token: write`), `attest-build-provenance` over `dist/*` (`attestations: write`). No GHCR push in this job, so dropping `packages: write` is correct. The Homebrew/winget pushes use their own PATs (`HOMEBREW_TAP_GITHUB_TOKEN` / `WINGET_PKGS_PUBLIC_REPO_TOKEN`), not `GITHUB_TOKEN`, so they need no workflow scope. ✔
- **`docker-{linux,mac,windows}`** — GHCR login + `build-push push: true` + cosign sign by digest + `attest-build-provenance push-to-registry: true` → `packages`/`id-token`/`attestations: write` + `contents: read`. ✔

Default-to-`contents: read` with per-job elevation is exactly the least-privilege shape you want. No functional change to artifacts. No concerns here.

### F-2 — Default session cap (512) ⚠️
Defaulting `--max-sessions` to a generous backstop instead of unlimited is a good safe-default and mirrors `defaultMaxCallCounterKeys`. The const + its doc comment are clear. One real gap (inline on `main.go`):

- **`listen.maxSessions: 0` no longer means "unlimited".** The flag→config merge (`main.go:601`) only lets a *positive* config value override the flag, so with the flag now defaulting to `512`, a config `0` silently yields `512` rather than unlimited. "0 disables the cap" only holds for `--max-sessions 0`.
- **Schema not updated.** `schemas/eunox-gateway-config.schema.json:64` still says `0 (default) means unlimited` — now wrong on both counts. Per repo conventions, the schema + docs should move with this behavior change (ideally with a precedence test).

Pick one: make `Listen.MaxSessions` an `*int` to preserve config-expressed unlimited (also cleans up the `!= 0` "is-set" check in `gateway_config.go:279`), or accept "unlimited = flag-only" and fix the schema/README wording to match. Either is fine; the silent mismatch is the thing to resolve.

### Nits
- README defaults-table row reads `pass 0 to disable`, which is flag-only — see inline.

Nothing here blocks the security intent; F-1 can ship as-is, and F-2 just needs the `0`-semantics/schema reconciliation.

---

eunolabs:
Good catch — fixed in e275bb0. I took the first option (preserve config-expressed unlimited) since it avoids a capability regression and matches the existing `StrictDrift *bool` "unset vs explicit" pattern:

- `listen.maxSessions` is now a `*int`; the flag→config merge applies the config value whenever the key is **present**, including `0`, so `maxSessions: 0` means unlimited again. Absent → the `--max-sessions` backstop default (512).
- This also fixes the pre-existing `!= 0` "is-it-set" check in `validate()` (it now uses `!= nil`), so a `maxSessions: 0` under `transport: stdio` is correctly rejected as a present listener key rather than treated as absent.
- Extracted `resolveMaxSessions` (mirrors `resolveUpstreamTimeout`) with a precedence table test, and added validation cases for `maxSessions: 0` under both transports.
- Reconciled `schemas/eunox-gateway-config.schema.json` (dropped the stale `default: 0`, clarified that an omitted key uses the proxy backstop) and the README row.

`go build`, `go vet`, and `go test -race ./cmd/eunox-mcp/` all pass. Changelog left untouched per request.

---

PR #583, opened 2026-06-15
Fix session rev-20260615T1510Z-369fe9 bugs

### LINKED ISSUE #578: [rev-20260615T1510Z-369fe9] token_cache: Put() samples clock before acquiring write lock, causing incorrect entry TTL under contention
**Location:** `pkg/capability/token_cache.go:188-211`

**Reference:** https://github.com/eunolabs/eunox/blob/add92231dd68f7fe1e24489722dbccd8652d74ff/pkg/capability/token_cache.go#L188-L211

**Severity:** Medium

**Description:**

`Get()` was fixed in #554 to sample `now := c.cfg.Now()` *under* the read-lock so that expired entries are never returned to a caller that observed a still-valid timestamp. `Put()` has the symmetric problem in reverse: it samples `now := c.cfg.Now()` at line 190, *before* `c.mu.Lock()` at line 211. Under lock contention the goroutine may be descheduled between those two points; when it is finally scheduled and acquires the lock, the `now` it computes `expiresAt` and `insertedAt` from is stale.

The practical effect is a slightly inflated TTL: the cached entry expires later than the wall clock at the time it was inserted. For short-lived tokens (e.g. `MaxEntryTTL` of a few seconds) the drift can represent a meaningful fraction of the entry lifetime, causing a just-expired token to remain cached past its true expiry.

The inconsistency also violates the design intent established by the #554 fix: `Get()` is now strictly conservative (clock under lock), while `Put()` is subtly optimistic (clock before lock). A future reader of the two methods will not expect the asymmetry.

**Trigger / repro:**

1. Configure `MaxEntryTTL` to a small value (e.g. 2 s).
2. Generate high write concurrency so `c.mu.Lock()` in `Put()` waits while many goroutines compete.
3. A goroutine that sampled `now` before contention but acquired the lock 1–2 s later computes `expiresAt = now + 2s` from the stale clock, effectively caching the token for up to 4 s total.
4. A subsequent `Get()` during the extra window serves the entry even though it should have been considered expired.

**Suggested fix:**

Move the clock sample to after the lock is acquired, mirroring the #554 fix on `Get()`:

```go
func (c *TokenCache) Put(tokenStr string, payload *TokenPayload) {
    key := tokenKey(tokenStr)

    c.mu.Lock()
    defer c.mu.Unlock()

    now := c.cfg.Now()   // sample clock under lock, consistent with Get()
    ...
}
```

Note: the early-return for an already-expired token (`tokenRemaining <= 0`) still works — it just moves inside the lock, which is the correct place to make that determination anyway.

**Session:** rev-20260615T1510Z-369fe9

### LINKED ISSUE #579: [rev-20260615T1510Z-369fe9] allowedExtensions: simple suffix (e.g. ".gz") silently matches compound-extension files (e.g. "archive.tar.gz"), creating undocumented security bypass
**Location:** `pkg/enforcement/handlers.go:539-675`

**Reference:** https://github.com/eunolabs/eunox/blob/add92231dd68f7fe1e24489722dbccd8652d74ff/pkg/enforcement/handlers.go#L539-L675

**Severity:** Medium

**Description:**

`handleAllowedExtensions` normalises each allowed entry to a dotted suffix and matches with `strings.HasSuffix(base, suffix)` (line 655). The comment at lines 606–610 correctly documents the compound→simple direction: `.tar.gz` in the allowlist matches `backup.tar.gz` but **not** bare `data.gz`. It does not document the inverse: a simple entry like `.gz` in the allowlist matches **both** `data.gz` **and** `archive.tar.gz`, because `.gz` is a suffix of `.tar.gz`.

This asymmetry creates a policy bypass an operator would not anticipate:

- Operator wants to allow plain gzip files and block tarballs: writes `extensions: [".gz"]`.
- The manifest **appears** to block `.tar.gz` because that entry is absent from the allowlist.
- In practice `archive.tar.gz` passes through, because `strings.HasSuffix("archive.tar.gz", ".gz")` is true.

The only defence available to the operator — explicitly adding `.tar.gz` to a *deny* list — is not possible with this condition type, which is allow-only. There is no configuration that says "allow `.gz` but not `.tar.gz`".

The comment at line 607–608 ("The leading dot is the segment boundary — `.gz` matches `data.gz` but not `datagz`") is accurate but incomplete: the "segment boundary" it describes is a *leading* dot, not a compound boundary. Readers naturally infer that `.gz` selects exactly the `.gz` extension; the `HasSuffix` semantics are broader.

**Trigger / repro:**

```yaml
allowedExtensions:
  argument: filePath
  extensions: [".gz"]
```

Call with `filePath = "exfil.tar.gz"`. The condition returns `allow`, despite `.tar.gz` being absent from the allowlist.

**Suggested fix:**

Two options:

1. **Document the behaviour** — add an explicit note to the YAML grammar doc and the inline comment: "A simple suffix like `.gz` also permits compound-extension files (`.tar.gz`, `.tar.br.gz`, etc.). To restrict to the exact last extension only, list the compound form explicitly (`.tar.gz`) instead."

2. **Change the match semantics** — use `filepath.Ext` for single-extension entries (`.gz` → must equal `filepath.Ext(base)`) and `HasSuffix` only for entries that contain more than one dot segment (`.tar.gz` → `HasSuffix`). This is a breaking grammar change and would need a migration note.

Option 1 is non-breaking and may be sufficient for now; option 2 is strictly correct but is a pre-1.0 breaking change.

**Session:** rev-20260615T1510Z-369fe9

### LINKED ISSUE #580: [rev-20260615T1510Z-369fe9] audit: Obligations []string field is unbounded, invalidating the "provably below 4 MiB" record-size invariant
**Location:** `cmd/eunox-mcp/audit.go:138, 641–643`

**Reference:** https://github.com/eunolabs/eunox/blob/add92231dd68f7fe1e24489722dbccd8652d74ff/cmd/eunox-mcp/audit.go#L138

**Severity:** Low

**Description:**

The `auditRecord` struct contains three size-sensitive fields:

- `Details map[string]interface{}` — capped at 1 MiB by `boundAuditDetails()`.
- Envelope fields (e.g. `ConditionType`, error messages) — capped at 8 KiB by `boundEnvelopeField()`.
- `Obligations []string` — **no cap applied** (line 643: `Obligations: slices.Clone(obligs)`).

`audit-verify` reads each line with a `bufio.Scanner` that has a 4 MiB buffer. A comment in the code asserts records are "provably far below" 4 MiB. That claim holds only if the obligations slice is small. `redactFields` directives can produce one obligation string per matched field; a manifest with many deeply nested redact paths (or a deliberately crafted one) can drive `Obligations` to an arbitrary size. A single record that grows past 4 MiB causes `bufio.ErrTooLong` in `audit-verify`, making the log unverifiable.

Beyond the verification failure, an unbounded field also breaks the property that the per-record size is predictable, which the drainer's comment relies on when reasoning about channel backpressure.

**Trigger / repro:**

1. Create a manifest with a tool that has `redactFields` listing thousands of long field paths (e.g. `$.a.b.c.d.e[0].f.g.h.i.j.k` × 10 000).
2. Issue a `tools/call` request that triggers the redact condition.
3. `buildAuditRecord` clones the full slice into `Obligations`.
4. The resulting JSONL line exceeds 4 MiB.
5. `audit-verify` returns `bufio.ErrTooLong` and cannot verify the log.

**Suggested fix:**

Add a `boundAuditObligations(obligs []string) []string` helper analogous to `boundAuditDetails`, with a cap such as 64 KiB total serialised length. When the slice exceeds the cap, truncate it and append a sentinel string like `"obligations_truncated:N"` (where N is the total omitted count) so the audit record remains structurally valid and the truncation is visible. Apply it at line 643:

```go
Obligations: boundAuditObligations(slices.Clone(obligs)),
```

The cap value should be documented and covered by a round-trip test.

**Session:** rev-20260615T1510Z-369fe9

### LINKED ISSUE #581: [rev-20260615T1510Z-369fe9] pdp_jwt: ValidateToken uses time.Now() directly, bypassing clock injection and making JWT expiry untestable
**Location:** `cmd/eunox-mcp/pdp_jwt.go:373`

**Reference:** https://github.com/eunolabs/eunox/blob/add92231dd68f7fe1e24489722dbccd8652d74ff/cmd/eunox-mcp/pdp_jwt.go#L373

**Severity:** Low

**Description:**

`ValidateToken` constructs the JWT validation context with:

```go
expected := jwt.Expected{Time: time.Now()}
```

Every other time-sensitive path in the codebase (the enforcement engine, `alwaysAllowPDP`, `TokenCache`, `Breaker`) uses an injectable `Clock` interface — either `enforcement.Clock` or the cache's `cfg.Now` field — so that tests can freeze time deterministically and CI does not rely on wall-clock timing.

This direct `time.Now()` call is the only production path that still escapes the injection pattern. Consequences:

1. **Untestable expiry logic.** Tests for JWT expiry must either use a real `time.Sleep` (slow, flaky) or sign tokens with absurd far-future `exp` claims that never exercise the expiry path at all. The fix in #569 that made `alwaysAllowPDP.wiretapAllow` clock-injectable explicitly cited this pattern; `ValidateToken` was not updated at the same time.

2. **Clock skew in multi-component tests.** When `TokenCache.Get` computes entry freshness with an injected clock and `ValidateToken` uses `time.Now()`, an integration test that freezes time sees the cache treat a token as live while the validator sees it as expired (or vice versa), producing a spurious test failure.

3. **Inconsistency.** The `JWTPDP` struct already receives a `JWTPDPOptions` struct at construction; adding a `Clock enforcement.Clock` field there is a one-line change that eliminates the divergence.

**Trigger / repro:**

Write a test that:
1. Injects a frozen clock into the enforcement engine and `TokenCache`.
2. Creates a JWT with `exp = frozenTime + 5s`.
3. Advances the fake clock past `exp`.
4. Calls `JWTPDP.Decide(...)`.

The test expects a denial (expired token), but `ValidateToken` reads `time.Now()` (real clock), which is before `exp`, so the token is accepted — test outcome diverges from the frozen-clock expectation.

**Suggested fix:**

Add a `Clock enforcement.Clock` field to `JWTPDPOptions` (nil → wall clock fallback). In `ValidateToken`, resolve the clock:

```go
now := time.Now()
if p.opts.Clock != nil {
    now = p.opts.Clock()
}
expected := jwt.Expected{Time: now}
```

This is identical to the pattern adopted in `alwaysAllowPDP` by #569 and requires no new abstractions.

**Session:** rev-20260615T1510Z-369fe9

### LINKED ISSUE #582: [rev-20260615T1510Z-369fe9] circuitbreaker: Stats() reports {State: Open, ConsecutiveFails: 0} after a HalfOpen probe failure — internally contradictory observability
**Location:** `pkg/circuitbreaker/breaker.go:343–397, 415–448`

**Reference:** https://github.com/eunolabs/eunox/blob/add92231dd68f7fe1e24489722dbccd8652d74ff/pkg/circuitbreaker/breaker.go#L343-L397

**Severity:** Low

**Description:**

`Stats()` returns a `ConsecutiveFails` field that has two different semantics depending on how the breaker reached `StateOpen`, and neither case is documented:

**Path A — Closed → Open (normal trip):**
`consecutiveFails` increments until it reaches `FailureThreshold`, then the breaker opens. `Stats()` returns `{State: Open, ConsecutiveFails: N}` where `N ≥ FailureThreshold`. This is meaningful: N is the count that tripped the breaker.

**Path B — HalfOpen → Open (failed probe):**
`recordFailure` at line 369 increments `consecutiveFails` by 1, then line 395 immediately resets it to 0 (fix for #428, which avoided `FailureThreshold+k` for k successive probe failures). `Stats()` then returns `{State: Open, ConsecutiveFails: 0}`.

The inconsistency:

- Path A: `State=Open, ConsecutiveFails=N` (N ≥ FailureThreshold) — looks "expected".
- Path B: `State=Open, ConsecutiveFails=0` — reads as "open with zero consecutive failures", which is contradictory. A breaker cannot be Open because of zero failures.

Any monitoring code, metrics exporter, or operator reading the stats field cannot distinguish "freshly tripped Open" (Path A) from "re-opened after a failed probe" (Path B), and the zero in Path B makes the Open state appear anomalous.

The `Stats()` projection already handles a parallel inconsistency (lines 422–430: when cooldown elapses the physical `consecutiveFails` is still `N`, but Stats projects 0 for the logical HalfOpen state so callers don't see "HalfOpen with N fails"). There is no analogous projection for Path B's `Open with 0 fails` window.

**Trigger / repro:**

```go
    FailureThreshold:        3,
    CooldownDuration:        100 * time.Millisecond,
    HalfOpenMaxProbes:       1,
    HalfOpenSuccessRequired: 1,
})

// Trip to Open (Path A)
for i := 0; i < 3; i++ { b.RecordFailure() }
// Stats: {State: Open, ConsecutiveFails: 3}  ← sensible

time.Sleep(150 * time.Millisecond) // wait for cooldown

// Consume the probe slot and fail it (Path B)
admitted, probe := b.AllowProbe()
// admitted == true (half-open probe)
probe.Failure()

s := b.Stats()
// s.State == StateOpen
// s.ConsecutiveFails == 0  ← contradictory
```

**Suggested fix:**

Two options:

1. **Preserve the pre-probe `consecutiveFails`** when re-opening from HalfOpen (instead of resetting to 0). The concern in #428 was `FailureThreshold+k` growing unboundedly; capping at `FailureThreshold` (not incrementing) avoids that while keeping the count meaningful.

2. **Extend the `Stats()` projection** to cover the Path B case: when `b.state == StateOpen && b.consecutiveFails == 0`, project `ConsecutiveFails` to the `FailureThreshold` value (since the breaker IS open, and the threshold is the reason it can be open). This is a read-side fix that requires no write-side change.

Option 1 avoids the projection asymmetry but changes write-side logic. Option 2 is contained to `Stats()` and makes the field semantics "the count of consecutive failures that last tripped the breaker" regardless of path.

**Session:** rev-20260615T1510Z-369fe9

eunolabs:
## Bugs addressed (session rev-20260615T1510Z-369fe9)

- Closes #578 — `TokenCache.Put` sampled the clock *before* acquiring the write lock, so under lock contention a stale `now` inflated the entry TTL and cached a just-expired token past its true expiry. Moved the clock sample under the lock, mirroring the #554 fix on `Get()`. The already-expired early-return now runs under the lock too (the correct place to make that determination).
- Closes #582 — `Breaker.Stats()` reported `{State: Open, ConsecutiveFails: 0}` after a HalfOpen probe failure (Path B), which reads as "open because of zero failures" — internally contradictory. Added a read-side projection in `Stats()`: when the breaker is physically Open with a zeroed counter (only possible via the HalfOpen→Open reopen, since Closed→Open always leaves `>= FailureThreshold`), project `ConsecutiveFails` to `FailureThreshold`. No write-side change, so the #428 fix is preserved.
- Closes #581 — `JWTPDP.ValidateToken` used `time.Now()` directly, the only time-sensitive production path escaping the injectable-clock pattern, making JWT expiry untestable and prone to clock skew in frozen-clock integration tests. Added a `Clock enforcement.Clock` field to `JWTPDPOptions` (nil → wall clock), wired through both constructors, and resolved it in `ValidateToken`. Matches the #569 `alwaysAllowPDP` pattern.
- Closes #580 — the audit `Obligations []string` field had no size cap, unlike `Details` (1 MiB) and the envelope fields (8 KiB), so a manifest with many `redactFields` paths could drive a single record past the 4 MiB `audit-verify` scanner buffer and make the log unverifiable. Added `boundAuditObligations` (64 KiB cap, keeps an in-order prefix and appends an `obligations_truncated:N` sentinel), applied before enqueue alongside `boundAuditDetails`. Threat model record-size invariant (§5/§6) updated.
- Closes #579 — `allowedExtensions` matches via `strings.HasSuffix`, so a simple entry like `.gz` silently also admits compound-extension files like `archive.tar.gz` (since `.gz` is a suffix of `.tar.gz`), with no way to allow `.gz` while denying `.tar.gz` in this allow-only condition. Documented the asymmetry in the inline handler comment and the manifest guide (the non-breaking option from the issue), and added a characterization test locking the documented semantics. The match behavior is unchanged.

## Testing
- New tests: `TestBreaker_StatsConsecutiveFailsAfterProbeReopen` (#582), `TestJWTPDP_ValidateToken_UsesInjectedClock` (#581), `TestBoundAuditObligations_*` / `TestRecordBoundsObligationsBeforeEnqueue` / `TestAuditVerify_LargeObligationsRoundTrip` (#580), `TestEngine_AllowedExtensions_CompoundSuffixSemantics` (#579). The #581 expiry test is discriminating: the token's `exp` is in the real future (wall clock would accept it) but the injected clock is advanced past `exp`, so it must be rejected.
- `make check-fmt`, `make check-license`, `go vet ./...`, and `golangci-lint run` on the changed packages all clean.

---


eunolabs (commented):
Self-review (acting as critical reviewer of my own diff).

Round 1 — one substantive finding:
- `pkg/capability/token_cache.go` (#578): the fix moved the clock sample under the lock, but the change shipped without a regression test. The bug only manifests under lock contention, so a single-threaded test can't catch a relapse. Added `TestTokenCache_Put_SamplesClockUnderLock` (commit fd41f25), the symmetric mirror of the existing `#554` `Get` test: it holds the write lock while `Put` blocks, advances the fake clock, releases, then asserts `expiresAt == start+advance+ttl`. The pre-fix code (clock sampled at function entry) would compute `start+ttl` and fail this test.

Round 2 — re-reviewed the full updated diff from the top; no further findings:
- #582 (`breaker.go`): read-side projection only; `Open + ConsecutiveFails==0` is reachable only via the HalfOpen->Open reopen (Closed->Open always leaves `>= FailureThreshold >= 1`), so projecting `FailureThreshold` is unambiguous. Existing #428/#525/#553 Stats tests still pass.
- #581 (`pdp_jwt.go`): `Clock` wired through both constructors (`NewJWTPDP` + `newJWTPDPWithCache`); nil → wall clock, so production wiring is unchanged. Test is discriminating (token `exp` in the real future, injected clock advanced past it).
- #580 (`audit.go`): `boundAuditObligations` keeps an in-order prefix + `obligations_truncated:N` sentinel; the "keep everything" branch is reachable only when the raw byte sum is already under the 64 KiB cap, bounding the worst-case serialized result to ~6x cap (JSON escaping) which is still orders of magnitude below the 4 MiB window. Threat-model §6.2 record-size invariant updated to include the new cap.
- #579 (`handlers.go` + guide): documentation-only (the non-breaking option the issue flagged as sufficient); match behavior unchanged, characterization test pins the documented semantics.

`go test -race ./...`, `make check-fmt`, `make check-license`, `go vet ./...`, and `golangci-lint run` on the changed packages are all clean.

---

eunolabs:
✅ Ready to merge — all session rev-20260615T1510Z-369fe9 bugs addressed (#578, #579, #580, #581, #582), self-review complete with no outstanding comments, tests passing (Go CI + Demo Integration green on fd41f25).

---

PR #584, opened 2026-06-15
Fix session rev-20260615T1410Z-2b0605 bugs

### LINKED ISSUE #574: [rev-20260615T1410Z-2b0605] ManifestPDP.findConstraint ignores actions when selecting best-matching constraint
**Location:** `cmd/eunox-mcp/pdp.go:381-409`

**Reference:** commit `cb74b92ea19f3fa00f48371f4b2c0d161e76e454`

**Severity:** High

**Description:**

`ManifestPDP.findConstraint` selects the highest-specificity constraint matching a target name but does **not** consider whether that constraint's `actions` list permits the requested operation. The action check happens afterward in `Decide()` via `containsAction`. This ordering is correct when there is only one matching constraint, but it produces a wrong result when two constraints match the same target with equal (or different) specificity and different `actions` lists.

Minimal scenario:

```yaml
tools:
  - name: "read_file"
    actions: [read]          # higher specificity wins — but wrong action
    constraints:
      - ...
  - name: "read_file"
    actions: [call]          # the actually-needed constraint
    constraints:
      - ...
```

`findConstraint` compares specificity scores and returns the first winner regardless of `actions`. If the `read`-action entry outscores (or ties) the `call`-action entry, `Decide()` then checks `containsAction(best, "call")`, finds false, and returns `CAPABILITY_DENIED` — even though the `call`-action constraint exists and would have allowed the request.

The engine (`pkg/enforcement/engine.go`, `findMatchingCapability`) does not have this problem because it calls `actionMatchScore` **inside** its scoring loop, so a constraint that doesn't match the action can never win.

```go
// pkg/enforcement/engine.go — correct
score := resourceSpecificity * resourceScoreWeight + actionMatchScore(c, action)
// action-mismatched constraint contributes 0 to score, never selected over a match

// cmd/eunox-mcp/pdp.go — incorrect
s := bareSpecificity(bare, target.Name)  // action not considered
if s > bestScore || (s == bestScore && hasPrincipal && !bestPrincipal) {
    bestScore = s
    best = i
    bestPrincipal = hasPrincipal
}
// action checked AFTER selection; wrong constraint may already be chosen
```

The practical impact is greatest when a manifest grants different capabilities per action (e.g., `read` with loose constraints and `call` with stricter constraints, or vice versa): the stricter constraints may be silently skipped, or — more seriously — the wrong set of conditions is evaluated before allowing the call.

**Trigger / Repro:**

1. Write a manifest with two constraints for the same tool, one with `actions: [call]` and one with `actions: [read]`, where the `read` entry has higher or equal name-specificity.
2. Issue a `tools/call` for that tool through the manifest PDP.
3. `Decide()` selects the `read`-action constraint, evaluates its (possibly weaker or absent) conditions, then fails with `CAPABILITY_DENIED` — even though the `call`-action constraint would have allowed the request.

**Suggested fix:**

Mirror `findMatchingCapability`: incorporate an action-match bonus (or filter) into `findConstraint`'s scoring loop so a constraint whose `actions` list does not include the requested action cannot win over one that does.

```go
// inside the findConstraint loop:
actionOK := containsAction(&p.manifest.Tools[i], target.Action)
if !actionOK {
    continue  // or assign zero action score and let a matching constraint win
}
```

Alternatively, factor out a shared `scoreConstraint(c, target)` helper used by both the cmd PDP and the pkg engine to ensure they stay in sync.

**Session:** rev-20260615T1410Z-2b0605

### LINKED ISSUE #575: [rev-20260615T1410Z-2b0605] buildConstraintsFromClaims re-parses all JWT capability claims on every Decide() call
**Location:** `cmd/eunox-mcp/pdp_jwt.go:775-792` (and `Decide()` caller)

**Reference:** commit `cb74b92ea19f3fa00f48371f4b2c0d161e76e454`

**Severity:** Medium

**Description:**

`JWTPDP.Decide()` calls `buildConstraintsFromClaims`, which iterates over every capability claim in the token and calls `parseV2Claim` on each one:

```go
// pdp_jwt.go:775-792 — called on every Decide()
func buildConstraintsFromClaims(caps []string, target EnforceTarget) []capability.Constraint {
    var out []capability.Constraint
    for _, claim := range caps {
        prefix, bareName, conds, err := parseV2Claim(claim)  // parse on every call
        ...
    }
    return out
}
```

`parseV2Claim` involves string splitting and condition unmarshalling. With C capability claims and N enforcement decisions per request (tools/call + any list calls), the total parsing work is O(C × N) — quadratic in the number of claims.

By contrast, the list-filtering path already has the correct structure: `parseCapClaims` parses each claim **once** before the loop, and each filter function then iterates over the pre-parsed result:

```go
// pdp_jwt.go — filterToolsListByClaims (correct)
parsed := parseCapClaims(caps)  // O(C) once
for _, tool := range tools {
    for _, pc := range parsed { ... }  // O(C) per tool
}
```

`Decide()` never received the same optimization. For a token with 50 capability claims and a gateway under moderate load, every `tools/call` forces 50 redundant claim parses on the hot path.

**Trigger / Repro:**

Issue any `tools/call` through `JWTPDP` with a token containing a large `mcp.capabilities` array. Each call re-parses every entry. Under profiling, `parseV2Claim` will appear proportionally in the `Decide` stack in direct proportion to the number of claims.

**Suggested fix:**

Parse claims once in `Decide()` (mirroring `parseCapClaims`) and pass the pre-parsed slice to `buildConstraintsFromClaims`:

```go
// In Decide():
parsed := parseCapClaims(caps)
constraints := buildConstraintsFromParsed(parsed, target)

// buildConstraintsFromParsed takes []parsedCap (already parsed) instead of []string
```

Alternatively, cache the parsed claims on the `JWTPDP` context object keyed by the raw claims slice, though the parse-once-per-request approach is simpler and sufficient.

**Session:** rev-20260615T1410Z-2b0605

### LINKED ISSUE #576: [rev-20260615T1410Z-2b0605] auditSink.Close() silently drops file.Close() error when Sync() also fails
**Location:** `cmd/eunox-mcp/audit.go:1098-1103`

**Reference:** commit `cb74b92ea19f3fa00f48371f4b2c0d161e76e454`

**Severity:** Low

**Description:**

`auditSink.Close()` (or the equivalent teardown path in the drain goroutine) runs:

```go
syncErr := s.f.Sync()
s.closeErr = s.f.Close()
if syncErr != nil {
    s.closeErr = syncErr
}
```

When both `s.f.Sync()` and `s.f.Close()` return errors, `s.closeErr` is set to the `Close()` error on line 2, then unconditionally overwritten by `syncErr` on line 4. The `Close()` error is silently discarded.

In practice this matters most when the underlying storage is under pressure: `Sync()` fails because the kernel cannot flush dirty pages, and `Close()` also fails (e.g., returns `EIO`) because the file descriptor state is corrupted. An operator debugging why a file descriptor is not being released would not see the `Close()` error; it is simply gone.

The audit sink is HMAC-signed and append-only; a partially-written tail that was not synced can be detected at verify time. But a descriptor leak (from a swallowed `Close()` error) is harder to diagnose and may accumulate across hot restarts.

**Trigger / Repro:**

Requires a filesystem that returns errors from both `Sync()` and `Close()` simultaneously — most easily simulated with a mock `*os.File` in a unit test. On a healthy system both calls succeed, so this path is rarely exercised.

**Suggested fix:**

Preserve both errors using `errors.Join` (Go 1.20+):

```go
syncErr := s.f.Sync()
closeErr := s.f.Close()
s.closeErr = errors.Join(syncErr, closeErr)
```

`errors.Join` returns `nil` when both inputs are `nil`, so the happy path is unchanged. Both errors are visible to callers and to any log sink that inspects `s.closeErr`.

**Session:** rev-20260615T1410Z-2b0605

eunolabs:
## Bugs addressed (session rev-20260615T1410Z-2b0605)

- Closes #574 — `ManifestPDP.findConstraint` selected the highest-specificity name match **ignoring actions**, so two entries for the same target differing only by action (e.g. `resource:foo actions:[subscribe]` and `resource:foo actions:[read]`) could resolve to the wrong one, and `Decide` would `CAPABILITY_DENY` a request a sibling entry permits. Made selection action-aware with a **two-tier** scan: a primary tier that considers only entries whose actions permit the required operation (so an action-matching sibling always wins), and a fallback tier — the exact pre-#574 selection — used only when no entry permits the action, so the lone-mismatch case still produces the precise `CAPABILITY_DENIED` (not `AUTHORIZATION_FAILED`). Only the buggy action-matching-sibling case changes behavior; lone-mismatch and no-match are unchanged. Replaces the `#409 A-3` "documented edge" with a real fix.
- Closes #575 — `buildConstraintsFromClaims` called `parseV2Claim` (which parses the expensive `?<conditions>` suffix: percent-decoding, per-pair map allocation) on **every** capability claim before filtering by target, so a token with many claims parsed every claim's conditions on every `Decide`. Reordered to a two-phase parse: the cheap namespace/bare-name match runs first, and `parseCondSuffix` runs only for claims that actually match the target. Behavior-preserving (same constraints, same fail-closed skip on a malformed matching suffix).
- Closes #576 — `auditSink.Close()` set `closeErr` to the `Close()` error then unconditionally overwrote it with `syncErr`, silently dropping the `Close()` error when both failed (a swallowed descriptor-release failure). Now `errors.Join(syncErr, closeErr)` preserves both; it returns nil when both are nil (happy path unchanged) and the single error unwrapped when only one fails.

## Testing
- New tests:
  - `TestFindConstraint_PrefersActionMatchingSibling` (#574) — both YAML orders select the action-matching sibling; existing `TestGap1_DecideResourceRead_WrongAction_Deny` confirms the lone-mismatch `CAPABILITY_DENIED` is preserved.
  - `TestJWTPDP_Decide_MatchingClaimConditionEnforcedAmongNoise` (#575) — a conditioned matching claim surrounded by non-matching conditioned claims is still parsed and enforced (allow when satisfied, deny when violated).
  - `TestAuditSink_Close_JoinsSyncAndCloseErrors` (#576) — forces both `Sync` and `Close` to fail (fd closed underneath) and asserts both errors are joined; fails on the old single-assignment code.

---


eunolabs (commented):
Self-review (critical pass over my own diff) — no substantive findings; all three fixes ship with regression tests.

Points I specifically checked:
- #574 (`pdp.go`): the `containsAction` check in `decideTarget` is **not** dead after the change — it is exactly what turns a fallback-tier (no action-permitting entry) match into `CAPABILITY_DENIED`, preserving `TestGap1_DecideResourceRead_WrongAction_Deny`. The change also fixes the symmetric list-filter case (a tool with a higher-specificity non-`call` entry plus a `call` entry is now correctly kept, not excluded), and `TestGap2_FilterToolsListResult_ReadActionNotCall_Excluded` (lone mismatch) still excludes. Where the newly-selected action-matching sibling has conditions that fail, the denial code becomes `CONDITION_FAILED` instead of `CAPABILITY_DENIED` — more accurate, still fail-closed.
- #575 (`pdp_jwt.go`): two-phase parse is observably identical — same constraints produced, same fail-closed skip when a *matching* claim's suffix is malformed; only non-matching claims now skip the (irrelevant) condition parse. The full JWT suite (allowlist, cond-chain, allowedvalues) passes unchanged.
- #576 (`audit.go`): `errors.Join` returns nil on the happy path and the single error unwrapped when only one fails, so the only behavior change is the both-fail case the issue describes.

No stale docs/comments reference the old `#409 A-3` "documented edge"; the `findConstraint` comment was rewritten. `go test -race ./...`, `make check-fmt`/`check-license`, `go vet`, and `golangci-lint` are clean.

---

eunolabs:
✅ Ready to merge — all session rev-20260615T1410Z-2b0605 bugs addressed (#574, #575, #576), self-review complete with no outstanding comments, tests passing (Go CI + Demo Integration green on f4d1e9a).

---

PR #587, opened 2026-06-15
Fix session rev-20260615T1325Z-1a88e1 bugs

### LINKED ISSUE #570: [rev-20260615T1325Z-1a88e1] audit.Close() returns nil when write failures occurred during operation
**Location:** `cmd/eunox-mcp/audit.go` lines 1072–1106 (`Close` method on `auditSink`)

**Reference:** https://github.com/eunolabs/eunox/blob/cb74b92ea19f3fa00f48371f4b2c0d161e76e454/cmd/eunox-mcp/audit.go#L1072-L1106

**Severity:** Medium

## Description

`Close()` sets `s.closeErr` only from `f.Sync()` and `f.Close()` failures (the final flush). Records lost *during operation* — tracked by the atomic `s.writeFailures` counter — do not cause `closeErr` to be set. A caller that checks `if err := sink.Close(); err != nil` receives `nil` even when the audit trail is incomplete due to mid-operation disk errors, channel overflow, or serialization failures. This breaks the fail-closed contract: the caller cannot distinguish a clean shutdown from one with lost audit records.

## How to trigger

Inject a disk error during `drain()` so that `writeRecord` returns an error and increments `s.writeFailures`. Then call `Close()`. If `f.Sync()` and `f.Close()` both succeed (e.g., the error was transient), `Close()` returns `nil` despite lost records.

## Suggested fix

Before returning `closeErr` from `Close()`, check `s.writeFailures.Load()`:

```go
if n := s.writeFailures.Load(); n > 0 {
    s.closeErr = errors.Join(s.closeErr, fmt.Errorf("audit: %d record(s) failed to write during operation", n))
}
return s.closeErr
```

This ensures any caller (e.g., a graceful-shutdown handler, a test harness) that checks the `Close()` error learns about data loss.

**Session:** rev-20260615T1325Z-1a88e1

### LINKED ISSUE #571: [rev-20260615T1325Z-1a88e1] Multi-value aud claim silently truncated to first element in TokenPayload
**Location:** `pkg/capability/jwks.go` lines 213–215 (`VerifyToken`)

**Reference:** https://github.com/eunolabs/eunox/blob/cb74b92ea19f3fa00f48371f4b2c0d161e76e454/pkg/capability/jwks.go#L213-L215

**Severity:** Medium

## Description

RFC 7519 §4.1.3 permits `aud` as either a string or a JSON array of strings. The token *validation* path correctly uses `AnyAudience` to check all values. However, when the parsed `TokenPayload` is returned to callers, the following assignment silently discards all audience values beyond the first:

```go
if p.Audience == "" && len(claims.Audience) > 0 {
    p.Audience = claims.Audience[0]   // claims.Audience[1:] dropped
}
```

Callers that use `TokenPayload.Audience` for audit logging, downstream authorization decisions, or display will see an incomplete audience. A multi-audience token is accepted correctly but its full audience is not preserved in the returned struct.

For example, a token with `"aud": ["service-a", "service-b"]` is validated against both values, but the returned `TokenPayload.Audience` is `"service-a"`. Audit records that include the audience claim will misrepresent the token.

## How to trigger

Issue a token with a JSON array `aud` containing two or more values. Verify the token through `JWKSClient.VerifyToken`. Inspect the returned `TokenPayload.Audience` — only the first array element is present.

## Suggested fix

Either:

1. Change `TokenPayload.Audience` from `string` to `[]string` and update all consumers (audit serialization, `ValidateWithLeeway` call, etc.), or
2. Add a parallel `Audiences []string` field and populate it alongside the existing `Audience` string (kept for compatibility):

```go
p.Audiences = claims.Audience
if p.Audience == "" && len(claims.Audience) > 0 {
    p.Audience = claims.Audience[0]
}
```

Option 1 is cleaner given the pre-1.0 no-compat-shims policy.

**Session:** rev-20260615T1325Z-1a88e1

### LINKED ISSUE #572: [rev-20260615T1325Z-1a88e1] fileSuffix reports wrong extension in denial message for dotfiles with compound suffixes
**Location:** `pkg/enforcement/handlers.go` lines 681–699 (`fileSuffix`)

**Reference:** https://github.com/eunolabs/eunox/blob/cb74b92ea19f3fa00f48371f4b2c0d161e76e454/pkg/enforcement/handlers.go#L681-L699

**Severity:** Low

## Description

`fileSuffix` is used to extract the file extension from an argument value and to construct the denial message ("file extension '...' is not allowed"). For a dotfile whose base name is itself a compound extension — e.g., `.tar.gz` — the function produces the wrong extension string, causing operator-facing error messages to reference the wrong suffix.

Walkthrough for `.tar.gz`:
1. `leading = 1` (strip the leading dot of the dotfile base name)
2. `trimmed = "tar.gz"` (the base after stripping the leading dot)
3. `i = strings.Index("tar.gz", ".")` → `i = 3`
4. Returns `base[leading+i:]` = `base[4:]` = `.gz`

So the denial message says `"file extension '.gz' is not allowed"` when the operator likely configured `.tar.gz` as the extension to block. An operator investigating why a request was denied sees `.gz` in the error, not `.tar.gz`, and may be confused about which manifest rule fired.

## How to trigger

Configure an `allowedExtensions` manifest entry that does not include `.tar.gz`. Submit a resource path whose basename is `.tar.gz` (a dotfile with no stem). The denial message will report `.gz` rather than `.tar.gz`.

## Suggested fix

When the file name is a dotfile (starts with `.`) and has no other stem besides the leading dot, return the full base name as the suffix rather than stripping the leading portion:

```go
func fileSuffix(name string) string {
    base := filepath.Base(name)
    if !strings.HasPrefix(base, ".") {
        i := strings.LastIndex(base, ".")
        if i <= 0 {
            return ""
        }
        return base[i:]
    }
    // Dotfile: return everything from the first dot after position 0
    // e.g. ".tar.gz" -> ".tar.gz", ".hidden" -> ""
    rest := base[1:]
    i := strings.Index(rest, ".")
    if i < 0 {
        return "" // plain dotfile, no extension
    }
    return base[1+i:] // includes the dot: ".gz" for ".tar.gz"
}
```

Alternatively, for a dotfile like `.tar.gz`, return the entire base name (`.tar.gz`) as the extension, which matches how most extension-blocklist operators think about compound dotfile names.

**Session:** rev-20260615T1325Z-1a88e1

### LINKED ISSUE #573: [rev-20260615T1325Z-1a88e1] AllowProbe() in StateClosed returns a Probe whose outcome is silently discarded on generation advance
**Location:** `pkg/circuitbreaker/breaker.go` lines 189–205 (`AllowProbe` / `allowClosed`) and lines 270–313 (`recordSuccess`)

**Reference:** https://github.com/eunolabs/eunox/blob/cb74b92ea19f3fa00f48371f4b2c0d161e76e454/pkg/circuitbreaker/breaker.go#L189-L205

**Severity:** Medium

## Description

`AllowProbe()` in `StateClosed` returns a `Probe` stamped with the current `b.halfOpenGen`. The `Do` wrapper in `do.go` documents that "the call must report exactly one outcome" and uses this Probe to call `probe.Success()` or `probe.Failure()` after the guarded function returns.

The problem: if the circuit breaker opens (due to concurrent failures on other goroutines) and then transitions to `HalfOpen` before the in-flight Probe reports its outcome, `halfOpenGen` is bumped. When the Probe finally calls `Success()`, `recordSuccess` silently drops it:

```go
// recordSuccess (simplified)
if guard && gen != b.halfOpenGen {
    return  // silent no-op: outcome discarded
}
```

The effect:
- The guarded call *succeeded*, but the breaker does not credit the success.
- `consecutiveFails` is not reset from this path (it was already cleared by the Open→HalfOpen transition, so this is not an immediate problem).
- The HalfOpen probe slot is consumed but never reported, leaving the HalfOpen window short one outcome. With `HalfOpenMaxProbes = 1`, the window is permanently stalled until the next timeout fires a new transition.
- **Specifically with `HalfOpenMaxProbes > 1`**: multiple probes can be in flight. A slow closed-state probe that reports after the transition consumes a slot without contributing a result, potentially leaving the breaker unable to close even when all actual HalfOpen probes succeeded.

The comment in `do.go` ("dropped rather than counted toward closing the new window") acknowledges the generation guard exists, but the case of a *closed-state* Probe being dropped is distinct from the intended half-open-overlap case and is not documented.

## How to trigger

1. Set `HalfOpenMaxProbes = 2`, `Threshold = 2`, `OpenTimeout = 10ms`.
2. Goroutine A calls `AllowProbe()` (StateClosed), gets `Probe{gen: 0}`.
3. Two concurrent failures trip the breaker → Open.
4. After `OpenTimeout`, breaker transitions to HalfOpen, `halfOpenGen = 1`.
5. Goroutine A's guarded function returns success. It calls `probe.Success()` with `gen = 0`.
6. `recordSuccess` sees `gen(0) != halfOpenGen(1)` → silent drop.
7. HalfOpen now has 0 of 2 probes reported. The window never closes.

## Suggested fix

Two options:

**Option A (documentation):** Add a comment in `AllowProbe`'s `StateClosed` branch and in `do.go` explicitly describing that closed-state probe outcomes may be silently discarded on generation advance, and document the consequence for `HalfOpenMaxProbes > 1`. This makes the behavior intentional and auditable.

**Option B (sentinel generation):** Use a sentinel value (e.g., `gen = -1`) for closed-state probes that bypasses the generation guard in `recordSuccess`, so the success is always credited toward resetting `consecutiveFails` without affecting HalfOpen probe counting:

```go
const closedStateGen = -1

// in allowClosed:
return true, Probe{b: b, gen: closedStateGen}

// in recordSuccess:
if guard && gen != closedStateGen && gen != b.halfOpenGen {
    return
}
```

**Session:** rev-20260615T1325Z-1a88e1

eunolabs:
## Bugs addressed (session rev-20260615T1325Z-1a88e1)

- Closes #570 — `auditSink.Close()` set `closeErr` only from the final `Sync`/`Close`, so records lost mid-operation (counted in `writeFailures`) left `Close()` returning `nil` even though the audit trail was incomplete — a caller could not distinguish a clean shutdown from a lossy one. `Close()` now joins a `writeFailures` error into `closeErr` (a clean run leaves it unchanged). Updated `TestAuditSink_WriteFailuresCounted`, which had pinned the old (buggy) "returns nil" behavior, and added `TestAuditSink_Close_ReportsWriteFailures`.
- Closes #571 — a multi-value `aud` claim (RFC 7519 §4.1.3 permits string **or** array) could not parse into `TokenPayload` (its `Audience` was a plain `string`), so a multi-aud native token was **rejected outright** at `UnsafeClaimsWithoutVerification` (not "accepted but truncated" as the issue framed it — the truncation line was effectively dead). Introduced a custom `Audience` type that unmarshals both forms and preserves every value (and marshals back to the compact form); `VerifyToken` now keeps the full audience. Repurposed the obsolete `#408` array-aud parse-error test into a `#571` acceptance test and added a separate malformed-payload test to keep `#408`'s error-attribution coverage.
- Closes #572 — `fileSuffix` reports `.gz` (not `.tar.gz`) for the stem-less dotfile `.tar.gz`. This follows from the single, tested rule (a dotfile's leading dot is its name prefix), and is **intentional and pinned** by `TestFileSuffix`; reporting the whole `.tar.gz` here but `.tar.gz` for `.archive.tar.gz` would require guessing which leading segment is a stem, which is not knowable from the name. Resolved by documenting the intent and noting the denial record always carries the full, unambiguous `filePath` in its Details — so this presentational suffix is never the operator's only handle on the file. (No behavior change; the suggested `LastIndex`-based fix would have broken compound-extension matching.)
- Closes #573 — a `Probe` returned by `AllowProbe()` in `StateClosed` can have its outcome dropped on a generation advance. Analysis shows this is **harmless by construction**: a closed-state admission reserves no half-open slot (`allow()` `StateClosed` does not touch `halfOpenProbes`), so dropping its late outcome cannot stall the new half-open window; `consecutiveFails` was already zeroed by the Open→HalfOpen transition; and the success/failure totals are counted before the guard. Crediting it instead (the issue's Option B) would be unsafe — a stale `Failure` would re-open a healthy window. Documented the semantics in `allow()` and corrected the inaccurate "consumed a probe slot" comment in `do.go`.

## Testing
- New/updated tests: `TestAuditSink_Close_ReportsWriteFailures` + `TestAuditSink_WriteFailuresCounted` (#570), `TestJWKSClient_VerifyToken_ArrayAudiencePreserved` + `TestJWKSClient_VerifyToken_MalformedPayloadIsParseError` (#571).

Note: #572 and #573 are resolved as documentation because each describes behavior that is intentional/harmless on inspection (and the suggested code changes would be incorrect or unsafe); details and reasoning are in the per-commit messages and code comments above.

---


eunolabs (commented):
Self-review (critical pass over my own diff) — no substantive findings remaining.

Specifically checked the riskiest change, #571 (`TokenPayload.Audience` string → custom `Audience` type):
- All ~7 `TokenPayload{Audience: ...}` literals updated; `JWKSClientConfig.Audience` (a separate string config field) is untouched.
- `MarshalJSON` emits a bare string for one value (preserving the prior wire shape) and an array for many; an empty/nil `Audience` marshals to `null` and round-trips back to nil (encoding/json skips `UnmarshalJSON` for JSON null). The full `./...` suite passes, so no test relied on the old `"aud":""` empty form.
- No production code compares `TokenPayload` by `==` (a slice field would break that) and the token cache keys on the raw token string, so the type change has no other blast radius.
- An existing test (`TestJWKSClient_VerifyToken_ArrayAudienceIsPayloadParseError`) had pinned the old reject-on-array behavior; it's repurposed into the acceptance test, and a new malformed-payload test preserves the `#408` error-attribution guarantee independently of the aud shape.

#570: the previously-passing `TestAuditSink_WriteFailuresCounted` asserted `Close()==nil` despite write failures — i.e. it encoded the bug. Updated to the new fail-closed contract.

#572/#573 are intentionally documentation-only: I verified `allow()` `StateClosed` does not touch `halfOpenProbes` (so a dropped closed-state outcome cannot stall a half-open window) and that the `.tar.gz`→`.gz` suffix is pinned by `TestFileSuffix` and cannot be changed consistently. The issues' suggested code changes (`LastIndex` for #572; sentinel-bypass for #573) would be incorrect/unsafe; reasoning is in the code comments.

---

eunolabs:
✅ Ready to merge — all session rev-20260615T1325Z-1a88e1 bugs addressed (#570, #571, #572, #573), self-review complete with no outstanding comments, tests passing (Go CI + Demo Integration green on 625658d).

---

PR #588, opened 2026-06-15
Fix session rev-20260615T1325Z-7905cf bugs

### LINKED ISSUE #567: [rev-20260615T1325Z-7905cf] maxCalls quota consumed for calls denied by subsequent conditions
## Location

`pkg/enforcement/handlers.go` — `handleMaxCalls` (the `IncrementIfBelow` call)
`pkg/enforcement/engine.go` — `runConditions` (condition evaluation loop)

## Reference

HEAD SHA at time of review: `d58a61c3df48303ef77eb48f8d8409c6b3f2d9b5`

## Severity

Medium

## Description

`handleMaxCalls` calls `limiter.IncrementIfBelow(ctx, key, mc.WindowSeconds, int64(mc.Count))` which, when it returns `admitted=true`, atomically appends a timestamp to the sliding window — consuming a quota slot — before `runConditions` returns control to the caller. If a condition listed *after* `maxCalls` in the same constraint's condition list fails (e.g. `allowedValues`, `timeWindow`, `ipRange`), `runConditions` returns an early-deny, but the quota slot recorded by `IncrementIfBelow` is never rolled back. The net effect is that a call the proxy ultimately denies silently burns a rate-limit slot, causing legitimate subsequent calls to be throttled.

The existing comment in `handleMaxCalls` correctly notes that calls *denied by `maxCalls` itself* do not record a timestamp (because `IncrementIfBelow` only appends on admission). The bug is the orthogonal case: a call *admitted by `maxCalls`* but *denied by a later condition*.

### Minimal reproduction manifest

```yaml
tools:
  - name: read_file
    allow: true
    conditions:
      - type: maxCalls
        count: 5
        windowSeconds: 60
      - type: allowedValues
        argument: path
        values: ["/safe/*"]
```

Sending `read_file(path="/etc/passwd")` five times in a row exhausts the quota. A subsequent call with `path="/safe/ok.txt"` — which would satisfy both conditions — is denied by `maxCalls` as the window is full, even though the five previous calls were all denied by `allowedValues` and never forwarded upstream.

## Trigger

Any constraint that lists `maxCalls` before at least one other condition that can fail independently. The window must be non-zero and the quota small enough for the phantom increments to matter.

## Suggested Fix

Three acceptable approaches, in order of invasiveness:

1. **Evaluate `maxCalls` last.** Document in `capability-manifest-guide.md` that `maxCalls` must be the final condition in its list (or enforce ordering at manifest-load time and return a validation error if it is not).

2. **Two-phase counter.** Split the handler into a "check" pass (read-only `Peek`) during `runConditions` and a "commit" pass after all conditions pass (before `collectObligations`). This requires an API extension on the `LimiterStore` interface or a new optional interface.

3. **Reorder within the engine.** Have `runConditions` evaluate all non-side-effecting conditions first, then side-effecting ones (`maxCalls`, `sequenceBlock`). This is invisible to manifest authors but requires classifying handlers by effect type.

Option 1 is lowest-risk for pre-1.0 and can be shipped as a manifest-validation rule with a clear error message.

## Session

rev-20260615T1325Z-7905cf

### LINKED ISSUE #568: [rev-20260615T1325Z-7905cf] expandHome silently misplaces audit log when os.UserHomeDir fails
## Location

`cmd/eunox-mcp/audit.go` — `expandHome` (last ~10 lines, around line 1780)

## Reference

HEAD SHA at time of review: `d58a61c3df48303ef77eb48f8d8409c6b3f2d9b5`

## Severity

Low

## Description

`expandHome` is a helper that expands a leading `~/` in a path to the user's home directory. When `os.UserHomeDir()` returns an error, the function silently returns the original `~/...` string unchanged:

```go
func expandHome(p string) string {
    if !filepath.IsAbs(p) && len(p) >= 2 && p[:2] == "~/" {
        home, err := os.UserHomeDir()
        if err == nil {
            return filepath.Join(home, p[2:])
        }
    }
    return p   // silently returns "~/.eunox/audit.jsonl" or "~/.eunox/audit.key"
}
```

`openAuditSink` calls `expandHome` for both `logPath` (default `~/.eunox/audit.jsonl`) and `keyPath` (default `~/.eunox/audit.key`), and then immediately calls `os.MkdirAll(filepath.Dir(path), 0700)`. When the home directory cannot be resolved — e.g. in a container running as an arbitrary UID, a systemd unit with `DynamicUser=yes`, or after `HOME` is unset — `MkdirAll` creates a directory literally named `~` relative to the process working directory and writes the audit log there. The operator receives no error or warning; the audit log appears to be working normally but is being written to an unexpected location.

Because the audit log is a security control (tamper-evident HMAC chain), silently misdirecting it degrades the security posture without detection.

## Trigger

Start `eunox-mcp proxy` in an environment where `os.UserHomeDir()` fails (no `HOME` env var, no `/etc/passwd` entry for the UID, container with arbitrary UID):

```
HOME= eunox-mcp proxy --config manifest.yaml
```

Audit records are silently written to `./~/.eunox/audit.jsonl` relative to CWD.

## Suggested Fix

Return an error from `expandHome` (or from `openAuditSink`) so the caller can fail closed with a diagnostic:

```go
func expandHome(p string) (string, error) {
    if !filepath.IsAbs(p) && len(p) >= 2 && p[:2] == "~/" {
        home, err := os.UserHomeDir()
        if err != nil {
            return "", fmt.Errorf("cannot expand %q: %w", p, err)
        }
        return filepath.Join(home, p[2:]), nil
    }
    return p, nil
}
```

`openAuditSink` (and any other callers) propagate the error up, which causes the proxy to refuse to start rather than silently misplace the audit log. This aligns with the fail-closed invariant.

## Session

rev-20260615T1325Z-7905cf

eunolabs:
## Bugs addressed (session rev-20260615T1325Z-7905cf)

- Closes #567 — a `maxCalls` slot was consumed for a call that a **later** condition denied: `handleMaxCalls`/`IncrementIfBelow` commits a sliding-window slot the instant it admits, and that commit was never rolled back when a subsequent condition (`allowedValues`, `timeWindow`, …) failed, so denied calls silently throttled legitimate ones. `runConditions` now evaluates in two passes — pure predicates first (in declared order), then the consuming `maxCalls` condition — so a slot is taken only after every other condition has passed (the issue's Option 3, invisible to manifest authors). `sequenceBlock` stays in the first pass since it only `Peek`s. Documented in the manifest guide; added `TestEngine_MaxCalls_NotConsumedWhenLaterConditionDenies`.
- Closes #568 — `expandHome` silently returned the literal `~/...` string when `os.UserHomeDir()` failed (HOME unset, arbitrary container UID, `DynamicUser=yes`), so `openAuditSink` would `MkdirAll` a directory literally named `~` under the CWD and write the tamper-evident audit log there with no diagnostic — degrading a security control undetected. `expandHome` now returns an error; `openAuditSink` propagates it so the proxy fails to start (fail closed). Updated the `suggest`/`audit-verify`/`stats` subcommands and the `doctor` diagnostic to handle the error, and the `expandHome` tests; added `TestExpandHome_HomeUnavailableFailsClosed`.

## Testing
- New tests: `TestEngine_MaxCalls_NotConsumedWhenLaterConditionDenies` (#567) verifies denied calls consume no quota and that `maxCalls` is still enforced for admitted calls; `TestExpandHome_HomeUnavailableFailsClosed` (#568) verifies the fail-closed error and empty return.
- The two-pass reorder was checked against the full enforcement suite (no regressions); the only intentional semantic change is that when a call would fail both `maxCalls` and another condition, the other condition's denial is reported (documented).

---


eunolabs (commented):
Self-review (critical pass over my own diff) — no substantive findings.

#567 (two-pass `runConditions`):
- `maxCalls` is the only commit-on-admit condition; `sequenceBlock` only `Peek`s during `Handle` (its antecedent marker is written post-allow by `recordSessionCall`), and `policy`/`allowedValues`/`timeWindow`/`ipRange`/… are pure — so deferring only `maxCalls` is sufficient and correct.
- The `evalCondition` extraction preserves the exact deny-response shape (unknown-type fail-closed + condition-error mapping); the full enforcement suite passes unchanged.
- Multiple `maxCalls` on one constraint (discouraged, and load-rejected when windows collide) is not worsened — if anything fewer slots are burned, since an intervening predicate now runs before both. The one intentional semantic change (when a call fails both `maxCalls` and another condition, the other condition's code is reported) is documented in the guide.

#568 (`expandHome` error):
- All six call sites updated: `openAuditSink` (log + key, the security-critical running-proxy path — fails closed), the `suggest`/`audit-verify`/`stats` subcommands (print + exit 1), and the `doctor` diagnostic (reports "cannot resolve" rather than crashing). The three existing `expandHome` tests were updated and a fail-closed test added.

---

eunolabs:
✅ Ready to merge — all session rev-20260615T1325Z-7905cf bugs addressed (#567, #568), self-review complete with no outstanding comments, tests passing (Go CI + Demo Integration green on 2668ec8).

---

PR #589, opened 2026-06-15
Fix session rev-20260615T1310Z-b3b12d bugs

### LINKED ISSUE #561: [rev-20260615T1310Z-b3b12d] JWTPDP.ValidateToken uses hardcoded time.Now() instead of injectable clock
**Location:** `cmd/eunox-mcp/pdp_jwt.go`, line 373

**Reference:** commit `187296e578b83e358f3fb7cc49ec1241590d48cf`

**Severity:** Medium

**Description:**

`JWTPDP.ValidateToken` constructs its JWT validation `Expected` struct with a hardcoded `time.Now()` call:

```go
// cmd/eunox-mcp/pdp_jwt.go:373
expected := jwt.Expected{Time: time.Now()}
```

In contrast, `JWKSClient.VerifyToken` in `pkg/capability/jwks.go` uses an injectable clock:

```go
// pkg/capability/jwks.go:183
expected := jwt.Expected{
    Time:        c.now(),
}
```

`JWTPDP` has no `now` field, and `JWTPDPOptions` has no `Now func() time.Time` option, so there is no way to override the clock. This creates two concrete problems:

1. **Untestable expiry logic.** Tests for near-expiry or just-expired tokens cannot control the wall clock, so they either sleep (flaky) or skip the expiry path entirely. The `JWKSClient` path is already tested with a fake clock (as evidenced by the `JWKSClientConfig.Now` field and the comment referencing issue #315); `JWTPDP` should match.

2. **Inconsistency inside a single token verification flow.** In JWT-intersection mode, `JWTPDP` calls `JWKSClient.VerifyToken` (which uses `c.now()`) AND separately calls `jwt.ParseSigned` + `claims.ValidateWithLeeway` with `time.Now()`. If the two clocks diverge even slightly (clock skew, test injection), a token could pass one check and fail the other, or vice versa — producing a non-deterministic allow/deny decision on a boundary token.

**Trigger / repro:**

Any test that needs to exercise JWT expiry in the JWTPDP code path (e.g., assert that a token with `exp = now-1s` is rejected) is forced to rely on real wall time. Under race-detector or slow CI, a token valid at test-start can expire mid-test.

**Suggested fix:**

1. Add `now func() time.Time` to the `JWTPDP` struct (defaulting to `time.Now` in the constructor).
2. Add `Now func() time.Time` to `JWTPDPOptions` and wire it in during construction.
3. Replace the hardcoded `time.Now()` at line 373 with `j.now()`.
4. Propagate the same clock into any other `time.Now()` calls in `JWTPDP` that touch token timestamps.

This mirrors the existing pattern in `JWKSClientConfig.Now` / `JWKSClient.now`.

**Session:** rev-20260615T1310Z-b3b12d

### LINKED ISSUE #562: [rev-20260615T1310Z-b3b12d] ForceRefreshForVerify error silently swallowed, masking JWKS unavailability as signature failure
**Location:**
- `pkg/capability/jwks.go`, lines 239–251 (`JWKSClient.VerifyToken`)
- `cmd/eunox-mcp/pdp_jwt.go`, lines 436–448 (`JWTPDP.ValidateToken`)

**Reference:** commit `187296e578b83e358f3fb7cc49ec1241590d48cf`

**Severity:** Low / Medium

**Description:**

Both `JWKSClient.VerifyToken` and `JWTPDP.ValidateToken` contain a kid-less key-rotation refresh path. When a JWT carries no `kid` and all cached keys fail signature verification, the code attempts a forced JWKS refresh via `ForceRefreshForVerify`. However, the error from that refresh is silently discarded:

```go
// pkg/capability/jwks.go:239-251 (identical pattern in pdp_jwt.go:436-448)
if kid == "" {
    if refreshed, ferr := c.cache.ForceRefreshForVerify(ctx); ferr == nil {
        refreshedKeys := FindKeys(refreshed, kid)
        payload, terminal, lastSig := tryKeys(refreshedKeys)
        if terminal != nil {
            return nil, terminal
        }
        if payload != nil {
            return payload, nil
        }
        if lastSig != nil {
            lastErr = lastSig
        }
    }
    // ferr != nil: silently ignored; execution falls through
}
```

When `ForceRefreshForVerify` returns an error (e.g., JWKS endpoint is down, circuit breaker is open, context cancelled), `ferr` is discarded and the function returns the original cached-key signature failure wrapped as `"verify signature: ..."`. The caller cannot distinguish between:

- "Every known key failed to verify this signature" (genuine bad token)
- "We couldn't reach the JWKS endpoint to check for a rotated key, so we don't actually know if the token is valid"

This matters for observability and for the fail-closed invariant: a transient JWKS outage during key rotation causes valid kid-less tokens to be rejected with a misleading error code, and operators see `"verify signature"` in logs rather than a JWKS-fetch error, delaying diagnosis.

**Trigger / repro:**

1. Deploy with a kid-less signing key.
2. Rotate the signing key at the JWKS endpoint.
3. Make the JWKS endpoint temporarily unavailable (e.g., network partition, circuit breaker tripped).
4. Send a request with a token signed by the new key.
5. The token is denied with `"verify signature: key ...: ..."` — indistinguishable from a forged token — rather than `"JWKS refresh failed: ..."`.

**Suggested fix:**

Capture `ferr` and propagate it when the refresh fails:

```go
if kid == "" {
    refreshed, ferr := c.cache.ForceRefreshForVerify(ctx)
    if ferr != nil {
        // Refresh failed; we cannot determine whether a rotated key would
        // have verified this token. Return the refresh error (not the
        // original signature failure) so callers can distinguish transient
        // JWKS unavailability from a genuinely invalid signature.
        return nil, fmt.Errorf("refresh JWKS for kid-less token: %w", ferr)
    }
    refreshedKeys := FindKeys(refreshed, kid)
    payload, terminal, lastSig := tryKeys(refreshedKeys)
    if terminal != nil {
        return nil, terminal
    }
    if payload != nil {
        return payload, nil
    }
    if lastSig != nil {
        lastErr = lastSig
    }
}
```

Apply the same fix in both `pkg/capability/jwks.go` and `cmd/eunox-mcp/pdp_jwt.go`. Update the corresponding audit error codes/log fields if the error message is used to classify the denial reason.

**Session:** rev-20260615T1310Z-b3b12d

eunolabs:
## Bugs addressed (session rev-20260615T1310Z-b3b12d)

- Closes #562 — the kid-less key-rotation path in both `JWKSClient.VerifyToken` (`pkg/capability/jwks.go`) and `JWTPDP.ValidateToken` (`cmd/eunox-mcp/pdp_jwt.go`) swallowed the error from `ForceRefreshForVerify` (the `if refreshed, ferr := …; ferr == nil` form), so a forced-refresh failure (JWKS endpoint down, circuit breaker open, context cancelled) fell through and the call returned the original cached-key signature failure. An operator then saw `"verify signature"`/`"signature verification failed"` — indistinguishable from a forged token — during a transient JWKS outage mid-rotation. Both paths now return the refresh error (`"refresh JWKS for kid-less token: …"`) when the refresh itself fails, so the failure is distinguishable in logs and the audit trail. Both paths still fail closed.

- Closes #561 — **already resolved.** `JWTPDP.ValidateToken` no longer uses a hardcoded `time.Now()`; the injectable `Clock` was added to `JWTPDPOptions`/`JWTPDP` and wired through both constructors in PR #583 (commit on `main`), and `ValidateToken` resolves `now` from it (`pdp_jwt.go:389–391`), with `TestJWTPDP_ValidateToken_UsesInjectedClock` covering it. There are no remaining `time.Now()` calls in `ValidateToken` that touch token timestamps. No code change is needed in this PR; closing the issue as fixed.

## Testing
- New tests: `TestJWKSClient_KidlessToken_RefreshFailureSurfaced` (#562, library path) and `TestJWTPDP_ValidateToken_KidlessRefreshFailureSurfaced` (#562, cmd path) — each primes the cache, takes the JWKS endpoint down, sends a kid-less token signed by a rotated/uncached key, and asserts the error surfaces the refresh failure (`"refresh JWKS"`) and is **not** masked as a signature failure.

---


eunolabs (commented):
Self-review (critical pass) — no substantive findings.

#562:
- The change only converts a swallowed `ferr` into an early `return`; both paths still fail closed (error → denial), and the only behavioral change is the error classification/message during a forced-refresh failure.
- The `#501` amplification protection is preserved: when `ForceRefreshForVerify` is rate-limited it returns `(cachedKeys, nil)`, so `ferr == nil` and the code still tries the cached set and falls through to the signature error — the new early-return fires only on a genuine fetch failure. `TestJWKSClient_KidlessToken_BadSignature_FetchRateLimited` still passes.
- Both the library (`jwks.go`) and cmd (`pdp_jwt.go`) paths are fixed identically and each has a dedicated regression test driving a switchable-down JWKS endpoint.

#561 was verified already-fixed on `main` (clock field + wiring from #583, used at `pdp_jwt.go:389–391`, with no remaining timestamp `time.Now()` and an existing injected-clock test), so this PR makes no code change for it.

---

eunolabs:
✅ Ready to merge — session rev-20260615T1310Z-b3b12d addressed (#562 fixed; #561 already resolved by #583), self-review complete with no outstanding comments, tests passing (Go CI + Demo Integration green on f4b73cd).

---

PR #590, opened 2026-06-15
Fix session rev-20260615T1309Z-685fd5 bugs

### LINKED ISSUE #558: [rev-20260615T1309Z-685fd5] Kid-less JWKS forced-refresh error swallowed in ValidateToken, masking JWKS outage as signature failure
**Location:** `cmd/eunox-mcp/pdp_jwt.go:435-450` and `pkg/capability/jwks.go:238-252`

**Reference:** https://github.com/eunolabs/eunox/blob/187296e578b83e358f3fb7cc49ec1241590d48cf/cmd/eunox-mcp/pdp_jwt.go#L435-L450

**Severity:** Medium

**Description:**

In `ValidateToken` (both in `pdp_jwt.go` and the identical pattern in `jwks.go:VerifyToken`), when a kid-less JWT (`kid == ""`) fails all cached-key signature checks, the code attempts a forced JWKS refresh:

```go
if kid == "" {
    if refreshed, ferr := p.cache.ForceRefreshForVerify(ctx); ferr == nil {
        // ... retry
    }
}
return ctx, fmt.Errorf("JWT signature verification failed: %w", lastErr)
```

The error `ferr` from `ForceRefreshForVerify` is **silently ignored** when non-nil. If the JWKS refresh fails (e.g., circuit breaker is open due to JWKS endpoint outage, or a network error), the caller receives only:

> `"JWT signature verification failed: <original key-mismatch error>"`

The actual root cause — `"JWKS fetch blocked by circuit breaker"`, `"JWKS request failed"`, etc. — is lost entirely.

By contrast, the kid-present path **does** surface the JWKS fetch error:
```go
keys, err = p.cache.ForceRefreshForKID(ctx, kid)
if err != nil {
    return ctx, fmt.Errorf("refreshing JWKS: %w", err)  // error is surfaced
}
```

The same silently-swallowed pattern exists in `pkg/capability/jwks.go:VerifyToken` at lines 238-252.

**Trigger / repro:**

1. Configure `--jwks-uri` with a kid-less signing key (i.e., JWKS has no `kid` headers, or the token omits `kid`).
2. Trigger a circuit breaker open state (e.g., kill the JWKS endpoint, let the breaker trip from prior errors).
3. Present a valid but kid-less JWT.

Expected: error includes `"JWKS fetch blocked by circuit breaker"` or similar.  
Actual: error is `"JWT signature verification failed: <stale key error>"`. The audit log records the wrong error code, and operators diagnose their token signing keys rather than the JWKS endpoint.

**Suggested fix:**

Check `ferr` and surface it when `ForceRefreshForVerify` fails, analogous to the kid-present path:

```go
if kid == "" {
    refreshed, ferr := p.cache.ForceRefreshForVerify(ctx)
    if ferr != nil {
        return ctx, fmt.Errorf("refreshing JWKS for kid-less token: %w", ferr)
    }
    refreshedCtx, terminal, lastSig := tryKeys(capability.FindKeys(refreshed, kid))
    // ... existing handling unchanged
}
```

Apply the same fix to `pkg/capability/jwks.go:VerifyToken`.

**Session:** rev-20260615T1309Z-685fd5

### LINKED ISSUE #559: [rev-20260615T1309Z-685fd5] circuitbreaker Stats() projects State=HalfOpen but not LastTransitionAt, causing "time in state" dashboards to overcount
**Location:** `pkg/circuitbreaker/breaker.go:415-439`

**Reference:** https://github.com/eunolabs/eunox/blob/187296e578b83e358f3fb7cc49ec1241590d48cf/pkg/circuitbreaker/breaker.go#L415-L439

**Severity:** Medium

**Description:**

`Stats()` applies a logical projection when the breaker is Open but the cooldown has elapsed: it reports `State=HalfOpen` and `ConsecutiveFails=0` (mirroring what `allow()` will do on the first probe). However, `LastTransitionAt` is **not** projected — it keeps the timestamp of the physical Closed→Open transition:

```go
    b.mu.RLock()
    defer b.mu.RUnlock()

    state := b.state
    consecutiveFails := b.consecutiveFails
        state = StateHalfOpen
        consecutiveFails = 0   // projected
    }
    return Stats{
        State:            state,            // projected to HalfOpen
        ConsecutiveFails: consecutiveFails, // projected to 0
        // ...
        LastTransitionAt: b.lastTransitionAt, // NOT projected — still Closed→Open time
    }
}
```

This means a monitoring system that computes:
```
timeInCurrentState = now - Stats().LastTransitionAt
```
would report a duration that includes the entire Open cooldown period, not just the logical HalfOpen duration. The snapshot is internally inconsistent: `State=HalfOpen` but `LastTransitionAt` marks when the breaker *opened*, not when it logically became HalfOpen.

The existing comment in `Stats()` (issue #525) documents the `ConsecutiveFails` projection but does not address `LastTransitionAt`. `State()` has the same gap (it also projects without mutating `lastTransitionAt`).

**Trigger / repro:**

1. Configure a breaker with `FailureThreshold=1, CooldownDuration=30s`.
2. Record one failure → breaker opens. Note `t_open`.
3. Wait 30 seconds (cooldown elapses but no `Allow()` is called).
4. Call `Stats()`: returns `{State: HalfOpen, LastTransitionAt: t_open}`.
5. `time.Since(Stats().LastTransitionAt)` = 30s+, but the breaker has been logically HalfOpen for ~0s.

A JWKS-specific scenario: the JWKS endpoint goes down, the breaker trips. A monitoring dashboard polling `Stats()` after cooldown shows `HalfOpen` with `LastTransitionAt` 30 seconds in the past, which looks like it has been stuck in HalfOpen — but no probe has been attempted yet. Alerting on "HalfOpen for >30 seconds" would fire spuriously.

**Suggested fix:**

Compute and project a logical `lastTransitionAt` in the Open→HalfOpen window, using the earliest time the breaker *could* have entered HalfOpen (i.e., `b.lastFailureTime + b.config.CooldownDuration`):

```go
    state = StateHalfOpen
    consecutiveFails = 0
    // Project the transition time to when the cooldown elapsed
    lastTransitionAt = b.lastFailureTime.Add(b.config.CooldownDuration)
}
```

This makes `Stats()` fully internally consistent: `State=HalfOpen` with `LastTransitionAt` reflecting when the state logically became HalfOpen.

**Session:** rev-20260615T1309Z-685fd5

### LINKED ISSUE #560: [rev-20260615T1309Z-685fd5] circuitbreaker TotalSuccesses/TotalFailures include stale-generation probe outcomes that were dropped by the state machine
**Location:** `pkg/circuitbreaker/breaker.go:273-276` (`recordSuccess`) and `pkg/circuitbreaker/breaker.go:347-350` (`recordFailure`)

**Reference:** https://github.com/eunolabs/eunox/blob/187296e578b83e358f3fb7cc49ec1241590d48cf/pkg/circuitbreaker/breaker.go#L270-L314

**Severity:** Low

**Description:**

In both `recordSuccess` and `recordFailure`, the observability counters (`totalSuccesses` / `totalFailures`) are incremented **before** the stale-generation guard:

```go
func (b *Breaker) recordSuccess(gen uint64, guard bool) {
    b.mu.Lock()
    defer b.mu.Unlock()

    b.totalSuccesses++           // ← incremented unconditionally

    if b.state == StateOpen {
        return
    }
    if guard && gen != b.halfOpenGen {
        return                   // ← stale probe dropped WITHOUT decrementing totalSuccesses
    }
    // ... state machine update
}
```

Identical structure in `recordFailure` for `totalFailures`.

The existing code comment for `recordFailure` documents the **Open-state** case: "A failure reported while the circuit is already open is counted in `TotalFailures` for observability." However, a **stale-generation** probe (from a prior HalfOpen window that has since been superseded — see the gen guard at #472) is also counted in the totals even though it is explicitly dropped and has no effect on the state machine.

This inflates `TotalSuccesses` and `TotalFailures` beyond the count of outcomes that actually affected the breaker. In particular, after a rapid trip-and-recover cycle under concurrent load (HalfOpenMaxProbes > 1), several in-flight probes from the previous HalfOpen window can all report and all increment the totals, while only one closed the breaker.

**Trigger / repro:**

1. Construct a breaker with `HalfOpenMaxProbes=3, FailureThreshold=1`.
2. Trip it (one failure). Cooldown elapses. Three probes are admitted in the HalfOpen window (gen=G).
3. All three succeed; the first to succeed closes the breaker (gen bumped to G+1).
4. The other two probes' `probe.Success()` calls land: `gen=G != halfOpenGen=G+1` → dropped.
5. `Stats().TotalSuccesses` = 4 (3 actual + 1 trip-close probe that succeeded + ...) even though only 1 HalfOpen success was needed and applied.

More critically: a closed-state probe admitted at gen=G that reports its **failure** after a full trip-recover cycle (gen=G+2) inflates `TotalFailures` without incrementing `consecutiveFails`. Monitoring computing `TotalFailures - TotalSuccesses` as a proxy for "net failures affecting the breaker" would overcount.

**Suggested fix:**

Move the totalSuccesses/totalFailures increments to after the guard checks, so they only count outcomes that were actually applied to the state machine:

```go
func (b *Breaker) recordSuccess(gen uint64, guard bool) {
    b.mu.Lock()
    defer b.mu.Unlock()

    if b.state == StateOpen {
        b.totalSuccesses++ // still count for observability (Open-state, documented)
        return
    }
    if guard && gen != b.halfOpenGen {
        return // stale: do NOT count in totalSuccesses
    }

    b.totalSuccesses++ // counts only outcomes applied to state machine
    // ...
}
```

Alternatively, add a separate `TotalDroppedOutcomes` counter to distinguish Open-state and stale-gen drops if full observability of dropped outcomes is desired.

Note: if counting **every** reported outcome in the totals is intentional (for "calls attempted through the breaker"), the existing behavior should be documented explicitly for stale-gen probes in both `recordSuccess` and `recordFailure`, similar to the Open-state comment in `recordFailure`.

**Session:** rev-20260615T1309Z-685fd5

eunolabs:
## Bugs addressed (session rev-20260615T1309Z-685fd5)

- Closes #560 — `TotalSuccesses`/`TotalFailures` count stale-generation probe outcomes that the state machine drops. This is **intentional and tested**: the counters are throughput metrics ("outcomes observed through the breaker"), documented in `recordFailure` as "observability-only and counts every reported failure", and pinned by existing tests (`breaker_test.go:741` "all reported failures counted", `:892` "TotalSuccesses must still count the reported success"). Changing them to exclude stale-gen (the issue's Option 1) would contradict the tested semantic and break monitoring that treats them as request counters. Resolved per the issue's Option 3: documented the count-every-outcome intent explicitly in **both** `recordSuccess` and `recordFailure` (the stale-generation case was previously only documented for the open-state case), and added `TestBreaker_StaleGenerationOutcomeStillCountedInTotals` to lock the behavior. The comments also note that net-state must not be derived from the totals (use `consecutiveFails`).

- Closes #559 — **already resolved.** `Stats()` projects `LastTransitionAt = lastFailureTime + CooldownDuration` in the Open→logical-HalfOpen window (`breaker.go:451`, issue #553), so the snapshot is internally consistent (`State=HalfOpen` with a matching transition time) and a "half-open since" dashboard no longer includes the cooldown. `State()` does not expose `LastTransitionAt`, so it has no analogous gap. No code change needed; closing as fixed.

- Closes #558 — **already resolved.** The kid-less forced-JWKS-refresh error is now surfaced (`"refresh JWKS for kid-less token: …"`) in both `JWKSClient.VerifyToken` (`jwks.go:253`) and `JWTPDP.ValidateToken` (`pdp_jwt.go:464`) — this is the same bug as #562, fixed in PR #589 (on `main`). No code change needed; closing as fixed.

## Testing
- New test: `TestBreaker_StaleGenerationOutcomeStillCountedInTotals` (#560) admits a probe, closes the window with a bare success (superseding the probe's generation), reports the now-stale probe success, and asserts it is still counted in `TotalSuccesses`.
- `make check-fmt`, `make check-license`, `go vet ./...` clean.

---


eunolabs (commented):

#560: verified the count-everything semantic is the established, *tested* contract (`breaker_test.go:741`/`:892`), so excluding stale-gen would be a regression, not a fix. The resolution documents the previously-implicit stale-generation case in both `recordSuccess` and `recordFailure` and adds a test that constructs a genuine stale-generation outcome (admit a probe, close the window with a bare `RecordSuccess` to bump the generation, then report the now-stale `probe.Success()`) and asserts it is still counted. The comment also steers consumers away from deriving net state from the totals.

#558 and #559 were verified already-fixed on `main` (#558 by the #562 fix in #589 — `jwks.go:253`/`pdp_jwt.go:464`; #559 by the #553 `LastTransitionAt` projection at `breaker.go:451`), so this PR makes no code change for them.

`go test -race ./...`, `make check-fmt`/`check-license`, and `go vet` are clean.

---

eunolabs:
✅ Ready to merge — session rev-20260615T1309Z-685fd5 addressed (#560 documented + test; #559 already fixed by #553; #558 already fixed by #589), self-review complete with no outstanding comments, Go CI green on 092407a (Demo Integration is path-filtered and not triggered for a pkg/circuitbreaker-only change).

---

PR #591, opened 2026-06-15
Fix session rev-20260615T1211Z-b4e0ad bugs

### LINKED ISSUE #547: [rev-20260615T1211Z-b4e0ad] JWT v0.2 resource capability claims with ? in URI are misparsed as condition suffixes

eunolabs:
## Bugs addressed (session rev-20260615T1211Z-b4e0ad)

- Closes #547 — a v0.2 JWT capability claim for an http(s) resource URL with a query string (e.g. `resource:https://api.example.com/search?q=widget`) was split at the URL's `?` by `strings.Cut(claim, "?")`, truncating the URL to `.../search` and reinterpreting `q=widget` as an `AllowedValues` condition on an argument `resources/read` never supplies — turning an explicit grant into a `MISSING_CONTEXT`/`AUTHORIZATION_FAILED` denial (and dropping the resource from `resources/list`). Added `splitV2Claim`, which leaves the `?query` attached to the name **for http(s) resource claims only**, and routed both the `Decide` path (`buildConstraintsFromClaims`) and the list-filter path (`parseV2Claim`/`parseCapClaims`) through it.

  Scoping note: the exception is intentionally limited to `http://`/`https://`. Other resource forms legitimately use `?` as the condition separator — notably `resource:file:///data/*?uri=…`, which constrains the synthesized `uri` argument (#410) — so `file://`, `urn:`, and the tool/prompt/system shorthands keep the original semantics. The tradeoff (an http(s) resource claim cannot also carry a condition suffix) is documented; there is no unambiguous delimiter that expresses both, and addressing the URL's own query string is the useful behavior.

## Testing
- `go test -race -count=1 ./...` — all packages pass, including the existing `TestJWT_ResourcePromptShorthandConditions` (#410, `file://`-with-condition), which the http(s) scoping preserves.
- New tests: `TestParseV2Claim_ResourceURIWithQueryString` (http(s) URLs keep their query; `file://`-with-condition still parses a condition; tool shorthand unchanged) and `TestJWTPDP_DecideResourceRead_URIWithQueryStringGranted` (end-to-end: a claim granting `https://…?q=widget` now allows the matching `resources/read`).
- `make check-fmt`, `make check-license`, `go vet ./...`, `golangci-lint run` on the changed package clean.

## Follow-up
The v0.2 claim-shorthand format is also documented in the spec repo (`eunolabs/mcp-capability-manifest`); the `?`-vs-URI-query clarification should be mirrored there as a separate docs PR with a migration note for IdP configs that embedded http(s) resource URLs.

---


eunolabs (commented):
Self-review — one finding, resolved before merge.

- My first pass detected `://` for *any* scheme and treated `?` as a URL query. The full suite caught that this broke `TestJWT_ResourcePromptShorthandConditions` (#410), which uses `resource:file:///data/*?uri=…` — a `file://` resource that legitimately uses `?` as a condition separator against the synthesized `uri` argument. Narrowed the exception to `http://`/`https://` only (`isHTTPResourceClaim`), which fixes #547 while preserving #410. Added a unit assertion pinning that `file://`-with-condition still parses a condition.
- Both claim-parsing entry points are covered: the `Decide` path (`buildConstraintsFromClaims`, which got its own inline `?`-split in #575) and the list-filter path (`parseV2Claim`/`parseCapClaims`) both route through `splitV2Claim`.
- Resource matching: the granted URL is used as a `path.Match` glob where `?` is a single-char wildcard, so the claim matches the exact target URL (the literal `?` is matched by the pattern `?`); the effect errs toward granting, which is the issue's intent.

`go test -race ./...`, `make check-fmt`/`check-license`, `go vet`, `golangci-lint` clean. The spec-repo mirror is noted as a follow-up in the PR description.

---

eunolabs:
✅ Ready to merge — session rev-20260615T1211Z-b4e0ad addressed (#547), self-review complete with no outstanding comments, tests passing (Go CI + Demo Integration green on a472bee).

---

PR #596, opened 2026-06-15
Fix session rev-20260615T1710Z-47e34e bugs

### LINKED ISSUE #592: [rev-20260615T1710Z-47e34e] tokenCacheEntry.insertedAt is set but never read — dead field wastes 24 bytes per entry
**Location:** `pkg/capability/token_cache.go:44–49`
**Reference:** https://github.com/eunolabs/eunox/blob/cb1e0dcf9317d73d774222baf13a485de4ae6caa/pkg/capability/token_cache.go#L44
**Severity:** Low — no correctness impact, but wastes 24 bytes per cached entry (8 bytes `time.Time` wall + 8 bytes ext + 8 bytes `*Location` pointer = 24 bytes per `time.Time`)

**Description:**

`tokenCacheEntry` carries an `insertedAt time.Time` field:

```go
type tokenCacheEntry struct {
    payload    *TokenPayload
    expiresAt  time.Time
    insertedAt time.Time   // set in Put() but never read
    listElem   *list.Element
}
```

`Put()` assigns it on every write:

```go
entry := &tokenCacheEntry{
    payload:    payload,
    expiresAt:  now.Add(entryTTL),
    insertedAt: now,          // assigned here
}
```

But `insertedAt` is not referenced in `Get()`, `Invalidate()`, `purgeExpired()`, `Len()`, `Stats()`, or any test — it is never read after being written. A search of the entire repository confirms no caller reads this field.

**Impact:**

At the default `MaxSize` of 4096 entries, this wastes 96 KiB of heap per cache instance (4096 × 24 bytes). In gateway mode with many routes, each route's token cache carries the overhead independently.

**Suggested fix:**

Remove the `insertedAt` field from `tokenCacheEntry` and its assignment in `Put()`. If per-entry age tracking is a future requirement, it can be re-added alongside the code that reads it.

**Session:** rev-20260615T1710Z-47e34e

### LINKED ISSUE #593: [rev-20260615T1710Z-47e34e] boundAuditObligations off-by-one: first JSON array element has no comma, causing 1-byte size overestimate
**Location:** `cmd/eunox-mcp/audit.go` — `boundAuditObligations`
**Reference:** https://github.com/eunolabs/eunox/blob/cb1e0dcf9317d73d774222baf13a485de4ae6caa/cmd/eunox-mcp/audit.go
**Severity:** Low-Medium — the obligations list in an audit record is silently truncated 1 entry earlier than the configured budget

**Description:**

`boundAuditObligations` estimates the serialized JSON size of the obligations slice to cap it at `maxBytes`. It applies a constant per-entry overhead:

```go
const perEntryOverhead = 3 // "two quotes + one comma separator"
used := len("[]")          // = 2
for i, o := range obligations {
    entrySize := perEntryOverhead + len(o)
    if used+entrySize > maxBytes {
        return obligations[:i]
    }
    used += entrySize
}
```

The comment says "two quotes + one comma separator" — correct for elements 2…N in a JSON array, where each element is preceded by a comma: `,\"value\"`. But the **first element** has no preceding comma; its actual serialized form is just `\"value\"` — overhead 2, not 3.

Actual JSON byte count for N entries:

```
len("[]") + sum(len(oi)) + 2*N + (N-1)   =   1 + 3N + sum(len(oi))
                                              ^^^
                                              brackets + quotes + commas
```

The function computes:

```
2 + 3*N + sum(len(oi))
```

This is 1 byte too large. When the budget is tight, the size check fires 1 byte early, and the last obligation that would have fit within the true budget is dropped.

**Concrete example:**

- `maxBytes = 10`, obligations `["ab", "cd"]`
- Actual JSON: `["ab","cd"]` = 10 bytes — fits exactly
- Function: `used=2`, entry 0: `2+3+2=7 ≤ 10` → keep; entry 1: `7+3+2=12 > 10` → truncate to `["ab"]`
- Correct answer: `["ab","cd"]`; produced answer: `["ab"]`

**Suggested fix:**

Account for the missing comma on the first element by subtracting 1 from the initial `used`, or by using `len("[")-1` as the seed and treating the first-element overhead as 3 (quotes + bracket-join cost) instead:

```go
used := len("[") - 1  // so first entry's perEntryOverhead=3 correctly nets to 2
```

Or equivalently, seed `used = 1` instead of `used = 2`.

**Session:** rev-20260615T1710Z-47e34e

### LINKED ISSUE #594: [rev-20260615T1710Z-47e34e] CircuitBreaker Stats/State report HalfOpen before halfOpenGen is bumped, creating observability inconsistency
**Location:** `pkg/circuitbreaker/breaker.go` — `Stats()` and `State()`
**Reference:** https://github.com/eunolabs/eunox/blob/cb1e0dcf9317d73d774222baf13a485de4ae6caa/pkg/circuitbreaker/breaker.go
**Severity:** Medium — correctness of observability data; does not affect traffic decisions, but causes dashboards/metrics to report contradictory state

**Description:**

`State()` and `Stats()` compute the logical state lazily without mutating the breaker's physical struct. When the circuit is in `StateOpen` and the cooldown has elapsed, both methods project `StateHalfOpen` under `RLock`:

```go
    b.mu.RLock()
    defer b.mu.RUnlock()
    // ...
        state = StateHalfOpen
        consecutiveFails = 0
        lastTransitionAt = b.lastFailureTime.Add(b.config.CooldownDuration) // projected
    }
    // ...
}
```

The physical struct is mutated to `StateHalfOpen` only inside `allow()` (which takes a write lock). `halfOpenGen` is bumped at that same point.

**The gap:** Between cooldown-elapsed and the first `Allow()` call, `Stats()` reports `StateHalfOpen` with the projected `lastTransitionAt`, but `halfOpenGen` has NOT been incremented yet. Any monitoring path that reads both `Stats()` and `halfOpenGen` in the same snapshot sees inconsistent data: the state says "half-open since T" but the generation counter still reads the Open value.

More concretely: if a caller reads `Stats()`, observes `StateHalfOpen`, and then calls `AllowProbe()` to get a generation-tagged `Probe`, the `Probe.gen` will be for the HalfOpen generation that was just created inside `allow()`. That is internally consistent. But a second caller that reads `Stats()` twice in a tight loop — once before `allow()` and once after — sees `lastTransitionAt` jump from the projected value to the actual value, making metric time-series show a duplicate HalfOpen transition event.

**Suggested fix:**

Option A (minimal): document the known gap in the `Stats()` and `State()` godoc, noting that `lastTransitionAt` is projected and may differ from the value written by the first `Allow()`.

Option B (stronger): promote the Open→HalfOpen transition to `Stats()`/`State()` by having them take a write lock when they detect cooldown-elapsed (double-checked locking pattern), performing the actual struct mutation so the physical and logical states agree from the moment cooldown expires — independent of whether `Allow()` has been called.

Option B aligns `Stats()`'s `lastTransitionAt` with `allow()`'s mutation and eliminates the dual-transition artifact, at the cost of converting some read-locks to write-locks on the observability path. Given that `Stats()` is typically called from monitoring goroutines at a low rate, the contention cost is negligible.

**Session:** rev-20260615T1710Z-47e34e

### LINKED ISSUE #595: [rev-20260615T1710Z-47e34e] Redis call-counter Lua script uses tonumber() on microsecond timestamps — future float64 precision hazard
**Location:** `pkg/callcounter/redis.go` — `incrIfBelowScript`
**Reference:** https://github.com/eunolabs/eunox/blob/cb1e0dcf9317d73d774222baf13a485de4ae6caa/pkg/callcounter/redis.go
**Severity:** Low — no precision loss today; latent defect that could produce wrong `retryAfter` estimates as Unix time grows

**Description:**

The Redis sliding-window counter's Lua script converts all arguments to numbers with `tonumber()`:

```lua
local limit     = tonumber(ARGV[1])
local windowUs  = tonumber(ARGV[2])   -- window size in microseconds
local nowUs     = tonumber(ARGV[3])   -- current time in Unix microseconds
local cutoff    = nowUs - windowUs
...
-- retryAfter calculation:
local oldest    = tonumber(pivot[2])  -- score = UnixMicro of oldest entry in window
local retryUs   = (oldest + windowUs) - nowUs
```

Lua's `number` type is IEEE 754 double-precision float64. The exact integer range is ±2^53 ≈ ±9.007 × 10^15.

**Current Unix time in microseconds (June 2026):** ~1.75 × 10^15 — safely within float64's exact range.

**The hazard:** The `retryAfter` arithmetic computes `oldest + windowUs`. If `windowUs` is large (e.g., a 24-hour window = 86,400,000,000 μs ≈ 8.64 × 10^10), the sum is ~1.75 × 10^15 + 8.64 × 10^10 ≈ 1.75 × 10^15, still fine today. But the safe headroom before float64 precision becomes an issue for a simple timestamp is only ~8 years from now (2^53 / 10^6 seconds ≈ year 2286), whereas sub-microsecond representation errors in the intermediate sum could appear much sooner with very large windows or when the two magnitudes interact unfavorably.

More immediately: if a future change uses `time.UnixNano()` (nanoseconds, ~1.75 × 10^18 today) instead of `time.UnixMicro()`, the values exceed float64's exact range immediately, causing silently wrong `retryAfter` values — a silent fail-open for rate limiting.

**Suggested fix:**

Replace `tonumber()` with integer coercion via `math.tointeger()` (available in Lua 5.3+, which Redis 7+ uses), which returns `nil` on non-integer or out-of-range inputs and can be checked explicitly:

```lua
local function toint(s)
    local n = math.tointeger(s)
    if n == nil then return redis.error_reply("non-integer argument: " .. tostring(s)) end
    return n
end
```

Add a guard that rejects any argument whose `tonumber()` value differs from its `math.tointeger()` value, ensuring precision loss is detected at call time rather than silently producing wrong results.

Also add a Go-side test that passes a timestamp near 2^53/1000 (scaled down for test speed) to verify the Lua path handles boundary values correctly.

**Session:** rev-20260615T1710Z-47e34e

eunolabs:
## Bugs addressed (session rev-20260615T1710Z-47e34e)

- Closes #592 — removed `tokenCacheEntry.insertedAt`, which was written by `Put()` but never read (eviction order comes from `insertOrder`), saving ~24 bytes/entry. Updated the `Put()` comment that referenced it.
- Closes #593 — corrected an off-by-one in `boundAuditObligations` (introduced in #580): seeding the byte budget with `len("[]")==2` over-counts the first array element (which has no preceding comma) by 1, so the size check could truncate one obligation earlier than the budget allows. Seed with `1` so the estimate `1 + sum(len)+3N` matches the true serialized size exactly. The safety invariant (result ≤ cap) is unchanged and still covered by `TestBoundAuditObligations_OverCapTruncatedWithSentinel`.
- Closes #594 — `Stats()`/`State()` project `lastTransitionAt = lastFailureTime + CooldownDuration` (the cooldown-elapsed instant, #553/#559) while the breaker is physically Open but past cooldown, but `allow()` wrote `lastTransitionAt = now` (the later first-probe instant) at the real transition, so the timestamp jumped forward when the first probe arrived. `allow()` now writes the same logical instant, so physical and projected agree and a "time in half-open" dashboard sees no spurious second transition. Still a single clock read on that path (#467). Updated `TestAllowOpenToHalfOpenTransitionTimestamp`; added `TestBreaker_LastTransitionAtStableAcrossFirstProbe`.
- Closes #595 — documented (in the `incrIfBelowScript` godoc) that timestamps must be passed in **microseconds**, not nanoseconds: Lua numbers are float64 (exact to 2^53 ≈ 9.0e15), and UnixMicro (~1.75e15, exact until ~year 2286) plus the `score + window` retry sum stay inside that range, whereas UnixNano (~1.75e18) would silently lose precision and produce a wrong `retryAfter`. The callers already use `UnixMicro`; the comment guards against a future regression. (A runtime `math.tointeger` guard was considered but not added: it introduces a new error path into the hot rate-limit script for a hazard that cannot occur with the current microsecond inputs.)

## Testing
- New/updated tests: `TestBreaker_LastTransitionAtStableAcrossFirstProbe` + updated `TestAllowOpenToHalfOpenTransitionTimestamp` (#594).
- `make check-fmt`, `make check-license`, `go vet ./...`, `golangci-lint run` clean.

Note: this session's issues were filed by the automated reviewer against the merged session-fix PRs (#593 refines #580; #594 builds on #553/#559).

---


eunolabs (commented):

- #592: confirmed `insertedAt` had no readers (eviction uses `insertOrder`); the full suite passes with the field removed.
- #593: the fix only shifts the *truncation boundary* by one byte (the fast-path already returns the slice unchanged when it marshals ≤ cap), so no fits-exactly behavior changes; the result-≤-cap invariant is unchanged. I did not add a dedicated exact-boundary test because the sentinel reservation plus the 64 KiB cap make a deterministic 1-byte-boundary case brittle; the existing cap test still guards safety.
- #594: the change is observability-only (`lastTransitionAt` is reporting state, not used in any decision), keeps a single clock read (#467), and makes `allow()` agree with the `#553/#559` projection. The `#525` `ConsecutiveFails`-projection and `#582` Path-B tests still pass.
- #595: documentation-only; the current `UnixMicro` callers are safe, and a runtime Lua guard would add a fail path to the hot script for a non-occurring case.

`go test -race ./...`, fmt, license, vet, and golangci-lint are clean.

---

eunolabs:
✅ Ready to merge — session rev-20260615T1710Z-47e34e addressed (#592, #593, #594, #595), self-review complete with no outstanding comments, tests passing (Go CI + Demo Integration green on 4624e79).

---

PR #597, opened 2026-06-15
Fix session rev-20260615T1610Z-501f67 bugs

### LINKED ISSUE #585: [rev-20260615T1610Z-501f67] killswitch.Redis.Reset() local cache clearing races with concurrent KillAgent/KillSession, causing temporary fail-open
**Location:** `pkg/killswitch/redis.go`, `Reset()` function

**Severity:** Medium — causes a temporary fail-open window (up to `defaultReconcileInterval`, currently 30 s) for newly-issued kills

---

## Description

`Reset()` performs three sequential phases without holding any lock across the whole operation:

1. Delete all Redis keys (`Del` global key, `deleteByPrefix` agent keys, `deleteByPrefix` session keys)
2. Publish `"reset"` to the pub/sub channel
3. Acquire `r.mu`, clear local cache (`globalActive = false`, `killedAgents = make(…)`, `killedSessions = make(…)`), release lock

A concurrent `KillAgent("X")` (or `KillSession`) can interleave as follows:

```
Reset goroutine                         KillAgent("X") goroutine
───────────────────────────────────     ──────────────────────────────────
deleteByPrefix(agentPrefix)   ←done
                                        SET killswitch:agent:X  (Redis)
                                        publish "agent:kill:X"
                                        r.mu.Lock()
                                          r.killedAgents["X"] = true
                                        r.mu.Unlock()
publish("reset")
r.mu.Lock()
  r.killedAgents = make(…)   ← WIPES "X"
r.mu.Unlock()
```

After this interleaving:
- **Redis** has `killswitch:agent:X` (kill is durable)
- **Local cache** says agent X is **not** killed

`ShouldBlock` reads the local cache first and short-circuits; it returns `false` for agent X. The kill is therefore silently bypassed until the next `reconcileLoop` tick (up to 30 s by default), at which point `refreshState` will pick up the Redis key and re-populate the cache.

The same window exists for `KillSession` interleaved with the session-prefix deletion step.

## Trigger / Repro

Race is timing-dependent but can be forced in tests:

1. Start a Redis-backed kill switch.
2. Call `Reset()` in one goroutine; immediately after `deleteByPrefix` for agents returns, call `KillAgent("victim")` in another goroutine.
3. Before the reconcile loop fires, assert `ShouldBlock(ctx, "victim", "")` — it returns `false` despite the kill being set in Redis.

## Suggested Fix

Extend the Redis-side deletion to run **after** the local cache is cleared, or perform the entire operation under a single distributed lock (or optimistic Redis transaction). A simpler alternative: after clearing the local cache in step 3, immediately call `refreshState` (or a targeted Redis read) to re-seed any kills that arrived after the delete sweep:

```go
func (r *Redis) Reset(ctx context.Context) error {
    r.client.Del(ctx, redisGlobalKey)
    r.deleteByPrefix(ctx, redisAgentPrefix)
    r.deleteByPrefix(ctx, redisSessionPfx)
    r.publish(ctx, "reset")

    r.mu.Lock()
    r.globalActive = false
    r.mu.Unlock()

    // Re-read Redis to pick up any kills that landed after the delete sweep.
    r.refreshState(ctx)
    return nil
}
```

`refreshState` already builds a consistent snapshot from Redis under a write lock, so a second call immediately after the cache clear closes the race window to the sub-millisecond delta between the re-read and any further concurrent writes.

---

*Session: rev-20260615T1610Z-501f67*

### LINKED ISSUE #586: [rev-20260615T1610Z-501f67] circuitbreaker: recordFailure resets consecutiveFails=0 on HalfOpen→Open re-trip, leaving internal state inconsistent
**Location:** `pkg/circuitbreaker/breaker.go`, `recordFailure()` and `Stats()`

**Severity:** Low — Stats() output is already papered over by a workaround (see `// Path B (#582)` comment); no wrong allow/deny decision results, but the internal invariant is broken and any future reader of `consecutiveFails` would be misled

---

## Description

When a breaker transitions **HalfOpen → Open** because a probe failed, `recordFailure()` resets `consecutiveFails` to 0 before (or as part of) the transition:

```go
// recordFailure — HalfOpen path (paraphrased from current code)
case StateHalfOpen:
    b.state = StateOpen
    b.consecutiveFails = 0   // ← reset, NOT incremented
    b.halfOpenGen++
    ...
```

This violates the invariant **"State=Open implies consecutiveFails ≥ FailureThreshold"**. Immediately after the transition, the breaker is Open with `consecutiveFails == 0`, which is contradictory: you cannot reach Open from HalfOpen without a failure.

### Workaround already in Stats()

A cosmetic patch exists in `Stats()` (referenced in-code as `#582`):

```go
} else if b.state == StateOpen && consecutiveFails == 0 {
    // Path B (#582): the breaker re-opened from HalfOpen on a failed probe.
    consecutiveFails = b.config.FailureThreshold
}
```

This fixes the *display* value but not the underlying field. Any code path that reads `b.consecutiveFails` directly—rather than going through `Stats()`—will observe the inconsistent value. Today that is only `Stats()` itself, but the field is exported via the struct and read without the lock in tests; future code could read it assuming the invariant holds.

## Trigger / Repro

```go
cb := New(Config{FailureThreshold: 3, CooldownPeriod: 10 * time.Millisecond})

// Trip the breaker (Closed → Open)
for i := 0; i < 3; i++ {
    cb.recordFailure()
}
// assert cb.state == StateOpen, cb.consecutiveFails == 3  ✓

// Advance past cooldown, allow one probe (Open → HalfOpen)
time.Sleep(15 * time.Millisecond)
cb.allow()  // transitions to HalfOpen

// Probe fails (HalfOpen → Open)
cb.recordFailure()

// BUG: consecutiveFails is 0, not ≥ FailureThreshold
// Stats() hides this via the workaround, but the field itself is wrong.
fmt.Println(cb.consecutiveFails)  // prints 0
```

## Suggested Fix

In the HalfOpen→Open branch of `recordFailure`, preserve or restore the failure count rather than zeroing it. The semantically correct value is 1 (this one probe failure triggered the re-open):

```go
case StateHalfOpen:
    b.state = StateOpen
    b.consecutiveFails = 1   // one probe just failed; do NOT reset to 0
    b.halfOpenGen++
    ...
```

Alternatively, set it to `b.config.FailureThreshold` to match what `Stats()` already projects (keeping external and internal views consistent). With either fix, remove the `// Path B` workaround from `Stats()`.

---

*Session: rev-20260615T1610Z-501f67*

eunolabs:
## Bugs addressed (session rev-20260615T1610Z-501f67)

- Closes #585 — `killswitch.Redis.Reset()` deletes the Redis keys, then clears the local cache, with no lock held across both phases. A concurrent `KillAgent`/`KillSession` could SET its Redis key after the delete sweep and then have its local-cache write wiped by the clear, leaving the kill durable in Redis but invisible to `ShouldBlock` (which reads the local cache) until the next reconcile tick — a temporary fail-open of up to the reconcile interval (~30 s). `Reset()` now calls `refreshState(ctx)` after clearing the cache to re-read Redis and re-seed any kill that raced in, shrinking the window to the sub-millisecond read→swap gap. It is best-effort (the deletions already succeeded; a transient refresh error is recorded in `lastRefreshErr` and caught by the next reconcile, and must not fail an otherwise-complete reset). The error path is unaffected — `refreshState` only runs after the deletions succeed and the cache is cleared.
- Closes #586 — on a HalfOpen→Open re-trip, `recordFailure` reset `consecutiveFails` to `0`, violating the invariant "State==Open ⟹ consecutiveFails ≥ FailureThreshold" and forcing the read-side `#582` projection in `Stats()`. The original `#428` reset-to-0 existed only to avoid `FailureThreshold+k` growth across repeated failed probes. Pinning to `FailureThreshold` instead satisfies both: it never grows across re-trips **and** keeps the field internally consistent, so the `#582` `Stats()` projection is removed (the field is now reported directly). The `#525` cooldown-elapsed projection (HalfOpen ⟹ 0) is unchanged.

## Testing
- New test: `TestRedis_Reset_ReseedsRacedKill` (#585) — a fake that surfaces an agent key only after the delete sweep confirms Reset's trailing `refreshState` re-seeds it so `ShouldBlock` blocks. The `fakeCmdable` was taught to model deletion (a deleted key no longer appears in later SCANs), which also keeps `TestRedis_Reset_Success` accurate now that Reset re-scans.
- Existing breaker tests cover #586: `TestBreaker_StatsConsecutiveFailsAfterProbeReopen` (#582) still sees `FailureThreshold` (now from the field, not a projection) and `TestBreaker_HalfOpenProbeFailureDoesNotInflateConsecutiveFails` (#428) still holds (`== FailureThreshold ≤ FailureThreshold`).
- `make check-fmt`, `make check-license`, `go vet ./...`, `golangci-lint run` clean.

Note: filed by the automated reviewer; #586 supersedes the `#582` read-side workaround with a write-side fix.

---


eunolabs (commented):

- #585: `refreshState` is self-locking and called after the cache-clear unlock, so no double-lock; it runs only on the success path (deletes errored ⟹ early return, cache preserved — `TestRedis_Reset_StatePreservedOnError` still passes). Best-effort by design: a failed re-read leaves the (already-cleared) cache empty and is recovered by reconcile, so it must not fail a reset whose deletions succeeded. The residual race window is the inherent read→swap gap the existing reconcile/pubsub paths already accept.
- #586: pinning to `FailureThreshold` (not `0`, not `1`) preserves both the `#428` no-growth property and the `Open ⟹ ≥ FailureThreshold` invariant, and matches what `#582` already projected — so removing the `Stats()` projection is behavior-preserving for external observers while fixing the internal field. The `#525` HalfOpen→0 projection is untouched and still tested.
- The `fakeCmdable` now models deletion (deleted keys drop out of later SCANs), which is strictly more realistic and was required because Reset now re-scans; all existing killswitch tests pass with it.

`go test -race ./...`, fmt, license, vet, golangci-lint clean.

---

eunolabs:
✅ Ready to merge — session rev-20260615T1610Z-501f67 addressed (#585, #586), self-review complete with no outstanding comments, Go CI green on c5c931f (Demo Integration is path-filtered, not triggered for a pkg-only change).

---

PR #604, opened 2026-06-15
Fix session rev-20260615T1910Z-5e54ed bugs

### LINKED ISSUE #602: [rev-20260615T1910Z-5e54ed] pkg/capability/jwks.go: VerifyToken never validates SchemaVersion against SupportedSchemaVersions
## Bug Report

**Session:** `rev-20260615T1910Z-5e54ed`
**Severity:** Medium
**Location:**
- [`pkg/capability/jwks.go` line 159 — `VerifyToken`](https://github.com/eunolabs/eunox/blob/71bdf4e0d2dad52f137f70b094fd1aa2ad346d28/pkg/capability/jwks.go#L159) (payload parsed, `SchemaVersion` populated, but never checked)
- [`pkg/capability/token.go` line 68 — `SupportedSchemaVersions`](https://github.com/eunolabs/eunox/blob/71bdf4e0d2dad52f137f70b094fd1aa2ad346d28/pkg/capability/token.go#L68) (defined, never read)

---

### Description

`pkg/capability/token.go` declares:

```go
var SupportedSchemaVersions = map[string]bool{"1.0": true}
```

`VerifyToken` in `pkg/capability/jwks.go` unmarshals the native `TokenPayload` (including its `SchemaVersion` field) at line 159:

```go
if err := tok.UnsafeClaimsWithoutVerification(&p); err != nil {
    return nil, fmt.Errorf("parse token payload: %w", err), nil
}
```

After that call, `p.SchemaVersion` holds whatever value was in the JWT. **`VerifyToken` never consults `SupportedSchemaVersions`** to validate it. The map is dead code: declared, imported by nothing, checked by nothing.

By contrast, the MCP-JWT path in `cmd/eunox-mcp/pdp_jwt.go` explicitly rejects an unknown schema version as a terminal error:

```go
if payload.MCP.Version != mcpClaimVersion {
    return nil, terminalError(...)
}
```

The native-token path has no analogous guard. Any `schemaVersion` value — including `""`, `"2.0"`, or an arbitrary string — is silently accepted and the token is interpreted under "1.0" semantics regardless.

### Impact

When a future schema version ("2.0") changes the semantics of existing fields — for example, altering how `capabilities` constraints are interpreted, how `authorizedBy` scoping works, or adding new required fields whose absence should deny — `VerifyToken` will accept tokens carrying `"schemaVersion": "2.0"` and interpret them as "1.0" tokens. Depending on the schema change, this can produce:

- **Incorrect allow decisions** — a "2.0" token intended to carry narrower restrictions than "1.0" is evaluated under the broader "1.0" logic.
- **Incorrect deny decisions** — a "2.0" field that grants additional access is ignored.
- **Silent regression** — the fail-closed invariant requires rejecting what the implementation does not understand; silently accepting an unknown schema version violates this.

The bug also applies to schema version `""` (a token where the issuer omitted `schemaVersion` entirely), which is arguably invalid but currently accepted.

### Trigger / Reproduction

Sign a `TokenPayload` with `"schemaVersion": "2.0"`. `VerifyToken` will return the payload without error, and the caller will enforce "1.0" semantics regardless of what the "2.0" schema intended.

### Suggested Fix

Add a schema-version check in `tryKeys`, immediately after `UnsafeClaimsWithoutVerification` succeeds, mirroring the MCP-JWT version gate:

```go
if err := tok.UnsafeClaimsWithoutVerification(&p); err != nil {
    return nil, fmt.Errorf("parse token payload: %w", err), nil
}
if !SupportedSchemaVersions[p.SchemaVersion] {
    return nil, fmt.Errorf("unsupported native token schemaVersion %q", p.SchemaVersion), nil
}
```

This is a terminal rejection (second return value), not a retryable signature failure, because an unsupported schema version means the proxy cannot interpret the token's claims at all — the same rationale as the MCP-JWT path. The empty-string case (`""`) is naturally rejected because `SupportedSchemaVersions[""]` is `false`.

**Session:** `rev-20260615T1910Z-5e54ed`

### LINKED ISSUE #603: [rev-20260615T1910Z-5e54ed] pkg/capability/token.go: Audience.MarshalJSON emits "aud":null for nil Audience, producing non-RFC-7519 tokens
## Bug Report

**Session:** `rev-20260615T1910Z-5e54ed`
**Severity:** Medium
**Location:** [`pkg/capability/token.go` lines 60–65 (`Audience.MarshalJSON`)](https://github.com/eunolabs/eunox/blob/71bdf4e0d2dad52f137f70b094fd1aa2ad346d28/pkg/capability/token.go#L60-L65) and [line 85 (`TokenPayload.Audience` field tag)](https://github.com/eunolabs/eunox/blob/71bdf4e0d2dad52f137f70b094fd1aa2ad346d28/pkg/capability/token.go#L85)

---

### Description

`Audience.MarshalJSON` is:

```go
func (a Audience) MarshalJSON() ([]byte, error) {
    if len(a) == 1 {
        return json.Marshal(a[0])
    }
    return json.Marshal([]string(a))
}
```

When `a` is nil (or empty), `len(a) == 0`, so neither branch is taken. The function falls through to `json.Marshal([]string(nil))`, which produces the JSON literal `null`.

`TokenPayload` declares the field without `omitempty`:

```go
Audience Audience `json:"aud"`
```

As a result, a `TokenPayload` with no `Audience` set marshals as `{"aud": null, ...}`. RFC 7519 §4.1.3 defines `aud` as a StringOrURI or array thereof; JSON `null` is not a valid value for this claim.

### Impact

**1. Any issued token with an unset `Audience` is permanently invalid.**

`VerifyToken` in `pkg/capability/jwks.go` always runs an audience check (line 181):

```go
expected := jwt.Expected{
    Time:        c.now(),
}
if err := claims.ValidateWithLeeway(expected, time.Minute); err != nil {
    return nil, fmt.Errorf("validate claims: %w", err), nil
}
```

`UnmarshalJSON` for `Audience` treats JSON `null` as a nil/empty slice (matching the "no audience" case in its comment). A token with `"aud": null` would therefore have an empty `Audience`, which fails the `AnyAudience` check — the token is issued successfully but can never be verified.

**2. Silent issuance of malformed tokens.**

Any code path that constructs a `TokenPayload` without setting `Audience` and then signs it produces a structurally invalid JWT. There is no error at marshal or sign time; the fault surfaces only at verification — potentially after the token has been delivered to a client.

**3. Interoperability.**

Third-party JWT libraries that strictly validate `aud` type (requiring string or string array) will reject tokens where `aud` is `null`, breaking cross-system delegation flows.

### Trigger / Reproduction

```go
tp := capability.TokenPayload{
    Issuer:    "https://example.com",
    Subject:   "agent-1",
    // Audience not set
}
data, _ := json.Marshal(tp)
// data contains: {..., "aud":null, ...}
```

Pass `data` as the payload of a signed JWT. `VerifyToken` will fail at the audience check even if the expected audience is `""`.

### Suggested Fix

Two complementary fixes are needed:

**Fix 1 — Handle the nil/empty case in `MarshalJSON`:**

The cleanest resolution is to emit a JSON array for the zero-value case (consistent with RFC 7519 §4.1.3's array form), or to return `null` and rely on `omitempty` to suppress the field. Since `omitempty` with a custom `MarshalJSON` requires the marshaler to return `[]byte("null")` when empty, the simplest approach is:

```go
func (a Audience) MarshalJSON() ([]byte, error) {
    if len(a) == 0 {
        // Return null so omitempty suppresses the field, rather than emitting
        // "[]" (an array with no members, which is also invalid per RFC 7519 §4.1.3).
        return []byte("null"), nil
    }
    if len(a) == 1 {
        return json.Marshal(a[0])
    }
    return json.Marshal([]string(a))
}
```

**Fix 2 — Add `omitempty` to the field tag:**

```go
Audience Audience `json:"aud,omitempty"`
```

With `omitempty`, when `MarshalJSON` returns `null` the field is omitted entirely, producing a valid JWT payload without an `aud` claim (RFC 7519 §4.1.3 is optional). This also matches the intent of `UnmarshalJSON`'s comment: "An empty string or a JSON null yields an empty Audience (treated as 'no audience'), matching how an absent aud claim is handled."

**Alternative:** Validate at token-creation time that `Audience` is always set before signing, making the zero-value case a programming error rather than a silent emission of an invalid claim.

**Session:** `rev-20260615T1910Z-5e54ed`

eunolabs:
## Bugs addressed (session rev-20260615T1910Z-5e54ed)

- Closes #603 — `Audience.MarshalJSON` fell through to `json.Marshal([]string(nil))` for the zero value, emitting `"aud":null`. JSON `null` is not a valid StringOrURI per RFC 7519 §4.1.3, and `UnmarshalJSON`/`VerifyToken` treat it as "no audience", so any token issued without an `Audience` set was structurally invalid and could never pass the audience check. `MarshalJSON` now returns the literal `null` for the empty case and the `TokenPayload.Audience` field gained `omitempty`, so an unset audience omits the `aud` claim entirely (a valid, optional-claim-absent payload) instead of emitting `"aud":null`.
- Closes #602 — `VerifyToken` populated `TokenPayload.SchemaVersion` but never validated it against `SupportedSchemaVersions`, so a native token carrying any `schemaVersion` (including `"2.0"` or `""`) was silently accepted and interpreted under "1.0" semantics — a fail-open that violates the fail-closed invariant. `tryKeys` now rejects an unsupported `schemaVersion` as a terminal error immediately after parsing the payload, mirroring the MCP-JWT version gate in `cmd/eunox-mcp/pdp_jwt.go`. The empty-string case is naturally rejected because `SupportedSchemaVersions[""]` is false.

## Testing
- `go test -race -count=1 ./pkg/capability/` — all pass.
- `go test -race -count=1 ./cmd/...` — all pass.
- New `TestAudience_MarshalJSON` (single/multiple/nil/empty forms) and `TestTokenPayload_OmitsNullAudience` (#603).
- New `TestJWKSClient_VerifyToken_UnsupportedSchemaVersion` (#602) — rejects `"2.0"`, `"0.9"`, and `""`; the supported version still verifies.
- Updated `TestJWKSClient_VerifyToken_MissingExp` to carry a valid `schemaVersion` so it still isolates the missing-exp path under the new gate.
- `gofmt -l` clean; `go vet ./pkg/capability/` clean.

---


eunolabs:
✅ Ready to merge — all session rev-20260615T1910Z-5e54ed bugs addressed (#603, #602), self-review complete with no outstanding comments, tests passing (`go test -race ./...` green, `make check-fmt` / `make check-license` clean).

---

PR #607, opened 2026-06-15
Fix session rev-20260615T1810Z-72c2ba bugs

### LINKED ISSUE #598: [rev-20260615T1810Z-72c2ba] runConditions: multi-window maxCalls burns quota on second-condition denial
## Bug Report

**Session:** `rev-20260615T1810Z-72c2ba`
**Severity:** Medium
**Location:** [`pkg/enforcement/engine.go` lines 457–464](https://github.com/eunolabs/eunox/blob/71bdf4e0d2dad52f137f70b094fd1aa2ad346d28/pkg/enforcement/engine.go#L457-L464)

---

### Description

`runConditions` uses a two-pass loop to ensure consuming conditions (those with commit-on-admit side effects) never burn their quota for a call that a *predicate* condition denies. The design comment is explicit:

> "Two-pass evaluation so a consuming condition never burns its quota for a call that a later condition denies." (line 441)

However, the invariant only holds between pass 1 (predicates) and pass 2 (consuming). **Within** pass 2, multiple `maxCalls` conditions on the same constraint still execute sequentially with no rollback: the first admitted `maxCalls` increments its sliding-window counter before the second `maxCalls` can deny, leaving the slot permanently consumed for a call that the proxy ultimately rejected.

### Trigger / Reproduction

Configure a single constraint with two `maxCalls` conditions on different windows, e.g.:

```yaml
conditions:
  - type: maxCalls
    window: 1h
    limit: 10
  - type: maxCalls
    window: 24h
    limit: 5
```

1. Make 5 successful calls (exhausts the 24-hour budget).
2. Make a 6th call within the same hour.

**Expected:** both counters stay at 5; the call is denied.  
**Actual:** the hourly counter increments to 6, then the daily `maxCalls` denies. The hourly slot is consumed for a rejected call. A legitimate call in that hour is now permanently displaced.

### Root Cause

[`runConditions`, second pass](https://github.com/eunolabs/eunox/blob/71bdf4e0d2dad52f137f70b094fd1aa2ad346d28/pkg/enforcement/engine.go#L457-L464):

```go
for _, cond := range matched.Conditions {
    if !isDeferredCondition(cond.ConditionType()) {
        continue
    }
    if deny := e.evalCondition(...); deny != nil {
        return deny  // first maxCalls already committed
    }
}
```

`handleMaxCalls` calls `IncrementIfBelow`, which commits atomically. When a second `maxCalls` then denies, there is no rollback.

### Suggested Fix

Two options:

**Option A (simpler):** Reject manifests that declare more than one `maxCalls` condition on a single constraint in `pkg/capability` validation. Document it as a well-formedness rule.

**Option B (general):** Collect *all* deferred conditions' outcomes speculatively before committing any. For `maxCalls` this means adding a separate `Peek`-only path (analogous to `callcounter.Peek`) and only calling `IncrementIfBelow` after every deferred condition has passed in a dry-run. This is more complex but handles future consuming conditions without new validation rules.

Option A is lower-risk and avoids the dual-counter pattern that is likely unintentional.

### LINKED ISSUE #599: [rev-20260615T1810Z-72c2ba] ValidateAction/EvaluateConditions: session history poisoned when collectObligations denies
## Bug Report

**Session:** `rev-20260615T1810Z-72c2ba`
**Severity:** Medium
**Locations:**
- [`pkg/enforcement/engine.go` lines 588–598 (`ValidateAction`)](https://github.com/eunolabs/eunox/blob/71bdf4e0d2dad52f137f70b094fd1aa2ad346d28/pkg/enforcement/engine.go#L588-L598)
- [`pkg/enforcement/engine.go` lines 771–779 (`EvaluateConditions`)](https://github.com/eunolabs/eunox/blob/71bdf4e0d2dad52f137f70b094fd1aa2ad346d28/pkg/enforcement/engine.go#L771-L779)

---

### Description

In both `ValidateAction` and `EvaluateConditions`, the call sequence on the allow path is:

1. `recordSessionCall` — writes the tool name into session history (used by `sequenceBlock`)
2. `collectObligations` — translates directives into post-allow obligations; **fails closed** with a hard deny (`ENFORCEMENT_ERROR`) when it encounters a directive type it cannot handle

If step 2 returns a deny, the function correctly returns a denial to the caller — but step 1 already executed. The session history now records the tool call as having occurred, even though the proxy denied the request and never forwarded it to the upstream.

This violates the `sequenceBlock` invariant: its `Peek` consults history to determine whether a prerequisite tool has run. A false positive in history allows a `sequenceBlock` condition on a *subsequent* tool to pass when it should deny (the prerequisite was never actually executed).

### Trigger / Reproduction

The failure path in `collectObligations` is a maintenance guard — it fires only when a new directive type is added to `capability.Constraint` but `collectObligations` is not updated. In that state:

1. Deploy with a manifest that includes the new directive type on a constraint for tool A (a prerequisite).
2. Call tool A. `runConditions` passes. `recordSessionCall` writes A to history. `collectObligations` hits the unknown-directive branch and returns a hard deny. The proxy rejects the call.
3. Now call tool B, which has a `sequenceBlock` requiring A to have run first. `sequenceBlock.Peek` finds A in history and passes. Tool B is allowed when it should be denied.

### Root Cause

```go
// ValidateAction (lines 588–598)
if err := e.recordSessionCall(ctx, req); err != nil {   // (1) writes session history
    return recordFailureDenial(...), nil
}
obligations, denyResp := e.collectObligations(matched, ...) // (2) can deny
if denyResp != nil {
    return *denyResp, nil  // history already poisoned
}
```

### Suggested Fix

Swap the order: call `collectObligations` first (it has no side effects — it only translates directives into obligation structs), then call `recordSessionCall` only after confirming that all obligations can be collected and the response will be allowed:

```go
obligations, denyResp := e.collectObligations(matched, requestID, now)
if denyResp != nil {
    return *denyResp, nil
}
if err := e.recordSessionCall(ctx, req); err != nil {
    return recordFailureDenial(requestID, now, matched.IsAuditOnly()), nil
}
```

`collectObligations` is a pure translation function (no I/O, no counters) so reordering it does not change its semantics. Apply the same swap in `EvaluateConditions`.

### LINKED ISSUE #600: [rev-20260615T1810Z-72c2ba] killswitch InMemory.Status returns agents/sessions in non-deterministic order
## Bug Report

**Session:** `rev-20260615T1810Z-72c2ba`
**Severity:** Low
**Location:** [`pkg/killswitch/memory.go` lines 128–136 (`Status`)](https://github.com/eunolabs/eunox/blob/71bdf4e0d2dad52f137f70b094fd1aa2ad346d28/pkg/killswitch/memory.go#L128-L136)

---

### Description

`InMemory.Status()` populates the `KilledAgents` and `KilledSessions` slices by iterating over Go maps:

```go
agents := make([]string, 0, len(m.killedAgents))
for id := range m.killedAgents {
    agents = append(agents, id)
}

sessions := make([]string, 0, len(m.killedSessions))
for id := range m.killedSessions {
    sessions = append(sessions, id)
}
```

Go randomises map iteration order at runtime. The returned slices therefore have a different order on each call.

### Impact

1. **Test flakiness** — any test that asserts on the slice order (e.g. `assert.Equal(t, []string{"a","b"}, status.KilledAgents)`) will fail intermittently rather than consistently, making failures hard to reproduce.
2. **Inconsistent API responses** — callers that compare successive `Status` snapshots to detect changes (e.g., for polling-based monitoring or audit diffing) will see spurious diffs caused purely by ordering.

The Redis backend (`pkg/killswitch/redis.go`) should be checked for the same pattern for consistency.

### Suggested Fix

Sort both slices before returning:

```go
import "sort"

agents := make([]string, 0, len(m.killedAgents))
for id := range m.killedAgents {
    agents = append(agents, id)
}
sort.Strings(agents)

sessions := make([]string, 0, len(m.killedSessions))
for id := range m.killedSessions {
    sessions = append(sessions, id)
}
sort.Strings(sessions)
```

Apply the same fix to the Redis backend's `Status` implementation if it exhibits the same pattern, so both backends return deterministic output.

### LINKED ISSUE #601: [rev-20260615T1810Z-72c2ba] expandHome silently ignores bare `~` audit log path
## Bug Report

**Session:** `rev-20260615T1810Z-72c2ba`
**Severity:** Low
**Location:** [`cmd/eunox-mcp/audit.go` lines 1882–1891 (`expandHome`)](https://github.com/eunolabs/eunox/blob/71bdf4e0d2dad52f137f70b094fd1aa2ad346d28/cmd/eunox-mcp/audit.go#L1882-L1891)

---

### Description

`expandHome` expands a leading `~/` in the audit log path to the user's home directory. The guard is:

```go
func expandHome(p string) (string, error) {
    if !filepath.IsAbs(p) && len(p) >= 2 && p[:2] == "~/" {
        home, err := os.UserHomeDir()
        // ...
        return filepath.Join(home, p[2:]), nil
    }
    return p, nil
}
```

The condition requires `len(p) >= 2 && p[:2] == "~/"`. A path of exactly `"~"` (the bare tilde, without a trailing slash) satisfies neither sub-expression: `len("~") == 1` and `"~"[:2]` would panic (though `len(p) >= 2` short-circuits it). The path is returned **as-is**, unexpanded.

The function's surrounding comment (line 1879) documents the motivation: "openAuditSink would otherwise MkdirAll a directory literally named `~` under the working directory and silently write the tamper-evident audit log there, degrading a security control with no diagnostic." That exact problem applies equally to a bare `~` — it is not caught by the current guard.

### Trigger / Reproduction

Configure the audit log path as `~` in a proxy config file:

```yaml
audit:
  path: "~"
```

`expandHome("~")` returns `"~"` unchanged. `openAuditSink` then calls `os.MkdirAll("~", ...)`, creating a directory literally named `~` in the working directory. The audit log is written there with no error and no diagnostic — the tamper-evident log is silently misdirected.

### Suggested Fix

Extend the guard to also handle a bare `"~"` as a synonym for the home directory:

```go
func expandHome(p string) (string, error) {
    if !filepath.IsAbs(p) {
        if p == "~" {
            home, err := os.UserHomeDir()
            if err != nil {
                return "", fmt.Errorf("cannot expand %q: home directory is unavailable: %w", p, err)
            }
            return home, nil
        }
        if len(p) >= 2 && p[:2] == "~/" {
            home, err := os.UserHomeDir()
            if err != nil {
                return "", fmt.Errorf("cannot expand %q: home directory is unavailable: %w", p, err)
            }
            return filepath.Join(home, p[2:]), nil
        }
    }
    return p, nil
}
```

Or, factor out the `os.UserHomeDir()` call and use `strings.TrimPrefix` on the resolved path to avoid duplication.

eunolabs:
## Bugs addressed (session rev-20260615T1810Z-72c2ba)

- Closes #601 — `expandHome` guarded only on `len(p) >= 2 && p[:2] == "~/"`, so a bare `~` (no trailing slash) was returned unexpanded. `openAuditSink` would then `MkdirAll` a directory literally named `~` under the CWD and silently misdirect the tamper-evident audit log there — the exact trap `#568` set out to close. `expandHome` now also treats a bare `~` as the home directory (and fails closed, like the `~/` path, when home is unresolvable).
- Closes #600 — `InMemory.Status` (and `Redis.Status`) built `KilledAgents`/`KilledSessions` by ranging over maps, so Go's randomised iteration order made the slices non-deterministic — flaky order-sensitive tests and spurious diffs for callers comparing successive snapshots. Both backends now `sort.Strings` the slices before returning.
- Closes #599 — in `ValidateAction` and `EvaluateConditions`, `recordSessionCall` ran before `collectObligations`. When `collectObligations` fails closed (an unhandled directive type), the call was already written to session history even though the proxy denied it and never forwarded it — poisoning a later `sequenceBlock` Peek into seeing the denied tool as "run". `collectObligations` (a pure, side-effect-free translation) now runs first; `recordSessionCall` runs only after the response is confirmed allowed.
- Closes #598 — a constraint with two `maxCalls` conditions on different windows (e.g. 10/hour + 5/day) burned the shorter window's quota when the longer window denied: the first `IncrementIfBelow` committed before the second denied, and the commit is not rolled back. The issue suggested rejecting multi-`maxCalls` constraints, but distinct-window rate limits are an **intentional, documented feature** (see `validateMaxCallsWindowsDistinct` and `#278`), so removing them would be a regression. Instead, `runConditions` now peek-checks **every** deferred condition (non-committing, via the window-accurate `CallHistory.Peek`) before committing any; if any window would deny, the call is denied with nothing consumed. A single `maxCalls` keeps its original atomic `IncrementIfBelow` path. Since distinct windows key independent buckets (same-window duplicates are already rejected at load), once all peeks pass every commit admits.

## Testing
- `make check-fmt`, `make check-license`, `go vet ./...`, `golangci-lint run` — all clean.
- New tests:
  - `TestExpandHome_BareTilde` (#601).
  - `TestInMemory_Status_DeterministicOrder` (#600) — 20 calls all return the same sorted order.
  - `TestObligationDeny_DoesNotPoisonSessionHistory` (#599) — both `ValidateAction` and `EvaluateConditions`; verified it fails (deny→allow) without the reorder.
  - `TestMaxCalls_MultiWindow_DenialDoesNotBurnSiblingQuota` (#598) — a `spyLimiter` proves a day-window denial consumes 0 hour-window slots; verified it fails (hourly count 6 vs 5) without the fix.

---


eunolabs:
✅ Ready to merge — all session rev-20260615T1810Z-72c2ba bugs addressed (#601, #600, #599, #598), self-review complete with no outstanding comments, tests passing (`go test -race ./...` green; `make check-fmt` / `make check-license` / `go vet` / `golangci-lint` clean). Note on #598: implemented a non-committing peek-check (Option B) rather than rejecting multi-`maxCalls` constraints (Option A), because distinct-window rate limits are an intentional, documented feature.

---

PR #608, opened 2026-06-15
Fix session rev-20260615T2010Z-a62f44 bugs

### LINKED ISSUE #605: [rev-20260615T2010Z-a62f44] parseTableArgument silently drops empty-string elements in array, violating fail-closed invariant
**Location:** `pkg/enforcement/handlers.go` — `parseTableArgument`, `case []interface{}:` branch

**Reference:** https://github.com/eunolabs/eunox/blob/682c3cb47598ea58815ed0b29d06d8b5ec6e11e2/pkg/enforcement/handlers.go

**Severity:** Low — inconsistency in fail-closed behavior; no direct security bypass, but violates the documented invariant that empty table names must always deny

**Description:**

`parseTableArgument` handles a top-level empty string correctly: `case string:` checks `strings.TrimSpace(t) != ""` before returning a result, and falls through to `return nil, nil` when the string is empty. The caller in `handleAllowedTables` then sees `len(tables) == 0` and returns a `MISSING_CONTEXT` denial. This is the fail-closed path.

However, when the argument is an array, empty-string elements take a different path:

```go
case []interface{}:
    for _, item := range t {
        switch item.(type) {
        case string, map[string]interface{}:
            parsed, err := parseTableArgument(item)  // item = ""
            if err != nil {
                return nil, fmt.Errorf("array element %s", err)
            }
            out = append(out, parsed...)  // parsed is nil → silently no-op
        ...
        }
    }
    return out, nil
```

When `item` is `""`:
1. It matches `case string, map[string]interface{}:` (it is a string)
2. `parseTableArgument("")` returns `nil, nil` (no error, empty slice)
3. `out = append(out, nil...)` is a no-op — the empty string is silently dropped
4. Processing continues with the remaining elements

**Concrete example:**

```go
req.Arguments["tables"] = []interface{}{"users", ""}
// handleAllowedTables receives: tables = [{Table: "users"}]
// Empty string was silently dropped; "users" is evaluated normally
```

This is inconsistent with the top-level string case:
- `arguments["tables"] = ""` → `MISSING_CONTEXT` denial (correct, fail-closed)
- `arguments["tables"] = []interface{}{"users", ""}` → `[{Table:"users"}]` (empty string dropped, evaluation continues)
- `arguments["tables"] = []interface{}{""}` → empty tables list → `MISSING_CONTEXT` denial (correct)

The middle case is the gap: a non-empty array where one element is an empty string is treated as if the empty string was never there.

**Fix:**

After `parseTableArgument(item)` returns with `err == nil`, check whether `parsed` is empty (nil/zero-length). A string element that survives the type switch but yields no table is structurally malformed — treat it as an error, not a no-op:

```go
case string, map[string]interface{}:
    parsed, err := parseTableArgument(item)
    if err != nil {
        return nil, fmt.Errorf("array element %s", err)
    }
    if len(parsed) == 0 {
        return nil, fmt.Errorf("array element is an empty or blank table name")
    }
    out = append(out, parsed...)
```

This makes `["users", ""]` consistently fail-closed: the error surfaces as `CONDITION_FAILED` with an accurate message, matching the behavior of non-string elements (which already return an error via the `default:` case added in #403).

**Session:** rev-20260615T2010Z-a62f44

### LINKED ISSUE #606: [rev-20260615T2010Z-a62f44] parseCapClaims calls parseV2Claim and discards conditions on every list-filter call; two-phase optimization from buildConstraintsFromClaims not ported
**Location:** `cmd/eunox-mcp/pdp_jwt.go` — `parseCapClaims` (lines 903–913)

**Reference:** https://github.com/eunolabs/eunox/blob/682c3cb47598ea58815ed0b29d06d8b5ec6e11e2/cmd/eunox-mcp/pdp_jwt.go

**Severity:** Low (performance) — no correctness impact; affects latency for every `tools/list`, `resources/list`, and `prompts/list` response when JWT mode is active

**Description:**

`parseCapClaims` is called on the hot path for all three list-filter functions (`filterToolsListByClaims`, `filterResourcesListByClaims`, `filterPromptsListByClaims`). It converts raw capability claim strings into `parsedCapClaim{prefix, bareName}` pairs:

```go
type parsedCapClaim struct {
    prefix   capability.TargetType
    bareName string
}

func parseCapClaims(caps []string) []parsedCapClaim {
    out := make([]parsedCapClaim, 0, len(caps))
    for _, claim := range caps {
        prefix, bareName, _, err := parseV2Claim(claim)  // conditions discarded (_)
        ...
    }
}
```

`parseV2Claim` does a full parse of the v0.2 claim string, including `parseCondSuffix`, which:
1. Splits the claim on `?` to isolate the condition suffix
2. Percent-decodes each key-value pair
3. Allocates a `map[string]string` per parsed pair

The resulting conditions (third return value) are immediately discarded — `parseCapClaims` only needs `prefix` and `bareName` for list matching.

**Contrast with `buildConstraintsFromClaims`:**

`buildConstraintsFromClaims` was optimized in issue #575 to use a two-phase approach: first extract just `prefix+bareName` cheaply (using `splitV2Claim`), then parse conditions only for claims that match the target. Its comment explicitly calls out the asymmetry:

```go
// Phase 1 — cheap name-only parse to find matching claims (#575)
// Phase 2 — full parse only for matching claims (conditions may be expensive)
```

`parseCapClaims` feeds list filtering, where conditions are never evaluated (list items are kept or dropped purely by name). It should use the same cheap phase-1 extraction, but it was not updated when #575 was landed.

**Cost:**

For a JWT with C capability claims, each with K condition pairs of average length L:
- Current: O(C × K × L) work per list-filter call (all conditions parsed and discarded)
- Optimal: O(C) work per list-filter call (name extraction only)

At 100 tool claims × 5 conditions each, this is ~500 unnecessary percent-decodes and map allocations on every `tools/list` response. With the caching note at #506 ("parse once, not per tool"), the per-claim optimization was recognized but the per-condition savings were overlooked.

**Fix:**

Replace `parseV2Claim` in `parseCapClaims` with the cheaper two-phase extraction — extract `prefix` and `bareName` without parsing conditions. The exact mechanism depends on whether `splitV2Claim` (or an equivalent) is exported; if not, inline the split:

```go
func parseCapClaims(caps []string) []parsedCapClaim {
    out := make([]parsedCapClaim, 0, len(caps))
    for _, claim := range caps {
        // Strip condition suffix (after '?') — conditions are never evaluated
        // for list filtering; parseV2Claim's full parse is wasted work here.
        bare := claim
        if i := strings.IndexByte(claim, '?'); i >= 0 {
            bare = claim[:i]
        }
        prefix, bareName, err := parseTarget(bare) // or equivalent name-only parse
        if err != nil {
            continue
        }
        out = append(out, parsedCapClaim{prefix: prefix, bareName: bareName})
    }
    return out
}
```

Alternatively, expose a `splitV2Claim`-style helper that stops after extracting the target and reuse it in both `buildConstraintsFromClaims` (phase 1) and `parseCapClaims`.

**Session:** rev-20260615T2010Z-a62f44

eunolabs:
## Bugs addressed (session rev-20260615T2010Z-a62f44)

- Closes #605 — `parseTableArgument` fail-closes a top-level empty string and a non-string array element, but an empty/blank **string** element inside a populated array (e.g. `["users", ""]`) was silently dropped: `parseTableArgument("")` returns `(nil, nil)` and `append(out, nil...)` is a no-op, so `["users", ""]` evaluated as just `["users"]`. The array branch now returns an error when a string/map element yields no table, so a blank element denies with `CONDITION_FAILED` — consistent with the top-level empty-string and non-string-element paths. The top-level empty-string → `MISSING_CONTEXT` behavior (intentional, #403) is unchanged: the new guard is scoped to array elements only.
- Closes #606 — `parseCapClaims` (the per-response list-filter hot path) called `parseV2Claim`, which runs the full v0.2 parse including `parseCondSuffix` (percent-decoding + per-pair map allocation), then discarded the conditions — list filtering matches purely on namespace+bareName (`anyCapCovers` never consults conditions). It now does the same phase-1 name-only parse `buildConstraintsFromClaims` uses (#575): `splitV2Claim` + `capability.ParseTarget`, skipping condition parsing entirely. Behavior is identical — every claim is already fully validated (suffix included) at token-validation time, so a claim reaching this path cannot have a malformed suffix, and `splitV2Claim` keeps an http(s) resource URI's `?query` attached to the name exactly as before (#547).

## Testing
- `make check-fmt`, `make check-license`, `go vet ./...`, `golangci-lint run` — all clean.
- New tests:
  - `TestEngine_AllowedTables_ArrayWithEmptyElement` (#605) — empty/blank string and blank-map elements all deny `CONDITION_FAILED`; verified it fails (allow instead of deny) without the fix.
  - `TestParseCapClaims` (#606) — strips condition suffixes, keeps http(s) URI query on the name, drops unparseable heads, and asserts prefix+bareName equivalence with `parseV2Claim`'s head. The existing `TestJWT_FilterToolsList_FiltersToClaims` (which uses a condition-suffixed claim) continues to pass.

---


eunolabs:
✅ Ready to merge — all session rev-20260615T2010Z-a62f44 bugs addressed (#605, #606), self-review complete with no outstanding comments, tests passing (`go test -race ./...` green; `make check-fmt` / `make check-license` / `go vet` / `golangci-lint` clean).

---

PR #620, opened 2026-06-16
Fix session rev-20260615T2110Z-dbd6ac bugs

### LINKED ISSUE #609: [rev-20260615T2110Z-dbd6ac] token_cache: purgeExpired samples clock before acquiring read lock
## Location

`pkg/capability/token_cache.go`, line 344 — `purgeExpired()` function.

## Reference

Commit `c1ccd67a6581d74c481d38f976d5055364dd0944`

## Severity

**Medium**

## Description

`purgeExpired()` samples the current time with `now := c.cfg.Now()` **before** acquiring `c.mu.RLock()`. This is directly inconsistent with the explicitly documented fix applied to both `Get()` (#554) and `Put()` (#578), where the clock is sampled **under** the lock.

The `Get()` code comment explains the invariant:
> "A pre-lock sample can go stale while Get waits on RLock contention (a Put/purgeExpired write lock in flight): an entry that expired during that wait would still pass `now.After` and be served past its TTL."

The same staleness window applies to `purgeExpired()`. If `purgeExpired()` samples `now` before the write lock from a concurrent `Put()` is released, the Phase 1 scan runs with a `now` that may be significantly in the past. This causes `purgeExpired()` to miss entries that expired during the time it waited on the read lock, leaving them in the cache past their TTL. Because the stale `now` is also reused in the Phase 2 deletion under the write lock, those entries survive the full purge cycle.

```go
// current code — incorrect
func (c *TokenCache) purgeExpired() {
    now := c.cfg.Now()  // sampled BEFORE RLock — stale if lock contended
    c.mu.RLock()
    // Phase 1: scan uses potentially stale now
    ...
    c.mu.RUnlock()
    c.mu.Lock()
    // Phase 2: delete also uses the same stale now
    ...
}
```

## Trigger / Repro

1. Hold a write lock on `TokenCache` (e.g., via a concurrent `Put()` with a long-lived token).
2. While that lock is held, advance the mock clock past the TTL of several existing entries.
3. Call `purgeExpired()` concurrently — it samples `now` before blocking on RLock.
4. After `Put()` releases its lock, `purgeExpired()` proceeds with the pre-lock `now`, which predates the expiry of those entries; they are not purged.

Under normal load this window is sub-millisecond, but it becomes material when `cfg.Now` is a test clock or when a large `Put()` holds the write lock for non-trivial time.

## Suggested Fix

Move the `now` sample to inside the RLock, matching the pattern already established by `Get()` and `Put()`:

```go
func (c *TokenCache) purgeExpired() {
    c.mu.RLock()
    now := c.cfg.Now()  // sample under lock — consistent with Get() / Put()
    // Phase 1 scan ...
    c.mu.RUnlock()
    c.mu.Lock()
    // Phase 2 re-check — re-sample now here too for correctness
    now = c.cfg.Now()
    // delete ...
    c.mu.Unlock()
}
```

Re-sampling `now` under the write lock in Phase 2 is also advisable: if significant time passes between Phase 1 (RLock) and Phase 2 (write Lock), a second sample ensures the deletion re-check is current.

## Session

`rev-20260615T2110Z-dbd6ac`

### LINKED ISSUE #610: [rev-20260615T2110Z-dbd6ac] killswitch: scanPrefix and deleteByPrefix silently no-op when client lacks Scanner interface
## Location

`pkg/killswitch/redis.go`, lines 451–478 — `scanPrefix()` and `deleteByPrefix()`.

## Reference

Commit `c1ccd67a6581d74c481d38f976d5055364dd0944`

## Severity

**High** — kill switch fails open for agent and session kills.

## Description

Both `scanPrefix()` and `deleteByPrefix()` perform a runtime type-assertion against an unexported `Scanner` interface. If the `redis.Cmdable` client does not implement that interface (e.g., a cluster client, ring client, test mock, or any future client wrapping), both functions **silently return nil (success) without performing any work**.

```go
// scanPrefix (lines 451-470) — simplified
func (r *RedisKillSwitch) scanPrefix(ctx context.Context, prefix string, dest map[string]bool) error {
    scanner, ok := r.client.(Scanner)
    if !ok {
        return nil  // silent no-op — dest is never populated
    }
    // ...
}

// deleteByPrefix (lines 474-478) — simplified
func (r *RedisKillSwitch) deleteByPrefix(ctx context.Context, prefix string) error {
    scanner, ok := r.client.(Scanner)
    if !ok {
        return nil  // silent no-op — keys are never deleted
    }
    // ...
}
```

**`scanPrefix` impact**: `refreshState()` calls `scanPrefix` to load the agent and session kill sets from Redis. When the type assertion fails, `refreshState()` loads an empty map, causing the in-process kill cache to show zero active kills even when Redis holds kill entries. Any agent or session that has been killed will appear alive to this instance.

**`deleteByPrefix` impact**: `Reset()` calls `deleteByPrefix` to remove all kill keys from Redis. When the type assertion fails, `Reset()` returns success but leaves all kill entries in Redis, corrupting the kill-switch state for all instances sharing that Redis.

Both paths violate the **fail-closed** invariant. The correct behavior on a client that cannot scan is to return an error (or panic at construction time), not to silently succeed with empty/no-op results.

## Trigger / Repro

1. Construct a `RedisKillSwitch` with a `redis.Cmdable` client that does not implement the `Scanner` interface (e.g., `redis.ClusterClient` without the scan adapter, or a hand-written test stub).
2. Call `KillAgent("agent-1")` followed by a restart/reconnect that triggers `refreshState()`.
3. Observe that `IsKilled("agent-1")` returns `false` — the kill is invisible.
4. Call `Reset()` and observe it returns nil while agent-1's kill key remains in Redis.

## Suggested Fix

Return an error (not nil) when the client does not implement `Scanner`, so callers fail closed:

```go
func (r *RedisKillSwitch) scanPrefix(ctx context.Context, prefix string, dest map[string]bool) error {
    scanner, ok := r.client.(Scanner)
    if !ok {
        return fmt.Errorf("killswitch: redis client %T does not implement Scanner; cannot enumerate kill-switch keys", r.client)
    }
    // ...
}
```

Alternatively, validate the interface at construction time in `NewRedisKillSwitch` and return an error early, preventing the silent-fail state from ever being reached.

## Session

`rev-20260615T2110Z-dbd6ac`

### LINKED ISSUE #611: [rev-20260615T2110Z-dbd6ac] killswitch: handlePubSubMessage reset case clears local state without re-reading Redis
## Location

`pkg/killswitch/redis.go`, lines 528–531 — `handlePubSubMessage()`, the `"reset"` case.

## Reference

Commit `c1ccd67a6581d74c481d38f976d5055364dd0944`

## Severity

**Medium**

## Description

The `"reset"` pub/sub handler clears the local in-memory kill state but does **not** call `refreshState()` afterward, unlike the direct `Reset()` method:

```go
// Reset() — correct: delete from Redis, then clear local state
func (r *RedisKillSwitch) Reset(ctx context.Context) error {
    if err := r.deleteByPrefix(ctx, r.keyPrefix); err != nil {
        return err
    }
    r.mu.Lock()
    r.mu.Unlock()
    return nil
}

// handlePubSubMessage "reset" case — incorrect: only clears local state
case "reset":
    r.mu.Lock()
    r.mu.Unlock()
    // refreshState() NOT called here
```

**Scenario that causes a security window:**

1. Replica A broadcasts a "reset" event after deleting all kill keys from Redis.
2. Concurrently, a separate call to `KillAgent("agent-x")` writes a new kill key to Redis between step 1's delete and step 1's pub/sub delivery.
3. Replica B receives the "reset" pub/sub event, clears its local kill map — and never re-reads Redis.
4. Replica B now shows `IsKilled("agent-x") == false`, even though that agent is killed in Redis and would be loaded by any new instance that calls `refreshState()`.

The window closes only at the next scheduled reconcile tick. Depending on the reconcile interval, this could leave a killed agent running for a non-trivial period on replicas that received the reset message.

## Trigger / Repro

1. Start two `RedisKillSwitch` instances sharing the same Redis, both subscribed to the pub/sub channel.
2. On instance A, call `Reset()` to clear all kills.
3. Immediately after `deleteByPrefix` completes on A (but before A publishes), inject a `KillAgent("victim")` write directly to Redis.
4. A publishes "reset"; B receives it, clears local state without refreshing.
5. Query `B.IsKilled("victim")` — returns false, but the kill is in Redis.

## Suggested Fix

Call `refreshState()` after clearing local state in the `"reset"` pub/sub handler, matching the semantic intent of a full reset (which should synchronize with what is now in Redis):

```go
case "reset":
    r.mu.Lock()
    r.mu.Unlock()
    // Re-read Redis so any kills that arrived after the publisher's delete
    // are visible without waiting for the next reconcile tick.
    _ = r.refreshState(context.Background())
```

Alternatively, make the "reset" event carry a logical clock or fence token so receivers can distinguish a true empty-state reset from a reset that raced with new kills.

## Session

`rev-20260615T2110Z-dbd6ac`

### LINKED ISSUE #612: [rev-20260615T2110Z-dbd6ac] circuitbreaker: Stats() and State() each call now() independently, causing split-brain state projection
## Location

`pkg/circuitbreaker/breaker.go` — `State()` and `Stats()` methods (projection logic for Open→HalfOpen transition).

## Reference

Commit `c1ccd67a6581d74c481d38f976d5055364dd0944`

## Severity

**Low** — observability inconsistency; does not affect admission decisions.

## Description

Both `State()` and `Stats()` independently project the logical circuit-breaker state by calling `b.now()` and comparing against `b.openedAt + b.config.CooldownPeriod`. Because they call `b.now()` separately, a caller that invokes both in rapid succession near the cooldown boundary can observe inconsistent results:

```go
// State() — calls b.now() once
    b.mu.RLock()
    defer b.mu.RUnlock()
    if b.state == StateOpen && !b.now().Before(b.openedAt.Add(b.config.CooldownPeriod)) {
        return StateHalfOpen  // projected
    }
    return b.state
}

// Stats() — calls b.now() again, independently
func (b *Breaker) Stats() BreakerStats {
    b.mu.RLock()
    defer b.mu.RUnlock()
    state := b.state
    if b.state == StateOpen && !b.now().Before(b.openedAt.Add(b.config.CooldownPeriod)) {
        state = StateHalfOpen  // projected with a different now() sample
    }
    // ...
}
```

Because `b.now()` is called twice under separate RLock acquisitions, `State()` and `Stats()` can disagree on whether the breaker has crossed the cooldown boundary. For example, if `State()` is called at `T = cooldown - 1ns` and `Stats()` is called at `T = cooldown + 1ns`, a caller that calls both in sequence sees `StateOpen` from `State()` and `StateHalfOpen` from `Stats()`. Dashboards, health checks, or metrics collectors that read both fields will surface contradictory data.

This is a projection-layer inconsistency only. The authoritative state transition is gated inside `allow()` under the write lock, which calls `b.now()` exactly once and transitions atomically. Admission control is not affected.

## Trigger / Repro

1. Open the breaker; configure a short `CooldownPeriod` (e.g., 1ms).
2. In a tight loop, call `b.State()` and `b.Stats()` concurrently near the cooldown boundary using a real clock.
3. Log the results and look for records where `State()` returns `StateOpen` and `Stats().State` returns `StateHalfOpen` (or vice versa) in the same iteration.

## Suggested Fix

Unify the projection into a private helper that samples `b.now()` once and returns both the projected state and a consistent timestamp. Both `State()` and `Stats()` call the helper:

```go
// projectedState must be called under b.mu.RLock().
func (b *Breaker) projectedState() (State, time.Time) {
    now := b.now()
    if b.state == StateOpen && !now.Before(b.openedAt.Add(b.config.CooldownPeriod)) {
        return StateHalfOpen, now
    }
    return b.state, now
}
```

This eliminates the dual `b.now()` calls and guarantees that `State()` and `Stats()` always agree on the projected state for a given lock acquisition.

## Session

`rev-20260615T2110Z-dbd6ac`

eunolabs:
## Bugs addressed (session rev-20260615T2110Z-dbd6ac)

- Closes #609 — `TokenCache.purgeExpired` sampled the clock *before* acquiring the read lock, so an entry that expired while the call waited on lock contention could survive past its TTL. Now samples `now` under the RLock (matching the #554/#578 fixes on `Get`/`Put`) and re-samples under the write lock in phase 2 so the deletion re-check is current.
- Closes #610 — `scanPrefix` and `deleteByPrefix` silently returned `nil` (success, no work) when the client failed the `Scan` type assertion, which would let `refreshState` load an empty kill set (fail-open) or `Reset` report success while leaving keys in Redis. They now return an error, failing closed per the project invariant. (Note: `redis.Cmdable` already mandates `Scan`, so the branch is defense-in-depth — see review note below.)
- Closes #611 — the `"reset"` pub/sub handler cleared local state but never re-read Redis, so a `KillAgent`/`KillSession` that raced the publisher's delete sweep stayed invisible on receiving replicas until the next reconcile tick (fail-open window up to the reconcile interval). It now sets `shouldRefresh = true`, mirroring `Reset()`'s trailing `refreshState`.
- Closes #612 — `State()` and `Stats()` each duplicated the Open→HalfOpen cooldown projection logic. Unified into a single `projectedState()` helper (one clock sample per call, one source of projection truth) so the two methods can never diverge in how they project the logical state.

## Testing
- `go test -race ./...` (full suite green)
- Added `TestStateAndStatsShareProjection` (circuitbreaker) and `TestRedis_HandlePubSubMessage_Reset_RefreshesFromRedis` (killswitch).
- `make check-fmt`, `go vet ./...`, `make build` all clean.

---


eunolabs:
✅ Ready to merge — all session rev-20260615T2110Z-dbd6ac bugs addressed (#609, #610, #611, #612), self-review complete with no outstanding comments, tests passing (`go test -race ./...`, `make check-fmt`, `go vet ./...`, `make build` all clean).

---

PR #621, opened 2026-06-16
Fix session rev-20260615T2107Z-630aee bugs

### LINKED ISSUE #613: [rev-20260615T2107Z-630aee] rotatedSiblings matches directories, allowing pruneRotated to silently fail or newestRotatedSibling to select a non-file
**Location:** `cmd/eunox-mcp/audit.go` — `rotatedSiblings` (lines ~98–114), `pruneRotated`, `newestRotatedSibling`

**Reference:** https://github.com/eunolabs/eunox/blob/c1ccd67a6581d74c481d38f976d5055364dd0944/cmd/eunox-mcp/audit.go

**Severity:** Medium

**Description:**

`rotatedSiblings` scans the directory containing the audit log file and returns every entry whose name has the base log filename as a prefix:

```go
func rotatedSiblings(logPath string) ([]string, error) {
    dir := filepath.Dir(logPath)
    base := filepath.Base(logPath)
    entries, err := os.ReadDir(dir)
    // ...
    for _, e := range entries {
        if strings.HasPrefix(e.Name(), base+".") {
            siblings = append(siblings, filepath.Join(dir, e.Name()))
        }
    }
    return siblings, nil
}
```

`os.ReadDir` returns both files and directories. The prefix check `strings.HasPrefix(e.Name(), base+".")` does not filter by type. A directory named `audit.log.20260614T120000Z` (e.g., created accidentally by a deployment tool or by a misconfigured log rotator) will be included in the returned slice.

This has two downstream failure modes:

**1. `pruneRotated` silently fails to prune:**
`pruneRotated` calls `rotatedSiblings`, then filters by `rotatedAuditRe` (a regexp that only matches timestamp-suffix filenames), sorts, and calls `os.Remove` on the oldest entries. `os.Remove` on a non-empty directory returns an error which is logged but not fatal — so the prune loop succeeds on the surface while failing to reclaim space. If the directory is empty, `os.Remove` silently deletes it, consuming one of the `keepRotated` retention slots and leaving a ghost entry in the sorted list without removing any actual log file.

**2. `newestRotatedSibling` may select a directory as the chain-resume source:**
`newestRotatedSibling` walks the sorted sibling list in reverse to find the last entry whose name matches `rotatedAuditRe`. If a matching directory appears later in ASCII-sort order than all valid rotated log files (e.g., if its name sorts after the newest real file), the walk-back will skip it (because it matches `rotatedAuditRe` but `os.Open` / reading it as a file would fail). However, if `newestRotatedSibling` is called on a directory entry that passes the regexp, `openAuditSink` will attempt to `os.Open` it; `os.Open` on a directory succeeds on Linux but subsequent `Read` calls return an error. The error surfaces as "could not read newest rotated sibling" and is logged, but the chain is broken — `prevHMAC` is reset to the zero value and the audit chain is silently interrupted.

**Trigger / repro:**

1. Configure an audit log at `/var/log/eunox/audit.log` with `keepRotated: 3`.
2. Create a directory `/var/log/eunox/audit.log.20260614T000000Z` (e.g., via `mkdir`).
3. Restart eunox-mcp so `openAuditSink` is called. Observe that `newestRotatedSibling` may select the directory, causing a chain break in the HMAC linkage.
4. Alternatively, rotate the log 4+ times and observe that `pruneRotated` fails to delete the directory and retains one fewer real log file than `keepRotated` specifies.

**Suggested fix:**

In `rotatedSiblings`, filter to regular files only using `e.Type().IsRegular()` (or exclude non-files):

```go
for _, e := range entries {
    if !e.Type().IsRegular() {
        continue
    }
    if strings.HasPrefix(e.Name(), base+".") {
        siblings = append(siblings, filepath.Join(dir, e.Name()))
    }
}
```

This ensures `pruneRotated`, `newestRotatedSibling`, and any other consumers of `rotatedSiblings` only operate on regular files, eliminating both failure modes.

**Session:** rev-20260615T2107Z-630aee

### LINKED ISSUE #614: [rev-20260615T2107Z-630aee] handleAllowedExtensions loop variable shadows Engine receiver, masking future bugs
**Location:** `pkg/enforcement/handlers.go` — `handleAllowedExtensions` method body

**Reference:** https://github.com/eunolabs/eunox/blob/c1ccd67a6581d74c481d38f976d5055364dd0944/pkg/enforcement/handlers.go

**Severity:** Low

**Description:**

The `handleAllowedExtensions` method on `*Engine` contains a range loop whose iteration variable reuses the name `e`, which is also the method receiver:

```go
func (e *Engine) handleAllowedExtensions(ctx context.Context, req Request, ae capability.AllowedExtensions, ...) Decision {
    // ...
    for _, e := range ae.Extensions {   // <-- e string shadows e *Engine
        if strings.HasSuffix(lower, "."+strings.ToLower(e)) {
            return allow(...)
        }
    }
    // ...
}
```

Within the loop body, `e` refers to the `string` iteration variable, not the `*Engine` receiver. Any access to `e.clock`, `e.counter`, `e.logger`, or any other Engine field or method inside the loop would silently operate on the loop variable (a string), causing a compile error only because `string` has no such fields — rather than the more dangerous case of accessing the wrong value.

This is a latent maintainability hazard: the next developer who needs to add a log call, a call-counter increment, or a clock read inside the extension-match loop will receive a confusing compile error ("e.SomeMethod undefined (type string has no field or method SomeMethod)") rather than the direct hint "rename the loop variable". In larger or more complex refactors, there is a real risk of the shadowed identifier being accessed via a different code path introduced in the same diff, where the shadowing is less obvious.

The variable shadowing also silently prevents the `go vet` shadow check from flagging the issue in CI unless `-shadow` is explicitly enabled (it is not a default vet analyzer).

**Trigger / repro:**

The bug is structural and does not manifest at runtime today because the loop body only uses the `e` string. It becomes a compile error the moment any code that needs `e *Engine` is added inside the loop.

To observe: add `_ = e.clock` anywhere inside the loop body. The compiler reports `e.clock undefined (type string has no field or method clock)` rather than the expected behavior.

**Suggested fix:**

Rename the loop variable to avoid shadowing:

```go
for _, ext := range ae.Extensions {
    if strings.HasSuffix(lower, "."+strings.ToLower(ext)) {
        return allow(...)
    }
}
```

This is a pure rename with no behavioral change and eliminates the shadow.

**Session:** rev-20260615T2107Z-630aee

### LINKED ISSUE #615: [rev-20260615T2107Z-630aee] handleMaxCalls check-only denial returns full-window retryAfter even when a tighter estimate is available
**Location:** `pkg/enforcement/handlers.go` — `handleMaxCalls` check-only branch (the `CallHistory` / `Peek` path)

**Reference:** https://github.com/eunolabs/eunox/blob/c1ccd67a6581d74c481d38f976d5055364dd0944/pkg/enforcement/handlers.go

**Severity:** Low

**Description:**

`handleMaxCalls` operates in two modes, selected by a two-pass mechanism in `runConditions`:

1. **Commit mode** (`RateLimiter` / `IncrementIfBelow`): atomically attempts the increment. On denial, the backend (`redis.go` Lua script) returns the exact timestamp of the oldest entry in the window, so the caller can compute a tight `retryAfter` value: `time until that entry ages out`.

2. **Check-only mode** (`CallHistory` / `Peek`): reads the current count without incrementing. On denial, the code returns:

```go
return deny(CONDITION_FAILED, "max calls exceeded", retryAfter(int64(mc.WindowSeconds)))
```

`mc.WindowSeconds` is the full window duration — e.g., 3600 seconds for an hourly limit — even though the `Peek` response from the backend contains timestamps of the existing entries. The oldest entry timestamp is available (or inferable from `CallHistory.History()`) and would give a tighter upper bound: `window_end = oldest_entry_timestamp + WindowSeconds`. The actual `retryAfter` could be as low as 1 second when the window is nearly expired.

This matters in two scenarios:

**A. Multiple `maxCalls` conditions on one constraint.** The two-pass design exists specifically for this: the check pass prevents committing any counter when ANY condition will deny. If condition 1 (hourly limit, 3600s window) denies with `retryAfter=3600` but condition 2 (per-minute limit, 60s window) would have denied with `retryAfter=15`, the host receives `retryAfter=3600` and waits the full hour rather than 15 seconds. The worst-case denial dominates regardless of which condition fired.

**B. Single `maxCalls` in check-only mode (when `CallHistory` rather than `FullStore` is configured).** The host waits up to the full window even when only seconds remain.

The commit path does not have this problem because the backend returns a tight estimate directly. The check-only path deliberately avoids calling the backend's commit operation, but could still compute a tight estimate from the timestamps returned by `Peek` / `History()`.

**Trigger / repro:**

1. Configure a tool constraint with two `maxCalls` conditions: `{windowSeconds: 3600, limit: 10}` and `{windowSeconds: 60, limit: 5}`.
2. Make 10 calls within the hour but only 4 within the last minute.
3. Attempt an 11th call. The check-only pass fires on the hourly condition first (or whichever condition is evaluated first). The denial JSON includes `retryAfter: 3600`.
4. The host should retry in ~seconds (hourly window nearly exhausted), but instead waits an hour.

**Suggested fix:**

In the check-only denial branch, compute the tight `retryAfter` from the oldest timestamp in the Peek result. The `CallHistory` interface already exposes enough information:

```go
// After Peek returns count >= mc.Limit:
history, err := hist.History(ctx, key, mc.WindowSeconds)
if err == nil && len(history) > 0 {
    oldest := history[0] // oldest timestamp, Unix nanoseconds or seconds
    windowEnd := oldest + int64(mc.WindowSeconds)  // adjust units as needed
    tight := windowEnd - e.clock.Now().Unix()
    if tight < 1 {
        tight = 1
    }
    return deny(CONDITION_FAILED, "max calls exceeded", retryAfter(tight))
}
// fallback to full window if history unavailable
return deny(CONDITION_FAILED, "max calls exceeded", retryAfter(int64(mc.WindowSeconds)))
```

Alternatively, `Peek` could be extended to return the oldest-entry timestamp alongside the count so a separate `History` call is not needed.

**Session:** rev-20260615T2107Z-630aee

eunolabs:
## Bugs addressed (session rev-20260615T2107Z-630aee)

- Closes #613 — `rotatedSiblings` returned every directory entry whose name had the log-file prefix, including directories. A directory named like a rotated log could make `pruneRotated` silently fail (or burn a retention slot) and `newestRotatedSibling` break the HMAC chain. Now filters to regular files via `e.Type().IsRegular()`.
- Closes #614 — the extension-normalization loop in `handleAllowedExtensions` used `e` as its iteration variable, shadowing the `*Engine` receiver. Renamed to `ext` (pure rename, no behavior change) so a future `e.<field>` access inside the loop is a clear error.
- Closes #615 — a check-only `maxCalls` denial (multiple `maxCalls` conditions on one constraint, or a read-only counter) could only advise the *full window* as `retryAfter`. Added a read-only `PeekRetryAfter` to both counter backends (`InMemory`, and a new read-only Redis Lua script) and a consumer-side `RateLimitPeeker` interface; the check-only path now reports the same tight estimate (time until the oldest over-limit entry ages out) the commit path does. Falls back to the full-window hint for counters that don't implement it, so there is no fail-open risk. Refactored the shared denial/seconds logic into `maxCallsRateLimited` / `retryAfterSeconds`.

## Testing
- Added cross-backend agreement tests `TestInMemory_PeekRetryAfter_MatchesIncrementIfBelow` and `TestRedis_PeekRetryAfter_MatchesIncrementIfBelow` (both report the identical 35s, read-only), and `TestEngine_MaxCalls_CheckOnlyDenialReportsTightRetryAfter` proving a check-only denial now advises ~10s instead of the full 3600s window.
- Compile-time `RateLimitPeeker` conformance assertions for both backends.
- `make check-fmt`, `make check-license`, `go vet ./...` clean; docs updated in `capability-manifest-guide.md`.

---


eunolabs:
✅ Ready to merge — all session rev-20260615T2107Z-630aee bugs addressed (#613, #614, #615), self-review complete with no outstanding comments, tests passing (`go test -race ./...`, `golangci-lint` 0 issues, `make check-fmt`/`check-license` clean).

---

Issue #481, opened 2026-06-15
[rev-20260615T0121Z-0bd5f2] redactFields re-serialization HTML-escapes the whole tool response, mutating non-redacted passthrough content


eunolabs:
**Location:** `cmd/eunox-mcp/pdp.go:979` (final `json.Marshal(result)` in `applyRedactObligs`) and `cmd/eunox-mcp/pdp.go:1019` (`json.Marshal(val)` in `redactJSONText`)

**Reference:** https://github.com/eunolabs/eunox/blob/e2dea12834a068d3129ba11bf116bed0c7672ba0/cmd/eunox-mcp/pdp.go#L924-L1024

**Severity:** Medium (data-fidelity / passthrough correctness; violates a documented invariant)

**Description:**

When a `redactFields` directive is configured for a tool, `applyRedactObligs` decodes the entire tools/call result into a generic map, applies redaction, and re-marshals the **whole** result with the standard library `json.Marshal`:

```go
out, err := json.Marshal(result)   // pdp.go:979
```

`redactJSONText` does the same for each JSON text content item:

```go
redacted, err := json.Marshal(val) // pdp.go:1019
```

Go's `encoding/json.Marshal` HTML-escapes `<`, `>`, `&` (and U+2028 / U+2029) by default. Confirmed:

```
input : {"url":"https://x?a=1&b=2","html":"<b>hi</b>"}
output: {"html":"<b>hi</b>","url":"https://x?a=1&b=2"}
```

Because the re-marshal runs over the **entire** response whenever *any* `redactFields` obligation is present, every text content item and the `structuredContent` object are rewritten — **including fields that were never redacted, and tools whose output simply happens to contain `&`, `<`, or `>`** (URLs with query params, HTML snippets, XML, code). The byte stream the host receives differs from what the upstream produced.

This directly contradicts the documented guarantee in `pkg/capability/directive.go:28` and the `applyRedactObligs` doc comment, which both promise that content the proxy does not redact is **"preserved unchanged."** It is not preserved unchanged for any value containing HTML-significant characters.

A strict JSON consumer that re-decodes the value is unaffected (`&` decodes back to `&`), but any host that string-matches, hashes, diffs, or displays the raw text content sees altered output — silently, and only when redaction is configured.

**Trigger / repro:**

1. Configure a `redactFields` directive (any path) on a `tool:` target.
[...]

PR #622, opened 2026-06-16
Fix session rev-20260615T2158Z-ae0f38 bugs

### LINKED ISSUE #616: [rev-20260615T2158Z-ae0f38] killswitch Redis handlePubSubMessage "reset" does not trigger reconcile, leaving other replicas fail-open for kills concurrent with reset
**Location:** `pkg/killswitch/redis.go` — `handlePubSubMessage` (the `"reset"` case), `Reset()`

**Reference:** https://github.com/eunolabs/eunox/blob/c1ccd67a6581d74c481d38f976d5055364dd0944/pkg/killswitch/redis.go

**Severity:** Medium

**Description:**

`handlePubSubMessage` processes kill-switch pub/sub events and updates the local in-memory cache. For the `"reset"` message it clears all local state:

```go
case payload == "reset":
    r.globalActive = false
```

For unrecognized messages it falls through to `shouldRefresh = true`, which causes a full reconcile from Redis. The `"reset"` case is explicitly recognized and never sets `shouldRefresh`.

`Reset()` itself calls `refreshState(ctx)` as its final step to pick up any kills that were established between `deleteByPrefix` and the end of the operation. This correctly handles the race on the local replica. But OTHER replicas that receive the `"reset"` pub/sub message only clear their local cache; they do not refresh from Redis.

**The race:**

1. `Reset()` calls `deleteByPrefix(redisAgentPrefix)` — no key yet for agent X.
2. A concurrent `KillAgent("X")` runs on any replica: sets `killswitch:agent:X` in Redis, publishes `"agent:kill:X"`.
3. `Reset()` publishes `"reset"`.
4. On replica B, pub/sub delivers `"agent:kill:X"` then `"reset"` (the server observed `KillAgent` first): replica B sets `killedAgents["X"]=true`, then clears it.
5. `killswitch:agent:X` is still in Redis (not deleted — `KillAgent` ran after `deleteByPrefix`).
6. `Reset()` calls `refreshState` on replica A — correctly sees and seeds the kill.
7. Replica B has `killedAgents["X"]=false` but Redis has the key → replica B fails open for agent X for up to one `reconcileInterval` (default 30 seconds).

In the reverse pub/sub ordering (`"reset"` before `"agent:kill:X"` reaches replica B) — replica B correctly ends up with the kill. But pub/sub ordering between two independent publishers is non-deterministic.

This means a kill issued concurrently with a reset can be invisible to non-initiating replicas for up to the full reconcile interval, during which those replicas fail open for the killed agent or session.

**Trigger / repro:**

1. Deploy two or more replicas sharing a Redis backend.
2. From replica A, call `KillAgent("target-agent")`.
3. From replica A (or any replica), call `Reset()` while `KillAgent` is still in progress — specifically, after `deleteByPrefix` completes but before `refreshState` is called.
4. On replica B, observe that `ShouldBlock(ctx, "target-agent", "")` returns `false` for up to 30 seconds, even though Redis still holds the `killswitch:agent:target-agent` key.

**Suggested fix:**

Set `shouldRefresh = true` in the `"reset"` case so replica B triggers a full Redis reconcile after clearing its local cache — mirroring the `default` case and shrinking the inconsistency window from the full reconcile interval to the sub-millisecond network round-trip:

```go
case payload == "reset":
    r.globalActive = false
    shouldRefresh = true  // re-read Redis to pick up concurrent kills
```

The refresh runs outside the lock (same pattern as the `default` case), so it does not block `ShouldBlock` callers during the Redis round-trip.

**Session:** rev-20260615T2158Z-ae0f38

### LINKED ISSUE #617: [rev-20260615T2158Z-ae0f38] redactJSONValue returns true for pure-scalar arrays, causing unnecessary re-marshaling that may change numeric formatting
**Location:** `cmd/eunox-mcp/pdp.go` — `redactJSONValue` and its caller `redactJSONText`

**Reference:** https://github.com/eunolabs/eunox/blob/c1ccd67a6581d74c481d38f976d5055364dd0944/cmd/eunox-mcp/pdp.go

**Severity:** Low

**Description:**

`redactJSONValue` is documented to:

> report whether the value was a redactable structure; scalars (string/number/bool/null) return false so the caller can leave the original text untouched.

Its implementation for arrays unconditionally returns `true`:

```go
case []interface{}:
    for _, item := range v {
        redactJSONValue(item, paths)  // recurse in place
    }
    return true   // ← always true, even for [1, 2, 3]
```

`redactJSONText` uses this return value to decide whether to re-marshal:

```go
if !redactJSONValue(val, paths) {
    // Scalar JSON: no structured field to redact.
    return text, nil
}
// ... re-marshal via marshalNoHTMLEscape
```

For a text content item whose body is a pure-scalar JSON array — e.g. `[1, 2, 3]` or `["a", "b"]` — `redactJSONValue` returns `true`, triggering a re-marshal via `marshalNoHTMLEscape`. This re-marshal produces semantically equivalent JSON but can differ in string representation from the original:

- Integer-valued floats are fine (`1.0` → `1` vs. input `1` — compatible)
- Scientific notation may diverge: `1e308` in the original text may marshal as `1e+308` (Go's `strconv.AppendFloat` adds the sign)
- Exponents close to boundary values may pick a different representation

The proxy's documented guarantee — "content the proxy does not redact is preserved unchanged" (issue #481) — is violated for tool responses that include a scalar array in a `text` content item. A host that hashes or byte-compares the tool response before and after proxying would observe a difference even when no redaction path matched any field in the response.

Note: this only affects `text` content items that:
1. Are valid JSON
2. Are a JSON array
3. Contain only scalars at every nesting level (no objects whose keys the redaction paths could address)

A tool returning e.g. `text: "[1234567890.123456789]"` (a precise numeric array) will have its text re-marshaled to `"[1.2345678901234568e+09]"` — a floating-point precision artifact of the parse-and-remarshal round-trip, not a value change, but a byte change.

**Trigger / repro:**

Configure a tool constraint with a `redactFields` directive. The tool's response carries a `text` content item whose body is a pure-scalar JSON array, e.g.:

```json
{"content":[{"type":"text","text":"[1, 2, 3]"}]}
```

Observe that the proxy returns `[1,2,3]` (whitespace removed) instead of the original `[1, 2, 3]`. For scientific notation, observe that `1e308` becomes `1e+308` in the returned response.

**Suggested fix:**

`redactJSONValue` should only return `true` for the array case if at least one element was itself a redactable structure (map or array containing a map):

```go
case []interface{}:
    redacted := false
    for _, item := range v {
        if redactJSONValue(item, paths) {
            redacted = true
        }
    }
    return redacted
```

With this change, a pure-scalar array (where every element returns `false`) returns `false`, causing `redactJSONText` to leave the original text bytes untouched — consistent with the documented "preserved unchanged" behavior and the godoc's intent for non-redactable structures.

An array of objects or an array containing an array of objects still returns `true` (because the recursive call returns `true` for those elements), so redaction continues to work correctly for all cases where it is actually needed.

**Session:** rev-20260615T2158Z-ae0f38

eunolabs:
## Bugs addressed (session rev-20260615T2158Z-ae0f38)

- Closes #616 — **already resolved in `main`.** This is the same killswitch-`reset` pub/sub bug as #611, which was fixed in PR #620 (commit `7044954`): the `"reset"` handler now sets `shouldRefresh = true` and re-reads Redis, exactly the fix #616 requests. The regression test `TestRedis_HandlePubSubMessage_Reset_RefreshesFromRedis` already covers it. No further code change is needed; this issue is closed for bookkeeping.
- Closes #617 — `redactJSONValue` returned `true` unconditionally for arrays, so a `text` content item that was a pure-scalar JSON array (`[1,2,3]`, `["a","b"]`, `[1e308]`) was needlessly parse-and-remarshaled, changing its bytes (whitespace stripped, `1e308` → `1e+308`) and violating the "content the proxy does not redact is preserved unchanged" guarantee (#481). It now returns `true` only when at least one element is itself a redactable structure (a map, or a nested array reaching one); a pure-scalar array returns `false`, so the caller leaves the original text untouched. Arrays of objects (and arrays of arrays of objects) still redact correctly.

## Testing
- Added `TestRedactFields_ScalarArrayJSON_BytePreserved` (scalar arrays — including nested scalar arrays and scientific notation — are byte-preserved); existing `TestRedactFields_TopLevelArray_IsRedacted` / `_NestedArrays` / `_TopLevelArrayOfArrays` confirm object-bearing arrays still redact.
- `make check-fmt` clean.

---


eunolabs:
✅ Ready to merge — session rev-20260615T2158Z-ae0f38 bugs addressed (#616 already resolved by the merged #611 fix in #620; #617 fixed here), self-review complete with no outstanding comments, tests passing (`go test -race ./...`, `make check-fmt` clean).

---

PR #626, opened 2026-06-16
Fix session rev-20260615T2207Z-10329c bugs

### LINKED ISSUE #618: [rev-20260615T2207Z-10329c] redactFields bypass via truncated JSON array with non-object first element
**Location:** `cmd/eunox-mcp/pdp.go:1027-1044` (`looksLikeUnparseableJSONContainer`)

**Reference:** https://github.com/eunolabs/eunox/blob/c1ccd67a6581d74c481d38f976d5055364dd0944/cmd/eunox-mcp/pdp.go#L1027-L1044

**Severity:** Medium (security — redaction directive bypass)

**Description:**

`looksLikeUnparseableJSONContainer` decides whether to fail-closed (block) or pass through a text content item that failed JSON parsing. For an array (`[`), it only checks whether the **first non-whitespace character after `[`** is `{` or `[`:

```go
case '[':
    rest := strings.TrimLeft(t[1:], " \t\r\n")
    if rest == "" {
        return false
    }
    return rest[0] == '{' || rest[0] == '['
```

The comment states "Only an object-bearing array (`[{` or `[[`) can carry a redaction target." This assumption is **incorrect**: an array like `[null, {"ssn":"123456789"}]` carries a redaction target, but its first element is `null`.

When the upstream response is a JSON array whose first element is a scalar/null (e.g., `[null, {"ssn":"..."}`) and the JSON is malformed or truncated:

1. `json.Unmarshal` fails.
2. `looksLikeUnparseableJSONContainer` checks `rest[0]` = `'n'` (for `null`) → returns `false`.
3. `redactJSONText` treats the text as free-form prose and returns it **unchanged**.
4. The sensitive field is forwarded unredacted.

If the same response were **well-formed** (`[null, {"ssn":"123456789"}]`), it would parse successfully, `redactJSONValue` would recurse into every element, and the SSN would be redacted. The bug exists only for the unparseable case.

**Trigger / repro:**

An adversarial upstream (or a legitimately buggy upstream) returns a text content item that is a truncated JSON array whose first element is `null`, `true`, `false`, a number, or a string — but a later element is a JSON object containing a field named in `redactFields`. Example:

```
[null, {"ssn":"123456789"}
```
(missing closing `}]`)

With a manifest entry such as:

```yaml
- target: tool:data_export
  actions: [call]
  directives:
    - type: redactFields
      fields: [ssn]
```

The proxy passes `[null, {"ssn":"123456789"}` through unchanged, exposing the SSN.

A deliberate adversarial bypass: the upstream controls the response and can craft it to start with `[null, ` or `[1, ` or `["ok", ` before the sensitive object, then truncate it (omit the closing bracket), to consistently evade the check.

**Suggested fix:**

Either:

1. **Fail-closed for any array whose text contains a `{`** after the `[` (guards against object-bearing mixed arrays; may have false positives on arrays of strings containing `{` but those are rare in practice and safer to block):

```go
case '[':
    rest := strings.TrimLeft(t[1:], " \t\r\n")
    if rest == "" {
        return false
    }
    // Fail closed for `[{` (object as first element) or `[[` (nested array),
    // AND for any array containing a `{` anywhere — a later object element can
    // carry a redaction target even when the first element is a scalar.
    if rest[0] == '{' || rest[0] == '[' {
        return true
    }
    return strings.ContainsRune(rest, '{')
```

2. **Fail-closed for ALL non-parseable arrays** and separately detect scalar-array log-line prefixes (e.g., bracket followed only by digits/uppercase) as exempt:

```go
case '[':
    rest := strings.TrimLeft(t[1:], " \t\r\n")
    if rest == "" {
        return false
    }
    // Only permit known log-line shapes ([ERROR], [2026-...]) through.
    // Any array shape that could carry a JSON object key fails closed.
    if len(rest) > 0 && (rest[0] >= 'A' && rest[0] <= 'Z' || rest[0] >= '0' && rest[0] <= '9') {
        return false
    }
    return true
```

Option 2 is more conservative and avoids the `ContainsRune` scan cost on long texts.

**Session:** rev-20260615T2207Z-10329c

### LINKED ISSUE #619: [rev-20260615T2207Z-10329c] boundAuditObligations: sentinel can exceed auditObligationsTotalCap by up to a few bytes
**Location:** `cmd/eunox-mcp/audit.go:833-868` (`boundAuditObligations`)

**Reference:** https://github.com/eunolabs/eunox/blob/c1ccd67a6581d74c481d38f976d5055364dd0944/cmd/eunox-mcp/audit.go#L833-L868

**Severity:** Low (correctness — minor invariant violation; no exploitable practical impact)

**Description:**

`boundAuditObligations` is meant to ensure that the serialized obligations slice fits within `auditObligationsTotalCap` (64 KiB). When the full slice exceeds the cap, it iterates through entries, reserving space for the truncation sentinel on each step:

```go
used := 1
kept := make([]string, 0, len(obligs))
for i, o := range obligs {
    omitted := len(obligs) - i
    sentinel := fmt.Sprintf("obligations_truncated:%d", omitted)
    reserve := len(sentinel) + perEntryOverhead  // len(sentinel) + 3
    if used+len(o)+perEntryOverhead+reserve > auditObligationsTotalCap {
        return append(kept, sentinel)
    }
    used += len(o) + perEntryOverhead
    kept = append(kept, o)
}
```

The truncation fires when: `used + len(o) + 3 + reserve > cap`

When truncation fires (dropping entry `o` and emitting the sentinel instead), the actual marshaled size of the result is:

```
used + len(sentinel) + 3
```

From the truncation condition: `used + len(o) + 3 + len(sentinel) + 3 > cap`, which rearranges to:

```
used + len(sentinel) + 3 > cap - len(o) - 3
```

Since `len(o) + 3 >= 3 > 0`, the actual output `used + len(sentinel) + 3` can be anywhere from `cap - 2` upward. When `len(o) = 0` (a zero-length obligation string), the truncation condition fires when `used + 6 + len(sentinel) > cap`, but the actual output is `used + len(sentinel) + 3` — which can exceed `cap` if `used > cap - len(sentinel) - 3`.

**Concrete example** with `auditObligationsTotalCap = 65536`:

Suppose we have `len(obligs) = 10000`, the first 9000 entries average 4.3 chars (`used = 65511` after 9000 entries), and entry 9000 is a zero-length string. The sentinel for the remaining 1000 entries is `"obligations_truncated:1000"` (26 chars):

- Truncation check: `65511 + 0 + 3 + 29 = 65543 > 65536` → truncate ✓
- Actual result size: `65511 + 26 + 3 = 65540 > 65536` ← **4 bytes over cap**

The overflow is bounded by `len(o) + 3` bytes — at most a handful of bytes for any realistic sentinel length — but it violates the invariant that `boundAuditObligations` guarantees output within the cap.

**Impact:**

In practice, an overflow of a few bytes out of 64 KiB out of the 4 MiB scanner buffer has no functional consequence; the audit log remains parseable and the record is well within `bufio`'s limit. However:

1. The documented invariant ("the bounded queue never retains an oversized obligations slice") is technically violated.
2. If `auditObligationsTotalCap` is ever reduced significantly, the overflow window grows.

**Trigger / repro:**

Create an `auditSink` and enqueue a record with a large number of short-to-medium-length obligations that total >64 KiB. The final serialized obligations slice returned by `boundAuditObligations` will exceed 65536 bytes by a small amount.

**Suggested fix:**

After appending the sentinel, verify the result fits, and if not, trim the last kept entry to make room. Alternatively, change the condition to subtract the actual sentinel cost from the cap before the per-entry budget check:

```go
if used+len(o)+perEntryOverhead > auditObligationsTotalCap-reserve {
    return append(kept, sentinel)
}
```

This changes the truncation condition to `used + len(o) + 3 > cap - reserve` (i.e., `used + len(o) + 3 + reserve > cap` but now also guarantees `used + reserve <= cap`). When truncation fires: `used + reserve = used + len(sentinel) + 3 <= cap` (because the condition ensures `used + len(o) + 3 <= cap - reserve <= cap - len(sentinel) - 3 + len(o) + 3`... let me verify:

With the new condition: truncation fires when `used + len(o) + 3 + reserve > cap`. The actual output is `used + reserve`. For this to be `<= cap`: `used + reserve <= cap`. The condition fires only when `used + len(o) + 3 + reserve > cap`, i.e., `used > cap - len(o) - 3 - reserve`. So `used + reserve > cap - len(o) - 3`. This still doesn't guarantee `used + reserve <= cap` unless `len(o) + 3 >= 0` (trivially true).

The simplest correct fix is a post-truncation check:

```go
result := append(kept, sentinel)
if encoded, err := json.Marshal(result); err == nil && len(encoded) <= auditObligationsTotalCap {
    return result
}
// Sentinel itself doesn't fit; drop the last kept entry to make room.
if len(kept) > 0 {
    return append(kept[:len(kept)-1], fmt.Sprintf("obligations_truncated:%d", len(obligs)-len(kept)+1))
}
return []string{fmt.Sprintf("obligations_truncated:%d", len(obligs))}
```

**Session:** rev-20260615T2207Z-10329c

eunolabs:
## Bugs addressed (session rev-20260615T2207Z-10329c)

- Closes #618 — `looksLikeUnparseableJSONContainer` decided pass-through vs fail-closed for an unparseable array by inspecting only the FIRST character after `[`, so a truncated array whose first element is a scalar but a later element is an object — `[null, {"ssn":"x"}`, `[1, {...}`, `["ok", {...}` — was waved through as prose and forwarded **unredacted** (an adversarial upstream could prefix the sensitive object with a scalar and truncate to evade `redactFields`). It now fails closed whenever the array text contains a `{` anywhere (an object can only appear via `{`). Scalar arrays and bracketed log-line prefixes (no `{`) still forward unchanged; the one accepted false positive is a malformed scalar-string array whose string embeds a literal `{`, which fails closed — the safe direction for an unverifiable body.
- Closes #619 — hardens the `boundAuditObligations` cap guarantee. A 200k-trial fuzz confirmed the existing arithmetic never exceeds the 64 KiB cap, **but the worst case lands exactly on the cap (zero headroom)** and nothing pinned the invariant, so any future change to the byte estimate could silently breach it. The truncation path now routes through `truncatedObligations`, which re-marshals and drops trailing kept entries (folding them into the sentinel count) until the result provably fits — making the cap invariant explicit and unconditional. (The issue's concrete "4 bytes over" example is not actually reachable — it would truncate one iteration earlier — but the hardening + regression test are the right resolution given the razor-thin margin.)

## Testing
- Added `TestRedactFields_MalformedArrayScalarFirstObjectLater_FailsClosed` (#618) and confirmed the existing scalar-array / log-prefix passthrough tests still pass.
- Added `TestBoundAuditObligations_NeverExceedsCap` (#619): a 3000-trial fuzz plus a boundary sweep (length-1 fillers + a trailing zero-length entry — the exact shape #619 raised) asserting the result never exceeds the cap and the sentinel count is consistent.
- `make check-fmt`, `make check-license`, `go vet ./...` clean.

---


eunolabs:
✅ Ready to merge — all session rev-20260615T2207Z-10329c bugs addressed (#618, #619), self-review complete with no outstanding comments, tests passing (`go test -race ./...`, `golangci-lint` 0 issues, `make check-fmt`/`check-license` clean).

---

PR #627, opened 2026-06-16
Fix session rev-20260616T0010Z-8ec038 bugs

### LINKED ISSUE #623: [rev-20260616T0010Z-8ec038] ForceRefreshForVerify marks negative-KID sentinel even after successful key-rotation fetch
## Location

`pkg/capability/jwkscache.go`, lines 239–249

## Reference

https://github.com/eunolabs/eunox/blob/b316488642a6647bfb32ba228b601e28adeb517e/pkg/capability/jwkscache.go#L239-L249

## Severity

Medium / Security

## Description

`ForceRefreshForVerify` unconditionally marks `verifyRetrySentinel` in the negative-KID cache after every forced refresh, regardless of whether the refresh actually found a newly rotated signing key:

```go
keys, err := c.ForceRefresh(ctx)
if err != nil {
    return nil, err
}
// BUG: sentinel is marked even when ForceRefresh returned a key set
// containing the newly rotated key — success and failure both suppress
// the next forced refresh for negativeKIDTTL (30 s).
c.markKIDAbsent(verifyRetrySentinel)
return keys, nil
```

The inline comment justifies this as "the same bounded false-negative window the kid path accepts", but the scopes differ:

- **Kid path** (`ForceRefreshForKID`): each unknown `kid` gets its own 30-second negative-cache slot. One blocked kid does not affect other kids or kid-less tokens.
- **Verify sentinel** (`ForceRefreshForVerify`): a single global slot covers *all* kid-less tokens. After one successful refresh, the entire kid-less verification path is suppressed for 30 seconds — not just the specific token that triggered the refresh.

## Trigger / repro

1. Issuer rotates signing key (rotation R1). A kid-less token signed with the new key triggers `ForceRefreshForVerify`, which fetches the new key and succeeds. `verifyRetrySentinel` is now marked for 30 s.
2. Within those 30 s, the issuer rotates again (rotation R2 — e.g., due to key compromise + emergency re-rotation).
3. A kid-less token signed with R2 arrives. `kidRecentlyAbsent(verifyRetrySentinel)` returns `true`, so `ForceRefreshForVerify` returns the *stale* cached key set from R1 without fetching. The token fails verification.
4. Every verification attempt for R2 tokens is rejected for the remainder of the 30-second window, effectively a partial outage for all kid-less token callers.

The window starts fresh each time the sentinel is marked, so continuous key rotation can extend the outage indefinitely.

## Suggested fix

Only mark the sentinel when the returned key set is identical to the previously cached set (i.e., no new key appeared and a subsequent retry would be pointless). If the refresh returned a *different* set, the suppression should not be activated:

```go
keys, err := c.ForceRefresh(ctx)
if err != nil {
    return nil, err
}
// Only suppress further forced refreshes when the fetch returned the
// same key set as the cache already held — a new key appeared, so the
// next rotation should not be blocked for negativeKIDTTL.
c.mu.RLock()
unchanged := c.jwks != nil && len(keys.Keys) == len(c.jwks.Keys)
c.mu.RUnlock()
if unchanged {
    c.markKIDAbsent(verifyRetrySentinel)
}
return keys, nil
```

A more precise check would compare key thumbprints, but the length guard is sufficient to open the fast path for the common rotation case.

## Session

`rev-20260616T0010Z-8ec038`

### LINKED ISSUE #624: [rev-20260616T0010Z-8ec038] ValidateToken calls time.Now() unconditionally before checking injected clock
## Location

`cmd/eunox-mcp/pdp_jwt.go`, lines 400–403

## Reference

https://github.com/eunolabs/eunox/blob/b316488642a6647bfb32ba228b601e28adeb517e/cmd/eunox-mcp/pdp_jwt.go#L400-L403

## Severity

Low

## Description

`ValidateToken` always executes a `time.Now()` syscall before checking whether an injected test clock is present, then silently discards the result when one is:

```go
now := time.Now()       // syscall always executed
if p.clock != nil {
    now = p.clock.Now() // result of line above is discarded
}
```

The preceding comment (lines 396–399) explicitly documents the intent as "Resolve 'now' from the injected clock when present … A nil clock falls back to the wall clock", which is the *opposite* of what the code does: it resolves the wall clock first and only conditionally overrides it.

This is on the hot path — `ValidateToken` is called for every incoming request when `JWTPDP` is active — so the dead syscall is incurred at production throughput, not just during tests.

## Trigger / repro

Any deployment using `JWTPDP` (i.e., a config with `auth.mode: jwt`). The extraneous `time.Now()` call is unconditional and cannot be avoided by callers.

## Suggested fix

```go
var now time.Time
if p.clock != nil {
    now = p.clock.Now()
} else {
    now = time.Now()
}
```

This matches the pattern used by the engine, `alwaysAllowPDP`, and `TokenCache` that the comment references.

## Session

`rev-20260616T0010Z-8ec038`

### LINKED ISSUE #625: [rev-20260616T0010Z-8ec038] Audit tail tampering detected at startup is not written as a structured OCSF record
## Location

`cmd/eunox-mcp/audit.go`, lines 455–469

## Reference

https://github.com/eunolabs/eunox/blob/b316488642a6647bfb32ba228b601e28adeb517e/cmd/eunox-mcp/audit.go#L455-L469

## Severity

Medium / Security

## Description

When the proxy starts and finds that the audit log's tail record fails HMAC verification, it emits a plain `fmt.Fprintf(os.Stderr, ...)` warning and resumes the chain from the suspect record:

```go
if prev.HMAC != "" {
        fmt.Fprintf(os.Stderr, "[eunox-mcp] WARNING: audit log tail failed HMAC "+
            "verification; the last record may have been tampered with or truncated. "+
            "Resuming the chain from it anyway — run 'eunox-mcp audit-verify' and "+
            "reconcile against your external sink.\n")
        // Falls through — s.seq and s.prevHMAC are set from the unverified record.
    }
}
s.seq = prev.Seq
s.prevHMAC = prev.HMAC
```

The comment documents this as a deliberate "warn rather than abort" tradeoff so a corrupted log does not block enforcement. That tradeoff is acceptable, but the detection event goes exclusively to `stderr` — a channel that may not be monitored, may be lost in container log rotation, and is explicitly not part of the tamper-evident audit trail.

This creates a gap: an operator relying on the audit log for compliance (rather than active stderr scraping) would never see a structured record of the integrity failure. The tamper event is itself not auditable.

Two specific problems follow:
1. A tampered-tail event has no OCSF representation; `audit-verify` identifies the break on the tail record but surfaces no distinct "integrity failure at startup" event type.
2. `s.prevHMAC` is set from the unverified record, so all subsequent legitimate records chain from an attacker-controlled HMAC value. This does not break `audit-verify` (it will still detect the broken link at the tail), but it means an attacker who controls the tail HMAC also controls the link anchor for all subsequent records written during the session.

## Trigger / repro

1. Truncate or modify the HMAC of the last line in an active audit log file.
2. Restart `eunox-mcp`.
3. The warning appears on stderr; nothing is written to the audit log until the first real audit event.
4. If the process runs in a container with ephemeral logs, the warning is silently discarded.

## Suggested fix

After detecting the HMAC failure and emitting the stderr warning, write a structured audit record before the first real event — analogous to a "drop marker" — with a distinct event type (e.g., `audit_integrity_failure`) and the fields:

- `prev_hmac_verified: false`
- `tail_seq: prev.Seq`
- `tail_hmac: prev.HMAC` (the suspect value)
- `message: "tail HMAC verification failed at startup; chain resumed from unverified tail"`

This record would itself be HMAC-signed (chained from `prev.HMAC` as currently), making it appear in `audit-verify` output and ensuring the event is present in any external sink that consumes the log.

Optionally, a non-zero exit code or a signal (SIGUSR1) could be added so orchestrators detect the condition without log parsing.

## Session

`rev-20260616T0010Z-8ec038`

eunolabs:
## Bugs addressed (session rev-20260616T0010Z-8ec038)

- Closes #623 — `ForceRefreshForVerify` marked the global kid-less negative sentinel after *every* successful forced refresh, including ones that pulled in a rotated key. That suppressed the entire kid-less verification path for `negativeKIDTTL`, so a token signed by an emergency re-rotated key was rejected for the rest of the window (a partial outage continuous rotation could extend indefinitely). It now snapshots the key set **before** the fetch and only marks the sentinel when the set is genuinely **unchanged** (by RFC 7638 thumbprint — order-independent, and unlike the issue's length-only suggestion it catches a same-length hard rotation; note the suggested post-`ForceRefresh` comparison is a no-op since `ForceRefresh` has already stored the new set). The DoS guard is preserved for the no-rotation case.
- Closes #624 — `ValidateToken` always ran a `time.Now()` syscall and then discarded it when an injected clock was present (hot path under `JWTPDP`). Reordered to resolve the injected clock first and fall back to the wall clock only when none is set, matching the documented intent and the engine/`TokenCache` pattern.
- Closes #625 — a tampered/truncated audit-log tail detected at startup was reported **only** via an `os.Stderr` warning — lost to container log rotation, and not part of the tamper-evident trail. The proxy now also writes a signed, structured `AUDIT_INTEGRITY_FAILURE` record (a synthetic deny marker, reusing `flushDropMarker`'s shape — no new top-level record field — with the suspect tail `seq`/`_hmac` in `details`) as the first record of the new session, so the detection event itself appears in `audit-verify` and any external sink.

## Testing
- `go test -race ./...` (full suite green), `golangci-lint` 0 issues, `make check-fmt`/`check-license` clean.
- Added `TestJWKSCache_ForceRefreshForVerify_DoesNotSuppressAfterRotation` and `_SuppressesWhenUnchanged` (#623); `TestOpenAuditSink_WritesIntegrityMarkerOnTamperedTail` (#625), which sign-and-verify the new marker round-trip.
- Threat model (`docs/threat-model-mcp.md` §Resume integrity) updated for the new marker.

## Note
This session's issues (#623–#625) were filed after the earlier sessions in this review batch; they are handled here in their own PR following the same one-PR-per-session flow.

---


eunolabs:
✅ Ready to merge — all session rev-20260616T0010Z-8ec038 bugs addressed (#623, #624, #625), self-review complete with no outstanding comments, tests passing (`go test -race ./...`, `golangci-lint` 0 issues, `make check-fmt`/`check-license` clean).

---

PR #652, opened 2026-06-16
Fix session rev-20260616T0305Z-d91098 bugs

### LINKED ISSUE #639: [rev-20260616T0305Z-d91098] Kill timer goroutine races cmd.Wait() in fetchLiveToolsStdio, can SIGKILL recycled PID
**Location:** `cmd/eunox-mcp/live_upstream.go:132-141`
**Reference:** https://github.com/eunolabs/eunox/blob/9edd245f0178db578452636c845fcff2a394c146/cmd/eunox-mcp/live_upstream.go#L132-L141
**Severity:** low

**Description:**

`fetchLiveToolsStdio` sets up a deferred cleanup that closes stdin, starts a 2-second `time.AfterFunc` kill timer, calls `cmd.Wait()`, and then calls `killTimer.Stop()`. The problem is the ordering:

```go
defer func() {
    _ = stdin.Close()
    killTimer := time.AfterFunc(stdioIntrospectShutdownMs*time.Millisecond, func() {
        if cmd.Process != nil {
            _ = cmd.Process.Kill()
        }
    })
    _ = cmd.Wait()      // process reaped here
    killTimer.Stop()    // timer goroutine may already be running
}()
```

When the subprocess exits normally in under 2 seconds:

1. `cmd.Wait()` returns and reaps the process (the PID is now freed).
2. In the window between `cmd.Wait()` returning and `killTimer.Stop()` executing, the `time.AfterFunc` goroutine can fire.
3. `cmd.Process` is non-nil (Go does not nil it after `Wait()`), so `cmd.Process.Kill()` sends `SIGKILL` to the now-freed PID.
4. If the OS has recycled that PID to an unrelated process between steps 1 and 3, the signal is delivered to the wrong target.

`time.AfterFunc` schedules the callback in a goroutine; `Stop()` returning `false` means the goroutine has already started executing and cannot be prevented. There is no safe drain mechanism for `AfterFunc` timers (only channel-based timers support `<-timer.C` drain).

**Trigger / repro:**

Run `eunox-mcp init --command <fast-exiting-subprocess>` (or `validate --live --config` against a stdio upstream that exits in under 2 seconds). The subprocess exits before the 2-second kill window. PID recycling is OS-dependent but is theoretically possible in a tight loop of rapid subprocess creation.

**Suggested fix:**

Swap the order: stop the timer *before* calling `cmd.Wait()`, then kill manually if Stop returns false:

```go
defer func() {
    _ = stdin.Close()
    killTimer := time.AfterFunc(stdioIntrospectShutdownMs*time.Millisecond, func() {
        if cmd.Process != nil {
            _ = cmd.Process.Kill()
        }
    })
    if !killTimer.Stop() {
        // Timer already fired; Wait will return once the kill lands.
    }
    _ = cmd.Wait()
}()
```

Alternatively, track whether `Wait` completed first and guard the kill with an atomic flag:

```go
var waited atomic.Bool
killTimer := time.AfterFunc(stdioIntrospectShutdownMs*time.Millisecond, func() {
    if !waited.Load() && cmd.Process != nil {
        _ = cmd.Process.Kill()
    }
})
_ = cmd.Wait()
waited.Store(true)
killTimer.Stop()
```

**Session:** rev-20260616T0305Z-d91098

### LINKED ISSUE #640: [rev-20260616T0305Z-d91098] writeDoctorAudit does not early-return after expandHome failure, calls os.Stat("") with confusing double error
**Location:** `cmd/eunox-mcp/doctor.go:383-420`
**Reference:** https://github.com/eunolabs/eunox/blob/9edd245f0178db578452636c845fcff2a394c146/cmd/eunox-mcp/doctor.go#L383-L420
**Severity:** low

**Description:**

`writeDoctorAudit` expands the log path's `~` prefix via `expandHome`, prints an error when that fails, and then falls through to `os.Stat(resolvedLog)` regardless:

```go
resolvedLog, logErr := expandHome(resolvedLog)
if logErr != nil {
    wf(w, "  log path:  (cannot resolve: %v)\n", logErr)
} else {
    wf(w, "  log path:  %s\n", resolvedLog)
}
// ... key path logic ...
info, err := os.Stat(resolvedLog)  // resolvedLog == "" when logErr != nil
switch {
case err != nil:
    wf(w, "  log file:  %v\n", err)
    return
```

When `expandHome` fails (e.g., no home directory in a container with `DynamicUser=yes`), it returns `("", error)`.  After the error message is written, `resolvedLog` is `""` and the subsequent `os.Stat("")` call fails with `stat : no such file or directory`, printing a second, confusing error to the bundle with an empty filename.

Compare to the key-path block in the same function, which correctly avoids the `os.Stat` call when `keyErr != nil` — the log-path block is inconsistent with that pattern.

**Trigger / repro:**

Run `eunox-mcp doctor --audit-log '~/.eunox/audit.jsonl'` in an environment where `os.UserHomeDir()` returns an error (no `HOME` env var, no `/etc/passwd` entry for the process UID, `DynamicUser=yes` systemd service). The bundle will contain:

```
  log path:  (cannot resolve: cannot expand "~/.eunox/audit.jsonl": home directory is unavailable: ...)
  log file:  stat : no such file or directory
```

The second line is spurious and misleading.

**Suggested fix:**

Mirror the key-path pattern: return (or skip the stat block) when `logErr != nil`:

```go
resolvedLog, logErr := expandHome(resolvedLog)
if logErr != nil {
    wf(w, "  log path:  (cannot resolve: %v)\n", logErr)
    return
}
wf(w, "  log path:  %s\n", resolvedLog)
```

**Session:** rev-20260616T0305Z-d91098

### LINKED ISSUE #641: [rev-20260616T0305Z-d91098] indentWriter.Write sets atLineStart=true before confirming the newline was actually written
**Location:** `cmd/eunox-mcp/doctor.go:553-562`
**Reference:** https://github.com/eunolabs/eunox/blob/9edd245f0178db578452636c845fcff2a394c146/cmd/eunox-mcp/doctor.go#L553-L562
**Severity:** low

**Description:**

In `indentWriter.Write`, when a newline is found in the input slice, the code writes the chunk up to and including the `\n`, then sets `atLineStart = true` before checking the error:

```go
n, err := iw.w.Write(p[:nl+1])
written += n
iw.atLineStart = true   // set unconditionally, before error check
if err != nil {
    return written, err
}
p = p[nl+1:]
```

If the underlying writer returns `(n < nl+1, err != nil)` — meaning it wrote some bytes but not the `\n` — `atLineStart` is set to `true` even though the newline was never written to the output. Any subsequent `Write` call (after the error is observed and the caller somehow retries) would incorrectly prepend an indent prefix at a position that is mid-line.

The same block also advances `p = p[nl+1:]` unconditionally (no early return before that assignment in the error path), which is fine because of the `return` on the `err != nil` branch. But `atLineStart` persists on the struct.

By contrast, the io.Writer contract requires that a non-nil error accompanies any `n < len(p)` result, so in practice the `iw.w` (an `io.File` or similar well-behaved writer) would never do this. The bug only surfaces with a buggy or partially-failing underlying writer (e.g. a test double or a network-backed writer under stress).

**Trigger / repro:**

Use a custom `io.Writer` that short-writes on lines containing a newline but returns `nil` error. Call `indentWriter.Write` with `p = "hello\nworld"`. If the write of `"hello\n"` returns `(n=5, err=nil)` without the `\n` actually reaching the underlying stream, the next `Write` will prepend a prefix to `"world"` at what is actually still the middle of the first logical line.

**Suggested fix:**

Guard `atLineStart` behind the actual write result:

```go
n, err := iw.w.Write(p[:nl+1])
written += n
if n == nl+1 {
    iw.atLineStart = true
}
if err != nil {
    return written, err
}
p = p[nl+1:]
```

This keeps the flag in sync with what was actually written rather than what was intended to be written.

**Session:** rev-20260616T0305Z-d91098

### LINKED ISSUE #642: [rev-20260616T0305Z-d91098] argumentSchema enum check short-circuits type constraint, allowing type-mismatched values to pass enforcement
**Location:** `pkg/enforcement/schema.go:110-119`
**Reference:** https://github.com/eunolabs/eunox/blob/9edd245f0178db578452636c845fcff2a394c146/pkg/enforcement/schema.go#L110-L119
**Severity:** medium

**Description:**

`schemaValidateValue` checks `enum` first and returns `nil` immediately when the value matches, skipping the `type` constraint check that follows:

```go
if len(schema.Enum) > 0 {
    for _, allowed := range schema.Enum {
        if reflect.DeepEqual(allowed, val) || numericEqual(allowed, val) {
            return nil   // ← returns without checking schema.Type
        }
    }
    return fmt.Errorf("%s: value not in enum", jsonPath)
}

// This block is never reached when the value matches an enum element:
if !schema.Type.IsZero() {
    if err := schemaCheckType(jsonPath, val, schema.Type); err != nil {
        return err
    }
}
```

A manifest that declares both `type` and `enum` expects both constraints to hold simultaneously. With the current code, a value that is in the enum silently bypasses the `type` check, regardless of how incompatible the types are.

**Concrete case:** A manifest capability uses `argumentSchema: {type: "string", enum: [1, 2]}` (contradictory but syntactically valid YAML). A request argument of `float64(1)` (JSON number `1`) passes `numericEqual(int(1), float64(1)) == true` and so clears enforcement without the `type: "string"` check ever running. The argument arrives at the upstream as a JSON number, contradicting the declared string type.

In JSON Schema (draft-07 and later), `enum` and `type` are independent keywords; a value must satisfy all applicable keywords. An `enum` match does not exempt the value from `type` validation.

**Trigger / repro:**

1. Write a manifest with a capability whose `argumentSchema` has both `type: "string"` and `enum: [1, 2]`.
2. Call the tool with `arguments: {"arg": 1}` (JSON integer).
3. Enforcement passes (`numericEqual` matches), no `CONDITION_FAILED` denial, and the upstream receives a JSON number where a string was required.

**Suggested fix:**

Do not return early from the enum match; continue to the `type` check:

```go
if len(schema.Enum) > 0 {
    matched := false
    for _, allowed := range schema.Enum {
        if reflect.DeepEqual(allowed, val) || numericEqual(allowed, val) {
            matched = true
            break
        }
    }
    if !matched {
        return fmt.Errorf("%s: value not in enum", jsonPath)
    }
    // fall through to type and keyword checks
}
```

Alternatively, restructure so all top-level keyword checks (`enum`, `type`, and the type-specific validators) run independently and accumulate errors, matching the JSON Schema multi-keyword validation model.

**Session:** rev-20260616T0305Z-d91098

eunolabs:
## Bugs addressed (session rev-20260616T0305Z-d91098)

- Closes #642 — `schemaValidateValue` returned early on an `enum` match, skipping the `type` (and keyword) checks. A value matching a numeric enum could bypass a declared `type: string`, reaching the upstream as the wrong JSON type. The enum match no longer short-circuits; type/keyword checks always run (independent keywords per JSON Schema).
- Closes #641 — `indentWriter.Write` set `atLineStart = true` before confirming the newline was written. On a short write (`n < nl+1`) the flag is now left unset, so a later `Write` cannot prepend an indent prefix mid-line.
- Closes #640 — `writeDoctorAudit` fell through to `os.Stat(resolvedLog)` after `expandHome` failed (`resolvedLog == ""`), printing a spurious second `log file:  stat : no such file...` line. It now skips the log-file stat when the path is unresolvable, mirroring the key-path block, while still emitting the key-path diagnostics.
- Closes #639 — `fetchLiveToolsStdio` started a `time.AfterFunc` kill timer that called `cmd.Process.Kill()`, racing `cmd.Wait()`: after the process was reaped the timer could SIGKILL a recycled PID. The backstop now cancels a child context bound to the subprocess; `os/exec` serializes the context-driven kill with `Wait` (Wait does not return until the watch goroutine that owns the kill has finished), so no signal is delivered to a reaped PID.

## Testing

- Added `TestValidateArgumentSchema_EnumDoesNotBypassType` + `_EnumMatchCorrectTypePasses` (pkg/enforcement).
- Added `TestIndentWriter_ShortWriteDoesNotMarkLineStart` (cmd/eunox-mcp).
- Added `TestWriteDoctorAudit_LogPathUnresolvable_NoSpuriousStat` (cmd/eunox-mcp).
- #639 is a structural concurrency fix covered by `TestFetchLiveToolsStdio_Subprocess` and `TestFetchLiveToolsStdio_BadCommand`, both green under `go test -race`.
- `gofmt`, `go vet`, `golangci-lint`, and the package tests all pass.

---


eunolabs:
✅ Ready to merge — all session rev-20260616T0305Z-d91098 bugs addressed (#642, #641, #640, #639), self-review complete with no outstanding comments, CI green (Test, Lint, builds, demos, OPA-comparison, enforcement-compliance).

---

PR #653, opened 2026-06-16
Fix session rev-20260616T0258Z-67a856 bugs

### LINKED ISSUE #638: [rev-20260616T0258Z-67a856] handleAllowedOperations silently accepts "*" in operations list, bypassing the condition
**Location:** `pkg/enforcement/handlers.go:590` (`handleAllowedOperations`)

**Reference:** https://github.com/eunolabs/eunox/blob/9edd245f0178db578452636c845fcff2a394c146/pkg/enforcement/handlers.go#L590

**Severity:** Medium — `operations: ["*"]` passes manifest validation and silently allows any operation verb at runtime, defeating the condition entirely; the `"*"` token is not documented as a valid operation and has no corresponding test

**Description:**

`handleAllowedOperations` contains a hard-coded `"*"` wildcard check that is absent from documentation, absent from tests, and not validated out by `validateAllowedOperations` at load time:

```go
for _, allowed := range ao.Operations {
    if strings.EqualFold(allowed, operation) || allowed == "*" {  // BUG
        return nil
    }
}
```

Any entry `"*"` in the `operations` list causes the condition to return `nil` (allow) for **any** operation verb, making the `allowedOperations` condition a no-op for that tool. Concretely:

```yaml
conditions:
  - type: allowedOperations
    argument: sql
    operations: ["*"]   # silently allows DROP, DELETE, TRUNCATE, …
```

At load time `validateAllowedOperations` checks only that `Argument` is non-empty and `Operations` is non-empty. `operations: ["*"]` passes both checks cleanly and is loaded without error or warning.

The `"*"` wildcard is legitimately documented for constraint **actions** (e.g., `actions: ["*"]` means "apply to all action types"), where `capability.ValidateActionForTargetType` explicitly includes it in the valid set. That same wildcard token appears to have been copied into `handleAllowedOperations` without an equivalent design intent or safeguard:

1. No documentation mentions `"*"` as a valid entry in `allowedOperations.operations`.
2. No test exercises the `allowed == "*"` branch.
3. `validateAllowedOperations` does not warn or reject it.
4. The `AllowedOperationsCondition` doc comment (`condition.go:62-65`) does not mention a wildcard.

**Trigger / repro:**

```yaml
capabilities:
  - target: tool:query_db
    actions: [call]
    conditions:
      - type: allowedOperations
        argument: sql
        operations: ["*"]   # intention: allow all ops (debugging aid)
```

With this manifest, `tools/call` for `query_db` with `{"sql": "DROP TABLE users"}` is **allowed**. There is no error at `eunox-mcp validate` time.

**Suggested fix:**

Two complementary changes:

1. **Remove or reject `"*"` from `handleAllowedOperations`**. The operations list is an explicit allowlist of verbs; a blanket wildcard defeats the purpose. Either remove the `allowed == "*"` branch entirely (fail closed) or have `validateAllowedOperations` reject it explicitly:

```go
for _, op := range v.Operations {
    if op == "*" {
        return fmt.Errorf("capability at index %d, condition %d: allowedOperations 'operations' list contains \"*\"; the wildcard is not a valid SQL/operation verb and would allow any operation — list the permitted verbs explicitly (e.g. [\"SELECT\"])", i, j)
    }
}
```

2. **Add a test** for the current (buggy) behavior so the fix is pinned by a regression test:

```go
func TestEngine_AllowedOperations_WildcardRejected(t *testing.T) {
    // operations: ["*"] must not silently allow any verb
    ...
}
```

**Session:** rev-20260616T0258Z-67a856

eunolabs:
## Bugs addressed (session rev-20260616T0258Z-67a856)

- Closes #638 — `handleAllowedOperations` matched a literal `"*"` entry in the `operations` list against any verb, so `operations: ["*"]` silently turned the condition into a no-op permitting every operation (`DROP`, `DELETE`, …). The token was undocumented, untested, and accepted at load. Two changes: the runtime handler no longer treats `"*"` as a wildcard (fail closed — only an explicit verb match allows), and `validateAllowedOperations` now rejects a `"*"` entry at manifest load with a message directing operators to list verbs explicitly.

## Testing

- Added `TestEngine_AllowedOperations_WildcardDoesNotAllowAnyVerb` (pkg/enforcement): `operations: ["*"]` no longer allows `DROP`.
- Added `allowedOperations wildcard rejected` case to `TestLoadManifest_ConditionArgumentValidation` (cmd/eunox-mcp): load-time rejection.
- Updated threat model (`docs/threat-model-mcp.md`, entry 1.26).
- `gofmt`, `go vet`, `golangci-lint`, and package tests all pass.

---


eunolabs:
✅ Ready to merge — session rev-20260616T0258Z-67a856 bug addressed (#638), self-review complete with no outstanding comments, CI green (Test, Lint, builds, demos, OPA-comparison, enforcement-compliance).

---

PR #654, opened 2026-06-16
Fix session rev-20260616T0206Z-1b5405 bugs

### LINKED ISSUE #636: [rev-20260616T0206Z-1b5405] circuitbreaker.Stats: omitempty tags on time.Time fields are silently ignored by encoding/json
**Location:** `pkg/circuitbreaker/breaker.go:517-524`

**Reference:** https://github.com/eunolabs/eunox/blob/9edd245f0178db578452636c845fcff2a394c146/pkg/circuitbreaker/breaker.go#L517-L524

**Severity:** Medium — consumers who JSON-serialize `Stats` (monitoring dashboards, health endpoints, log shippers) receive `"0001-01-01T00:00:00Z"` for fields they expect to be absent, which can trigger spurious "failure at year-0001" alerts and contradicts the intent expressed by the `omitempty` tag

**Description:**

The exported `Stats` struct carries `omitempty` tags on its two `time.Time` fields:

```go
type Stats struct {
    State            State     `json:"state"`
    ConsecutiveFails int       `json:"consecutiveFails"`
    TotalFailures    int64     `json:"totalFailures"`
    TotalSuccesses   int64     `json:"totalSuccesses"`
    LastFailureTime  time.Time `json:"lastFailureTime,omitempty"`   // BUG
    LastTransitionAt time.Time `json:"lastTransitionAt,omitempty"`  // BUG
}
```

Go's `encoding/json` v1 defines "empty" as: `false`, `0`, nil pointer, nil interface, nil/empty slice or map, or empty string. Struct types are **never** considered empty, regardless of whether they implement `json.Marshaler`. `time.Time{}` (the zero value) therefore serializes as `"0001-01-01T00:00:00Z"` even with `omitempty`, not as an absent field.

Concretely:

```go
b := circuitbreaker.New(circuitbreaker.DefaultConfig())
// No failures yet — LastFailureTime and LastTransitionAt are time.Time{}
data, _ := json.Marshal(b.Stats())
// Actual:   {"state":"closed","consecutiveFails":0,...,"lastFailureTime":"0001-01-01T00:00:00Z","lastTransitionAt":"0001-01-01T00:00:00Z"}
// Expected: {"state":"closed","consecutiveFails":0,...}  (fields absent)
```

After `Reset()`, `LastFailureTime` is explicitly zeroed:
```go
b.lastFailureTime = time.Time{}
```
If Stats is serialized after a Reset, the "year 1" timestamp reappears, potentially misleading operators into thinking the breaker last failed in year 1.

**Trigger / repro:**

Any code that JSON-serializes `b.Stats()` before any failures have occurred, or after `Reset()`, will see the two fields present with `"0001-01-01T00:00:00Z"` values despite the `omitempty` tags.

**Suggested fix:**

Two options:

1. **Use pointer fields** so that a nil pointer IS omitted:
```go
LastFailureTime  *time.Time `json:"lastFailureTime,omitempty"`
LastTransitionAt *time.Time `json:"lastTransitionAt,omitempty"`
```
`Stats()` sets these to `nil` when the underlying `time.Time` is zero, and to `&b.lastFailureTime` otherwise.

2. **Implement `MarshalJSON` on `Stats`** that explicitly skips zero-time fields.

3. **Switch to `json:"...,omitzero"`** when migrating to `encoding/json/v2` (which adds `omitzero` for types implementing `IsZero() bool`).

Option 1 is the cleanest fix for the current stdlib without changing the field semantics.

**Session:** rev-20260616T0206Z-1b5405

### LINKED ISSUE #637: [rev-20260616T0206Z-1b5405] Corrupt audit-log tail (JSON parse failure) at startup has no in-band signed marker, unlike HMAC failure
**Location:** `cmd/eunox-mcp/audit.go:458-503` (`openAuditSink`, tail chain-resume block)

**Reference:** https://github.com/eunolabs/eunox/blob/9edd245f0178db578452636c845fcff2a394c146/cmd/eunox-mcp/audit.go#L458

**Severity:** Medium — compliance-focused operators who rely on the audit log (not stderr scraping) have no in-band evidence that the prior session ended with a structurally corrupt record; the HMAC-failure case at the same code site does write a signed marker, creating an asymmetric detection capability

**Description:**

When `openAuditSink` resumes the tamper-evident chain at startup it reads the last line of the existing audit log and branches on two distinct failure modes:

**Case A — HMAC verification failure** (line ~479):
```go
    fmt.Fprintf(os.Stderr, "[eunox-mcp] WARNING: audit log tail failed HMAC verification; ...")
    tailIntegrityFailed = true   // <── flag set
    tailSeq  = prev.Seq
    tailHMAC = prev.HMAC
}
```
Later (line ~508):
```go
if tailIntegrityFailed {
    s.writeTailIntegrityMarker(tailSeq, tailHMAC)  // signed in-band record written
}
```

**Case B — JSON parse failure** (line ~459):
```go
if err := json.Unmarshal([]byte(last), &prev); err != nil {
    fmt.Fprintf(os.Stderr, "[eunox-mcp] WARNING: audit log tail failed to parse (%v); ...")
    // tailIntegrityFailed is NOT set — no in-band marker is written
}
```

Case B is the *more severe* integrity event: the last record is structurally corrupt (power loss, partial write, disk corruption), yet it receives only a transient stderr warning. Case A (HMAC mismatch — content tampered but structurally valid) correctly writes a signed `AUDIT_INTEGRITY_FAILURE` marker that is part of the tamper-evident chain, surfaced by `eunox-mcp audit-verify`, and forwarded to any external sink.

An operator running in a container where stderr is ephemeral (or relying on the audit log alone for compliance) has **no evidence** in the audit trail that a JSON parse failure occurred. If an attacker deliberately truncates the last record to erase a chain break, that truncation event is also silent in the log.

The code comment at line ~469 acknowledges the gap: *"this case is more severe (structurally corrupt, not merely tampered) yet was silent. Warn so an operator can reconcile promptly"* — but the fix was only the stderr warn, not an in-band marker, leaving the audit log itself incomplete.

**Trigger / repro:**

1. Write a few records to the audit log.
2. Truncate the last line mid-record (e.g., `truncate -s -50 ~/.eunox/audit.jsonl`) to simulate a partial write at shutdown.
3. Restart `eunox-mcp`.
4. Check stderr — warning appears. ✓
5. Run `eunox-mcp audit-verify` on the log — **no `AUDIT_INTEGRITY_FAILURE` marker appears in the new session's records**; the chain simply restarts from genesis with no annotated reason.
6. For comparison, corrupt the last record's content (keeping it valid JSON but wrong HMAC) — `AUDIT_INTEGRITY_FAILURE` marker appears correctly.

**Suggested fix:**

Introduce a parallel `tailParseFailure` flag alongside `tailIntegrityFailed` and write an in-band marker for the JSON parse failure case, mirroring what #625 added for HMAC failure:

```go
var (
    tailIntegrityFailed bool
    tailParseFailure    bool
    tailSeq             uint64
    tailHMAC            string
)
if last != "" {
    var prev auditRecord
    if err := json.Unmarshal([]byte(last), &prev); err != nil {
        fmt.Fprintf(os.Stderr, "[eunox-mcp] WARNING: ...")
        tailParseFailure = true          // NEW
    } else {
        // ... existing HMAC check ...
    }
}
if tailParseFailure {
    s.writeTailParseFailureMarker()      // NEW — similar to writeTailIntegrityMarker
}
if tailIntegrityFailed {
    s.writeTailIntegrityMarker(tailSeq, tailHMAC)
}
```

The parse-failure marker can reuse the `AUDIT_INTEGRITY_FAILURE` denial code with a `"kind": "tail_parse_failure"` detail, or introduce a distinct code (`AUDIT_TAIL_CORRUPT`) for SIEM disambiguation. It should chain from genesis (no valid predecessor HMAC) and appear as the first signed record of the new session, making the gap visible to `audit-verify` and any external compliance sink.

**Session:** rev-20260616T0206Z-1b5405

eunolabs:
## Bugs addressed (session rev-20260616T0206Z-1b5405)

- Closes #637 — A corrupt audit-log tail that fails to **parse** as JSON (truncated/partial write, power loss, disk corruption, or a deliberate truncation that erases a chain break) only warned to stderr, while the HMAC-mismatch case wrote a signed in-band `AUDIT_INTEGRITY_FAILURE` marker (#625) — an asymmetric detection gap for operators relying on the audit log alone. `openAuditSink` now writes a signed marker for the parse-failure case too, chained from genesis (the corrupt tail yields no resumable predecessor) so it is the first signed record of the new session and surfaces in `audit-verify` and any external sink. It reuses the existing record shape (no new top-level field); a `"kind":"tail_parse_failure"` Details value disambiguates it from the HMAC-mismatch marker.
- Closes #636 — `circuitbreaker.Stats` carried `omitempty` on two `time.Time` fields, but `encoding/json` never treats a struct as empty, so a fresh or just-`Reset()` breaker serialized `lastFailureTime`/`lastTransitionAt` as `"0001-01-01T00:00:00Z"`, tripping "failure in year 1" alerts. Added a `Stats.MarshalJSON` that omits each time field when zero (field types unchanged, so existing consumers and the `projectedState` logic are unaffected).

## Testing

- Added `TestOpenAuditSink_WritesMarkerOnUnparseableTail` (sign-and-verify round-trip on the new marker) and updated `TestOpenAuditSinkWarnsOnUnparseableTail` for the genesis-chained marker (cmd/eunox-mcp).
- Added `TestStats_JSON_OmitsZeroTimes` (pkg/circuitbreaker): zero times omitted, real values present after a trip.
- Updated threat model (`docs/threat-model-mcp.md`, entry 1.27).
- `gofmt`, `go vet`, `golangci-lint`, and `go test ./...` all pass.

---


eunolabs:
✅ Ready to merge — all session rev-20260616T0206Z-1b5405 bugs addressed (#637, #636), self-review complete with no outstanding comments, CI green (Test, Lint, builds, demos, OPA-comparison, enforcement-compliance).

---

PR #655, opened 2026-06-16
Fix session rev-20260616T0210Z-d7fa63 bugs

### LINKED ISSUE #632: [rev-20260616T0210Z-d7fa63] boundAuditObligations slow path underestimates JSON size for strings with escape sequences
**Location:** `cmd/eunox-mcp/audit.go` lines ~897–932 (`boundAuditObligations`, slow path)

**Reference:** https://github.com/eunolabs/eunox/blob/9edd245f0178db578452636c845fcff2a394c146/cmd/eunox-mcp/audit.go#L897

**Severity:** Medium — `auditObligationsTotalCap` (64 KiB) can be exceeded in the audit record when obligation strings contain characters that expand under JSON encoding

**Description:**

`boundAuditObligations` uses two paths to decide how many obligation entries fit within `auditObligationsTotalCap`:

- **Fast path** (few entries): calls `json.Marshal` on the full obligations map and checks the actual serialized byte length. Correct.
- **Slow path** (many entries or fast-path failure): iterates over obligation strings and maintains a running byte estimate using `len(o) + perEntryOverhead`, where `perEntryOverhead = 3` (for `"`:` ` and `,`).

The slow path uses `len(o)` — the raw UTF-8 byte length of the Go string — as a proxy for the JSON-encoded byte length. This is wrong for strings containing characters that `json.Marshal` expands:

| character | raw bytes | JSON bytes |
|-----------|-----------|------------|
| `"` | 1 | 2 (`\"`) |
| `\` | 1 | 2 (`\\`) |
| `\n` | 1 | 2 (`\n`) |
| U+0000–U+001F control chars | 1 | 6 (`\uXXXX`) |
| U+0080–U+009F C1 controls (if in string) | 2 | 6 (`\uXXXX`) |

If obligation strings are heavy with such characters (e.g., SQL fragments, file paths on Windows, or user-controlled input surfaced through the policy), the slow path will underestimate the serialized size, include more entries than fit in 64 KiB, and the final `json.Marshal` of the audit record will produce a `details.obligations` field larger than `auditObligationsTotalCap`.

The bug is silent: no error is returned from `boundAuditObligations`; the oversized result is written to the audit log as-is. The audit record is not truncated and not flagged.

**Trigger / repro:**

1. Configure a constraint with obligations whose string values contain many `"` or `\` characters (e.g., a `redactFields` obligation containing paths like `"a\\b\\c"`, or obligation values derived from SQL fragments).
2. Generate enough obligation entries that the fast path fails (the fast-marshal call returns an error, or there are > N entries), forcing the slow path.
3. Observe that the serialized `details.obligations` in the audit record exceeds 64 KiB (verify with `jq '.details.obligations | @json | length'` on the audit JSONL).

**Suggested fix:**

Replace `len(o) + perEntryOverhead` with the actual JSON-encoded length. Since `perEntryOverhead` already handles map overhead, the simplest correction is to escape-estimate the value with `len(strconv.AppendQuoteToASCII(nil, o))` (or a direct `json.Marshal` of just the string). If a full `json.Marshal` per entry is too expensive in the slow path, a conservative over-estimate (e.g., `len(o)*6 + perEntryOverhead`) bounds the size correctly at the cost of accepting fewer entries. Either way, the slow path must never let the total exceed `auditObligationsTotalCap` after actual serialization.

**Session:** rev-20260616T0210Z-d7fa63

### LINKED ISSUE #633: [rev-20260616T0210Z-d7fa63] ForceRefreshForVerify TOCTOU causes negative-KID sentinel to be wrongly suppressed after concurrent key rotation
**Location:** `pkg/capability/jwkscache.go` lines 232–268 (`ForceRefreshForVerify`)

**Reference:** https://github.com/eunolabs/eunox/blob/9edd245f0178db578452636c845fcff2a394c146/pkg/capability/jwkscache.go#L232

**Severity:** Low — bounded by `negativeKIDTTL` (30 s); allows one extra JWT verification attempt against stale keys before the KID sentinel re-fires

**Description:**

`ForceRefreshForVerify` is called when a kid-less JWT has failed against all known keys and a forced JWKS refresh should detect key rotation:

```go
func (c *JWKSCache) ForceRefreshForVerify(ctx context.Context) error {
    c.mu.RLock()
    before := c.jwks        // snapshot taken outside ForceRefresh
    c.mu.RUnlock()

    if err := c.ForceRefresh(ctx); err != nil {
        return err
    }

    c.mu.RLock()
    keys := c.jwks
    c.mu.RUnlock()

    if jwksKeysUnchanged(before, keys) {
        c.markKIDAbsent(sentinel)
    }
    return nil
}
```

The TOCTOU window: goroutine G1 reads `before = c.jwks` (snapshot of the old key set). Concurrently, goroutine G2 has already completed a `ForceRefresh` that rotated the keys. When G1 enters `ForceRefresh`, `singleflight.Do("refresh", ...)` joins an in-flight call or starts a new one and immediately returns because `lastForceRefresh` debounce says the cache is fresh. G1 reads `keys = c.jwks` — which is the already-rotated key set. Now `jwksKeysUnchanged(before, keys)` may return `false` (keys differ), so G1 does NOT mark the sentinel absent.

But `before` was the *old* key set, and `keys` is the *new* key set (rotated by G2). The keys changed — yet G1 never tried to verify the JWT against the new keys. The sentinel is not marked, which is correct in intent, but the *reason* `jwksKeysUnchanged` returns false here is the rotation G2 already handled — not a rotation G1 detected. G1 will re-enter `VerifyToken`'s retry loop, try the new keys, and likely succeed. This scenario is benign.

The problematic case is the reverse: G2 has just done a ForceRefresh that *did not change the keys* (the IdP returned the same key set). G1 reads `before`, calls `ForceRefresh` (which now debounces because G2 just finished), reads `keys == before`. `jwksKeysUnchanged(before, keys)` = true, so G1 marks the sentinel absent — as intended.

However, if G2's refresh happened *between* G1's `before` read and G1's `ForceRefresh` call, and that refresh *did change the keys*, then G1's `before` is the pre-G2 key set and `keys` is the post-G2 key set. `jwksKeysUnchanged` returns false, the sentinel is not marked — correct. No bug in this path.

The actual bug manifests when G2's refresh completes *after* G1 enters ForceRefresh but *before* G1 reads `keys`: `before` = old set, G1's ForceRefresh joins G2 in-flight and gets the new set, `keys` = new set. `jwksKeysUnchanged` = false (keys changed). Sentinel not marked. G1 retries with new keys. If those keys also don't verify the JWT (legitimately invalid token), the sentinel is never set, and re-verification is attempted on every subsequent kid-less JWT for up to `negativeKIDTTL` seconds.

**Trigger / repro:**

1. Two goroutines simultaneously call `VerifyToken` with a kid-less JWT that is invalid (wrong signature).
2. Both hit `ForceRefreshForVerify` concurrently; a real key rotation happens between one goroutine's `before` read and the completion of the shared `ForceRefresh`.
3. `jwksKeysUnchanged` returns false for both (keys changed).
4. Neither goroutine marks the sentinel absent.
5. Subsequent invalid kid-less JWTs trigger a full JWKS fetch on every call for up to 30 s.

**Suggested fix:**

Move the `before` snapshot inside the `ForceRefresh` lock scope (or take it immediately before calling `singleflight.Do` inside `ForceRefresh`). The cleanest fix: return the pre-refresh key set from `ForceRefresh` itself so callers can compare atomically, or move the unchanged-check inside the singleflight closure where it executes under the write lock:

```go
// inside ForceRefresh singleflight closure, under write lock after update:
if jwksKeysUnchanged(oldKeys, newKeys) {
    c.markKIDAbsent(sentinel)
}
```

This eliminates the window between the `before` read and the refresh completion.

**Session:** rev-20260616T0210Z-d7fa63

### LINKED ISSUE #634: [rev-20260616T0210Z-d7fa63] uniqueRotatedPath loop counter has no upper bound, theoretical integer overflow
**Location:** `cmd/eunox-mcp/audit.go` lines ~1766–1781 (`uniqueRotatedPath`)

**Reference:** https://github.com/eunolabs/eunox/blob/9edd245f0178db578452636c845fcff2a394c146/cmd/eunox-mcp/audit.go#L1766

**Severity:** Low — only reachable after `math.MaxInt` iterations against a filesystem that accepts every candidate filename; practical risk negligible, but violates the rule that loops must terminate

**Description:**

`uniqueRotatedPath` generates a unique destination path for log rotation by incrementing a suffix counter until it finds a name that does not exist:

```go
for i := 1; ; i++ {
    candidate := fmt.Sprintf("%s.%d", base, i)
    if _, err := os.Stat(candidate); os.IsNotExist(err) {
        return candidate, nil
    }
    if err != nil {
        return "", err
    }
    // i continues unbounded
}
```

`i` is typed as `int`. On 64-bit systems, `int` is 64 bits, so the counter wraps to `math.MinInt` (negative) after `math.MaxInt64` iterations. `fmt.Sprintf("%s.%d", base, i)` will then produce a filename with a negative suffix (e.g., `audit.log.-9223372036854775808`). Most filesystems accept such names, so `os.Stat` returns `IsNotExist`, the function returns a negative-suffixed path, and rotation "succeeds" — producing a file that is invisible to any tooling expecting `audit.log.N` where N is a positive integer. On a 32-bit system, the overflow happens far earlier (after ~2 billion iterations).

The practical attack vector requires an adversary or a runaway process to pre-create `math.MaxInt64` files with names `audit.log.1` through `audit.log.9223372036854775807`, which is not feasible. However, the loop is also unbounded in the presence of a filesystem that consistently returns non-`IsNotExist` errors for all candidates (e.g., a fuse filesystem returning `EACCES` for every stat call): the error arm returns immediately on the first such error, so in practice the loop does terminate on filesystem errors. The theoretical concern is an infinite loop on a pathological filesystem that alternates errors and non-exists.

**Trigger / repro:**

Theoretical only for normal filesystems. On a mock/fuse filesystem that returns `EACCES` for even suffixes and `ENOENT` for odd suffixes, the loop terminates immediately on the first error. The integer overflow requires `math.MaxInt` pre-existing files.

**Suggested fix:**

Add a maximum iteration limit (e.g., `maxRotateSuffix = 10_000`) and return an error if exceeded:

```go
const maxRotateSuffix = 10_000
for i := 1; i <= maxRotateSuffix; i++ {
    candidate := fmt.Sprintf("%s.%d", base, i)
    if _, err := os.Stat(candidate); os.IsNotExist(err) {
        return candidate, nil
    }
    if err != nil {
        return "", err
    }
}
return "", fmt.Errorf("uniqueRotatedPath: no free slot found after %d attempts for %q", maxRotateSuffix, base)
```

This keeps the loop finite, produces a clear error rather than a silent integer-wraparound filename, and gives operators a signal that the rotation directory is unexpectedly cluttered.

**Session:** rev-20260616T0210Z-d7fa63

### LINKED ISSUE #635: [rev-20260616T0210Z-d7fa63] evaluateJWTConditions scan-all SQL mode logs non-deterministic matchedOp due to Go map iteration order
**Location:** `cmd/eunox-mcp/pdp_jwt.go` lines ~1181–1210 (`evaluateJWTConditions`, scan-all SQL branch)

**Reference:** https://github.com/eunolabs/eunox/blob/9edd245f0178db578452636c845fcff2a394c146/cmd/eunox-mcp/pdp_jwt.go#L1181

**Severity:** Low — allow/deny outcome is deterministic; only the `matched_op` field logged in the audit record is non-deterministic, making post-incident analysis unreliable

**Description:**

In scan-all SQL mode, `evaluateJWTConditions` iterates over `args` (a `map[string]interface{}`) to find arguments that contain SQL verbs:

```go
var matchedOp string
for argName, argVal := range args {           // Go map: non-deterministic order
    s, ok := argVal.(string)
    if !ok {
        continue
    }
    verb, found := isSQLVerb(s)
    if !found {
        continue
    }
    matchedOp = verb                           // last match wins
    if !permittedVerbs[verb] {
        return deny(...)
    }
}
```

Go's map iteration order is randomized per run. If multiple arguments in a single `tools/call` request contain SQL verbs (e.g., both a `query` argument and a `fallback_query` argument), the loop visits them in a different order on each invocation. Whichever argument is visited last sets `matchedOp`. The deny/allow outcome is still correct (any disallowed verb denies; all allowed verbs allow), but `matchedOp` — the value written into the audit record as the matched SQL operation — is non-deterministic across runs for the same request.

This violates the audit invariant that structured audit fields must be deterministic for a given input: the same request should produce the same `matched_op` value. A post-incident analyst running `audit-verify` or replaying the log cannot determine whether `matched_op=SELECT` and `matched_op=UPDATE` for the same request represent different calls or the same call observed under different map orderings.

**Trigger / repro:**

1. Configure a JWTPDP in scan-all SQL mode with `permittedVerbs = {SELECT: true}`.
2. Send a `tools/call` request where two arguments both contain SQL verbs, e.g., `{"query": "SELECT ...", "explain": "EXPLAIN SELECT ..."}`.
3. Run the same request 100 times; observe that `matched_op` in the audit JSONL alternates between `SELECT` and `EXPLAIN` (depending on which argument the Go runtime iterates last).

**Suggested fix:**

Collect all matching verbs into a deterministic structure before assigning `matchedOp`. The simplest fix: sort the argument names before iterating:

```go
argNames := make([]string, 0, len(args))
for k := range args {
    argNames = append(argNames, k)
}
sort.Strings(argNames)

for _, argName := range argNames {
    argVal := args[argName]
    // ... rest of logic unchanged
}
```

Alternatively, prefer the first-matching verb (break on first match) rather than last-wins. Either approach makes `matchedOp` deterministic for a given request shape. The sort approach is preferable because it also makes the deny/allow logic independent of map iteration order for multi-verb requests.

**Session:** rev-20260616T0210Z-d7fa63

eunolabs:
## Bugs addressed (session rev-20260616T0210Z-d7fa63)

- Closes #635 — In the JWT scan-all-args SQL mode, `evaluateJWTConditions` iterated the arguments map in Go's randomized order. When several arguments each carried a disallowed verb, which one the denial named (and the internally tracked `matchedOp`) varied across runs for one fixed request, making the audit text non-deterministic. The argument keys are now sorted before iteration.
- Closes #634 — `uniqueRotatedPath`'s `.N` collision search loop had no upper bound; an unbounded counter could (theoretically) wrap to a negative suffix. Bounded at `maxRotateSuffix` (10000); on exhaustion it returns an error and `rotate()` fails closed (keeps the existing fd, backs off) rather than overwriting a sibling with a wrapped-counter name. Split out a `uniqueRotatedPathBounded` seam so the exhaustion path is testable.
- Closes #633 — `ForceRefreshForVerify` snapshotted the cached key set *before* the forced fetch, then compared it to the post-fetch set; a concurrent refresh landing in that window made the comparison observe another goroutine's rotation instead of this fetch's own transition, so the negative-cache sentinel was wrongly left clear (extra fetches for up to `negativeKIDTTL`). The change determination is now computed inside the singleflight closure, under the same lock that installs the new set, and shared by every joiner.
- Closes #632 — `boundAuditObligations`' slow path sized each entry by its raw byte length, but it is only reached once the actual (escaped) encoding already exceeds the cap, so the "all entries fit" return could emit a slice serializing past `auditObligationsTotalCap`. It now accounts each entry by its exact JSON-encoded length (`jsonEncodedStringLen`), so the running total equals the true marshaled size and the cap holds.

## Testing

- Added: `TestJWTPDP_Decide_MultiVerbDenialIsDeterministic`, `TestUniqueRotatedPath_ExhaustionReturnsError`, `TestBoundAuditObligations_EscapeHeavyRespectsCap`; updated rotation-path tests for the new `(string, error)` signatures.
- #633 is a structural concurrency fix covered by the existing `TestJWKSCache_ForceRefreshForVerify_{DoesNotSuppressAfterRotation,SuppressesWhenUnchanged}` tests, green under `go test -race`.
- `gofmt`, `go vet`, `golangci-lint`, and `go test ./...` all pass.

---


eunolabs:
✅ Ready to merge — all session rev-20260616T0210Z-d7fa63 bugs addressed (#635, #634, #633, #632), self-review complete with no outstanding comments, CI green (Test, Lint, builds, demos, OPA-comparison, enforcement-compliance).

---

PR #656, opened 2026-06-16
Fix session rev-20260616T0110Z-3b0e07 bugs

### LINKED ISSUE #628: [rev-20260616T0110Z-3b0e07] sequenceBlock fails open when tool name carries trailing whitespace
**Location:** `pkg/enforcement/handlers.go` (`handleSequenceBlock`) and `pkg/enforcement/engine.go` (`recordSessionCall`)

**Reference:** Commit `9edd245f0178db578452636c845fcff2a394c146`

**Severity:** Medium — sequenceBlock condition silently permits tool calls that it should deny

**Description:**

`handleSequenceBlock` and `recordSessionCall` build their session-history keys from the tool name, but they apply whitespace normalization inconsistently:

- `handleSequenceBlock` trims each `afterTools` entry before building the **lookup** key:
  ```go
  priorTool = strings.TrimSpace(priorTool)
  // then: sequenceHistoryKey(req.SessionID, priorType, priorTool)
  ```
- `recordSessionCall` does **not** trim `req.ToolName` before building the **recording** key:
  ```go
  tool = splitEnginePrefix(req.ToolName)   // no TrimSpace
  // then: sequenceHistoryKey(req.SessionID, targetType, tool)
  ```

When a MCP client sends a tool name with trailing whitespace (e.g., `"read_file "` instead of `"read_file"`), the engine records the call under the key `"read_file "` (with space) but the condition looks it up under `"read_file"` (without space). The lookup returns count=0, so `handleSequenceBlock` concludes the prior tool has not yet run and does not block the call — a fail-open.

**Trigger / repro:**

1. Configure a `sequenceBlock` condition that requires tool A to run before tool B.
2. Have the MCP client send a `tools/call` request for tool A where the tool name has a trailing space (`"tool_a "`).
3. Then send a `tools/call` request for tool B.
4. Expected: tool B denied because tool A has not been called with the expected name. Actual: tool B allowed because the recording key (`"tool_a "`) does not match the lookup key (`"tool_a"`).

**Suggested fix:**

Apply `strings.TrimSpace` to `tool` in `recordSessionCall` (or canonicalize `req.ToolName` upstream at the transport layer before it enters the engine), consistent with how `handleSequenceBlock` already normalizes manifest entries. The canonical form should be decided in one place; the recording side is the cheaper fix.

**Session:** rev-20260616T0110Z-3b0e07

### LINKED ISSUE #629: [rev-20260616T0110Z-3b0e07] verifyAuditLog SEQ GAP check suppressed when sequence counter wraps through zero
**Location:** `cmd/eunox-mcp/audit.go` (~line 1515, `verifyAuditLog`)

**Reference:** Commit `9edd245f0178db578452636c845fcff2a394c146`

**Severity:** Low — HMAC chain still catches tampering; only the SEQ GAP diagnostic message is suppressed

**Description:**

`verifyAuditLog` includes a SEQ GAP check to detect injected or deleted records:

```go
if rec.Seq != prevSeq+1 && rec.Seq > 0 && prevSeq > 0 {
    // report gap
}
```

The `prevSeq > 0` guard is correct for the legacy-records case: two consecutive pre-signing records both have `Seq=0`, so the guard suppresses the false positive `0 != 0+1`.

However, the same guard creates a blind spot when the `uint64` sequence counter wraps around. When `prevSeq` is `math.MaxUint64`, the counter wraps to `0` for the next record. After processing that wrapped record, `prevSeq` is set to `0`. The record that follows has `prevSeq=0`, so `prevSeq > 0` evaluates to `false`, and the SEQ GAP check is silently skipped for that one record — giving an attacker who can insert a record immediately after the wrap point a window where the diagnostic goes unreported.

The HMAC chain (`prev_hmac` check) continues to detect any insertion correctly, so this is an observability/diagnostic gap rather than an authentication failure.

**Trigger / repro:**

1. Produce an audit log long enough for `seq` to reach `math.MaxUint64` (theoretical for normal operation; practically reproducible by patching the seq counter or constructing a synthetic log with `seq=18446744073709551615` followed by `seq=0` followed by an injected record with `seq=0`).
2. Run `eunox-mcp audit-verify` on the synthetic log.
3. Expected: SEQ GAP warning for the injected record. Actual: no SEQ GAP warning; only the HMAC chain error fires (if the injected record's `prev_hmac` is wrong).

**Suggested fix:**

Replace the `prevSeq > 0` test with a separate `signedSeen bool` (or `postWrap bool`) flag that tracks whether the verifier has observed at least one signed record with a nonzero seq. Set it to `true` after processing the first signed record with `Seq > 0`. Use it instead of `prevSeq > 0` in the gap check, so the check is not suppressed by a post-wrap `prevSeq=0`.

**Session:** rev-20260616T0110Z-3b0e07

### LINKED ISSUE #630: [rev-20260616T0110Z-3b0e07] allowedExtensions and recipientDomain silently skip blank array elements, inconsistent with allowedTables
**Location:** `pkg/enforcement/handlers.go` lines ~646–648 (`handleAllowedExtensions`) and ~953–955 (`handleRecipientDomain`)

**Reference:** Commit `9edd245f0178db578452636c845fcff2a394c146`

**Severity:** Low — no direct security bypass (empty strings fail extension/domain checks anyway), but violates the fail-closed design principle and produces a misleading error code

**Description:**

When an argument is an array of strings, `handleAllowedExtensions` and `handleRecipientDomain` silently drop any blank (empty or whitespace-only) element:

```go
// handleAllowedExtensions, line ~646
if s = strings.TrimSpace(s); s != "" {
    paths = append(paths, s)  // blank element simply vanished
}

// handleRecipientDomain, line ~953
if s = strings.TrimSpace(s); s != "" {
    recipients = append(recipients, s)  // same
}
```

`parseTableArgument` (used by `handleAllowedTables`) applies the opposite policy — it explicitly fails closed:

```go
case "":
    return nil, fmt.Errorf("array element is an empty or blank table name")
```

The inconsistency matters in two concrete ways:

1. **Misleading error code on all-blank arrays.** An array like `["", "  "]` for an `allowedExtensions` argument produces no `paths`, and the subsequent `len(paths) == 0` guard returns `MISSING_CONTEXT` ("argument is missing or empty"). The argument is not missing — it was supplied but contained only blank entries. `allowedTables` returns `CONDITION_FAILED` for the same shape.

2. **Silent masking of malformed client input.** A buggy MCP client that sends `["/data/report.pdf", ""]` has its blank entry silently dropped. The condition then evaluates only `/data/report.pdf`. While the security outcome is the same (the empty string would fail the extension check anyway), the blank element is never surfaced in the audit log or error response, making it invisible for debugging.

Note: this is distinct from closed issue #208 / fix `d2f017e`, which addressed *non-string typed* array elements. Empty strings are correctly typed; the issue is the skip-rather-than-reject policy.

**Suggested fix:**

Apply the same fail-closed policy as `allowedTables`: in the `case []interface{}:` branch of both handlers, after the `item.(string)` assertion succeeds, check `strings.TrimSpace(s) == ""` and return `CONDITION_FAILED` with an explicit message ("argument array contains a blank element") rather than silently skipping.

**Session:** rev-20260616T0110Z-3b0e07

### LINKED ISSUE #631: [rev-20260616T0110Z-3b0e07] JWKSCache.refresh performs spurious extra HTTP fetch when TTL check races with a concurrent refresh
**Location:** `pkg/capability/jwkscache.go` (~lines 344–398, `refresh` / non-forced path)

**Reference:** Commit `9edd245f0178db578452636c845fcff2a394c146`

**Severity:** Low — extra JWKS round-trip, not a security bypass; circuit breaker absorbs the failure case

**Description:**

The non-forced `refresh` flow has a TOCTOU window between its staleness check and its entry into `singleflight.Do`. The staleness check runs outside the singleflight closure:

```go
// outside singleflight
c.mu.RLock()
stale := time.Since(c.lastFetch) > c.ttl
c.mu.RUnlock()
if !stale {
    return nil
}

// singleflight.Do (only one goroutine executes the closure)
c.sfGroup.Do(key, func() (interface{}, error) {
    // no re-check of staleness here
    resp, err := http.Get(c.jwksURL)   // always fetches
    ...
})
```

If goroutine G1 sees a stale cache and enters `singleflight.Do`, and concurrently goroutine G2 also passes the staleness check before G1's fetch completes, G2 will block on singleflight (correct: it waits for G1). However, if G1 completes its refresh *between* G2's staleness check and G2's `singleflight.Do` call, G2 enters the singleflight closure as a new leader and issues a second, unnecessary HTTP request even though the cache is now fresh.

The `GetKeys` caller path has an additional pre-check, but it suffers from the same window: the cache can be refreshed between `GetKeys`'s pre-check and the call to `refresh`, and then `refresh` fetches again.

The result is an extra JWKS round-trip during concurrent JWT validation bursts (e.g., a spike of requests after a cold start or key rotation event). Under the circuit breaker, a second consecutive failed fetch could count as a second failure toward the open threshold, which is usually harmless but is unnecessary.

**Trigger / repro:**

1. Two goroutines both call `GetKeys` / `refresh` when the cache is stale.
2. G1 enters `singleflight.Do` and fetches JWKS (cache becomes fresh).
3. G2 passed the staleness check before G1 completed, so it enters `singleflight.Do` as the new leader and issues another HTTP GET.
4. Expected: G2 notices the cache is now fresh inside the closure and returns without fetching. Actual: G2 fetches anyway.

**Suggested fix:**

Add a staleness re-check inside the `singleflight.Do` closure (double-checked locking):

```go
c.sfGroup.Do(key, func() (interface{}, error) {
    c.mu.RLock()
    fresh := time.Since(c.lastFetch) <= c.ttl
    c.mu.RUnlock()
    if fresh {
        return nil, nil  // another goroutine already refreshed
    }
    // proceed with HTTP fetch
    ...
})
```

This is the standard double-checked locking pattern used by the forced-refresh path's `jwksKeysUnchanged` guard and is safe under the existing `RWMutex`.

**Session:** rev-20260616T0110Z-3b0e07

eunolabs:
## Bugs addressed (session rev-20260616T0110Z-3b0e07)

- Closes #631 — `JWKSCache.refresh`'s non-forced TTL check runs outside the singleflight, so a goroutine could pass it while stale and, if a concurrent refresh freshened the cache before it became the singleflight leader, issue a redundant HTTP fetch. Added a double-checked staleness re-check inside the singleflight closure: a non-forced refresh whose cache became fresh while it waited returns the cached set with no round-trip.
- Closes #630 — `handleAllowedExtensions` and `handleRecipientDomain` silently dropped blank (empty/whitespace) array elements, so an all-blank array produced a misleading `MISSING_CONTEXT` and malformed client input was hidden. Both now fail closed with `CONDITION_FAILED` ("array contains a blank element"), matching `parseTableArgument` (`allowedTables`).
- Closes #629 — `verifyAuditLog`'s SEQ GAP check was gated on `prevSeq > 0`; after the uint64 seq counter wraps, the record following the wrapped seq-0 record has `prevSeq == 0`, so a missing/renumbered record right after the wrap escaped the diagnostic. Gated on a `signedSeen` latch instead, which stays true across the wrap while still suppressing the legacy-record false positives.
- Closes #628 — `recordSessionCall` keyed session history from `req.ToolName` without trimming, while `handleSequenceBlock` trims its `afterTools` lookup entries. A tool name with trailing whitespace (matched via a glob cap) was recorded under a key the lookup never matched, so a `sequenceBlock`-gated tool failed OPEN. `recordSessionCall` now `TrimSpace`s the tool name (both the prefix-derived and target-fallback sources).

## Testing

- Added: `TestJWKSCache_Refresh_DoubleCheckSkipsRedundantFetch`, `TestEngine_AllowedExtensions_BlankArrayElementFailsClosed` + `TestEngine_RecipientDomain_BlankArrayElementFailsClosed`, `TestVerifyAuditLog_SeqGapAfterWrap_StillReported`, `TestSequenceBlock_AntecedentWithTrailingWhitespaceStillBlocks`.
- `gofmt`, `go vet`, `golangci-lint`, and `go test ./...` all pass (cache concurrency paths green under `-race`).

---


eunolabs:
✅ Ready to merge — all session rev-20260616T0110Z-3b0e07 bugs addressed (#631, #630, #629, #628), self-review complete with no outstanding comments, CI green (Test, Lint, builds, demos, OPA-comparison, enforcement-compliance).

---

PR #657, opened 2026-06-16
Fix session rev-20260616T0310Z-cdb798 bugs

### LINKED ISSUE #648: stdio+HTTP: host responses to server-initiated requests are dropped, breaking sampling round-trip when allowed
**Session:** `rev-20260616T0310Z-cdb798`

## Summary

When the PDP **allows** a server-initiated `sampling/createMessage` request (or any other upstream-initiated request such as `roots/list`), both transports forward the request to the host but **never route the host's response back to the upstream**. The upstream blocks indefinitely waiting for a response that never arrives.

### `cmd/eunox-mcp/stdio.go` — `serveHost` (lines 358–384)

```go
func (p *StdioProxy) serveHost(ctx context.Context) {
    var wg sync.WaitGroup
    for {
        msg, err := p.hostReader.Read()
        if err != nil { break }
        if msg.isNotification() {
            if msg.Method != "notifications/initialized" {
                _ = p.upWriter.Write(msg)   // forwarded ✓
            }
            continue
        }
        if msg.isRequest() {
            wg.Add(1)
            go func(m rpcMsg) { defer wg.Done(); p.handleHostRequest(ctx, m) }(msg)
        }
        // Ignore malformed messages.   ← host responses silently land here and are dropped
    }
    wg.Wait()
}
```

A JSON-RPC response (`id` present, `method` absent) satisfies neither `isNotification()` nor `isRequest()`, so it falls through and is silently discarded. The comment "Ignore malformed messages" applies to genuinely malformed input, but structurally valid host responses also reach that path.

### `cmd/eunox-mcp/http.go` — `handleMCPPost` (lines 701–703)

```go
if !msg.isRequest() {
    w.WriteHeader(http.StatusAccepted)
    return
}
```

When the host POSTs a JSON-RPC response to a server-initiated upstream request (e.g. the LLM result for `sampling/createMessage`), `!msg.isRequest()` is true and the handler returns 202 without forwarding the response to the upstream.

## Reproduction scenario (stdio)

1. Upstream sends `{"jsonrpc":"2.0","id":42,"method":"sampling/createMessage","params":{...}}`.
2. `readUpstream` → `handleUpstreamRequest`: PDP allows → `p.hostWriter.Write(msg)` forwards the request to the host.
3. Host calls the LLM and sends `{"jsonrpc":"2.0","id":42,"result":{"role":"assistant","content":{...}}}` back through the proxy's stdin.
4. `serveHost` reads the message: `isNotification()=false`, `isRequest()=false` → **dropped**.
5. Upstream is permanently blocked waiting for response id=42. No response arrives.

The same sequence occurs for `roots/list` and any future server-initiated method (e.g. `elicitation/create`).

## Impact

* An MCP server that issues `sampling/createMessage` and receives an **allow** from the proxy will hang indefinitely awaiting the LLM result. Since `handleUpstreamRequest` is called synchronously inside the `readUpstream` goroutine loop, a hung upstream response also prevents subsequent upstream messages from being processed.
* In HTTP mode the host's POST response is silently discarded (202 returned) for the same reason, leaving the upstream subprocess in the same hung state.
* The bug only manifests on the **allow** path. Denied sampling correctly returns an error response to the upstream (lines 743–750 in `stdio.go`; lines 1105–1110 in `http.go`), so the deny path is unaffected.

## Relationship to conformance.md

The conformance matrix states `sampling/createMessage` is "Enforced (local subprocess upstream only)" and points to Known Gaps for transport caveats. The Known Gaps table covers scenarios where enforcement is absent (remote HTTP upstream, stdio host + HTTP upstream) but does not document the missing response-routing for the local-subprocess case. The bug therefore exists on the only path the matrix currently marks as fully enforced.

## Notes on a fix

The fix requires two symmetric changes:

**stdio transport:** `serveHost` needs an `isResponse()` branch that forwards the host's response to the upstream writer. Because the upstream's server-initiated request IDs and the host's tool-call IDs share the same namespace, the proxy must track which IDs originated from server-initiated requests (a small map keyed by ID, populated in `handleUpstreamRequest`) so it can route only those responses to the upstream and let the rest fall through.

**HTTP transport:** `handleMCPPost` needs to detect incoming host responses (non-request, non-notification POSTs that carry an `id`) and forward them to the upstream subprocess via `sess.upWriter`. The same ID-tracking approach used for the stdio fix applies here.

Both fixes need corresponding tests: a full allowed-sampling round-trip test (upstream request → host receives → host responds → upstream receives result) for each transport.

eunolabs:
## Bugs addressed (session rev-20260616T0310Z-cdb798)

- Closes #648 — When the PDP **allowed** a server-initiated request (`sampling/createMessage`, or pass-through ones like `roots/list`), both transports forwarded the request to the host but never routed the host's **response** back to the upstream, so the upstream blocked forever awaiting a reply that was silently dropped.
  - **stdio** (`serveHost`): host responses satisfy neither `isNotification()` nor `isRequest()` and fell through to "ignore malformed messages". Added an `isResponse()` branch that routes the response to the upstream — but only for IDs the proxy actually forwarded (tracked in a new per-proxy `serverReqIDs` set populated by `handleUpstreamRequest`). Untracked responses are still ignored.
  - **HTTP** (`handleMCPPost`): a host response POST (`!isRequest()`) returned a bare `202` and was dropped. It now forwards tracked server-request responses to `sess.upWriter`; server-initiated requests broadcast over SSE are tracked per-session via `broadcastServerRequest`. Remote-mode sessions issue no server-initiated requests, so nothing is routed there.

  ID tracking bounds which host responses are forwarded (server-initiated request IDs and host tool-call IDs share a namespace), so a stray host reply cannot be injected to the upstream.

## Testing

- stdio: `TestStdioSamplingRoundTrip_HostResponseRoutedToUpstream` (full allow round-trip: upstream request → host → host response → upstream) and `TestStdioServeHost_UntrackedResponseIgnored`.
- HTTP: `TestHTTPSamplingRoundTrip_HostResponseRoutedToUpstream` and `TestHTTPHandleMCPPost_UntrackedResponseIgnored`.
- Existing sampling tests still pass; `go test -race ./cmd/eunox-mcp/` is clean.
- Updated `docs/conformance.md` (matrix + by-design no-audit notes).
- `gofmt`, `go vet`, `golangci-lint` all pass.

---


eunolabs:
✅ Ready to merge — session rev-20260616T0310Z-cdb798 bug addressed (#648), self-review complete with no outstanding comments, CI green (Test, Lint, builds, demos, OPA-comparison, enforcement-compliance, CodeQL) and `go test -race` clean.

---

PR #662, opened 2026-06-16
test: add comprehensive end-to-end integration suite

eunolabs:
## Summary

Adds a complete black-box end-to-end test suite (`demo/e2e/`) that runs the real compiled `eunox-mcp` binary against mock MCP server and host implementations, exercising every enforced MCP method, condition type, directive, and transport across five test legs with a shared signed audit tape.

## Why

The existing unit tests cannot catch a class of critical bugs: CLI/config wiring, transport framing, subprocess lifecycle, `*/list` filtering, response redaction, server-initiated sampling round-trips, kill-switch behavior, and audit-log integrity across process restarts. This suite drives the actual binary as a black box over real transports (stdio and HTTP) against a realistic upstream, asserting policy outcomes in positive, negative, and edge cases.

## Type of change

- New feature (comprehensive integration test suite)
- Test infrastructure

## What it tests

**Five legs, all writing to one shared signed audit tape:**

1. **Full enforcement matrix (stdio)** — every target type (tool, resource, prompt, system), every condition type (allowedValues, allowedOperations, allowedExtensions, recipientDomain, sequenceBlock, maxCalls, timeWindow), the redactFields directive, and server-initiated sampling ALLOW.

2. **Sampling DENY (stdio)** — a manifest without the `system:sampling/createMessage` opt-in denies server-initiated sampling requests.

3. **HTTP gateway suite** — transport, session isolation, concurrency, kill-switch (`/control/kill`), malformed request body, 404 routes, and per-route policy isolation (two upstreams with different manifests).

4. **Audit-verify** — the HMAC-SHA256 chain over the shared audit tape stays intact across all three proxy invocations (cross-restart continuity).

5. **Audit assertions** — decisions, denial codes, condition types, target types, and obligations are recorded as expected.

## Components added

- **`mock-server/main.go`** (558 lines) — upstream MCP server implementing the full enforced surface: `tools/list`, `tools/call` (with redactable and malformed payloads), `resources/list`, `resources/read`, `resources/subscribe`, `resources/templates/list`, `prompts/list`, `prompts/get`, `sampling/createMessage` (server-initiated), `ping`, `completion/complete`. Speaks both stdio and HTTP transports.

- **`mock-host/main.go`** (969 lines) — MCP host (client) that drives the proxy as a black box. Three modes: `--mode stdio` spawns the proxy and talks over pipes; `--mode http` connects to a running HTTP gateway; `--mode audit` parses the audit JSONL and asserts specific records. Includes assertion helpers (`expectAllow`, `expectDeny`, `expectErrorCode`) and result inspection utilities.

- **`run.sh`** (172 lines) — orchestration script that builds all three binaries, generates config files with absolute paths, spawns the mock server and proxy, runs the five test legs sequentially, and exits 0 only if all assertions pass.

- **Policy manifests:**
  - `policy.yaml` — comprehensive manifest covering every target type, condition, and directive.
  - `policy-no-sampling.yaml` — identical minus the `system:sampling/createMessage` opt-in (for sampling DENY leg).
  - `policy-db.yaml` — minimal per-route manifest (query_db SELECT only) for HTTP gateway cross-route isolation.

- **Tests:**
  - `mock-server/main_test.go` — unit tests for tool dispatch, redactable payload fixtures, and list methods.
  - `mock-host/main_test.go` — unit tests for result inspection helpers (toolText, firstContentJSON, listNames, etc.).

- **Documentation:**
  - `README.md` — overview of the suite, what it proves, and how to run it.

- **CI integration:**
  - `.github/workflows/demo-integration.yml` — new `e2e-suite` job that runs `make -C demo ci-test-e2e`.

- **Makefile:**
  - `demo/Makefile` — new `e2

https://claude.ai/code/session_01UnahJry1FF5WRYxXvfnNrR


demo/e2e/mock-host/main.go
```diff
@@ -0,0 +1,969 @@
...
+	s.expectDeny("foo/bar -> DENY (unknown method)", "AUTHORIZATION_FAILED", "", m, e)
+
+	// ── id-type preservation ──
+	m, e = c.callRawID(`"abc-123"`, "tools/call", map[string]interface{}{"name": "read_file", "arguments": map[string]interface{}{"path": "/reports/q3.pdf"}})
+	if e == nil && m.ID != nil && string(*m.ID) == `"abc-123"` {
+		s.ok(`id preservation: string id "abc-123" echoed unchanged`)
+	} else {
+		s.bad("id preservation (string)", fmt.Sprintf("err=%v id=%v", e, idStr(m.ID)))
+	}
+	m, e = c.callRawID(`90071992547409`, "tools/call", map[string]interface{}{"name": "read_file", "arguments": map[string]interface{}{"path": "/reports/q3.pdf"}})
```
eunolabs: `90071992547409` is ~9e13, well below 2^53 (`9007199254740992`), so it's exactly representable as a float64 and round-trips even through a parser that decodes JSON numbers to float64. This test therefore won't catch the classic JSON-number id-precision regression it appears to target.

To actually exercise big-integer id preservation, use a value **above** 2^53 — e.g. `9007199254740993` (which does *not* round-trip through float64).

---

demo/e2e/mock-host/main.go
```diff
@@ -0,0 +1,969 @@
...
+	if e == nil && m.Error == nil && strings.Contains(toolText(m), "sampling:allowed") {
+		s.ok("sampling/createMessage (opted in) -> ALLOW (round-trip reached host)")
+	} else {
+		s.bad("sampling allow", fmt.Sprintf("err=%v msg=%v text=%q", e, m.Error, toolText(m)))
+	}
+
+	// ── large + unicode payload pass through ──
+	big := strings.Repeat("A", 256*1024)
+	m, e = tc("read_file", map[string]interface{}{"path": "/reports/q3.pdf", "note": big, "emoji": "héllo-世界"})
+	s.expectAllow("tools/call with 256KiB + unicode argument -> ALLOW (no truncation of MCP response)", m, e)
```
eunolabs: This exercises a large *request argument* — `note`/`emoji` are dropped by the upstream `read_file` handler, so the response that comes back is tiny. The label's "(no truncation of MCP **response**)" isn't actually verified here.

Two options: reword to "large request argument", or (more valuable) add an upstream tool that returns a payload larger than the response path's interesting limit — the 8 MiB scanner cap (`sc.Buffer(make([]byte, 1<<20), 8<<20)`) is what a large-response test would actually stress.

---

demo/e2e/mock-host/main.go
```diff
@@ -0,0 +1,969 @@
...
+		s.ok("http initialize /mcp/db -> session established")
+		m, e = httpToolCall(db, "query_db", map[string]interface{}{"query": "SELECT 1"})
+		s.expectAllow("http /mcp/db query_db SELECT -> ALLOW", m, e)
+		m, e = httpToolCall(db, "read_file", map[string]interface{}{"path": "/reports/q3.pdf"})
+		s.expectDeny("http /mcp/db read_file -> DENY (read_file not in db-policy; per-route isolation)", "AUTHORIZATION_FAILED", "", m, e)
+	}
+
+	// ── concurrent-session isolation: sequenceBlock armed in one session must
+	//    not affect another ──
+	runHTTPSessionIsolation(base, s)
```
eunolabs: Nit: this (and the README's leg-3 row) call it "concurrent-session isolation", but sessions A and B run sequentially/interleaved here — there's no actual concurrency. It does correctly prove *session* isolation.

Since the proxy runs as a separate compiled process (no `-race`), spawning real goroutines wouldn't add much detection value, so I'd just soften the wording to "session isolation" rather than parallelize.

---

demo/e2e/mock-server/main.go
```diff
@@ -0,0 +1,558 @@
...
+		ID:      &id,
+		Method:  "sampling/createMessage",
+		Params: json.RawMessage(`{"messages":[{"role":"user","content":{"type":"text","text":"e2e ping"}}],` +
+			`"maxTokens":16,"systemPrompt":"e2e"}`),
+	})
+
+	// The next inbound line is either the host's sampling reply (proxy allowed
+	// and forwarded it) or the proxy's JSON-RPC error (proxy denied it). The
+	// host serializes its requests, so nothing else can interleave here.
+	line, ok := conn.readLine()
```
eunolabs: Optional hardening: this treats the next inbound line as the sampling reply/denial unconditionally, classifying purely on `reply.Error`. It's safe today because the host serializes its requests and the proxy emits nothing else on stdout — but a stray notification or out-of-band message would be misclassified as the sampling verdict.

Guarding on `reply.ID == samplingReqID` (and looping past anything that doesn't match) would make the round-trip robust without changing the happy path.

---

eunolabs (commented):
## Review: end-to-end integration suite

Thorough, well-structured black-box suite, and CI confirms it actually runs green (`demo / e2e` ✅) — which is the real test of this PR, since a passing run validates the config keys, denial codes, condition grammar, and CLI flags against the live binary. A few things are genuinely strong:

- **Non-vacuous redaction asserts.** The redaction checks are guarded by fixture tests (`TestGetSecretRecord_RedactablePayload`, `TestGetMalformed_LooksLikeJSON`) so a silent payload change can't make the redaction assertion pass trivially. That's the right instinct.
- **Real cross-restart audit verification.** Sharing one signed audit tape across three *separate* proxy processes and then `audit-verify`-ing it exercises HMAC chain continuity across restarts, not just within a single process. The explicit "kill + wait so the drainer flushes" before verify shows the non-blocking audit sink was accounted for.
- **Good fail-closed coverage.** Deny-by-default for unmapped methods (`ping`, `completion/complete`, `resources/templates/list`, unknown methods), the missing-arg (`MISSING_CONTEXT`) vs wrong-type (`CONDITION_FAILED`) distinction, and the malformed-JSON redaction failing closed (`-32603`) are all exercised. `expectAllow`/`expectDeny`/`expectErrorCode` cleanly separate transport errors, structured denials, and bare internal errors.

I left four inline notes — mostly cases where an assertion's *label* claims more than it verifies (the "large numeric id" and "large response" ones are the substantive ones). None are blocking.

### Minor / optional (not worth inline)
- `setEqual` is order-sensitive slice equality (callers pre-sort), not true set semantics — `slicesEqual` would read truer and avoid surprise if a future list contains duplicates.
- `samplingReqID` is a fixed constant — fine today (one `trigger_sampling` per leg), but two calls in one session would reuse the id.
- The redaction fail-closed path returns a bare JSON-RPC `-32603` with no structured denial data, unlike the `AUTHORIZATION_FAILED`/`CONDITION_FAILED` shape elsewhere. The suite documents this faithfully; flagging only so it's a conscious choice that response-processing failures aren't surfaced as structured denials — a proxy-behavior question, not a test issue.

Nice work — this fills a real gap that in-process unit tests can't reach.

---

PR #663, opened 2026-06-16
Refactor server-request tracking and add live-introspection timeouts

eunolabs:
## Summary

Consolidates duplicated server-initiated request tracking logic into a shared `serverReqTracker` type, bounds the tracked ID set to prevent unbounded growth, adds context timeouts to live introspection (validate/doctor/init --live), and refactors JWT PDP list filtering to eliminate code duplication.

## Why

**Server-request tracking:** Both `StdioProxy` and `HTTPProxy` independently reimplemented the same logic to track JSON-RPC IDs of server-initiated requests (sampling/createMessage, roots/list) forwarded to the host, so responses could be routed back to the upstream. This duplication risked divergence and left the tracked set unbounded — a host that never responds would grow the map for the session lifetime.

**Live introspection hangs:** The validate, doctor, and init subcommands with `--live` spawn a stdio upstream and run the MCP handshake to introspect its capabilities. If the upstream completes initialize but never answers tools/list, the command blocks forever. Without a context deadline, there's no way to unwedge it.

**JWT PDP duplication:** The three list-filtering methods (`FilterToolsList`, `FilterResourcesList`, `FilterPromptsList`) in `JWTPDP` had identical control flow, differing only in which inner filter they called.

**Sampling opt-in on HTTP upstreams:** A manifest that grants `system:sampling/createMessage` cannot be enforced for a remote HTTP upstream — eunox only reads server-initiated requests from subprocess upstreams. Loading such a config silently ships an inert grant, which is a fail-open risk.

## Type of change

- Refactor or chore (no behavior change) — server-request tracking consolidation, JWT PDP list-filter refactoring
- Bug fix — bounded server-request tracking prevents unbounded growth; live-introspection timeouts prevent hangs
- Security hardening — reject HTTP upstreams with sampling opt-ins to prevent silent inert grants

## How it was tested

```
make test    # All existing tests pass; new tests added:
             # - TestTrackServerReqID: verifies bounded insert and eviction
             # - TestFetchLiveToolsStdio_ContextTimeout: proves deadline unwedges a non-answering stdio upstream
             # - TestBuildRoutes_SamplingOptInOnHTTPUpstreamRejected: sampling opt-in rejected for HTTP
             # - TestBuildRoutes_SamplingOptInOnStdioUpstreamAllowed: sampling opt-in allowed for stdio
```

## Security considerations

**Fail-closed preserved:** The sampling opt-in check in `loadUpstreamPDP` rejects HTTP upstreams with a manifest grant at config load time, preventing silent inert enforcement. Live introspection timeouts prevent a wedged upstream from hanging the validate/doctor/init commands indefinitely.

**Bounded tracking:** The `maxTrackedServerReqs` cap (1024) prevents a non-responding host from growing the server-request ID map unbounded. When the cap is reached, an arbitrary old ID is evicted; a late response to a forgotten ID is simply not routed (same as any untracked ID), so the loss fails safe.

**No audit impact:** Server-request tracking is internal request/response correlation; audit records are unaffected.

**No secrets impact:** No new token, env var, or header handling.

**No authority widening:** The sampling opt-in rejection is a restriction (deny HTTP upstreams with sampling), not a widening.

## Checklist

- [x] `make lint` passes — `go vet` + golangci-lint
- [x] Docs updated — `docs/capability-manifest-guide.md` clarified `allowedOperations` glob semantics; threat model updated with sampling opt-in enforcement scope
- [x] Schema unchanged — no new config fields
- [x] PR description explains why (duplication, unbounded growth, hangs, silent inert grants)

https://claude.ai/code/session_01Pb45H186UdCDHGFmGHrj4J


cmd/eunox-mcp/jsonrpc.go
```diff
@@ -54,6 +54,63 @@ func msgKey(id *json.RawMessage) string {
...
+	if ids == nil {
+		ids = make(map[string]struct{})
+	}
+	if _, exists := ids[key]; !exists && len(ids) >= maxTrackedServerReqs {
+		for k := range ids {
+			delete(ids, k)
+			break
+		}
+	}
+	ids[key] = struct{}{}
```
eunolabs, line 88: Good fix for the unbounded-growth bug — the cap plus a shared tracker is the right call, and the test covers the bound/eviction/dup cases well.

One nuance worth recording in the comment itself: eviction picks an **arbitrary** tracked ID (`for k := range ids { … break }`). The "fails safe" framing holds for eunox (bounded memory, no authority leak), but the evicted ID is a *real* in-flight server-initiated request whose upstream subprocess is blocked awaiting that exact response. Dropping it re-introduces the #648 hang for that one call until the session tears down (or an upstream-side timeout fires) — so it isn't strictly equivalent to an untracked ID, where nothing is waiting.

At a 1024 cap this only triggers under a pathological flood (a host ignoring 1024+ outstanding `sampling`/`roots` requests, in which case the upstream is already wedged), so I think the trade-off is fine as-is. If you wanted eviction to be least-harmful, dropping the **oldest** entry (FIFO via a tiny ring of keys) sheds the request most likely already abandoned rather than possibly the one about to be answered. Optional, given the trigger is pathological.

---

cmd/eunox-mcp/route.go
```diff
@@ -236,6 +236,15 @@ func loadUpstreamPDP(u *UpstreamConfig, counter callcounter.Store, ks killswitch
 		return nil, nil, "", "", fmt.Errorf("upstream %q: manifest version %q does not match pinned expectVersion %q", u.Name, merged.Version, u.ExpectVersion)
 	}
 
+	// A system:sampling/createMessage opt-in cannot be enforced for a remote HTTP
+	// upstream: eunox reads server-initiated requests only from a subprocess
+	// upstream, so a remote upstream's sampling/createMessage is never seen,
+	// enforced, or audited. Rather than load a grant that is silently inert,
+	// fail closed and tell the operator (issue surfaced in transport review).
+	if u.Transport == hostTransportHTTP && manifestHasSamplingOptIn(merged) {
```
eunolabs, line 244: Good hardening, and nicely placed: because the check lives in `loadUpstreamPDP`, it covers **both** callers — the gateway path (`BuildRoutes`, route.go:151) and the non-gateway stdio host (main.go:762). I verified that a stdio host fronting a remote HTTP upstream hits this same rejection, so the silent-inert-grant gap is closed for that combination too, not just the gateway. (It also runs on the *merged* manifest, so a sampling grant introduced by any policy file is caught.)

Minor: `manifestHasSamplingOptIn` returns true for *any* `system:sampling/createMessage` entry regardless of its `actions`. If a `deny` on a `system:` target is expressible in the grammar, an operator who explicitly *denies* sampling on an http upstream would also be rejected here — harmless (it's inert on this transport anyway) but slightly surprising. If `system:` targets are allow-only by validation, ignore this; otherwise gating on the action would avoid the false positive.

---

pkg/enforcement/handlers.go
```diff
@@ -207,7 +207,7 @@ func (e *Engine) handleTimeWindow(_ context.Context, cond capability.Condition,
 				Message:       "request is before the allowed time window",
 				Details: map[string]interface{}{
 					"notBefore": tw.NotBefore,
-					"now":       now.Format(time.RFC3339),
+					"now":       now.Format(time.RFC3339Nano),
```
eunolabs, line 210: Sensible: this lines the `now` detail up with the RFC3339Nano precision already used to *parse* `notBefore`/`notAfter` (L195/L217), so time-window denials keep sub-second precision in the audit detail. It's outside the PR's stated scope, though — worth a line in the description so it doesn't surprise anyone in `git blame`, and a quick check that no golden / sign-and-verify round-trip test pins the old second-precision `now` string.

---

eunolabs (commented):
## Review: refactor server-request tracking + live-introspection timeouts

Solid, well-documented PR. The refactors preserve behavior, the two bug fixes (bounded tracking, live-introspection timeouts) are real, and the sampling-opt-in rejection is a genuine fail-closed improvement. I checked out the branch and verified the load-bearing claims against the source rather than reading the diff alone. **No blockers** — a few notes below and inline.

### What I verified
- **Timeout mechanism is sound.** `fetchLiveToolsStdio` binds the subprocess via `exec.CommandContext(procCtx, …)` with a deferred stdin-close → `Wait` → `AfterFunc(cancel)` backstop, so a parent-ctx deadline cancels `procCtx` → subprocess is killed → stdout closes → the blocked `r.Read()` returns → `readResponseWithID` surfaces `ctx.Err()`. `TestFetchLiveToolsStdio_ContextTimeout` exercises exactly this path.
- **Sampling rejection covers both transports.** The check lives in `loadUpstreamPDP`, called from both `BuildRoutes` (gateway) and the non-gateway stdio host (`main.go:762`), and runs on the *merged* manifest — so the stdio-host→remote-HTTP combination is covered, not just the gateway. (inline on `route.go`)
- **Nil `pending` on remote sessions is safe.** The only map write (`s.pending[key] = ch`) is in `callSubprocessUpstream`, reachable only when `s.upstreamURL == ""`; `callUpstream` routes remote sessions to `callRemoteUpstream` first, and the subprocess reader goroutine that touches `pending` isn't started for remote sessions. The `http_remote.go` comment is accurate.
- **`matchResource`/`resSpecificity` relocation is clean.** They're now referenced only from `drift.go`, so moving them out of `pdp.go` and the "used by the drift checks here" comment are both accurate.
- **`allowedValues` doc-sync matches code.** The new "glob-only, never exact-first" comment in `condition.go` matches `handlers.go` (`MatchValueGlob` / `path.Match`, ~L1109). Good to have the type doc agree with the handler.
- **JWT `filterList`/`innerFilter` refactor is behavior-preserving** — the consolidated flow is identical to the three originals, and the method-expression selector (`ListFilterer.FilterToolsList`) type-checks cleanly against `sel`.

### One note I couldn't pin inline (pre-existing line, outside the diff)
`main.go:789` — the `stdio host with HTTP upstream … sampling deny control is inactive` NOTICE is now slightly stale. With the new hard rejection at `route.go:244`, a sampling opt-in on an http upstream can no longer reach this line (it errors at load), so "deny control is inactive" reads as "would silently no-op" when it's actually a startup error. Worth tightening toward `roots/list` / the no-opt-in case so the two messages don't appear to contradict.

### Minor
- The PR description says docs were updated for "`allowedOperations` glob semantics," but the `condition.go` change is about **`allowedValues`** (the guide change is the separate `allowedOperations` first-token caveat). Both doc changes are real — the description just conflates them.
- The `RFC3339` → `RFC3339Nano` change in `handlers.go` isn't called out in the summary (inline note with a test-coverage check).

Nice work overall — the inline rationale is unusually thorough, and the "limit of the in-chain guarantee" and `hasSubscribers` trade-off notes are the kind of thing that saves a future reader real time.

---

cmd/eunox-mcp/route.go
eunolabs: Confirmed `system:` targets are allow-only: `validActionsFor(TargetTypeSystem)` returns `["allow", "*"]`, and manifest validation additionally rejects an `enforcement` keyword on a system target ("the opt-in is binary"). A `deny` on `system:sampling/createMessage` isn't expressible, so `manifestHasSamplingOptIn` has no false-positive path — left presence-based as-is.

---

cmd/eunox-mcp/jsonrpc.go
eunolabs: Good nuance — clarified the comment in 87351ea: a forgotten ID is a real in-flight request whose upstream is blocked on that response, so eviction can re-hang that one call (the #648 hang) until teardown, and it's bounded to the pathological 1024+ flood where the upstream is already wedged. Kept arbitrary eviction and noted FIFO was considered but a key ring isn't worth the complexity at this trigger.

---

eunolabs (commented):
## Re-review (head `3392dfe`) — feedback addressed, looks good to merge

Took another pass over the three commits added since my first review (`87351ead`, the merge of `main`, and `3392dfed`). All of my earlier points are resolved, and the new changes are clean.

### My previous comments — all addressed
- **Eviction comment (`jsonrpc.go`)** ✅ — rewritten to state exactly the nuance: a forgotten ID is a *real* in-flight request whose upstream is blocked on that response, so eviction can re-hang that one call (the #648 hang), bounded to a pathological 1024+ flood where the upstream is already wedged. FIFO-considered-and-skipped is noted. Accurate now.
- **Stale NOTICE (`main.go:789`)** ✅ — reworded from "sampling deny control is inactive" to "server-initiated messages … are not consumed or relayed … a manifest opt-in for it on an HTTP upstream is rejected at load." No longer contradicts the new load-time rejection.
- **`RFC3339Nano` scope (`handlers.go`)** ✅ — the standalone commit message documents the rationale (aligns the `now` detail with `DecidedAt`/drop-markers; comparison already used full precision, so no decision changes), and the green `Test` check confirms no golden/round-trip test pinned the old second-precision string.
- **`manifestHasSamplingOptIn` deny-action edge case** ✅ — resolved by validation, no code needed: `validActionsFor(TargetTypeSystem)` is `["allow", "*"]`, so a `system:` target can never carry `deny`. Presence is unambiguously an opt-in, so the check cannot misfire.

### New work since last review
- **`3392dfed` (e2e overlay split)** is a correct fix for a real downstream consequence of the hardening: the shared `demo/e2e/policy.yaml` carried the sampling opt-in and was reused by the HTTP gateway leg, which now fails closed at load. Splitting the opt-in into `policy-sampling.yaml` and layering it only on the stdio leg is the right shape. The `gen_stdio_config` rewrite (`shift` + per-file loop, 6-space YAML list indentation) is correct, and the header comments explain the split clearly. Good catch — and the green `demo / gateway` + `demo / e2e` checks confirm the rejection fires and the suite still passes end-to-end.

### Status
CI is fully green (43/43 checks, including `Test`, `Lint`, `Enforcement compliance`, `govulncheck`, CodeQL, and all demo legs), and `mergeable_state` is now `clean`. No remaining concerns from me — this is in good shape to merge.

---

PR #664, opened 2026-06-16
Release-readiness fixes: kill-switch fail-closed default, /control/kill auth, supply-chain consistency

eunolabs:
## Summary

Fixes the findings from the release-readiness audit, in one PR. Two of these are deliberate, security-relevant **behavior changes** (flagged `!` / BREAKING below) that were chosen explicitly; the rest are consistency, supply-chain, and test-coverage cleanups.

Full suite is green (`go test -race ./...`, 1815+ tests, 0 skips), `golangci-lint` 0 issues, `go vet` clean, `gofmt`/license/NOTICE/stale-syntax checks pass, coverage floors hold (pkg 97.7%, cmd 92.8%).

## Changes (5 commits)

**`feat(http)!`: require an auto-generated control token on `/control/kill` (MED-4 / SEC-07)**
- The emergency-stop endpoint was previously guarded only by the loopback-source check and the host-facing `listen.authToken` (unset by default, and mutually exclusive with `--jwks-uri`), so any same-host process — including a compromised upstream subprocess — could trigger a global kill (self-DoS) just by reaching the loopback endpoint.
- The proxy now mints a 256-bit token at startup, writes it `0600` to `~/.eunox/control.token` (`--control-token-path` to override), and requires it via the `X-Eunox-Control-Token` header (constant-time compare), independent of `authToken`/JWT mode. No token configured ⇒ endpoint refuses (503, fail closed).
- `eunox-mcp kill` reads the file automatically (`--control-token` / `EUNOX_CONTROL_TOKEN` override).
- **BREAKING:** `eunox-mcp kill` over HTTP now requires the control token.

**`feat(killswitch)!`: fail closed by default during a Redis outage (MED-3 / ADR-0003)**
- The Redis kill switch previously failed **open** when Redis was unreachable, silently bypassing revocations issued during a partition — the single documented exception to eunox's fail-closed posture.
- Now defaults to **fail closed**: once a refresh fails, `ShouldBlock` denies any request it cannot confirm (`KILL_SWITCH_ERROR`, via the static `ErrBackendUnreachable` sentinel so the Redis dial error isn't leaked to clients). A kill already cached before the outage still blocks normally.
- Opt back in with `--killswitch-fail-open` (`Redis.WithFailOpen(true)`); active mode logged at startup. In-memory/single-instance unaffected.
- **BREAKING:** `--redis-addr` deployments now deny during a Redis outage unless `--killswitch-fail-open` is set.

**`chore`: align CODEOWNERS, dependabot, CI, registry, and docs with the repo**
- CODEOWNERS: replaced the invalid `-` default-owner pattern with `*`, removed dead multi-service paths, and added an owner for `cmd/eunox-mcp/` (the network-facing code previously had none).
- dependabot: extended docker coverage to the two demo images built in CI.
- CI: pinned `golangci-lint` to `v2.12.2` (was `latest`).
- Makefile: docker targets point at `ghcr.io/eunolabs/eunox-mcp` (matching release/README).
- `.dockerignore`: `\*.md` → `*.md`; release-checklist cosign filename → `checksums.txt.sigstore.json`.

**`fix(docker)`: build release images with the tested Go 1.25.11 toolchain**
- Both Dockerfiles built `FROM golang:1.26.4` while go.mod and all CI use `1.25.11` (and the comment claimed 1.25.11). Pinned the builder images to `1.25.11`.

**`test(jwt)`: pin algorithm-confusion defenses (reject HS256 and `alg:none`)**
- Added negative tests at both the capability (`VerifyToken`) and IdP (`ValidateToken`) layers asserting symmetric and unsecured tokens are rejected — previously the asymmetric-only allowlist had no explicit guard test.

## Docs updated in-PR
ADR-0003 (rewritten), threat-model (§3.3/§3.7/§5.3, L-2, revs 1.27–1.28), README (kill-switch ops note + `kill` command), CHANGELOG (Added / Changed / Security).

## Deferred (not in this PR)
- **Docker base-image digest pinning** (audit L5): resolving real registry digests needs network access this environment lacks; inventing digests would break the build. Dependabot (now covering all docker dirs) + the release-time Trivy gate mitigate.
- **macOS code signing** and a **production ops runbook**: out of scope for a code PR (need signing infra / product decisions).

https://claude.ai/code/session_0161dAw1qKVT3ohSnTpNkcVK

---
_Generated by [Claude Code](https://claude.ai/code/session_0161dAw1qKVT3ohSnTpNkcVK)_


pkg/killswitch/redis.go
```diff
@@ -232,6 +291,15 @@ func (r *Redis) ShouldBlock(_ context.Context, agentID, sessionID string) (bool,
 	if sessionID != "" && r.killedSessions[sessionID] {
 		return true, nil
 	}
+	// Nothing in the local cache matches. In the default fail-closed degraded
+	// mode, a failed most-recent refresh means the cache may be stale -- a kill
+	// issued during the outage would not be reflected -- so deny rather than admit
+	// an unconfirmed request. The reconcile loop clears lastRefreshErr once Redis
+	// is reachable again, bounding the block window. Opt-in fail-open serves the
+	// cache as-is for availability.
+	if r.lastRefreshErr != nil && !r.failOpen {
```
eunolabs, line 304: The wiring is correct — `killCheck` (cmd/eunox-mcp/pdp.go:117-120) checks the error *before* the bool and maps `ErrBackendUnreachable` to a `KILL_SWITCH_ERROR` deny, and the sentinel carries no host:port, so this fails closed cleanly. Verified `pkg/killswitch` and the new `redis_failclosed_test.go` pass under `-race`.

One operational consequence worth surfacing for fail-closed deployments: recovery is bounded by the reconcile interval, not Redis itself. Once `lastRefreshErr` is set, it's only cleared on the next *successful* `refreshState` — and during an outage pub/sub is also down, so that's effectively the next reconcile tick. A brief blip observed at a tick therefore yields up to ~`defaultReconcileInterval` (30s) of full data-plane denial *after* Redis has already recovered. That's the documented trade-off (ADR-0003), but `WithReconcileInterval` is only reachable from the library — there's no `--killswitch-reconcile-interval` flag — so a binary operator who picks fail-closed can't shorten that window. Worth exposing the interval as a flag alongside `--killswitch-fail-open`, since the two knobs are jointly what makes fail-closed tunable in production.

---

cmd/eunox-mcp/control_token.go
```diff
@@ -0,0 +1,88 @@
...
+	expanded, err := expandHome(path)
+	if err != nil {
+		return "", err
+	}
+	if dir := filepath.Dir(expanded); dir != "" && dir != "." {
+		if err := os.MkdirAll(dir, 0o700); err != nil { //nolint:gosec // G301: 0700 is the intended restrictive mode
+			return "", fmt.Errorf("creating control-token directory: %w", err)
+		}
+	}
+	if err := os.WriteFile(expanded, []byte(token+"\n"), 0o600); err != nil { //nolint:gosec // G306: 0600 is the intended restrictive mode for a secret
```
eunolabs: Minor hardening (low severity, given the file holds an ephemeral per-start secret): `os.WriteFile` applies the `0600` mode only when it *creates* the file. If `~/.eunox/control.token` already exists with looser perms, the mode is left untouched and the secret is written into, e.g., a `0644` file. The directory is `MkdirAll`'d `0700`, which makes this hard for a different local user to reach in practice, so this is informational — but it isn't quite what the comment above addresses (that's about replay/atomicity, not mode-on-existing-file).

Separately, truncate-then-write is non-atomic against a concurrent `eunox-mcp kill` reading this same file during a proxy restart: the reader can observe an empty/partial token. `resolveControlToken` rejects an empty file, and a partial token just fails the constant-time compare (401, operator retries), so it's not a security hole — just a transient flake.

A write-temp-then-`os.Rename` would close both at once: the rename is atomic, and the temp file is a fresh inode created `0600` regardless of any pre-existing file's mode. If you keep the direct write, an explicit `os.Chmod(expanded, 0600)` after `WriteFile` covers the mode case.

---

eunolabs (commented):
## Review — release-readiness fixes

Reviewed the full diff (27 files, +991/−205) and traced the security-relevant paths against the actual tree at `66313e8`. This is a high-quality, well-documented PR; the two behavior changes are correctly implemented and the docs/threat-model updates are thorough. Findings below are minor/hardening — nothing blocking.

### What it does
Five logically-separate changes: (1) `feat(http)!` an auto-generated 256-bit control token gating `POST /control/kill` (SEC-07); (2) `feat(killswitch)!` Redis kill switch fails **closed** by default during an outage, opt-out via `--killswitch-fail-open` (MED-3/ADR-0003); (3) `chore` CODEOWNERS/dependabot/CI/registry alignment; (4) `fix(docker)` pin builder images to the tested `go 1.25.11`; (5) `test(jwt)` algorithm-confusion negative tests.

### Verification performed
- **Fail-closed is correctly wired (not a no-op).** `killCheck` (`cmd/eunox-mcp/pdp.go:117-120`) inspects the `ShouldBlock` error *before* the bool and maps it to a `KILL_SWITCH_ERROR` deny. `(false, ErrBackendUnreachable)` therefore denies. ✅
- **No backend-detail leak.** `ErrBackendUnreachable` is a static sentinel; `redis_failclosed_test.go` asserts the dial error / host:port never reaches the client message. ✅
- **Control token always present in production.** The only non-test caller path is `serveHTTPGateway → NewHTTPProxyGateway`, which generates + passes the token; there is no production `serveHTTP` that could leave it empty (which would 503 the endpoint). ✅
- **`eunox-mcp kill --redis-addr` does *not* require the control token** — `cmdKill` returns on the Redis branch (`main.go:1441`) before `resolveControlToken`. ✅
- **Dependabot dirs are real** — all three (`deploy/docker`, `demo/opa-comparison/server`, `demo/mock-mcp-server`) contain Dockerfiles. ✅
- **Ran the key new tests under `-race`:** `pkg/killswitch` (fail-closed/fail-open) and `cmd/eunox-mcp` (control-token, SEC-07, kill cmd, JWT alg-confusion) — all pass. ✅

### Strengths
- Constant-time token compare (`hmac.Equal`), fail-closed on `rand.Read` error, 503 when unconfigured rather than reverting to no-auth.
- ADR-0003 rewrite, threat-model §3.3/§3.7/§5.3 + L-2, and the `1.27/1.28` security-revision rows are exemplary — the "why", the residual risk, and the migration are all spelled out.
- Both `!`/BREAKING changes are explicitly flagged with migration notes.

### Findings (2 inline)
- `pkg/killswitch/redis.go` — fail-closed recovery is bounded by the 30s reconcile interval, which isn't exposed as a CLI flag; consider adding one alongside `--killswitch-fail-open`.
- `cmd/eunox-mcp/control_token.go` — `os.WriteFile` won't tighten perms on a pre-existing file and isn't atomic vs a concurrent `kill` read; a temp-then-rename would harden both. Low severity.

### Additional notes (non-blocking)
- **Multi-instance, same host, default token path:** each HTTP proxy rewrites `~/.eunox/control.token` at startup, so with >1 proxy under one `$HOME` only the last-started instance's token survives, and `kill --port <earlier-instance>` would 401. The `--control-token-path` override covers it; a one-line doc breadcrumb (or per-port default) would save a debugging session. The primary multi-instance story is `--redis-addr` + Redis-side kill, which is unaffected.
- **Mergeability:** GitHub currently reports this branch as conflicting with `main` (`mergeable_state: dirty`) — it'll need a rebase/merge before it can land.
- The fail-closed default puts Redis on the data-plane critical path for `--redis-addr` deployments (a Redis outage denies new `initialize`/all enforced calls until recovery). This is the intended, documented trade-off — flagging it only so it's a conscious operational expectation, not a surprise.

Overall: looks good to merge once rebased; the inline items are optional polish.

---

cmd/eunox-mcp/control_token.go
eunolabs: Fixed in 0cd94ba. `writeControlTokenFile` now writes to a fresh `os.CreateTemp` file (created `0600`, with an explicit `Chmod(0600)` to be umask-proof) and `os.Rename`s it into place — so the secret never lands in a looser-mode pre-existing file, and a concurrent `eunox-mcp kill` reader observes either the old or the new token, never an empty/partial one. Added a regression test that overwriting a `0644` file still lands the secret at `0600`.

---

pkg/killswitch/redis.go
eunolabs: Good call — exposed in 2c9574d as `--killswitch-reconcile-interval` (0 = default 30s), the tuning companion to `--killswitch-fail-open`. Wired to `WithReconcileInterval`, logged at startup when set, and documented in the flag help, the `WithReconcileInterval` doc comment, and ADR-0003's consequences (noting recovery is bounded by this interval, not Redis itself).

---

PR #673, opened 2026-06-16
Fix session rev-20260616T0707Z-b374a0 bugs

### LINKED ISSUE #665: [rev-20260616T0707Z-b374a0] FindKeys panics on nil *jose.JSONWebKeySet receiver
**Location:** `pkg/capability/jwkscache.go:492-497`

**Reference:** https://github.com/eunolabs/eunox/blob/172dc5f7dcae101f23de48fec0eab1466b04e897/pkg/capability/jwkscache.go#L492-L497

**Severity:** Medium

**Description:**

`FindKeys` is an exported function that accesses the JWKS receiver without a nil guard:

```go
func FindKeys(jwks *jose.JSONWebKeySet, kid string) []jose.JSONWebKey {
    if kid != "" {
        return jwks.Key(kid)   // panics if jwks == nil
    }
    return jwks.Keys           // panics if jwks == nil
}
```

Calling `FindKeys(nil, kid)` causes a nil pointer dereference panic on either branch. Both `jwks.Key(kid)` (method call) and `jwks.Keys` (field access) dereference the receiver; Go will panic with a nil pointer dereference before either body executes.

All current callers in the codebase pass non-nil JWKS obtained from successful cache fetches, so no current caller triggers this in production. However, as an **exported** function with no documentation constraint against nil, an external package consumer calling `FindKeys` with a nil set — or a future internal caller that forgets to check the fetch error first — would crash the process. This is especially dangerous in the audit/enforcement hot path where a panic is unrecovered.

**Trigger / repro:**

```go
// Any caller passing nil explicitly or failing to check an error return:
keys := capability.FindKeys(nil, "some-kid")
// → panic: runtime error: invalid memory address or nil pointer dereference
```

The path most likely to accidentally trigger this in the future is any new caller that calls `FindKeys` on a `*jose.JSONWebKeySet` returned from a function that can return `(nil, err)` without first checking the error:

```go
set, err := someJWKSFetch(ctx)
// forgot: if err != nil { return ... }
keys := capability.FindKeys(set, kid)  // set may be nil → panic
```

**Suggested fix:**

Add an explicit nil guard and return an empty slice (matching the "no keys found" semantics a caller expects when the key set is absent):

```go
func FindKeys(jwks *jose.JSONWebKeySet, kid string) []jose.JSONWebKey {
    if jwks == nil {
        return nil
    }
    if kid != "" {
        return jwks.Key(kid)
    }
    return jwks.Keys
}
```

Alternatively, document the precondition (`// jwks must be non-nil`) in a godoc comment so the contract is explicit for external callers — but a nil guard is safer for an exported API.

**Session:** rev-20260616T0707Z-b374a0

### LINKED ISSUE #666: [rev-20260616T0707Z-b374a0] retryAfterSeconds ceiling-rounds negative sub-second durations to 1 instead of falling back to the full window
**Location:** `pkg/enforcement/handlers.go:504-513`

**Reference:** https://github.com/eunolabs/eunox/blob/172dc5f7dcae101f23de48fec0eab1466b04e897/pkg/enforcement/handlers.go#L504-L513

**Severity:** Low

**Description:**

The function comment documents that `retryAfterSeconds` falls back to the full `windowSec` whenever the backend retry-after estimate is "unavailable (`<= 0`)":

```go
// retryAfterSeconds converts a backend retry-after estimate to whole seconds
// (rounded up), falling back to the full window when the estimate is unavailable
// (<= 0). Shared by the commit and check-only maxCalls denial paths so both
// advise the hint the same way.
func retryAfterSeconds(d time.Duration, windowSec int) int64 {
    secs := int64(d / time.Second)
    if d%time.Second != 0 {
        secs++
    }
    if secs < 1 {
        secs = int64(windowSec)
    }
    return secs
}
```

The implementation has a subtle gap: for **negative durations with magnitude strictly less than one second** (e.g., `-1ns` through `-999ms`), Go's integer truncation-toward-zero produces `secs = 0`, and the `d%time.Second != 0` fractional test then increments it to `1`. The `secs < 1` guard never triggers, and the function returns `1` instead of `windowSec`.

Concrete examples:

| `d`      | `d / time.Second` | `d % time.Second` | `secs` after ceil | final return |
|----------|-------------------|--------------------|-------------------|--------------|
| `-1ns`   | `0`               | `-1ns` (≠ 0)       | `1`               | `1` ← **wrong** |
| `-500ms` | `0`               | `-500ms` (≠ 0)     | `1`               | `1` ← **wrong** |
| `-999ms` | `0`               | `-999ms` (≠ 0)     | `1`               | `1` ← **wrong** |
| `-1s`    | `-1`              | `0`                | `-1`              | `windowSec` ✓ |
| `0`      | `0`               | `0`                | `0`               | `windowSec` ✓ |
| `1500ms` | `1`               | `500ms` (≠ 0)      | `2`               | `2` ✓ |

**Current backend behavior:** Both `InMemory.IncrementIfBelow` and `Redis.IncrementIfBelow` return `retryAfter >= 0` in practice (the denied-call path computes a non-negative time until a slot frees). So this bug is currently unreachable with the shipped backends. However, it violates the documented contract ("falls back when estimate is unavailable, i.e. `<= 0`") and would silently activate against a future backend or library consumer that returns a negative duration to indicate "no estimate available."

**Impact when triggered:** The rate-limited error surfaced to the calling MCP host would carry `retryAfter: 1` (second) instead of `retryAfter: windowSec`. A well-behaved client would retry after just 1 second, get denied again immediately (the quota window is still full), and repeat — generating unnecessary deny noise and audit records for the full window duration rather than backing off for the intended window length.

**Suggested fix:**

Guard against negative inputs before the ceiling computation:

```go
func retryAfterSeconds(d time.Duration, windowSec int) int64 {
    if d <= 0 {
        return int64(windowSec)
    }
    secs := int64(d / time.Second)
    if d%time.Second != 0 {
        secs++
    }
    if secs < 1 {
        secs = int64(windowSec)
    }
    return secs
}
```

The early-return on `d <= 0` matches the documented intent exactly and eliminates the surprising ceiling behaviour for negative sub-second values.

**Session:** rev-20260616T0707Z-b374a0

### LINKED ISSUE #667: [rev-20260616T0707Z-b374a0] writeRecord's post-rotate s.f == nil guard is unreachable dead code
**Location:** `cmd/eunox-mcp/audit.go:1255-1261`

**Reference:** https://github.com/eunolabs/eunox/blob/172dc5f7dcae101f23de48fec0eab1466b04e897/cmd/eunox-mcp/audit.go#L1255-L1261

**Severity:** Low

**Description:**

`writeRecord` contains two consecutive `s.f == nil` guards, one before and one after the `rotate()` call:

```go
// audit.go:1249-1261
if s.f == nil {
    // File was lost during a failed rotation — skip this record. Defensive:
    // rotate() always retains a valid fd, but count the loss if it is ever hit.
    s.writeFailures.Add(1)
    return
}
    s.rotate()
}
if s.f == nil {         // ← this check is unreachable
    s.writeFailures.Add(1)
    return
}
```

The first guard (line 1249) is explicitly documented as defensive ("rotate() always retains a valid fd"). The **second** guard (line 1258) is dead code: `rotate()` never sets `s.f = nil`. Tracing every exit path in `rotate()`:

| Outcome | `s.f` after `rotate()` returns |
|---------|-------------------------------|
| rename fails → early return | unchanged (original fd, still valid) |
| new file open fails → early return | unchanged (original fd, now at renamed path) |
| success → closes old fd, opens new | `s.f = f` (new non-nil `*os.File`) |

No path through `rotate()` assigns `nil` to `s.f`. The post-rotate nil check therefore can never be entered, yet it increments `writeFailures` — a metric operators alert on as "audit trail is incomplete" — for an impossible scenario.

**Why this matters:**

1. **Dead code is a maintenance hazard.** Someone adding a code path to `rotate()` (e.g. a future error path that clears the fd) might observe the guard and incorrectly conclude it is already tested and covered, when it is not.

2. **Misleading `writeFailures` accounting.** If the branch were ever reachable, a `writeFailures` increment without a written marker would silently inflate the metric. The correct handling for a nil post-rotate fd would be to also emit a stderr warning (as the first guard's caller chain does on a disk-write failure).

3. **The comment on the first guard is subtly misleading** — it says the guard is there because "rotate() always retains a valid fd, but count the loss if it is ever hit." That sentence describes both guards identically, but the second has no comment at all, making it easy to miss during review.

**Suggested fix:**

Remove the second (post-rotate) nil check. The first (pre-rotate) guard already handles a nil fd inherited from startup or any other path not involving `rotate()`. If `rotate()` is ever modified to leave `s.f` nil on a new error path, that path should handle `writeFailures` and stderr in `rotate()` itself (where the context is clear), not silently in `writeRecord`.

```go
// Before:
if s.f == nil {
    s.writeFailures.Add(1)
    return
}
    s.rotate()
}
if s.f == nil {          // dead — remove
    s.writeFailures.Add(1)
    return
}

// After:
if s.f == nil {
    // File was lost during a failed rotation — skip this record. Defensive:
    // rotate() always retains a valid fd, but count the loss if it is ever hit.
    s.writeFailures.Add(1)
    return
}
    s.rotate()
}
```

**Session:** rev-20260616T0707Z-b374a0

eunolabs:
## Bugs addressed (session rev-20260616T0707Z-b374a0)

- Closes #666 — `retryAfterSeconds` (`pkg/enforcement/handlers.go`) ceiling-rounded negative sub-second durations (e.g. `-500ms`) to `1` instead of falling back to the full window, contradicting its documented `<= 0` fallback contract. Added an early `d <= 0 -> windowSec` guard before the ceiling computation.
- Closes #665 — `FindKeys` (`pkg/capability/jwkscache.go`) is exported and dereferenced its `*jose.JSONWebKeySet` receiver with no nil guard, so `FindKeys(nil, kid)` panicked. Added a nil guard returning no keys.
- Closes #667 — `writeRecord` (`cmd/eunox-mcp/audit.go`) had an unreachable second `s.f == nil` guard after `rotate()`; no path through `rotate()` assigns nil to `s.f`, so the dead branch could only ever misreport `writeFailures`. Removed it (pre-rotate guard retained), with a comment documenting the invariant.

## Testing

- Added `TestRetryAfterSeconds` (table-driven: positive ceiling, zero, and negative sub-second/whole-second/large cases all fall back correctly).
- Added `TestFindKeys_NilSet` (nil set returns nil, no panic, for both empty and non-empty kid).
- `go build ./...`, `go vet ./...`, `make check-fmt`, `make check-license`, and targeted `-race` tests for enforcement/capability/audit all pass.

---


eunolabs:
✅ Ready to merge — all session rev-20260616T0707Z-b374a0 bugs addressed (#665, #666, #667), self-review complete with no outstanding comments, tests passing (`go test -race ./pkg/enforcement/ ./pkg/capability/ ./cmd/eunox-mcp/`, plus `make check-fmt`/`check-license`/`go vet`).

---

PR #674, opened 2026-06-16
Fix session rev-20260616T0610Z-f2bd35 bugs

### LINKED ISSUE #659: [rev-20260616T0610Z-f2bd35] VerifyToken silently accepts tokens with co-signatures without verifying them (fail-open inconsistency with DPoP rejection)
**Location:** `pkg/capability/jwks.go:109-307`
**Reference:** https://github.com/eunolabs/eunox/blob/5e3f461/pkg/capability/jwks.go#L109-L307
**Severity:** Medium (security)

**Description:**

`JWKSClient.VerifyToken` verifies the primary JWT signature and standard claims, but does **not** check whether the token carries co-signatures (`IssuanceProofs.Signatures`) and silently accepts any such token without calling `VerifyCoSignatures`.

The `IssuanceProofs` enforcement contract in `pkg/capability/token.go` is explicit:

> "When Signatures is non-empty, ALL co-signatures MUST be verified before a token is considered valid."
> "This is NOT purely decorative metadata: the presence of co-signatures changes the verification requirements for the token."

The DPoP `cnf.jkt` field is handled consistently with this contract — `VerifyToken` **rejects** tokens carrying it with a terminal error (lines 234–236):

```go
if p.Confirmation != nil && p.Confirmation.JKT != "" {
    return nil, fmt.Errorf("validate claims: sender-constrained token (cnf.jkt) requires proof-of-possession …"), nil
}
```

The rationale given there is exactly what applies here: the proxy cannot verify the constraint, so accepting the token without verification silently downgrades the security model. Co-signatures have no analogous guard; `p.Proofs` is never inspected after the payload is unpacked, so a locally-issued token that carries non-empty `IssuanceProofs.Signatures` is returned as valid without any of those signatures being checked.

**Trigger / repro:**

1. Issue a native capability token whose `proofs.signatures` array is non-empty (any `IssuerSignature` records, regardless of validity).
2. Call `JWKSClient.VerifyToken` with this token.
3. The call succeeds and returns the payload — the co-signatures are never passed to `VerifyCoSignatures`.

A co-issuer might have imposed restrictions via a co-signature; those restrictions are silently skipped. More critically, the stated contract ("ALL co-signatures MUST be verified") is violated even though `VerifyCoSignatures` and the full verification infrastructure exist and are exported.

**Suggested fix:**

Add a terminal error for locally-issued tokens that carry co-signatures, parallel to the DPoP rejection, in the `tryKeys` closure just after the DPoP check:

```go
if p.Proofs != nil && len(p.Proofs.Signatures) > 0 {
    return nil, fmt.Errorf(
        "validate claims: token carries %d co-signature(s) that require VerifyCoSignatures "+
            "with a PublicKeyResolver; refusing to accept unverified co-signatures "+
            "(fail closed — use the federated token path if co-signatures are intended)",
        len(p.Proofs.Signatures)), nil
}
```

This matches the DPoP precedent and the documented enforcement contract. When the proxy gains a co-signature verification path the check becomes a gate that routes to `VerifyCoSignatures` instead of a hard rejection.

**Session:** rev-20260616T0610Z-f2bd35

### LINKED ISSUE #660: [rev-20260616T0610Z-f2bd35] audit record SessionID field is not size-bounded unlike Target and Method
## Location

`cmd/eunox-mcp/audit.go` — `Record(...)` call site where `auditRecord` is constructed; compare with `boundEnvelopeField` calls on `Target` and `Method`.

## Reference

https://github.com/eunolabs/eunox/blob/5e3f461/cmd/eunox-mcp/audit.go

## Severity

Low-Medium (defense-in-depth gap; low exploitability in practice)

## Description

`Target` and `Method` in every audit record are explicitly capped by `boundEnvelopeField` (limit: `auditEnvelopeFieldCap` = 8 KiB) before the record is serialized and HMAC-signed. The comment in the code notes this is specifically to prevent an attacker-controlled field from producing a record large enough to overflow the `bufio.Scanner` buffer (4 MiB) used by `verifyAuditLog`.

`SessionID` is stored verbatim with no analogous bound:

```go
rec := auditRecord{
    ...
    Target:    boundEnvelopeField(target),   // capped
    Method:    boundEnvelopeField(method),   // capped
    SessionID: sessionID,                    // NOT capped
    ...
}
```

In HTTP mode, `sessionID` originates from the `Mcp-Session-Id` request header, which a client fully controls. The HTTP proxy stores the session ID at session-creation time and stamps it on every subsequent audit record. If a client presents a pathologically large `Mcp-Session-Id` value (e.g., several MiB), the resulting JSON record can exceed 4 MiB and will cause `bufio.Scanner` to return `bufio.ErrTooLong` during `audit-verify`, silently truncating or aborting log verification for any session with that ID.

The inconsistency is particularly notable because the threat model explicitly sizes the scanner buffer (`auditScannerBufSize = 4 * 1024 * 1024`) as the backstop, and `boundEnvelopeField` exists precisely to ensure no single field can breach it — yet `SessionID` is exempt.

## Trigger / repro

1. Run eunox in HTTP gateway mode.
2. Open a session presenting `Mcp-Session-Id: <4 MiB random string>`.
3. Make any tool call through that session (generates one audit record with the oversized session ID).
4. Run `eunox-mcp audit-verify` against the log file.
5. `bufio.Scanner` returns `ErrTooLong`; verification of subsequent records in the file is aborted.

In stdio mode the session ID is either absent or operator-controlled, making this unexploitable there.

## Suggested fix

Apply `boundEnvelopeField` (or a separate, appropriately-sized cap) to `SessionID` at record construction time, mirroring the treatment of `Target` and `Method`. A separate cap constant (e.g., `auditSessionIDCap = 256`) would also prevent log pollution from legitimate but accidentally long session IDs and aligns with the UUIDv4 values the server normally assigns.

## Session

rev-20260616T0610Z-f2bd35

### LINKED ISSUE #661: [rev-20260616T0610Z-f2bd35] verifyAuditLog postWrap detection fails for legitimate seq=0 as first record of a rotated log file
## Location

`cmd/eunox-mcp/audit.go` — `verifyAuditLog`, specifically the `postWrap` / `forgedSeq0` detection block.

## Reference

https://github.com/eunolabs/eunox/blob/5e3f461/cmd/eunox-mcp/audit.go

## Severity

Low (astronomically rare trigger; logic error in audit-verify, not in runtime enforcement)

## Description

`verifyAuditLog` distinguishes a legitimate seq=0 record (a genuine wrap from `math.MaxUint64`) from a forged decoy (an injected record crafted to look like a chain-start):

```go
postWrap := havePrev && prevSeq == math.MaxUint64
forgedSeq0 := rec.Seq == 0 && rec.HMAC != "" && !postWrap
```

The exemption only fires when `havePrev == true` (i.e., at least one previous record has been processed in the current file). When a rotated log file begins with the wrap record — seq=0 immediately following seq=`math.MaxUint64` in the previous file — `havePrev` is `false` for the very first line, so `postWrap` is always `false`, and the record is flagged as a forged decoy (`forgedSeq0 = true`).

The chain state (`prevSeq`, `havePrev`) is local to one `verifyAuditLog` invocation and is not threaded across file boundaries. If the operator passes multiple log files to audit-verify (or if the tool is designed to scan a directory), the wrap record at the head of a rotated file is misevaluated as a forgery.

Concretely: after exactly 2^64 written records, the first record of the new file has seq=0, HMAC set (it was signed normally by `writeRecord`), and `havePrev=false` — it is incorrectly classified as a forged decoy and increments the invalid-record counter.

## Trigger / repro

This is not practically triggerable in any near-term deployment (requires 2^64 records). The bug is a latent logic error in the chain-verification invariant that would surface if:

1. A single eunox instance runs long enough to write `math.MaxUint64` + 1 records (wrap at seq=0), AND
2. `audit-verify` is invoked on the rotated file beginning at that wrap record as the first argument (so `havePrev` starts as `false` when the seq=0 record is read).

## Suggested fix

Carry chain state across files when verifying a sequence of log files, or treat `havePrev=false && rec.Seq==0` as an ambiguous case (neither confidently legitimate nor confidently forged) rather than defaulting to `forgedSeq0=true`. One option: emit a warning for the cross-file wrap case rather than an error, and document that the cross-file continuity check requires the caller to pass files in chronological order.

Alternatively, reserve seq=0 as a sentinel only used for the very first record ever written (outside of any wrap), and use a distinct wrap-marker field so the chain is unambiguous across file boundaries.

## Session

rev-20260616T0610Z-f2bd35

eunolabs:
## Bugs addressed (session rev-20260616T0610Z-f2bd35)

- Closes #659 — `JWKSClient.VerifyToken` (`pkg/capability/jwks.go`) verified only the primary issuer signature and never inspected `IssuanceProofs.Signatures`, so a locally-issued token carrying co-signatures was accepted with them unverified — violating the `IssuanceProofs` enforcement contract and silently skipping a co-issuer's restrictions. Now rejects any verified token with a non-empty `proofs.signatures` slice as a terminal error, parallel to the existing `cnf.jkt` rejection. Updated the `IssuanceProofs` doc comment and threat model.
- Closes #660 — the audit `session_id` field was written unbounded, unlike `target`/`method`. In HTTP mode it comes verbatim from the client-controlled `Mcp-Session-Id` header, so a multi-MiB value could push a record past the 4 MiB `audit-verify` scanner buffer and abort verification. Now capped at 256 bytes (`auditSessionIDCap`) with a visible marker at record construction; UUIDv4 session IDs pass through untouched.
- Closes #661 — when a rotated log file *begins* with a legitimate post-wrap `seq 0` record and is verified in isolation, `verifyAuditLog` flagged it with a misleading "forged seq-0 decoy" message. The fail-closed verdict is correct and deliberately kept (a positional first-record exemption would re-open the #446 genesis-substitution vector), but the first-record diagnostic now explains the record cannot be confirmed in isolation and directs the operator to re-verify with the chronologically preceding file prepended, so a genuine cross-file wrap is verifiable via chain continuity.

## Testing

- Added `TestJWKSClient_VerifyToken_CoSignaturesRejected` (reject with co-signatures; sanity-verify without).
- Added `TestRecordOversizedSessionIDBounded` (~4 MiB session id → bounded, record still HMAC-verifies).
- Added `TestVerifyAuditLog_PostWrapSeq0_FirstRecordInIsolation_Flagged` (isolation → flagged invalid with the cross-file remedy in the message); existing forged/wrap regression tests still pass.
- `go test -race ./pkg/capability/ ./cmd/eunox-mcp/`, `make check-fmt`, `make check-license`, `go vet` all pass.

---


eunolabs:
✅ Ready to merge — all session rev-20260616T0610Z-f2bd35 bugs addressed (#659, #660, #661), self-review complete with no outstanding comments, tests passing (`go test -race ./pkg/capability/ ./cmd/eunox-mcp/`, `make check-fmt`/`check-license`/`go vet`). Threat model updated with entries 1.28–1.30.

---

PR #675, opened 2026-06-16
Fix session rev-20260616T0710Z-99fcb4 bugs

### LINKED ISSUE #668: [rev-20260616T0710Z-99fcb4] descriptionHash enforced only at tools/list filter, not at tools/call — description-changed tools remain callable
**Location:** `cmd/eunox-mcp/pdp.go:741-756` (`filterToolsListResult`) vs `decideTarget` (tools/call enforcement path)

**Reference:** https://github.com/eunolabs/eunox/blob/172dc5f7dcae101f23de48fec0eab1466b04e897/cmd/eunox-mcp/pdp.go#L741-L756

**Severity:** Medium (security — prompt-injection protection bypass)

**Description:**

`descriptionHash` is a manifest field that pins a tool's description text. The feature exists specifically to guard against prompt-injection attacks where an upstream server silently changes a tool's description to manipulate the MCP host into calling it with attacker-controlled arguments.

The current implementation only enforces `descriptionHash` inside `filterToolsListResult`, which is called when processing a `tools/list` response:

```go
// filterToolsListResult — called for tools/list only
func (p *ManifestPDP) filterToolsListResult(...) {
    for _, tool := range tools {
        if constraint.DescriptionHash != "" {
            actual := sha256OfDescription(tool.Description)
            if actual != constraint.DescriptionHash {
                // tool is hidden from the list
                continue
            }
        }
        permitted = append(permitted, tool)
    }
}
```

`decideTarget`, which is called for every `tools/call` request, never checks `descriptionHash`:

```go
// decideTarget — called for tools/call (and resources/read, prompts/get)
func (p *ManifestPDP) decideTarget(...) (Decision, error) {
    // kill-switch check, findConstraint, action check, schema validation,
    // evaluateAndRecord — no descriptionHash verification anywhere
}
```

**Impact:**

A tool whose description changes after the manifest was authored is correctly hidden from `tools/list` — the MCP host won't see it in the tool list. However, a host that issues a `tools/call` directly (by name, without consulting the current tool list) bypasses the filter entirely. The tool call is evaluated against conditions and forwarded to the upstream, even though the description no longer matches the pinned hash.

The attack scenario the feature targets — an upstream that rewrites a tool description to inject a prompt — can therefore still complete the `tools/call` leg of the attack. The `tools/list` filter gives false assurance that the pin is enforced end-to-end.

**Trigger / repro:**

1. Manifest has `descriptionHash: <sha256 of original description>` for `my_tool`.
2. Upstream silently changes `my_tool`'s description (e.g. prompt-injection payload added).
3. `tools/list` → `my_tool` is hidden from the filtered list (correct).
4. Host issues `tools/call my_tool` directly (any well-known tool the host has cached, or an attacker-controlled host that ignores the list).
5. `decideTarget` allows the call — the changed description never triggers a deny.

**Suggested fix:**

At the start of `decideTarget` (or inside `evaluateAndRecord`), when a matching constraint carries a non-empty `DescriptionHash`, fetch the live tool description from the upstream's `tools/list` cache (already available in the proxy's upstream state) and compare:

```go
if constraint.DescriptionHash != "" {
    liveDesc, ok := p.liveDescription(target)
    if !ok || sha256OfDescription(liveDesc) != constraint.DescriptionHash {
        return Deny, authorizationFailedError("description hash mismatch for tool %q", target)
    }
}
```

Alternatively, the proxy can cache each tool's last-seen description at `tools/list` time and use that snapshot for the `decideTarget` check, so no extra upstream round-trip is needed.

Either path must be taken before conditions are evaluated, so a description change always produces a deny (fail-closed).

**Session:** rev-20260616T0710Z-99fcb4

### LINKED ISSUE #669: [rev-20260616T0710Z-99fcb4] checkLimit has no upper bound — limits above 2^53 lose precision silently in Lua float64 arithmetic
**Location:** `pkg/callcounter/callcounter.go:224-229` + `pkg/callcounter/redis.go:225-238`

**Reference:** https://github.com/eunolabs/eunox/blob/172dc5f7dcae101f23de48fec0eab1466b04e897/pkg/callcounter/callcounter.go#L224-L229

**Severity:** Low

**Description:**

`checkLimit` validates that a caller-supplied `limit` is at least 1, but imposes no upper bound:

```go
// pkg/callcounter/callcounter.go
func checkLimit(limit int64) error {
    if limit < 1 {
        return fmt.Errorf("callcounter: limit must be >= 1, got %d", limit)
    }
    return nil
}
```

The `limit` value is then forwarded to the Redis backend's `incrIfBelowScript` Lua script as a string argument (`ARGV[4]`), where it is converted back to a number via Lua's `tonumber()`:

```go
// pkg/callcounter/redis.go — incrIfBelowScript (simplified)
-- ARGV[4] = limit (string)
local limit = tonumber(ARGV[4])
if count < limit then
    -- admit
end
```

Lua's `tonumber()` produces an IEEE-754 double-precision float64. Go's `int64` can represent values up to `2^63 - 1 ≈ 9.2 × 10^18`; float64 can represent integers exactly only up to `2^53 = 9007199254740992 ≈ 9.0 × 10^15`. Any `limit` above `2^53` is rounded to the nearest representable float64 — silently, with no error from either Lua or Go.

**Concrete example:**

| Go `limit` | Lua `tonumber()` result | difference |
|---|---|---|
| `9007199254740993` (2^53+1) | `9007199254740992.0` | -1 (rounded down) |
| `9007199254740994` (2^53+2) | `9007199254740994.0` | exact (even, representable) |
| `9007199254741000` | `9007199254741000.0` | exact |
| `9223372036854775807` (MaxInt64) | `9.223372036854776e+18` | off by ~512 |

For most practical rate limits this precision loss is harmless — no operator sets a per-call limit above a few billion. However:

1. The function signature accepts `int64` with no documented ceiling.
2. The validation explicitly promises to fail only on `< 1`, implying all other values are valid.
3. A manifest author or API consumer who supplies a very large limit (e.g. `math.MaxInt64` as a "no effective limit" sentinel) would receive silent enforcement of a slightly different number, with no diagnostic.

**Suggested fix:**

Add an upper bound in `checkLimit` equal to `2^53` (the largest integer exactly representable as float64), and document the constraint:

```go
// maxLimitForLua is the largest integer exactly representable as a Lua number
// (IEEE-754 float64). Limits above this would be silently rounded in the Redis
// Lua script, producing a different enforcement threshold than requested.
const maxLimitForLua = 1 << 53

func checkLimit(limit int64) error {
    if limit < 1 {
        return fmt.Errorf("callcounter: limit must be >= 1, got %d", limit)
    }
    if limit > maxLimitForLua {
        return fmt.Errorf("callcounter: limit %d exceeds maximum representable value in Lua (%d)", limit, maxLimitForLua)
    }
    return nil
}
```

Alternatively, the Lua script could accept the limit as two 32-bit integers and reconstruct it as a 64-bit integer using integer arithmetic, but the cap is simpler and sufficient in practice.

**Session:** rev-20260616T0710Z-99fcb4

### LINKED ISSUE #670: [rev-20260616T0710Z-99fcb4] boundEnvelopeField violates its own documented invariant when the truncation marker itself exceeds auditEnvelopeFieldCap
**Location:** `cmd/eunox-mcp/audit.go:1083-1095`

**Reference:** https://github.com/eunolabs/eunox/blob/172dc5f7dcae101f23de48fec0eab1466b04e897/cmd/eunox-mcp/audit.go#L1083-L1095

**Severity:** Low

**Description:**

`boundEnvelopeField` is documented with an explicit invariant:

```go
// boundEnvelopeField truncates s to auditEnvelopeFieldCap bytes, appending a
// truncation marker so the reader knows the field was cut.
// The returned string is always <= auditEnvelopeFieldCap bytes.
func boundEnvelopeField(s string) string {
    if len(s) <= auditEnvelopeFieldCap {
        return s
    }
    const marker = " [truncated]"
    keep := auditEnvelopeFieldCap - len(marker)
    if keep < 0 {
        return marker   // <-- returns marker regardless of its length vs cap
    }
    return s[:keep] + marker
}
```

When `keep < 0` — i.e., `len(marker) > auditEnvelopeFieldCap` — the function returns `marker` alone. The marker (`" [truncated]"`, 12 bytes) is hardcoded and `auditEnvelopeFieldCap` is currently 8192 bytes, so `keep < 0` is unreachable today. However the documented invariant **"The returned string is always <= auditEnvelopeFieldCap bytes"** is only conditionally true:

- `auditEnvelopeFieldCap >= len(marker)`: invariant holds (normal path).
- `auditEnvelopeFieldCap < len(marker)`: the `keep < 0` branch returns `marker` whose length (`len(marker)`) exceeds `auditEnvelopeFieldCap` — **invariant violated**.

This is a latent correctness issue. If either the marker string is ever made longer (e.g. to include a byte count: `" [truncated, original 12345 bytes]"`) or `auditEnvelopeFieldCap` is ever reduced below `len(marker)` (e.g. for a stricter log-line budget), the `keep < 0` branch silently emits an over-budget field. Downstream consumers — in particular `verifyAuditLog`'s scanner, which is sized to `auditScannerBufSize = 4 MiB` — could eventually be affected if many such fields compound into a record larger than the scanner can handle.

The bug is particularly subtle because:
1. The comment says the invariant holds unconditionally ("always").
2. The `keep < 0` path looks defensive/safe — it returns something shorter than the original — but it does not actually honour the cap.
3. Tests that exercise the normal (large-string) truncation path would not cover this branch.

**Suggested fix:**

In the `keep < 0` branch, truncate the marker to exactly `auditEnvelopeFieldCap` bytes rather than returning it verbatim, or assert that the cap is at least `len(marker)` at startup (via a `func init()` or a build-time constant comparison):

```go
func boundEnvelopeField(s string) string {
    if len(s) <= auditEnvelopeFieldCap {
        return s
    }
    const marker = " [truncated]"
    keep := auditEnvelopeFieldCap - len(marker)
    if keep < 0 {
        // Marker itself exceeds cap — truncate the marker to the cap.
        // This path is unreachable with current constants but the invariant
        // must hold regardless.
        return marker[:auditEnvelopeFieldCap]
    }
    return s[:keep] + marker
}
```

Alternatively, add a package-level `init` that panics if the invariant would be violated:

```go
func init() {
    const marker = " [truncated]"
    if len(marker) > auditEnvelopeFieldCap {
        panic("boundEnvelopeField: truncation marker exceeds auditEnvelopeFieldCap")
    }
}
```

Either approach makes the invariant unconditionally true and protects future maintainers who might adjust either constant.

**Session:** rev-20260616T0710Z-99fcb4

eunolabs:
## Bugs addressed (session rev-20260616T0710Z-99fcb4)

- Closes #668 — `descriptionHash` was enforced only at the `tools/list` filter, so a host issuing `tools/call` by name bypassed the pin and a description-changed tool was still forwarded (prompt-injection bypass). `ManifestPDP` now records each tool's live description hash whenever a `tools/list` response is filtered, and `decideTarget` denies a `tools/call` (`AUTHORIZATION_FAILED`) when a pinned tool's most recently observed description no longer matches the pin. An as-yet-unobserved description is **not** denied — the startup drift probe already verifies every pin against the live upstream (fatal on mismatch) before the session is handed to the host, so a legitimate `initialize → tools/call` flow is not broken.
- Closes #669 — `checkLimit` (`pkg/callcounter`) enforced `limit >= 1` but no ceiling, while the Redis Lua backend converts the limit via `tonumber()` to a float64 (exact only up to 2^53), so a larger limit was silently rounded to a different threshold. Now rejects `limit > 2^53` (`MaxLimit`) up front with a structured error.
- Closes #670 — `boundFieldTo` (shared by `boundEnvelopeField` and the new `session_id` cap) returned the whole truncation marker when the marker exceeded the limit, violating its "`<= limit` bytes" invariant. Unreachable with current constants, but a tighter future budget would let an over-budget field through. The `keep < 0` branch now truncates the marker to the limit.

## Testing

- `TestDescriptionHash_EnforcedAtToolsCall` — unobserved → allow; observed-matching → allow; observed-changed → deny (`AUTHORIZATION_FAILED`); restored → allow.
- `TestInMemory_IncrementIfBelow_LimitAboveMaxFailsClosed` — `limit > MaxLimit` fails closed with no state; boundary `MaxLimit` admits.
- `TestBoundFieldToHonorsLimitBelowMarker` — tiny limits still yield `<= limit` bytes.
- `go test -race ./pkg/callcounter/ ./cmd/eunox-mcp/`, `make check-fmt`/`check-license`, `go vet` all pass. Threat model updated (§3.7 rewritten; entries 1.31–1.33).

---


eunolabs:
✅ Ready to merge — all session rev-20260616T0710Z-99fcb4 bugs addressed (#668, #669, #670), self-review complete with no outstanding comments, tests passing (`go test -race ./pkg/callcounter/ ./cmd/eunox-mcp/`, `make check-fmt`/`check-license`/`go vet`). Verified `checkLimit` is enforced on both callcounter backends and the descriptionHash cache is shared by the inner ManifestPDP across JWT/gateway modes. Threat model §3.7 rewritten + entries 1.31–1.33.

---

PR #676, opened 2026-06-16
Fix session rev-20260616T0800Z-c7d91a bugs

### LINKED ISSUE #671: [rev-20260616T0800Z-c7d91a] pdp.go: blanket AuditOnly defer in decideTarget stamps stray-argumentSchema ENFORCEMENT_ERROR as forwarded in audit mode
## Location

`cmd/eunox-mcp/pdp.go` — `ManifestPDP.decideTarget`, lines 255–363

## Session ID

`rev-20260616T0800Z-c7d91a`

## Severity

Medium — security logic contradicts documented invariant. Practical impact is currently gated behind programmatic manifest construction (the YAML loader's invariant prevents it via normal usage), but the bug was introduced by the B-1 refactoring from issue #539 and violates the explicit "fail closed even in audit mode" guarantee stated in two separate code comments.

## Description

`decideTarget` uses a named return value and a blanket defer to stamp every denial from an audit-mode constraint as `AuditOnly = true`:

```go
func (p *ManifestPDP) decideTarget(...) (resp capability.EnforceResponse) {
    // ...
    matched := p.findConstraint(target, claims)
    if matched.IsAuditOnly() {
        defer func() { resp.AuditOnly = true }()   // ← stamps ALL returns
    }
    // ...
    if matched.ArgumentSchema != nil {
        if schema == validateSchema {
            // INVALID_PARAMS path — AuditOnly stamp is correct here
        } else {
            // Defense-in-depth guard for a stray argumentSchema.
            // The comment says: "refusing to forward a request guarded by an
            // unenforced schema (fail closed)".
            resp = capability.EnforceResponse{
                    Code:    capability.ErrCodeEnforcementError,
                    Message: fmt.Sprintf("...refusing to forward... (fail closed)"),
                },
            }
            return resp   // ← defer fires here, stamps AuditOnly = true
        }
    }
```

When `matched.IsAuditOnly()` and `schema == rejectSchema` (resources/read or prompts/get path) and the matched constraint somehow carries an `argumentSchema`, the ENFORCEMENT_ERROR denial is stamped `AuditOnly = true` at function exit. The transport sees `Decision=Deny, AuditOnly=true` and **logs the denial and forwards the request** — the exact opposite of "refusing to forward… fail closed."

## Contrast with `collectObligations`

The analogous engine-bug guard in `pkg/enforcement/engine.go` (`collectObligations`) explicitly calls out the same invariant and deliberately avoids the AuditOnly stamp:

```go
// The deny is a hard ENFORCEMENT_ERROR and is deliberately NOT marked AuditOnly
// even when matched.IsAuditOnly(): an unwired directive is an engine bug, not a
// policy verdict, so "fail closed on ambiguity" wins over "audit never blocks".
return nil, &capability.EnforceResponse{
    RequestID: requestID,
    DecidedAt: now,
        Code:    capability.ErrCodeEnforcementError,
        ...
    },
}
```

`decideTarget`'s stray-argumentSchema path is the same class of engine-bug error and should behave identically.

## Trigger

1. An `enforcement: audit` (audit-only) manifest constraint for a `resource:` or `prompt:` target  
2. That constraint has an `argumentSchema` field set  
3. A request arrives for that resource/prompt target

Condition 2 is impossible through the YAML manifest loader (it rejects `argumentSchema` on non-tool targets at load time). It is reachable by constructing a `capability.Constraint` programmatically without going through `LoadManifest`, which is a supported use case for embedded callers and tests.

## Root Cause

The B-1 refactoring (issue #539) extracted `decideTarget` and used a blanket `defer` to stamp `AuditOnly` on all returns that follow constraint matching. The INVALID_PARAMS path (schema validation failure for tools/call) correctly gets this treatment — audit mode should downgrade an INVALID_PARAMS denial. The stray-argumentSchema ENFORCEMENT_ERROR path should NOT, but the defer cannot distinguish them.

## Suggested Fix

Break the stray-argumentSchema return out of the defer's scope by constructing and returning the response WITHOUT relying on the named return, or add an explicit early return before the defer is registered (for the schema mode check). The simplest fix is to stamp `AuditOnly = false` explicitly on the stray-schema response before returning, overriding the defer:

```go
} else {
    // Defense-in-depth fail-closed guard. Engine bug, not a policy verdict:
    // AuditOnly must NOT be set even when the constraint is audit-only.
    resp = capability.EnforceResponse{
        // ... existing Denial ...
    }
    // Prevent the enclosing defer from stamping AuditOnly on this engine-error
    // path. See collectObligations for the same deliberate choice.
    // The defer fires after this return and would set AuditOnly=true, so we
    // override it by returning a fresh value that won't be mutated by the defer.
        Denial:   resp.Denial,
        // AuditOnly deliberately omitted (false)
    }
}
```

Alternatively, restructure `decideTarget` so the schema-mode check and the defer registration order are swapped, placing the stray-schema guard BEFORE the defer registration:

```go
// schema-mode guard runs BEFORE the audit-mode defer is registered
if matched.ArgumentSchema != nil && schema != validateSchema {
    resp = enforcementErrorResp(...)
    recordAuditModeAntecedent(...)
    return resp   // defer not yet registered; AuditOnly stays false
}
if matched.IsAuditOnly() {
}
```

This is the cleanest fix: it preserves the fail-closed ordering, eliminates the implicit coupling between the defer and every subsequent return, and makes the non-audit-mode stray-schema path a compile-time guarantee rather than a runtime invariant.

## Session ID reminder

`gh issue list --search "rev-20260616T0800Z-c7d91a"`

eunolabs:
## Bugs addressed (session rev-20260616T0800Z-c7d91a)

- Closes #671 — `decideTarget` uses a named return + a blanket `defer` that stamps `AuditOnly=true` on every denial from an audit-mode constraint. The defense-in-depth guard that refuses a `resource:`/`prompt:` constraint carrying a tool-only `argumentSchema` (an `ENFORCEMENT_ERROR`, reachable only via a programmatically-built manifest since the loader rejects it) returned from inside the defer's scope, so it was stamped `AuditOnly=true` — the transport then logged the denial and **forwarded** the request guarded by an unenforced schema, the exact opposite of the documented "fail closed even in audit mode" guarantee. The guard now runs **before** the audit-mode defer is registered, so the `ENFORCEMENT_ERROR` stays `AuditOnly=false` and the request is refused regardless of mode — matching `collectObligations`, which deliberately does not mark its analogous unwired-directive deny `AuditOnly`. The downgradable policy verdicts (action check, `INVALID_PARAMS`) deliberately remain inside the defer's scope.

## Testing

- Added `TestStrayArgumentSchema_AuditMode_FailsClosed` — a programmatically-built `resource:` constraint carrying an `argumentSchema`, in both enforce and audit mode, yields a deny with `ENFORCEMENT_ERROR` and `AuditOnly=false`.
- Existing argument-schema, audit-mode, and resource/prompt tests unchanged and passing.
- `go test -race ./cmd/eunox-mcp/`, `make check-fmt`/`check-license`, `go vet` all pass. Threat model entry 1.34 added.

---


eunolabs:
✅ Ready to merge — session rev-20260616T0800Z-c7d91a bug addressed (#671), self-review complete with no outstanding comments, tests passing (`go test -race ./cmd/eunox-mcp/`, `make check-fmt`/`check-license`/`go vet`). The fail-closed guard now returns before the audit-mode defer (AuditOnly stays false), and the antecedent record is correctly dropped on the refused path. Threat model entry 1.34 added.

---

PR #703, opened 2026-06-16
Fix session rev-20260616T0910Z-b3fe7c bugs

### LINKED ISSUE #677: [rev-20260616T0910Z-b3fe7c] Non-sampling server-initiated requests forwarded with no audit record
**Location:** `cmd/eunox-mcp/http.go`, `handleHTTPUpstreamRequest`, lines 1101–1106

**Reference:** https://github.com/eunolabs/eunox/blob/5b21c2a8666a4626b8aebb6a124fb37985b65e6c/cmd/eunox-mcp/http.go#L1101-L1106

**Severity:** High

**Description:**

`handleHTTPUpstreamRequest` is the handler for server-initiated JSON-RPC requests that arrive from the upstream (e.g. `roots/list`, `elicitation/create`, `logging/setLevel`). When the method is anything other than `sampling/createMessage`, the function calls `sess.broadcastServerRequest(msg)` and returns immediately — no audit record is written.

```go
func (p *HTTPProxy) handleHTTPUpstreamRequest(ctx context.Context, sess *httpSession, msg rpcMsg) {
    if msg.Method != "sampling/createMessage" {
        sess.broadcastServerRequest(msg)  // request forwarded — no audit record
        return
    }
    // ... sampling enforcement + audit below ...
}
```

The architecture doc and `CONTRIBUTING.md` both state that every forwarded request must produce an audit record. The `sampling/createMessage` path immediately below this block writes one; the other methods do not. An operator auditing the OCSF log will see no evidence that a `roots/list` or `elicitation/create` call was ever forwarded, even though the host received and executed it.

**Trigger / repro:**

1. Run the proxy in HTTP mode with an upstream that issues `roots/list` server→client requests.
2. Observe the audit log — no record appears for the forwarded `roots/list` request.
3. Contrast with `sampling/createMessage`, which always produces an audit record.

**Suggested fix:**

Before the early return, emit an audit record with action `"allow"` (and `audit_only=false`). Mirror the pattern used in `handleHTTPToolsCall` for the non-audit-mode path. If the method is entirely unenforced (no PDP decision), use `"allow"` with a note that no policy was evaluated, rather than silently omitting the record, to satisfy the append-only invariant.

**Session:** rev-20260616T0910Z-b3fe7c

### LINKED ISSUE #678: [rev-20260616T0910Z-b3fe7c] Audit-mode sampling path writes deny record but no allow record after forwarding
**Location:** `cmd/eunox-mcp/http.go`, `handleHTTPUpstreamRequest` audit-mode sampling branch, lines 1135–1145

**Reference:** https://github.com/eunolabs/eunox/blob/5b21c2a8666a4626b8aebb6a124fb37985b65e6c/cmd/eunox-mcp/http.go#L1135-L1145

**Severity:** Medium

**Description:**

In audit / wiretap mode, when a `sampling/createMessage` request would be denied under the policy but the proxy is configured to observe-only, `handleHTTPUpstreamRequest` forwards the request via `broadcastServerRequest` and then writes a single `deny` record with `audit_only=true`. It then returns without writing any subsequent record.

```go
// audit mode observe path:
sess.broadcastServerRequest(msg)
if rt.sink != nil {
    rt.sink.Record(ctx, sess.id, "sampling/createMessage", "sampling/createMessage",
        "deny", ..., true /* audit_only */)
}
return   // no allow record written after forward
```

Compare with `handleHTTPToolsCall`, which in the equivalent audit-mode path writes **two** records: first the `deny` (audit-only), then an `allow` record to confirm the request was actually forwarded and processed. A reader of the audit log for the sampling path sees only a `deny`, with no record that the request was in fact forwarded. This breaks the correlation an auditor needs to reconstruct what the LLM actually sent and received.

**Trigger / repro:**

1. Configure the proxy in audit/wiretap mode (e.g. `alwaysAllowPDP`) with a manifest that would deny `sampling/createMessage`.
2. Trigger a `sampling/createMessage` from the upstream.
3. Inspect the audit log — only one record appears (the `deny`/audit-only record). No `allow` record follows, even though the request was forwarded.

**Suggested fix:**

After the `deny` audit record is written and `broadcastServerRequest` has been called, emit a second record with action `"allow"` and `audit_only=true`, mirroring `handleHTTPToolsCall`'s two-record pattern. This ensures every forwarded request has an allow record, consistent with the rest of the proxy.

**Session:** rev-20260616T0910Z-b3fe7c

### LINKED ISSUE #679: [rev-20260616T0910Z-b3fe7c] JWT shorthand conditions always produce string values, so numeric/boolean tool args are never matched
**Location:** `cmd/eunox-mcp/pdp_jwt.go`, `buildV2Constraint` / `evaluateJWTConditions`; cross-reference `pkg/enforcement/handlers.go`, `MatchAllowedValue`

**Reference:** https://github.com/eunolabs/eunox/blob/5b21c2a8666a4626b8aebb6a124fb37985b65e6c/cmd/eunox-mcp/pdp_jwt.go

**Severity:** Medium

**Description:**

JWT v0.2 shorthand conditions are expressed as claim-value suffixes, e.g. a claim `tool:query_db?limit=100` is parsed by `splitV2Claim` into key `limit` and value `100` (as a percent-decoded **string**). `buildV2Constraint` unconditionally builds an `AllowedValuesCondition` whose `Values` slice contains that string:

```go
// buildV2Constraint (pdp_jwt.go)
return &Constraint{
    Conditions: []Condition{{
        Type:   "allowedValues",
        Arg:    cp.key,
        Values: []interface{}{cp.value},  // cp.value is always string
    }},
}, nil
```

`MatchAllowedValue` in `handlers.go` handles `AllowedValuesCondition`. When the actual tool argument is a non-string (e.g. the JSON integer `100` or boolean `true`), the handler reaches the `numericEqual` path, which compares `interface{}(int64(100))` against `interface{}(string("100"))`. These are never equal:

```go
func numericEqual(a, b interface{}) bool {
    // converts both to float64 and compares — but string "100" cannot be
    // parsed as a number when it is already interface{}(string), so the
    // conversion path differs and the comparison returns false
}
```

The result: a JWT that grants `tool:query_db?limit=100` **always denies** a call with `{"limit": 100}` (integer) even though the intent is clearly to allow it. Operators who write JWT claims for tools that accept numeric or boolean parameters get unexpected denials with no obvious diagnostics.

**Trigger / repro:**

1. Issue a JWT with claim `tool:query_db?limit=100`.
2. Call `query_db` with argument `limit` as JSON integer `100`.
3. Observe denial — the JWT condition never matches a non-string argument value.
4. Calling with `limit` as JSON string `"100"` succeeds, confirming the type mismatch.

**Suggested fix:**

In `buildV2Constraint` (or in `evaluateJWTConditions` before building the condition), attempt to parse the percent-decoded string value as a JSON scalar (number, boolean) using `json.Number` / `strconv`. Store the parsed typed value in `Values` so that `MatchAllowedValue`'s `numericEqual` path receives a comparable type. Alternatively, extend `MatchAllowedValue` to coerce string patterns to numbers when the actual argument is numeric, mirroring how glob patterns already special-case strings.

**Session:** rev-20260616T0910Z-b3fe7c

### LINKED ISSUE #680: [rev-20260616T0910Z-b3fe7c] Negative KID cache TTL resets on every presentation of an already-known-absent kid
**Location:** `pkg/capability/jwkscache.go`, `markKIDAbsent`, lines 333–346

**Reference:** https://github.com/eunolabs/eunox/blob/5b21c2a8666a4626b8aebb6a124fb37985b65e6c/pkg/capability/jwkscache.go#L333-L346

**Severity:** Low

**Description:**

`markKIDAbsent` records the time a `kid` was first found absent from the JWKS, so `kidRecentlyAbsent` can suppress forced refreshes for that kid for `negativeKIDTTL` (30 s). The implementation unconditionally overwrites the timestamp even when the kid is already in the map:

```go
    c.negMu.Lock()
    defer c.negMu.Unlock()
    now := c.now()
    for k, at := range c.negKIDs {
        if now.Sub(at) >= negativeKIDTTL { delete(c.negKIDs, k) }
    }
    if _, ok := c.negKIDs[kid]; !ok && len(c.negKIDs) >= negativeKIDMaxLen {
        return
    }
    c.negKIDs[kid] = now   // always overwrites, resetting the clock
}
```

Every call to `markKIDAbsent` for an already-cached kid resets its timestamp to `now`, extending the effective suppress window by another 30 s. A client that continuously presents tokens with a stale or rotated `kid` (e.g. a misconfigured client that cached an old token) will never allow that `kid` to leave the negative cache, even after the JWKS has been updated to include it. The intended behaviour is that the suppress window is anchored to the **first** absent observation, not the last.

This is most harmful after a key rotation: if the old `kid` was absent-cached and the new JWKS now reuses the same `kid` value (e.g. a rolling key naming scheme), the proxy will refuse forced refreshes for that `kid` indefinitely as long as any client keeps presenting old tokens.

**Trigger / repro:**

1. Configure `JWKSCache` with a `negativeKIDTTL` of 30 s.
2. Present tokens with `kid="k1"` (absent from JWKS) repeatedly, one per second.
3. After 31 s, the key rotation puts `k1` into the live JWKS.
4. Subsequent tokens with `k1` are still suppressed — `kidRecentlyAbsent` returns true because `markKIDAbsent` kept refreshing the timestamp.

**Suggested fix:**

Guard the assignment with an existence check so that only **new** absent kids get timestamped; existing entries retain their original insertion time:

```go
if _, ok := c.negKIDs[kid]; !ok {
        return
    }
    c.negKIDs[kid] = now
}
```

**Session:** rev-20260616T0910Z-b3fe7c

### LINKED ISSUE #681: [rev-20260616T0910Z-b3fe7c] Multi-maxCalls TOCTOU allows concurrent requests to exceed the configured call limit
**Location:** `pkg/enforcement/engine.go`, `runConditions`, lines 519–544

**Reference:** https://github.com/eunolabs/eunox/blob/5b21c2a8666a4626b8aebb6a124fb37985b65e6c/pkg/enforcement/engine.go#L519-L544

**Severity:** Medium

**Description:**

`runConditions` uses a two-pass strategy when a constraint has multiple `maxCalls` conditions: a non-committing "check-only" pass to confirm all counters have headroom, followed by a committing pass that actually increments them. The gap between the two passes is not guarded by any lock or compare-and-swap:

```go
if len(deferred) > 1 {
    checkCtx := withMaxCallsCheckOnly(ctx)
    for _, cond := range deferred {
        if deny := e.evalCondition(checkCtx, ...); deny != nil { return deny }
    }
}
// commit pass — not atomic with the check above
for _, cond := range deferred {
    if deny := e.evalCondition(ctx, ...); deny != nil { return deny }
}
```

Two concurrent requests for the same session with the same capability can both pass the check-only pass (both see counter < limit), then both enter the commit pass and each increment the counter. The result is that the combined committed count exceeds the limit by the number of concurrent racing requests. For a `maxCalls: 1` guard, two simultaneous requests both succeed where only one should.

The code contains an inline comment acknowledging this as a TOCTOU, but it is not reflected in any public issue, the threat model, or the capability-manifest guide, so operators using multiple `maxCalls` conditions on a single constraint have no signal that the limit is advisory under concurrent load.

**Trigger / repro:**

1. Define a capability with two `maxCalls` conditions (different windows or scopes) both with limit 1.
2. Fire two concurrent requests for the same session targeting that capability.
3. Both requests succeed; the per-window counter ends at 2, exceeding the limit.

**Suggested fix:**

Options in increasing robustness:
- **Document only (minimal):** Expand the inline TOCTOU comment into a public note in the capability-manifest guide and threat model, warning operators that multiple `maxCalls` conditions on a single constraint are subject to a small concurrent-request race window.
- **Atomic multi-counter CAS (preferred):** Extend the `CallCounter` interface with a `CheckAndIncrementAll(keys []string, limits []int64) error` method so the backend can perform the multi-counter check-and-commit atomically (Redis `MULTI`/`EXEC` or Lua script; in-memory via a single mutex covering all keys in one operation).
- **Single-condition workaround:** Document that the TOCTOU window collapses to zero when there is exactly one `maxCalls` condition (the existing single-condition path goes straight to commit without a check pass), and recommend operators avoid multiple `maxCalls` on the same constraint until the atomic path lands.

**Session:** rev-20260616T0910Z-b3fe7c

eunolabs:
## Bugs addressed (session rev-20260616T0910Z-b3fe7c)

- Closes #677 — Non-sampling server-initiated requests (`roots/list`, `elicitation/create`, …) were forwarded with **no audit record**, violating the "every forwarded request produces an audit record" invariant. `handleHTTPUpstreamRequest` (and the stdio `handleUpstreamRequest` mirror — same root cause) now emit an `allow` record before forwarding (no policy evaluated → empty denial code/condition type).
- Closes #678 — In route audit mode, a would-be-denied `sampling/createMessage` was forwarded but only a `deny` record was written, not the following `allow`. Both handlers now write the second `allow` record, matching `enforcedForward`'s two-record (deny, then allow) pattern so an auditor can correlate that the forwarded request was actually processed.
- Closes #679 — JWT v0.2 shorthand condition values are always percent-decoded **strings**, so `tool:query_db?limit=100` could never match the JSON number `100` (a string allowed-value is treated as a glob, which cannot match a non-string argument). `buildV2Constraint` now keeps the raw string (so the string `"100"` and glob patterns still match) **and** appends the parsed JSON scalar when the value parses as a number or boolean, so the native-typed argument matches via `MatchAllowedValue`'s exact/numeric path.
- Closes #680 — `markKIDAbsent` overwrote the negative-cache timestamp on every presentation, sliding the `negativeKIDTTL` window forward indefinitely; a client repeatedly presenting a stale `kid` could pin it in the negative cache even after a JWKS update re-added it. The window is now anchored to the **first** absent observation (an already-tracked `kid` keeps its original timestamp).
- Closes #681 — Documented the multi-`maxCalls` concurrent-request TOCTOU. A single `maxCalls` is admitted atomically, but two-or-more `maxCalls` on one constraint use a non-atomic two-phase (check-all, then commit-all) admission, so concurrent same-session requests can exceed the limit by the number of racers. Expanded the inline comment in `runConditions` and documented the limitation in `docs/capability-manifest-guide.md` and `docs/threat-model-mcp.md`, recommending a single `maxCalls` per constraint for a hard ceiling under concurrent load. (This is the issue's explicitly-sanctioned "document only" option; the atomic multi-bucket check-and-increment is noted as deferred future work since it changes the security-critical counter backend.)

## Testing
- New regression tests:
  - `TestHTTPHandleUpstreamRequest_NonSampling_WritesAuditRecord`, `TestStdioHandleUpstreamRequest_NonSampling_WritesAuditRecord` (#677)
  - `TestHTTPHandleUpstreamRequest_SamplingAuditMode_RecordsDenyThenAllow`, `TestStdioHandleUpstreamRequest_SamplingAuditMode_RecordsDenyThenAllow` (#678)
  - `TestJWTShorthandValuesMatchTypedArgs` (numeric/boolean/string/glob cases) (#679)
  - `TestJWKSCache_MarkKIDAbsent_AnchorsToFirstObservation` (#680)
- `pkg/` coverage well above the 80% CI gate.

---


eunolabs (commented):
## Self-review (round 1)

Reviewed the full diff for: each fix resolving its issue, regressions, missed edge cases, test strength, and convention consistency.

**Finding (resolved in cfe1268):** `docs/threat-model-mcp.md` §6.1 enumerates exactly which methods produce audit records. The #677 fix adds an `allow` record for forwarded non-sampling server-initiated requests (`roots/list`, `elicitation/create`, …), but §6.1's list was not updated — CLAUDE.md requires a docs update for behavior changes. Updated §6.1 to include the new coverage.

**Verified, no action needed:**
- Both transports dispatch notifications separately (`isNotification()` → broadcast, no audit) and only route `isRequest()` to the handler, so the #677 `allow` record fires only for genuine server-initiated *requests* — no spurious per-notification records.
- #677 uses `auditOnly = rt.audit`/`p.audit` (consistent with the `*/list` allow records); #678's second `allow` is in a branch already gated on audit mode, matching `enforcedForward`'s deny-then-allow pattern.
- #679 keeps the raw string (preserving glob `#388` semantics and string-arg matches) and only *adds* a typed scalar for JSON number/boolean values; it never relaxes argument-name matching. Empty/leading-zero/whitespace-trailing values fail `json.Unmarshal` and stay string-only.
- #680 anchors the window to the first observation; the prune sweep still bounds the map, and a kid is re-timestamped only after it expires (one retry per TTL window).
- #681 is the issue's explicitly-sanctioned "document only" option; the atomic multi-bucket fix is noted as deferred.

Round-2 re-review of the updated diff produced no new findings. `go test -race ./...`, `go vet`, `golangci-lint run` (0 issues), `make check-license`, `gofmt -l` all clean.

---

eunolabs:
✅ Ready to merge — all session rev-20260616T0910Z-b3fe7c bugs addressed (#677, #678, #679, #680, #681), self-review complete with no outstanding comments, tests passing (`go test -race ./...` locally + CI Test/Lint/Build and all demo/OPA/security checks green).

---

PR #705, opened 2026-06-16
Fix session rev-20260616T1010Z-eace32 bugs

### LINKED ISSUE #682: [rev-20260616T1010Z-eace32] cnf.jwk and cnf.kid sender-constraint methods silently ignored, accepting sender-constrained tokens without proof-of-possession
**Location:** `pkg/capability/token.go:135-137`, `pkg/capability/jwks.go:234-236`

**Reference:** https://github.com/eunolabs/eunox/blob/5b21c2a8666a4626b8aebb6a124fb37985b65e6c/pkg/capability/token.go#L135-L137

**Severity:** High

**Description:**

The `Confirmation` struct only declares a `JKT` field:

```go
// pkg/capability/token.go:135-137
type Confirmation struct {
    JKT string `json:"jkt"`
}
```

RFC 7800 defines three proof-of-possession methods for the `cnf` claim: `jkt` (thumbprint), `jwk` (embedded public key), and `kid` (key reference). The sender-constraint guard in `VerifyToken` only checks `cnf.jkt`:

```go
// pkg/capability/jwks.go:234-236
if p.Confirmation != nil && p.Confirmation.JKT != "" {
    return nil, fmt.Errorf("validate claims: sender-constrained token (cnf.jkt) requires proof-of-possession..."), nil
}
```

Because Go's `json.Unmarshal` silently discards unknown struct fields, a JWT whose `cnf` claim carries `jwk` or `kid` (but not `jkt`) is decoded into `Confirmation{JKT: ""}`. The guard evaluates to false (JKT is empty) and the token is **accepted as a plain bearer token** without any proof-of-possession check.

This violates the fail-closed principle stated in CONTRIBUTING.md: "On any ambiguity… deny." A token explicitly declaring a sender constraint via `cnf.jwk` or `cnf.kid` is not ambiguous — it must not be honoured without proof-of-possession verification. Accepting it silently downgrades a security mechanism and can allow a captured sender-constrained token to be replayed by any bearer.

**Trigger / repro:**

1. Issue a JWT with `"cnf": {"jwk": {"kty": "EC", "crv": "P-256", ...}}` (a sender constraint using the embedded-JWK form).
2. Present that token to a `JWKSClient`-protected endpoint (no DPoP proof).
3. The token passes `VerifyToken` because `p.Confirmation.JKT == ""`, so the cnf guard is not triggered.
4. Contrast: the same token with `"cnf": {"jkt": "...thumbprint..."}` is correctly rejected.

**Suggested fix:**

Replace the single-field `Confirmation` struct with one that captures all three RFC 7800 PoP methods, then reject any token whose `cnf` claim contains ANY non-empty proof-of-possession method:

```go
type Confirmation struct {
    JKT string           `json:"jkt"` // RFC 7800 §3.3 thumbprint
    JWK *json.RawMessage `json:"jwk"` // RFC 7800 §3.2 embedded key
    KID string           `json:"kid"` // RFC 7800 §3.4 key reference
}

// In VerifyToken:
if p.Confirmation != nil &&
    (p.Confirmation.JKT != "" || p.Confirmation.JWK != nil || p.Confirmation.KID != "") {
    return nil, fmt.Errorf("validate claims: sender-constrained token requires " +
        "proof-of-possession (cnf.jkt/cnf.jwk/cnf.kid), which the proxy does not yet verify; " +
        "refusing to accept as a plain bearer token (fail closed)"), nil
}
```

This ensures all sender-constraint forms are rejected with a clear error until a PoP verification path is implemented, consistent with the existing `cnf.jkt` handling (issue #536).

**Session:** rev-20260616T1010Z-eace32

### LINKED ISSUE #683: [rev-20260616T1010Z-eace32] VerifyToken does not retry after signature failure for kid-bearing tokens, causing 5-minute outage on key rotation with stable kid
**Location:** `pkg/capability/jwks.go:131-321`

**Reference:** https://github.com/eunolabs/eunox/blob/5b21c2a8666a4624b8aebb6a124fb37985b65e6c/pkg/capability/jwks.go#L131-L321

**Severity:** Medium

**Description:**

`VerifyToken` has two forced-refresh retry paths:

1. **Kid-miss path** (line 132–145): when `FindKeys(cachedKeys, kid)` returns zero keys, `ForceRefreshForKID` is called and the lookup retried. This handles the common rotation pattern where a new kid appears.
2. **Kid-less path** (line 296–319): when `kid == ""` and every cached key fails signature verification, `ForceRefreshForVerify` is called to pull a freshly-rotated key.

There is **no retry path** for the case where `kid != ""`, matching keys ARE found in the cache, but ALL of them fail signature verification. The code falls through directly to the terminal error:

```go
payload, terminal, lastErr := tryKeys(matchingKeys)
// ...
// kid != "" → the kid-less ForceRefreshForVerify branch is skipped
```

The inline comment rationalises this as "A kid-bearing token already went through the kid-miss forced refresh above, so its failure is terminal." That reasoning is only valid when `matchingKeys` was empty on the first lookup (the kid-miss path ran). When `matchingKeys` is non-empty — keys exist in the cache for that kid — no forced refresh happened at all.

This causes a silent outage when an issuer rotates key material while reusing the same `kid` string (a common practice in rolling-update schemes where the kid is a slot name like `"current"`):

1. JWKS cache holds Key-A tagged `kid="current"`.
2. Issuer rotates to Key-B, same `kid="current"`. Old Key-A is removed from the JWKS.
3. Token signed with Key-B, `kid="current"`.
4. `matchingKeys = [Key-A]` (cache hit, non-empty).
5. `tryKeys([Key-A])` → signature verification fails.
6. No retry, no `ForceRefreshForKID` call. Error returned to caller.
7. All tokens from this issuer are rejected for up to the default 5-minute cache TTL.

**Trigger / repro:**

1. Seed the `JWKSCache` with a JWKS containing Key-A, `kid="slot-0"`.
2. Rotate the JWKS to Key-B with the same `kid="slot-0"`, removing Key-A.
3. Call `VerifyToken` with a token signed by Key-B, `kid="slot-0"` in the header.
4. Observe "verify signature" error — no forced refresh attempted, even though the cache holds a stale key.
5. Wait for the cache TTL to elapse; calls succeed again.

**Suggested fix:**

After `tryKeys(matchingKeys)` returns `(nil, nil, lastSig)` for a kid-bearing token (`kid != ""`), call `ForceRefreshForKID(ctx, kid)` and retry against the refreshed key set before returning the terminal error. Guard it with the existing suppression logic in `ForceRefreshForKID` so a flood of genuinely-bad-signature tokens does not amplify into unbounded upstream fetches:

```go
// Every matching key failed sig verification for a kid-bearing token.
// The cached keys may be stale (key rotation with a stable kid reuses the
// same kid string for new key material). Attempt one forced refresh.
if kid != "" {
    refreshed, ferr := c.cache.ForceRefreshForKID(ctx, kid)
    if ferr == nil {
        if payload, terminal, lastSig2 := tryKeys(FindKeys(refreshed, kid)); terminal != nil {
            return nil, terminal
        } else if payload != nil {
            return payload, nil
        } else if lastSig2 != nil {
            lastErr = lastSig2
        }
    }
}
```

**Session:** rev-20260616T1010Z-eace32

### LINKED ISSUE #684: [rev-20260616T1010Z-eace32] Audit HMAC signing key written without trailing newline, breaking shell-based key rotation
**Location:** `cmd/eunox-mcp/audit.go:1968-2053` (`generateAndPersistAuditKey`)

**Reference:** https://github.com/eunolabs/eunox/blob/5b21c2a8666a4626b8aebb6a124fb37985b65e6c/cmd/eunox-mcp/audit.go#L1968-L2053

**Severity:** Medium

**Description:**

`generateAndPersistAuditKey` writes the 64-character hex-encoded key to the key file without a trailing newline:

```go
encoded := make([]byte, hex.EncodedLen(len(key))) // 64 bytes
hex.Encode(encoded, key)
// ...
if _, err := tmp.Write(encoded); err != nil { ... }
```

The key file format documented in `loadOrCreateAuditKeys` is "one 64-hex-character key per line; the first key line is the active signing key; further key lines are retired keys." Key rotation therefore requires the operator to add the new active key as the first line, shifting the current key to a retired line.

When the generated key file lacks a trailing newline, a standard Unix append-then-move rotation (`echo "newkey" >> keyfile`) produces a file where the old and new keys are concatenated **on the same line** (128 hex chars with no separator). `parseAuditKeys` then calls `hex.Decode` on the 128-character line, which decodes to 64 bytes. The length check `if n != 32` then fails and the proxy refuses to start:

```
audit key file "~/.eunox/audit.key": a key decoded to 64 bytes (want 32);
file may be truncated or corrupt
```

This is a latent operational bug. The proxy starts and runs correctly with a single key (no trailing newline required), so the problem is invisible until a human or automation attempts to rotate the key using the most natural shell idiom.

**Trigger / repro:**

```bash
# Simulate initial state: key file as generated (no trailing newline)
printf '%s' "$(python3 -c 'import os,binascii; print(binascii.hexlify(os.urandom(32)).decode())')" > ~/.eunox/audit.key

# Attempt key rotation using shell append
NEWKEY=$(python3 -c 'import os,binascii; print(binascii.hexlify(os.urandom(32)).decode())')
echo "$NEWKEY" >> ~/.eunox/audit.key
# File now contains 128 hex chars before the newline — parseAuditKeys fails.
eunox-mcp proxy  # → "a key decoded to 64 bytes (want 32)"
```

**Suggested fix:**

Append a newline when writing the key so that the file conforms to the one-key-per-line contract and standard shell append operations are safe:

```go
// In generateAndPersistAuditKey, replace:
if _, err := tmp.Write(encoded); err != nil { ... }

// With:
if _, err := tmp.Write(append(encoded, '\n')); err != nil { ... }
```

`parseAuditKeys` already calls `strings.TrimSpace` on each line, so the trailing newline is consumed correctly on read. No migration is needed for existing key files — `parseAuditKeys` handles the no-newline form — but future generated files will be safe to append to.

**Session:** rev-20260616T1010Z-eace32

### LINKED ISSUE #685: [rev-20260616T1010Z-eace32] audit-verify INVALID output susceptible to newline injection via attacker-controlled audit record fields
**Location:** `cmd/eunox-mcp/audit.go:1736-1739` (`verifyAuditLog`)

**Reference:** https://github.com/eunolabs/eunox/blob/5b21c2a8666a4626b8aebb6a124fb37985b65e6c/cmd/eunox-mcp/audit.go#L1736-L1739

**Severity:** Low

**Description:**

When `audit-verify` prints an `INVALID` line for a record whose HMAC does not match, it emits the `request_id`, `session_id`, and `target` fields from the decoded record verbatim using `%s`:

```go
// cmd/eunox-mcp/audit.go:1737-1739
_, _ = fmt.Fprintf(out, "INVALID  seq=%d request_id=%s session_id=%s target=%s\n",
    rec.Seq, rec.RequestID, rec.SessionID, rec.Target)
```

These fields are attacker-controlled:

- `target` derives from the MCP tool name, resource URI, or prompt name supplied by the upstream server — bounded to 8 KiB by `boundEnvelopeField` but **not newline-sanitized**.
- `session_id` is the client-supplied `Mcp-Session-Id` header value — bounded to 256 bytes, also not newline-sanitized.
- `request_id` is process-generated and safe, but printed alongside the others.

A tool name containing a literal newline (`\n`) is stored as `\n` (the JSON escape sequence) in the JSONL record. When `json.Unmarshal` decodes the record, the field value contains a literal newline character. Printing it with `%s` inserts the newline into the output.

An adversary who controls a tool name (i.e., controls an upstream MCP server or can influence its tool list) can craft a target like:

```
legit-tool\nINVALID  seq=99 request_id=req-forged session_id= target=no-such-tool
```

The resulting `audit-verify` output contains a spurious second `INVALID` line that appears structurally identical to a real finding. A SIEM or CI check that parses `audit-verify` output line-by-line to count or categorise `INVALID` records would be confused by injected lines.

The HMAC chain is not affected — the injected content still resides in a genuinely-invalid record — but the **display output** of the forensic tool is corrupted. In incident response contexts this can mislead investigators or cause automated alert systems to miscount integrity failures.

**Trigger / repro:**

1. Register a tool whose name contains `\nCHAIN BREAK at seq 1:` with an upstream MCP server.
2. Make a call through the proxy; the audit record's `target` field contains the embedded newline.
3. Tamper with (or corrupt) that record so that it fails HMAC verification.
4. Run `eunox-mcp audit-verify`; observe the spurious `CHAIN BREAK` diagnostic line in the output.

**Suggested fix:**

Sanitize the printed fields by replacing control characters (at minimum `\r`, `\n`, `\t`) with a safe placeholder before interpolating them. A concise helper suffices:

```go
func sanitizeAuditField(s string) string {
    return strings.Map(func(r rune) rune {
        if r == '\n' || r == '\r' || r == '\t' {
            return ' '
        }
        return r
    }, s)
}

// In verifyAuditLog:
_, _ = fmt.Fprintf(out, "INVALID  seq=%d request_id=%s session_id=%s target=%s\n",
    rec.Seq,
    sanitizeAuditField(rec.RequestID),
    sanitizeAuditField(rec.SessionID),
    sanitizeAuditField(rec.Target))
```

Apply the same sanitization to the `CHAIN BREAK` and `SEQ GAP` output lines that also print `rec.Seq` — those are safe (uint64) but the pattern should be consistent across all output paths. Alternatively, switch to `%q` for the string fields so they are always JSON-quoted in the output, making injection visually obvious.

**Session:** rev-20260616T1010Z-eace32

eunolabs:
## Bugs addressed (session rev-20260616T1010Z-eace32)

- Closes #682 — **(High)** The `Confirmation` (`cnf`) struct declared only `jkt`, so `json.Unmarshal` silently dropped the other two RFC 7800 proof-of-possession methods (`cnf.jwk`, `cnf.kid`): a token sender-constrained via `jwk`/`kid` decoded to an empty `jkt` and was accepted as a plain bearer token, defeating the #536 fail-closed guard for those forms. `Confirmation` now captures all three methods (`IsSenderConstrained()`), and `VerifyToken` rejects a token declaring ANY of them as a terminal error.
- Closes #683 — **(Medium)** `VerifyToken` retried a forced JWKS refresh only on a kid-miss or a kid-less signature failure; a kid-**bearing** token whose cached keys all failed (an issuer rotating key material while reusing the same `kid`, e.g. a rolling-update slot name) took no retry and was rejected for up to a full cache TTL. It now forces one (rate-limited) refresh for the `kid` and retries, guarded by a `refreshedForKid` flag so a token that already refreshed via the kid-miss path is not refetched. The kid-less path is unchanged.
- Closes #684 — **(Medium)** `generateAndPersistAuditKey` wrote the hex key with no trailing newline, so a shell-append key rotation (`echo "$NEWKEY" >> keyfile`) concatenated the keys onto one 128-hex-char line that `parseAuditKeys` rejected (decodes to 64 bytes, want 32), preventing the proxy from starting. The key is now written with a trailing newline, conforming to the documented one-key-per-line format. Existing files still parse (no migration).
- Closes #685 — **(Low)** `audit-verify`'s `INVALID` line printed the attacker-influenceable `target` (upstream tool/resource/prompt name) and `session_id` (client `Mcp-Session-Id` header) fields verbatim, so a field with an embedded newline could inject a spurious `INVALID`/`CHAIN BREAK` line into the output and mislead a SIEM parsing it line by line. A `sanitizeAuditField` helper now replaces control characters (newlines, CR, tab, terminal escapes) with spaces before interpolation.

## Testing
- New regression tests:
  - `TestJWKSClient_VerifyToken_AllSenderConstraintFormsRejected` (cnf.jwk + cnf.kid) (#682)
  - `TestJWKSClient_VerifyToken_StableKidRotationRetry` (#683)
  - `TestGenerateAndPersistAuditKey_TrailingNewlineEnablesShellAppendRotation` (#684)
  - `TestSanitizeAuditField` + `TestVerifyAuditLog_SanitizesControlCharsInInvalidLine` (#685)
  - Existing `TestJWKSClient_VerifyToken_SenderConstrainedRejected` (cnf.jkt) and the `Confirmation` round-trip test still pass (the struct now uses `omitempty`, preserving the jkt-only wire form).
- Threat-model changelog updated (entry 1.37) documenting all four fixes.

---


eunolabs (commented):
## Self-review (round 1)

Reviewed the full diff for: each fix resolving its issue, regressions, missed edge cases, test strength, and convention consistency.

**Substantive finding (resolved in b96aed1):** The #683 kid-bearing retry as first written (`pkg/capability/jwks.go`) called `ForceRefreshForKID`, whose suppression only fires for an *absent* kid (`kidRecentlyAbsent`). For a kid that is **present but stale** (the exact stable-kid-rotation case #683 targets), the kid is never marked absent, so the suppression never triggers and `ForceRefresh` has no sequential rate limit — a flood of bad-signature tokens carrying a valid kid would force one JWKS fetch *each*, an amplification the issue explicitly warned against. Fixed by routing both the kid-less and kid-bearing retries through `ForceRefreshForVerify`, which marks a sentinel when a forced fetch returns an unchanged set and suppresses further forced fetches for `negativeKIDTTL`, while a real rotation (changed set) is never suppressed. Added `TestJWKSClient_VerifyToken_BadSignatureFloodDoesNotAmplifyFetches` asserting the fetch count stays bounded under a 20-token bad-signature flood.

**Verified, no action needed:**
- #682: `IsSenderConstrained()` handles a nil receiver; the call site `p.Confirmation.IsSenderConstrained()` is safe when `cnf` is absent. `omitempty` on the new tags preserves the jkt-only wire form (round-trip test passes). The only `cnf` consumer in the codebase is this verifier — the IdP MCP-JWT path uses `mcp.*` claims, not `cnf`, so the fix is correctly scoped.
- #683: a kid-bearing token that already refreshed via the kid-miss path (`refreshedForKid == true`) is not retried again; the kid-less path is behaviorally unchanged (same `ForceRefreshForVerify`, same `FindKeys` selection).
- #684: `hex.Encode` writes 64 bytes into the 65-byte buffer, index 64 set to `\n`; `parseAuditKeys` trims, so existing newline-less files still parse.
- #685: the INVALID line was the only `%s`-with-attacker-data output path; CHAIN BREAK uses `%q` and SEQ GAP prints only integers.

Round-2 re-review of the updated diff produced no new findings. `go test -race ./...`, `go vet`, `golangci-lint run` (0 issues), `make check-license`, `gofmt -l` all clean.

---

eunolabs:
✅ Ready to merge — all session rev-20260616T1010Z-eace32 bugs addressed (#682, #683, #684, #685), self-review complete (one substantive finding — the #683 fetch-amplification — found and fixed in b96aed1, re-review clean), tests passing (`go test -race ./...` locally + CI Test/Lint/Build and all demo/OPA/security checks green).

---

PR #706, opened 2026-06-16
Fix session rev-20260616T1110Z-7bb934 bugs

### LINKED ISSUE #687: [rev-20260616T1110Z-7bb934] JWT custom claims silently lose precision for integers > 2^53
**Location:** `cmd/eunox-mcp/pdp_jwt.go`, `ValidateToken` / `UnsafeClaimsWithoutVerification` call

**Reference:** https://github.com/eunolabs/eunox/blob/5b21c2a8666a4626b8aebb6a124fb37985b65e6c/cmd/eunox-mcp/pdp_jwt.go#L373-L380

**Severity:** Medium-High

## Description

`tok.UnsafeClaimsWithoutVerification(&payload, &rawClaims)` uses the standard `encoding/json` decoder under the hood. The `rawClaims` value (which backs `idpJWTPayload.Extra` and is passed as `input.claims` to OPA/Cedar policy evaluators) is decoded into `map[string]interface{}`. Go's `encoding/json` decodes all JSON numbers as `float64` by default when the target type is `interface{}`.

Any custom JWT claim carrying an integer larger than 2^53 (e.g., `user_id`, `tenant_id`, `account_id` in enterprise IdP tokens) will be silently truncated to the nearest representable `float64`. For example, `9007199254740993` (2^53 + 1) becomes `9007199254740992`.

## Trigger / Repro

1. Issue a JWT with a custom claim `"tenant_id": 9007199254740993`.
2. Write a Cedar/OPA policy condition that checks `input.claims.tenant_id == 9007199254740993`.
3. The condition silently evaluates to `false` because `rawClaims["tenant_id"]` is `float64(9007199254740992)`.
4. Depending on the policy, this produces a silent allow or deny — both incorrect.

The fail-closed invariant means a deny here is "safe" but misleading; an allow (e.g., where the policy grants access only to `tenant_id == 9007199254740993`) is a security bypass.

## Suggested Fix

Decode `rawClaims` using `json.NewDecoder` with `UseNumber()` so large integers are preserved as `json.Number` strings rather than lossy `float64`:

```go
// instead of the UnsafeClaimsWithoutVerification path for rawClaims,
// decode the raw JWT payload bytes with UseNumber:
dec := json.NewDecoder(bytes.NewReader(rawPayloadBytes))
dec.UseNumber()
if err := dec.Decode(&rawClaims); err != nil {
    return nil, fmt.Errorf("decoding raw claims: %w", err)
}
```

Policy evaluators that receive `json.Number` values need to handle them accordingly (OPA's Go SDK already does; Cedar's SDK should be verified).

**Session:** rev-20260616T1110Z-7bb934

### LINKED ISSUE #688: [rev-20260616T1110Z-7bb934] readLastAuditLine silently swallows I/O errors, causing audit chain reset
**Location:** `cmd/eunox-mcp/audit.go`, `readLastAuditLine`

**Reference:** https://github.com/eunolabs/eunox/blob/5b21c2a8666a4626b8aebb6a124fb37985b65e6c/cmd/eunox-mcp/audit.go#L636-L672

**Severity:** Medium

## Description

`readLastAuditLine` returns `""` (empty string) for **both** of the following cases:

1. The audit file is genuinely empty (no records yet).
2. An I/O read error occurred (e.g., transient `EINTR`, `EIO`, NFS `ESTALE`, permission change mid-read).

The caller in `openAuditSink` treats `""` as "file is empty — try rotated siblings; if none exist, start the chain from genesis." It has no way to distinguish the two cases.

If a transient I/O error occurs on a non-empty audit file and no rotated siblings are available, the chain silently resets: `seq` restarts at 0, `prevHMAC` is `""`, and the new record chain is unlinked from all prior records. This creates an undetected gap in the audit trail.

By contrast, every other integrity-breaking event (`VerifyRecord` failure, parse failure, tail parse failure on startup) writes a signed, in-band integrity marker (`tail_parse_failure` or `tail_integrity_failure`) before the new chain begins. I/O errors during resume are the only path that resets the chain without any in-band marker.

## Trigger / Repro

1. Write at least one audit record so the `.jsonl` file is non-empty.
2. Simulate an I/O error on the second-to-last `ReadAt`/`Read` call inside `readLastAuditLine` (e.g., mock the underlying file, inject fault at filesystem level with a faulty NFS mount).
3. `openAuditSink` will start the new session as if the file is empty: `seq=0`, `prevHMAC=""`.
4. `eunox-mcp audit-verify` will report a gap/chain break at the boundary, but the boundary itself has no signed marker explaining why.

## Suggested Fix

Return a distinct sentinel or error from `readLastAuditLine` to indicate a read error vs. a genuinely empty file:

```go
// return ("", false, nil) for empty; ("", false, err) for I/O error; (line, true, nil) for success
func readLastAuditLine(f *os.File) (string, bool, error) { ... }
```

In `openAuditSink`, when the function returns an I/O error, write a signed `chain_resume_failed` integrity marker before beginning the new chain — consistent with the existing `tail_parse_failure` path — so `audit-verify` can distinguish a legitimate genesis from an error-induced reset.

**Session:** rev-20260616T1110Z-7bb934

### LINKED ISSUE #689: [rev-20260616T1110Z-7bb934] ValidateToken gives misleading error when mcp claim is absent vs wrong version
**Location:** `cmd/eunox-mcp/pdp_jwt.go`, `ValidateToken`, version check

**Reference:** https://github.com/eunolabs/eunox/blob/5b21c2a8666a4626b8aebb6a124fb37985b65e6c/cmd/eunox-mcp/pdp_jwt.go#L415-L422

**Severity:** Low

## Description

When a JWT is issued without an `mcp` claim at all (field entirely absent), `payload.MCP` is the zero value of `mcpClaimSet`, so `payload.MCP.Version` is `""`. The version check:

```go
if payload.MCP.Version != mcpClaimVersion {
    return nil, fmt.Errorf("unsupported mcp claim version %q (want %q)", payload.MCP.Version, mcpClaimVersion)
}
```

produces the error:

```
unsupported mcp claim version "" (want "0.2")
```

This is misleading for operators: the message implies there is an `mcp` claim present with an unrecognised version string, not that the claim is absent entirely. In practice, the most common cause of this error during IdP integration is a misconfigured JWT template that omits the `mcp` block entirely — and the current message sends operators in the wrong direction (looking for a version field rather than the missing claim block).

The error surfaces in audit records and operator logs, so it affects debuggability.

## Trigger / Repro

Issue a JWT where the `mcp` key is absent from the payload (not just `mcp.v` — the entire `mcp` object). `ValidateToken` will return `unsupported mcp claim version "" (want "0.2")`.

## Suggested Fix

Add an explicit presence check before the version check:

```go
if payload.MCP.Version == "" {
    return nil, fmt.Errorf("jwt is missing required mcp capability claim (expected mcp.v=%q)", mcpClaimVersion)
}
if payload.MCP.Version != mcpClaimVersion {
    return nil, fmt.Errorf("unsupported mcp claim version %q (want %q)", payload.MCP.Version, mcpClaimVersion)
}
```

This gives operators an immediately actionable error message pointing to the missing claim block.

**Session:** rev-20260616T1110Z-7bb934

eunolabs:
## Bugs addressed (session rev-20260616T1110Z-7bb934)

- Closes #687 — **(Medium-High)** The raw top-level JWT claim map exposed to policy evaluators as `input.claims` was decoded via go-jose's default decoder (every JSON number → `float64`), so a custom claim carrying an integer above 2^53 (`tenant_id`, `account_id`, … in enterprise IdP tokens) was silently rounded — `9007199254740993` (2^53+1) became `9007199254740992` — which could flip an OPA/Cedar equality check and allow or deny incorrectly. `ValidateToken` now re-decodes the verified payload segment with `json.Number` (`decodeJWTClaimsPreservingNumbers`) so every integer is preserved exactly. The canonical identity fields (`sub`/`iss`/`task_id`/`agent_id`) are unaffected, and string/array claims pass through unchanged.
- Closes #688 — **(Medium)** `readLastAuditLine` returned `""` for both a genuinely empty log and an I/O read error, so a transient `EIO`/`ESTALE`/permission failure on a **non-empty** log silently reset the tamper-evident chain (`seq=0`, `prev_hmac=""`) with no in-band evidence — the only chain-reset path lacking a signed marker. It now returns `(string, error)` (a *missing* file stays the normal empty case); when a tail read errors, `openAuditSink` writes a signed `AUDIT_INTEGRITY_FAILURE` marker (`"kind":"chain_resume_failed"`, carrying the underlying error as `reason`) before restarting from genesis — symmetric with the existing parse-failure and HMAC-mismatch markers.
- Closes #689 — **(Low)** A token with no `mcp` claim block left `payload.MCP.Version == ""`, so the version check reported `unsupported mcp claim version ""` — misdirecting operators to a version field rather than the absent claim block (the most common IdP-template misconfiguration). An explicit presence check now returns an actionable `missing the required mcp capability claim` error before the version-mismatch check.

## Testing
- New regression tests:
  - `TestJWTPDP_ValidateToken_LargeIntClaimPreserved` (2^53+1 survives as `json.Number`) (#687)
  - `TestReadLastAuditLineReportsIOError` + `TestWriteChainResumeFailedMarker` (signed, HMAC-valid, kind=chain_resume_failed, genesis-chained) (#688)
  - `TestJWTPDP_ValidateToken_MissingMCPClaimGivesActionableError` (#689)
  - Updated `TestJWTPDP_ValidateToken_BadClaimVersion` to split the empty-version case (now the actionable "missing" message) from non-empty unsupported versions; existing `readLastAuditLine` callers migrated to the new signature via `mustReadLastAuditLine`.
- Threat-model §3.4 and changelog (entry 1.38) updated for the new marker kind and the precision/error-message changes.

---


eunolabs (commented):
## Self-review

**Verified:**
- #687: `decodeJWTClaimsPreservingNumbers` re-decodes the *same* payload segment go-jose verified (signature checked first), differing only in numeric representation (the fix). `jwtClaimsAsMap` copies claim values verbatim and the OPA evaluator marshals them, and `json.Number` round-trips to identical JSON — so normal small-int claims are unaffected and no downstream `float64` type assertion exists on claim values (the `float64` asserts in `schema.go` are for tool arguments, not claims). Duplicate-key semantics match go-jose (last wins). Decode failure fails closed.
- #688: on an I/O read error `openAuditSink` deliberately does NOT fall back to a rotated sibling (the read is unreliable) and writes the genesis-chained `chain_resume_failed` marker; a genuinely *missing* file stays the normal empty case. The three marker kinds are mutually exclusive (a read failure leaves `last==""`, so no parse/HMAC attempt). All `readLastAuditLine` callers migrated to the `(string, error)` signature.
- #689: empty `mcp.v` (absent block or present-with-empty-v) now yields the actionable message; the existing `BadClaimVersion` table was split so non-empty unsupported versions still assert the version-mismatch message.

Tests: `go test -race ./...`, `go vet`, `golangci-lint run` (0 issues), `make check-license`, `gofmt -l` all clean. New regression tests cover all three issues plus the new marker's HMAC validity. Re-review produced no further comments.

---

eunolabs:
✅ Ready to merge — all session rev-20260616T1110Z-7bb934 bugs addressed (#687, #688, #689), self-review complete with no outstanding comments, tests passing (`go test -race ./...` locally + CI Test/Lint/Build and all demo/OPA/security checks green).

---

PR #709, opened 2026-06-16
Fix session rev-20260616T1211Z-d7bc35 bugs

### LINKED ISSUE #690: [rev-20260616T1211Z-d7bc35] Trailing `?` in JWT capability claim silently accepted as unconditioned grant
**Location:** `cmd/eunox-mcp/pdp_jwt.go`, `splitV2Claim` (line 769) and `parseV2Claim`/inner loop (lines 739, 888)

**Reference:** https://github.com/eunolabs/eunox/blob/ad4d03b7cee1532eceb3bcdf1f229539b4899579/cmd/eunox-mcp/pdp_jwt.go#L769-L774

**Severity:** Medium

## Description

`splitV2Claim` splits a JWT capability claim at the first `?` to separate the name part from the condition suffix:

```go
func splitV2Claim(claim string) (namepart, condpart string) {
    if i := strings.IndexByte(claim, '?'); i >= 0 && !isHTTPResourceClaim(claim[:i]) {
        return claim[:i], claim[i+1:]
    }
    return claim, ""
}
```

For a claim like `"tool:read_file?"` (trailing `?`, no condition), this returns `namepart = "tool:read_file"`, `condpart = ""`.

Both call sites (`parseV2Claim` at token validation time, and the inner matching loop at decision time) guard `parseCondSuffix` with `if condpart != ""`:

```go
if condpart != "" {
    conds, parseErr = parseCondSuffix(condpart)
    ...
}
```

With `condpart = ""`, `parseCondSuffix` is never called. The trailing `?` is silently treated as if no conditions were specified. The claim `"tool:read_file?"` grants `tool:read_file` with **no conditions**, which is almost certainly not the token issuer's intent.

`parseCondSuffix("")` would return an error — `"condition \"\" is missing '=' separator"` — but it is never called for an empty condpart, so that validation is bypassed.

## Why this violates fail-closed

CONTRIBUTING.md: _"On any ambiguity — missing manifest, unmapped method, malformed JWT, unset condition argument — deny, with a structured error code."_

A trailing `?` with no condition pairs is a structurally malformed claim. It should cause `ValidateToken` to reject the JWT outright. Instead it is silently normalised to an unconditioned grant, which is maximally permissive.

## Trigger / Repro

1. Issue a JWT with claim `"tool:read_file?"` (trailing `?`, nothing after).
2. Present the token to `JWTPDP`; call `tools/call` for `read_file` with any arguments.
3. The JWTPDP evaluates zero conditions (no `evaluateJWTConditions` is run) and allows the call unconditionally — the trailing `?` was silently dropped.
4. Replace `"tool:read_file?"` with `"tool:read_file?path=/etc/shadow"` and verify conditions ARE applied, confirming the contrast.

## Suggested Fix

In `parseV2Claim` (and the inner matching loop), change the guard from `if condpart != ""` to also reject an empty condpart when a `?` was present. The easiest way is for `splitV2Claim` to return a boolean `hadSep` alongside `condpart`, and treat `hadSep && condpart == ""` as a parse error:

```go
func splitV2Claim(claim string) (namepart, condpart string, hadSep bool) {
    if i := strings.IndexByte(claim, '?'); i >= 0 && !isHTTPResourceClaim(claim[:i]) {
        return claim[:i], claim[i+1:], true
    }
    return claim, "", false
}

// In parseV2Claim:
namepart, condpart, hadSep := splitV2Claim(claim)
if hadSep && condpart == "" {
    return "", "", nil, fmt.Errorf("JWT capability claim %q: trailing '?' with no condition pairs", claim)
}
```

This causes `ValidateToken` to reject any JWT containing a bare-`?` claim, closing the silent-normalisation path fail-closed.

**Session:** rev-20260616T1211Z-d7bc35

### LINKED ISSUE #691: [rev-20260616T1211Z-d7bc35] `boundFieldTo` truncates at byte boundary, producing invalid UTF-8 in audit envelope fields
**Location:** `cmd/eunox-mcp/audit.go`, `boundFieldTo`, line 1119

**Reference:** https://github.com/eunolabs/eunox/blob/ad4d03b7cee1532eceb3bcdf1f229539b4899579/cmd/eunox-mcp/audit.go#L1100-L1120

**Severity:** Low

## Description

`boundFieldTo` caps audit envelope fields (SessionID, Target, Method, RequestID) to a configured byte limit. When a field value exceeds the limit, the function retains the first `keep` bytes and appends a truncation marker:

```go
func boundFieldTo(s string, limit int) string {
    if len(s) <= limit {
        return s
    }
    marker := fmt.Sprintf("...[eunox: truncated, %d bytes]", len(s))
    keep := limit - len(marker)
    if keep < 0 { ... }
    return s[:keep] + marker   // ← byte-level truncation, not rune-boundary
}
```

`s[:keep]` slices `s` at a byte offset, not at a UTF-8 rune boundary. If `keep` falls in the middle of a multi-byte rune (e.g. a 3-byte CJK character, 4-byte emoji in a tool name), the resulting string contains one or more orphaned continuation bytes: invalid UTF-8.

### What happens downstream

`encoding/json` (used in `writeRecord`) replaces invalid UTF-8 bytes with the Unicode replacement character `�`. This means:
1. The in-memory `auditRecord` struct field contains invalid UTF-8.
2. The serialised JSONL audit file contains `�` in place of the broken bytes — the file is valid JSON.
3. `audit-verify` decodes the field back to a string with `�` bytes.
4. When the decoded field is printed with `fmt.Fprintf` (e.g. in `INVALID` diagnostic lines), operators see replacement characters rather than the original rune.

The HMAC chain is unaffected — the MAC is computed over the JSON bytes after the `�` substitution — but audit fields for non-ASCII tool names or session IDs that are truncated at a rune boundary will contain garbage bytes in memory and replacement characters in the output, making forensic work harder.

## Trigger / Repro

1. Register an MCP tool whose name is at least `auditEnvelopeFieldCap` bytes long and contains multi-byte UTF-8 characters (e.g. a Japanese name repeated until it crosses the cap boundary such that the cap lands mid-rune).
2. Call the tool through the proxy.
3. Inspect the raw audit record in memory just before `json.Marshal` — the `target` field ends with `\xNN` orphaned bytes.
4. Read the JSONL file back: the `target` field ends with `�` replacement characters rather than the original rune.

## Suggested Fix

After computing `keep`, walk backwards to find the last complete rune boundary before `keep`:

```go
// Ensure truncation falls on a UTF-8 rune boundary.
for keep > 0 && !utf8.RuneStart(s[keep]) {
    keep--
}
return s[:keep] + marker
```

`utf8.RuneStart(b)` returns true when `b` is the first byte of a rune (top bit 0, or top two bits 11). This shrinks `keep` by at most 3 bytes (the longest continuation sequence), keeping the returned string within the limit.

Import: `unicode/utf8` (already in the standard library).

**Session:** rev-20260616T1211Z-d7bc35

### LINKED ISSUE #692: [rev-20260616T1211Z-d7bc35] `isHTTPResourceClaim` is case-sensitive, misclassifying uppercase-scheme resource claims
**Location:** `cmd/eunox-mcp/pdp_jwt.go`, `isHTTPResourceClaim`, lines 779–786

**Reference:** https://github.com/eunolabs/eunox/blob/ad4d03b7cee1532eceb3bcdf1f229539b4899579/cmd/eunox-mcp/pdp_jwt.go#L779-L786

**Severity:** Low

## Description

`isHTTPResourceClaim` determines whether a JWT capability claim's name part is an http(s) resource URL, to decide whether `?` should be treated as a URL query-string delimiter (not a condition separator):

```go
func isHTTPResourceClaim(head string) bool {
    const ns = "resource:"
    if !strings.HasPrefix(head, ns) {
        return false
    }
    v := head[len(ns):]
    return strings.HasPrefix(v, "http://") || strings.HasPrefix(v, "https://")
}
```

The comparison is case-sensitive. RFC 3986 §3.1 specifies that URI schemes are case-insensitive. A JWT issuer that emits `resource:HTTPS://api.example.com/data?q=1` (uppercase scheme) causes `isHTTPResourceClaim` to return `false`. `splitV2Claim` then treats the `?` as a condition separator, splitting the claim into:

- namepart: `resource:HTTPS://api.example.com/data`
- condpart: `q=1`

The condition suffix `q=1` is then parsed by `parseCondSuffix` as if it were a condition pair. Depending on the claim set, this either causes an unexpected condition evaluation (the `q` argument is checked against value `1`) or a parse error that silently drops the matching claim (the inner loop's `continue` at line 893).

In the parse-error case the JWT claim effectively grants nothing for that resource (fail-closed, so safe), but it is invisible to the operator: no error is surfaced, the token is accepted, and the resource is simply unreachable via JWT despite a seemingly valid claim.

## Trigger / Repro

1. Issue a JWT with the claim `"resource:HTTPS://api.example.com/data?q=1"` (uppercase HTTPS).
2. Call `resources/read` for URI `HTTPS://api.example.com/data?q=1`.
3. Observe: the claim is parsed with `condpart = "q=1"`, which is evaluated as a JWT condition — an `allowedValues` condition on argument `q` — not as part of the resource URI. The resource read either fails with a condition error (if `q != "1"`) or is mismatched (target vs. stored namepart mismatch).
4. Contrast with `"resource:https://api.example.com/data?q=1"` (lowercase), which correctly keeps the full URL as the claim name.

## Suggested Fix

Normalise the scheme to lowercase before the prefix check:

```go
func isHTTPResourceClaim(head string) bool {
    const ns = "resource:"
    if !strings.HasPrefix(head, ns) {
        return false
    }
    v := strings.ToLower(head[len(ns):])
    return strings.HasPrefix(v, "http://") || strings.HasPrefix(v, "https://")
}
```

Only the scheme detection needs lowercasing; the actual namepart returned by `splitV2Claim` remains unchanged (preserving the original case for manifest lookup). This aligns with RFC 3986's scheme case-insensitivity without altering the stored claim value.

**Session:** rev-20260616T1211Z-d7bc35

### LINKED ISSUE #693: [rev-20260616T1211Z-d7bc35] Empty column allowlist in `AllowedTablesCondition` manifest is not rejected, silently creating a permanently unfulfillable condition
**Location:** `pkg/enforcement/handlers.go`, `handleAllowedTables`, lines 900–930; `pkg/capability/condition.go`, `AllowedTablesCondition`

**Reference:** https://github.com/eunolabs/eunox/blob/ad4d03b7cee1532eceb3bcdf1f229539b4899579/pkg/enforcement/handlers.go#L900-L930

**Severity:** Low–Medium

## Description

`AllowedTablesCondition` supports per-table column restrictions via a `Columns map[string][]string` field. When the manifest specifies a table key with an empty column slice, e.g.:

```yaml
conditions:
  - type: allowedTables
    argument: table
    tables: [users]
    columns:
      users: []   # empty allowlist
```

`handleAllowedTables` sees `allowedCols = []string{}` and `hasColumnRestriction = true` for key `"users"`:

```go
allowedCols, hasColumnRestriction := columnsByTable[tableKey]
if hasColumnRestriction {
    if len(access.Columns) == 0 {
        return &ConditionError{Code: ErrCodeMissingContext, ...}  // always fires
    }
    colSet := make(map[string]bool, len(allowedCols))  // empty map
    for _, c := range allowedCols { colSet[...] = true }
    for _, col := range access.Columns {
        if !colSet[...] {                                // always true (colSet empty)
            return &ConditionError{Code: ErrCodeConditionFailed, ...}  // always fires
        }
    }
}
```

**Both error paths always fire for any request involving that table:**
- Request sends no columns → `ErrCodeMissingContext: "column list required for table users"`
- Request sends any columns → `ErrCodeConditionFailed: "column X on table users is not allowed"`

The table is listed in `tables` (so table-name access IS allowed), but every attempt to actually access it fails unconditionally. This is a permanently unfulfillable condition.

### Root cause

There is no manifest validation pass that checks `AllowedTablesCondition`. The empty column slice is accepted at load time without error. An operator who writes `users: []` meaning "no column restriction for users" (instead of the correct form: omit the key from `Columns`) discovers the problem only at request time, with no indication that the manifest itself is misconfigured.

### Contrast with similar validations

Other handlers enforce early-failure semantics for impossible configurations: `handleSequenceBlock` denies immediately when `AfterTools` is empty (line 1295–1300), `handleAllowedExtensions` requires a non-empty `Extensions` list. `AllowedTablesCondition` has no equivalent check.

## Trigger / Repro

1. Write a manifest with `AllowedTablesCondition` where `columns: {users: []}`.
2. Start the proxy; no error at load time.
3. Call `tools/call` for the guarded tool with `{"table": {"table": "users", "columns": ["id"]}}`.
4. Receive `CONDITION_FAILED: column "id" on table "users" is not allowed` — a confusing error implying the manifest should list `"id"` in the column allowlist, when the real fix is to either remove the `users` key or add actual columns.

## Suggested Fix

Add a validation check — either at manifest load time or in `handleAllowedTables` on first invocation — that rejects an empty column allowlist for any table key:

```go
// In handleAllowedTables, before the per-table loop, or in a future
// validateAllowedTablesCondition called at engine construction:
for tableName, cols := range at.Columns {
    if len(cols) == 0 {
        return &ConditionError{
            Message: fmt.Sprintf("allowedTables condition has an empty column allowlist for table %q; "+
                "to allow any column omit the table from 'columns', "+
                "to deny all column access remove the table from 'tables'", tableName),
        }
    }
}
```

Alternatively, add this check to a manifest validation pass (a `ValidateConditions` function in `pkg/capability`) called from `loadUpstreamPDP`, so the operator gets a clear error at startup rather than at request time.

**Session:** rev-20260616T1211Z-d7bc35

eunolabs:
## Bugs addressed (session rev-20260616T1211Z-d7bc35)

- Closes #690 — **(Medium)** A JWT capability claim with a trailing `?` and no condition pairs (`tool:read_file?`) was silently normalised to an *unconditioned*, maximally permissive grant: `splitV2Claim` returned an empty `condpart`, and the `if condpart != ""` guard skipped `parseCondSuffix` entirely. `splitV2Claim` now returns a `hadSep` flag; `parseV2Claim` rejects a bare-`?` claim with a structured error (so `ValidateToken` rejects the token, fail closed), and `buildConstraintsFromClaims` skips it at decision time.
- Closes #692 — **(Low)** `isHTTPResourceClaim` was case-sensitive, so an uppercase/mixed-case scheme (`resource:HTTPS://api.example.com/data?q=1`) was misclassified as a non-URL resource — `splitV2Claim` then treated the URL's `?q=1` query as a condition suffix, silently misconditioning or dropping the grant. Scheme detection now lowercases before the prefix check (RFC 3986 §3.1 scheme case-insensitivity); the stored claim name keeps its original case for manifest lookup.
- Closes #693 — **(Low-Medium)** An `allowedTables` condition with an empty per-table column allowlist (`columns: { users: [] }`) is permanently unfulfillable — at runtime a request with no columns fails `MISSING_CONTEXT` and any request *with* columns fails (none are in the empty allowlist), so every access to the table is denied even though it's listed in `tables`. `validateAllowedTables` now rejects it at `validate`/load time with actionable guidance (omit the key to allow any column, or remove the table from `tables`).
- Closes #691 — **(Low)** `boundFieldTo` truncated audit envelope fields (`session_id`, `target`, `method`) at a byte offset, which could split a multi-byte UTF-8 rune and leave orphaned continuation bytes that `json.Marshal` rewrites to U+FFFD — corrupting non-ASCII fields in the on-disk record and in `audit-verify` output. Truncation now walks back to a rune boundary (`utf8.RuneStart`), dropping at most 3 bytes and staying within the limit.

## Testing
- New regression tests:
  - `TestParseV2Claim_TrailingQuestionMarkRejected` (#690)
  - `TestParseV2Claim_UppercaseHTTPSchemeResource` (#692)
  - `TestLoadManifest_ConditionArgumentValidation/allowedTables_empty_column_allowlist_rejected` (#693)
  - `TestBoundFieldToTruncatesOnRuneBoundary` (valid UTF-8 across a sweep of mid-rune limits) (#691)
- Docs: capability-manifest guide updated for the empty-column-allowlist rule (#693); threat-model changelog entry 1.39 covers all four. (No `mcp-capability-manifest` spec change — #693 is a validation tightening of an existing field, not a grammar/schema change.)

---


eunolabs (commented):
## Self-review

**Verified:**
- #690: the `hadSep && condpart == ""` rejection fires only for a NON-http(s) trailing `?` (the condition-separator case). An http(s) resource with a trailing `?` (`resource:https://api/data?`) is correctly left intact — `isHTTPResourceClaim` is true, so `splitV2Claim` doesn't split and `hadSep` is false; the empty query stays part of the URL (valid RFC 3986). The decision-time matcher skips a bare-`?` claim too (defensive; `ValidateToken` already rejects the token). All four `splitV2Claim` call sites updated for the new signature.
- #692: only the scheme *detection* is lowercased (a local copy); `splitV2Claim` returns the original-case `claim[:i]`, so the stored name preserves case for manifest lookup — the test asserts `HTTPS://…` round-trips uppercase.
- #693: rejected at load time (the issue's preferred option), giving operators a startup error; a programmatically-built manifest that bypasses validation still fails closed at runtime (deny), so there's no security regression. Same call coverage as the existing empty-`tables` check.
- #691: the rune-boundary walk runs only in the `keep >= 0` path (past the marker-longer-than-limit branch), `keep < len(s)` is guaranteed, and it shrinks `keep` by ≤3 bytes so the result stays within `limit`; the test sweeps mid-rune limits and asserts `utf8.ValidString`.

Tests: `go test -race ./...`, `go vet`, `golangci-lint run` (0 issues), `make check-license`, `gofmt -l` all clean. Re-review produced no further comments.

---

eunolabs:
✅ Ready to merge — all session rev-20260616T1211Z-d7bc35 bugs addressed (#690, #691, #692, #693), self-review complete with no outstanding comments.

CI note: the first run's `Test` job hit a **flaky, unrelated** failure — `TestRedis_MultiInstance_KillPropagatesAcrossInstances` in `pkg/killswitch` failed with `connection refused` on random ports (a Redis pub/sub test-infra hiccup). This PR touches only `pdp_jwt.go`, `audit.go`, and `manifest.go` — nothing in `pkg/killswitch` — and the test passes locally. I re-ran the failed job and it passed (coverage: pkg/ 97.8%, cmd/ 92.8%, total 82.6%, all above the 80% gate). All other checks (Lint, Build × all platforms, demos, OPA, CodeQL, govulncheck, Trivy) are green.

---

PR #710, opened 2026-06-16
Fix session rev-20260616T1241Z-361bc1 bugs

### LINKED ISSUE #699: [rev-20260616T1241Z-361bc1] Non-float64 numeric args silently bypass minimum/maximum schema validation
**Location:** `pkg/enforcement/schema.go:139-148`

**Reference:** https://github.com/eunolabs/eunox/blob/62e6ba756dc3fbc043e00a4aa6ad33c70322cffe/pkg/enforcement/schema.go#L139-L148

**Severity:** Medium

**Description:**
`schemaValidateValue` dispatches on the Go type of `val` in a type switch. The only numeric case handled is `case float64`, which is the JSON decoder default. However, programmatic and embedded callers can pass `int`, `int32`, `int64`, `uint`, or other integer types directly (the parameter type is `interface{}`). When `schema.Type` is not set (or a `schemaCheckType` call is skipped), such values fall through the switch to `return nil` — silently passing validation without evaluating `minimum`, `maximum`, or any other numeric keyword.

`schemaJSONTypeOf` also returns `"unknown"` for Go integer types (they are absent from its switch), so a declared `type: integer` or `type: number` constraint in the manifest likewise fails to catch them. The fail-closed invariant is violated: a disallowed numeric value reaches the upstream instead of being denied.

**Trigger / repro:**
1. Define a capability with an argument schema specifying `maximum: 10`.
2. Call the engine from an embedded/programmatic context (not from a JSON-decoded MCP request), passing `args["count"] = int(999)` (Go `int`, not `float64`).
3. `schemaValidateValue` enters the `case float64` branch only for actual `float64` values; `int(999)` matches no case, drops to `return nil`, and the call is allowed.

Real-world risk: any adapter or test harness that builds `map[string]interface{}` args from typed Go values rather than JSON decoding can bypass schema-enforced numeric bounds.

**Suggested fix:**
Add a `case int, int64, uint64` (or use `reflect` on `val`) branch in `schemaValidateValue` that converts to `float64` and delegates to `schemaValidateNumber`. Alternatively, normalize incoming `args` to JSON-decoded types (all numbers as `float64`) at the engine entry point before schema validation. Also update `schemaJSONTypeOf` to return `"number"` for Go integer types so `type: integer/number` constraints also fire.

**Session:** rev-20260616T1241Z-361bc1

### LINKED ISSUE #700: [rev-20260616T1241Z-361bc1] handleSequenceBlock records empty blockedTool in audit when req.ToolName is unset
**Location:** `pkg/enforcement/handlers.go` — `handleSequenceBlock`, lines ~1376-1381

**Reference:** https://github.com/eunolabs/eunox/blob/62e6ba756dc3fbc043e00a4aa6ad33c70322cffe/pkg/enforcement/handlers.go#L1376-L1381

**Severity:** Low-Medium

**Description:**
`handleSequenceBlock` derives the blocked tool identity for audit details as:

```go
blockedType, blockedName := splitEnginePrefix(req.ToolName)
    blockedType = req.Target.Type
}
blockedTarget := blockedType + ":" + blockedName
```

`blockedName` comes solely from `splitEnginePrefix(req.ToolName)`. When `req.ToolName` is empty — which occurs for resource and prompt requests routed through a shared engine path — `splitEnginePrefix("")` returns `("", "")` and `blockedName` stays `""`. The code overrides `blockedType` from `req.Target.Type` but never overrides `blockedName` from `req.Target.Name`. The resulting `blockedTarget` in the denial audit record is `"tool:"` (or `"resource:"`) rather than the actual resource or prompt name.

By contrast, `recordSessionCall` (the session-counter bookkeeping function in `engine.go`) does fall back to `req.Target.Name` when `req.ToolName` is empty, so the session counter is keyed on the real name while the denial audit record says `""`. This inconsistency makes audit correlation and post-incident review unreliable for the affected request types.

**Trigger / repro:**
1. Configure a `sequenceBlock` condition on a resource capability.
2. Issue a `resources/read` request that violates the sequence block — `req.ToolName` will be `""`, `req.Target.Name` will be the resource URI.
3. Inspect the resulting denial audit record: `details.blockedTool` (or equivalent) will contain `"resource:"` instead of `"resource:<uri>"`.

**Suggested fix:**
After the `blockedType` override, add a symmetric override for `blockedName`:

```go
blockedType, blockedName := splitEnginePrefix(req.ToolName)
if req.Target != nil {
    if req.Target.Type != "" {
        blockedType = req.Target.Type
    }
    if req.Target.Name != "" {
        blockedName = req.Target.Name
    }
}
blockedTarget := blockedType + ":" + blockedName
```

**Session:** rev-20260616T1241Z-361bc1

### LINKED ISSUE #701: [rev-20260616T1241Z-361bc1] handleMaxCalls denies with misleading error when req.ToolName is empty but req.Target.Name is set
**Location:** `pkg/enforcement/handlers.go` — `handleMaxCalls`, lines ~406-422

**Reference:** https://github.com/eunolabs/eunox/blob/62e6ba756dc3fbc043e00a4aa6ad33c70322cffe/pkg/enforcement/handlers.go#L406-L422

**Severity:** Low

**Description:**
`handleMaxCalls` resolves the tool name for counter keying as:

```go
if toolName == "" {
    return &ConditionError{
        Code:    capability.ErrCodeMissingContext,
        Message: "tool name is required for maxCalls condition",
    }
}
```

When `req.ToolName` is empty — which is the case for `resources/read`, `resources/subscribe`, `prompts/get`, and `sampling/createMessage` requests processed through a shared code path — the condition handler returns a hard `MISSING_CONTEXT` denial even though `req.Target.Name` carries the correct identifier. This fires *before* any call-count check, so the call is denied for the wrong reason and the counter is never incremented.

`recordSessionCall` in `engine.go` handles the identical situation correctly: it falls back to `req.Target.Name` when `req.ToolName` is empty. The asymmetry means `maxCalls` cannot be applied to non-tool requests, and the error message ("tool name is required") is misleading for resource/prompt/sampling requests.

**Trigger / repro:**
1. Configure a `maxCalls` condition on a resource capability (e.g., `resources/read` for a specific URI).
2. Issue a `resources/read` request — `req.ToolName` will be `""`, `req.Target.Name` will hold the URI.
3. The request is denied immediately with `ErrCodeMissingContext` rather than being allowed up to the configured limit.

**Suggested fix:**
Mirror `recordSessionCall`'s fallback logic:

```go
if toolName == "" && req.Target != nil {
    toolName = req.Target.Name
}
if toolName == "" {
    return &ConditionError{
        Code:    capability.ErrCodeMissingContext,
        Message: "tool or resource name is required for maxCalls condition",
    }
}
```

**Session:** rev-20260616T1241Z-361bc1

### LINKED ISSUE #702: [rev-20260616T1241Z-361bc1] Post-Close() record drops may be undercounted in shutdown warning
**Location:** `cmd/eunox-mcp/audit.go` — `record()` / `Close()`, lines ~1355-1358

**Reference:** https://github.com/eunolabs/eunox/blob/62e6ba756dc3fbc043e00a4aa6ad33c70322cffe/cmd/eunox-mcp/audit.go#L1355-L1358

**Severity:** Low

**Description:**
In `record()`, when the sink is already closed the drop counter is incremented:

```go
if s.closed {
    s.dropped.Add(1)
    return
}
```

In `Close()`, the shutdown warning is emitted (approximately):

```go
if n := s.dropped.Load(); n > 0 {
    fmt.Fprintf(os.Stderr, "audit: %d records dropped\n", n)
}
```

There is a window between when `Close()` reads `s.dropped` and when it sets `s.closed = true`. Goroutines that call `record()` concurrently during shutdown can race to increment `s.dropped` *after* `Close()` has already sampled the counter but *before* it sets `s.closed`. These drops are silently lost from the warning message.

More concretely: `Close()` drains the channel, sets `s.closed = true`, then reads `s.dropped`. But if `Close()` reads `s.dropped` before setting `s.closed`, concurrent callers of `record()` that race through the `!s.closed` branch (and successfully send to a full/closed channel, or are preempted just after the check) may not be accounted for in the warning.

The `DroppedRecords()` metric counter is accurate because it counts all increments; only the stderr warning at shutdown may be understated.

**Trigger / repro:**
High-throughput graceful shutdown where multiple goroutines are calling `record()` while `Close()` executes. The likelihood is proportional to shutdown concurrency and queue depth. The consequence is a misleadingly low (or zero) drop count in the shutdown warning, potentially hiding data loss from operators.

**Suggested fix:**
Emit the dropped-records warning *after* setting `s.closed = true` and after all concurrent `record()` calls that could still be in-flight have had an opportunity to observe the closed state and increment the counter. A simple approach: read `s.dropped` at the very end of `Close()`, after the drainer goroutine has joined (via the done channel) and after `s.closed` has been set. This ensures all concurrent `record()` callers that raced through the pre-close path have finished their `dropped.Add(1)` before the warning is printed.

**Session:** rev-20260616T1241Z-361bc1

eunolabs:
## Bugs addressed (session rev-20260616T1241Z-361bc1)

- Closes #699 — **(Medium)** `schemaValidateValue` dispatched only on `float64`, so a non-float64 numeric argument (`int`/`int64`/`uint`/… from a programmatic or embedded caller, not a JSON-decoded request) matched no `case` and silently bypassed `minimum`/`maximum` and the `type` keyword — a disallowed value reaching the upstream (fail-open). A non-float64 numeric is now normalized to `float64` (the type a JSON-decoded arg always has) at the top of `schemaValidateValue`, so type/enum/keyword checks all fire identically.
- Closes #701 — **(Low)** `handleMaxCalls` keyed only on `req.ToolName`, so a `maxCalls` on a non-tool capability (resources/read, prompts/get, …) — whose identifier is in `req.Target.Name` — denied every call with a misleading `MISSING_CONTEXT` "tool name is required" before any count check, and never keyed the counter. It now falls back to `req.Target.Name`, mirroring `recordSessionCall`.
- Closes #700 — **(Low-Medium)** `handleSequenceBlock` derived the blocked target name from `req.ToolName` only, so a resource/prompt denial recorded a bare `resource:`/`prompt:` (no name) in the audit details while `recordSessionCall` keyed the call on the real name — breaking audit correlation. `blockedName` now falls back to `req.Target.Name`.
- Closes #702 — **(Low)** The close-time read of the audit drop counter ran without the lock, so a producer racing `Close` (incrementing the closed-drop counter under `record()`'s read lock) could increment just after the unsynchronized sample, undercounting drops in the shutdown warning. The counter is now sampled under the write lock as a barrier; `DroppedRecords()` remains the authoritative metric.

## Testing
- New regression tests:
  - `TestArgumentSchema_ValidateUnit` cases for non-float64 int/int64 args (maximum + integer-type) (#699)
  - `TestMaxCalls_ResourceTarget_NoToolName` (#701)
  - `TestSequenceBlock_ResourceTarget_BlockedNameInDetails` (#700)
  - `TestAuditSink_ConcurrentRecordDuringClose` (`-race` deadlock/consistency guard) (#702)
- Threat-model changelog entry 1.40 covers all four.

---


eunolabs (commented):
## Self-review

**Verified:**
- #699: the float64 fast-path is unchanged (JSON-decoded args already float64). Only non-float64 *numerics* are coerced (`toFloat64` rejects bool/string/etc.); nested numerics are normalized too since `schemaValidateObject`/`Array` recurse through `schemaValidateValue`. The coercion runs before enum/type/keyword checks, so all fire consistently; an `int` satisfying `type:number` now correctly passes (was wrongly rejected), and no fail-open is introduced.
- #701: falls back to `stripEnginePrefix(req.Target.Name)` only when `req.ToolName` is empty; a resource URI carries no engine prefix so the strip is a no-op, and `targetType` already keys the bucket distinctly. With neither set it still denies (renamed message).
- #700: mirrors the `blockedType` override, gated on `blockedName == ""`, so tool calls (which set `req.ToolName`) are unaffected.
- #702: the new `s.mu.Lock()` runs after `wg.Wait()` (drainer gone) and only pairs with `record()`'s brief `RLock` — no nesting, no deadlock (confirmed by `TestAuditSink_ConcurrentRecordDuringClose` under `-race`).

Tests: `go test -race ./...`, `go vet`, `golangci-lint run` (0 issues), `make check-license`, `gofmt -l` all clean. Re-review produced no further comments.

---

eunolabs:
✅ Ready to merge — all session rev-20260616T1241Z-361bc1 bugs addressed (#699, #700, #701, #702), self-review complete with no outstanding comments, tests passing (`go test -race ./...` locally + CI Test/Lint/Build and all demo/OPA/security checks green).

---

PR #711, opened 2026-06-16
Fix session rev-20260616T1259Z-331ef0 bugs

### LINKED ISSUE #704: [rev-20260616T1259Z-331ef0] verifyEdDSA panics instead of erroring on wrong-length Ed25519 public key
**Location:** `pkg/capability/proofs.go:319-328`

**Reference:** https://github.com/eunolabs/eunox/blob/62e6ba756dc3fbc043e00a4aa6ad33c70322cffe/pkg/capability/proofs.go#L319-L328

**Severity:** Medium

**Description:**
`verifyEdDSA` type-asserts the key to `ed25519.PublicKey` and then calls `ed25519.Verify` without checking the key's byte length:

```go
func verifyEdDSA(message, sigBytes []byte, key gocrypto.PublicKey) error {
    edKey, ok := key.(ed25519.PublicKey)
    if !ok {
        return errors.New("key type mismatch: expected ed25519.PublicKey")
    }
    if !ed25519.Verify(edKey, message, sigBytes) {
        return errors.New("EdDSA (Ed25519) verification failed")
    }
    return nil
}
```

Go's `ed25519.Verify` **panics** (does not return an error) when `len(publicKey) != ed25519.PublicKeySize` (32 bytes):

```
panic: ed25519: bad public key length: N
```

A `PublicKeyResolver` can return an `ed25519.PublicKey` with a wrong length due to:
1. A bug in the resolver implementation that does not validate key material
2. A malformed DID document (truncated base64, wrong key type stored in an Ed25519 slot)
3. An adversary controlling the DID resolution endpoint who deliberately returns a wrong-length key

In case 3 this is a process-level DoS: an attacker crafts a token carrying an EdDSA co-signature whose `issuerDID` resolves to a wrong-length key, causing every call to `VerifyCoSignatures` for that token to crash the process.

The sibling functions `verifyRSAPKCS1v15` and `verifyRSAPSS` do not share this vulnerability (`rsa.VerifyPKCS1v15` and `rsa.VerifyPSS` return errors on bad inputs). `verifyECDSA` is also safe because `ecdsa.Verify` does not panic on a wrong-sized key (it returns false). EdDSA is the only special case because `ed25519.Verify` explicitly panics on the wrong key length by design.

**Trigger / repro:**
1. Register a co-issuer DID backed by a resolver that returns `ed25519.PublicKey([]byte{0x01, 0x02})` (wrong length).
2. Present a token that carries an EdDSA co-signature for that DID.
3. `VerifyCoSignatures` calls `verifyOneSig` → `verifyRawSignature` → `verifyEdDSA` → `ed25519.Verify` → **panic**, process crashes.

**Suggested fix:**
Add a length guard before calling `ed25519.Verify`:

```go
func verifyEdDSA(message, sigBytes []byte, key gocrypto.PublicKey) error {
    edKey, ok := key.(ed25519.PublicKey)
    if !ok {
        return errors.New("key type mismatch: expected ed25519.PublicKey")
    }
    if len(edKey) != ed25519.PublicKeySize {
        return fmt.Errorf("invalid Ed25519 key length: got %d, want %d", len(edKey), ed25519.PublicKeySize)
    }
    if !ed25519.Verify(edKey, message, sigBytes) {
        return errors.New("EdDSA (Ed25519) verification failed")
    }
    return nil
}
```

This converts the panic into a structured error that `verifyOneSig` surfaces as a normal verification failure, consistent with all other `verify*` functions.

**Session:** rev-20260616T1259Z-331ef0

eunolabs:
## Bugs addressed (session rev-20260616T1259Z-331ef0)

- Closes #704 — **(Medium)** `verifyEdDSA` type-asserted the resolved key to `ed25519.PublicKey` and called `ed25519.Verify` without a length check. Go's `ed25519.Verify` **panics** (by design) when the key is not exactly `ed25519.PublicKeySize` (32) bytes, so a `PublicKeyResolver` returning a wrong-length Ed25519 key — from a buggy implementation, a malformed/truncated DID document, or an adversary controlling DID resolution — crashed the process for every co-signed token naming that DID (a process-level DoS). The function now rejects a wrong-length key with a structured error, surfaced as an ordinary verification failure like the RSA/ECDSA verifiers (which never had this issue — their stdlib funcs return an error or `false`).

## Testing
- New regression test `TestVerifyCoSignatures_EdDSA_WrongKeyLength` drives a resolver returning a 2-byte Ed25519 key through `VerifyCoSignatures` and asserts a structured error with **no panic** (`require.NotPanics`).
- `golangci-lint run` (0 issues), `make check-license`, `gofmt -l` — all clean.
- Threat-model changelog entry 1.41.

---


eunolabs (commented):
## Self-review

Single, isolated change: a length guard added before `ed25519.Verify`, returning a structured error on the wrong-length key. No behavior change for valid keys (a correct 32-byte key passes the guard unchanged), no new fail-open (the guard *adds* a rejection), and the RSA/ECDSA paths are untouched. The regression test asserts both no-panic (`require.NotPanics`) and a structured error. `go test -race ./...`, `golangci-lint` (0 issues), `make check-license`, `gofmt` all clean. No further comments.

---

eunolabs:
✅ Ready to merge — session rev-20260616T1259Z-331ef0 bug addressed (#704), self-review complete with no outstanding comments. The first `Test` run hit the same flaky, unrelated `pkg/killswitch` Redis pub/sub failure (`connection refused` on random ports — this PR only touches `pkg/capability/proofs.go`); the re-run passed. All checks green.

---

PR #712, opened 2026-06-16
Fix session rev-20260616T1342Z-9a3b6f bugs

### LINKED ISSUE #707: [rev-20260616T1342Z-9a3b6f] WithDryRun silently bypasses sequenceBlock in --audit mode, contradicting its own documentation
## Bug Report

**Location:**
- `pkg/enforcement/engine.go:289–297` (`WithDryRun` function and its godoc)
- `pkg/enforcement/handlers.go:1335–1340` (`handleSequenceBlock` dry-run early-return)
- `cmd/eunox-mcp/stdio.go:479–481` (audit-mode context decoration in `handleToolsCall`)

**Reference:**
- engine.go: https://github.com/eunolabs/eunox/blob/62e6ba756dc3fbc043e00a4aa6ad33c70322cffe/pkg/enforcement/engine.go#L289-L297
- handlers.go (sequenceBlock dry-run): https://github.com/eunolabs/eunox/blob/62e6ba756dc3fbc043e00a4aa6ad33c70322cffe/pkg/enforcement/handlers.go#L1335-L1340
- stdio.go audit decoration: https://github.com/eunolabs/eunox/blob/62e6ba756dc3fbc043e00a4aa6ad33c70322cffe/cmd/eunox-mcp/stdio.go#L479-L481

**Severity:** Medium — misleads operators, does not affect enforcement-mode security

---

### Description

`WithDryRun` is documented to skip only `MaxCalls`:

```go
// WithDryRun does NOT skip any other condition type — the decision
// reflects actual policy except that MaxCalls is treated as always-passing.
```

In reality, `handleSequenceBlock` *also* returns nil (always-passing) when `isDryRun(ctx)` is true:

```go
// In dry-run mode (preflight /validate) session history is not consulted —
// the verdict depends on runtime call order, which validation does not model.
// Treat the condition as satisfied so validation reflects structure, not state.
if isDryRun(ctx) {
    return nil
}
```

Additionally, `recordSessionCall` in `engine.go` (line 377) checks `isDryRun(ctx)` and skips writing session-history markers entirely when dry-run is set. This is the correct design for preflight `/validate` paths where no actual calls are being made. However, the stdio and HTTP transports also use `enforcement.WithDryRun` to set the dry-run flag when the proxy runs with the `--audit` flag:

```go
// cmd/eunox-mcp/stdio.go  handleToolsCall
decCtx := ctx
if p.audit {
    decCtx = enforcement.WithDryRun(ctx)  // <-- sets isDryRun for ALL conditions
}
dec := p.pdp.Decide(decCtx, ...)
```

The consequence is that in full proxy audit mode (`--audit`):

1. Tool A is called and forwarded (audit forwards everything).
2. `recordSessionCall` skips writing A's history marker because `isDryRun(decCtx) == true`.
3. Tool B (which has `sequenceBlock: afterTools: [A]`) is called.
4. `handleSequenceBlock` sees `isDryRun == true` → returns nil → always passes.
5. Audit log records B as "allow (observed)" when the policy would actually deny it in enforcement mode.

An operator using `--audit` to evaluate their `sequenceBlock` policy will always observe "allowed", giving false confidence that the condition is non-functional, when in fact it correctly denies in enforcement mode.

### Trigger / Reproduction

1. Configure a manifest with:
   ```yaml
   capabilities:
     - target: tool:A
       actions: [call]
     - target: tool:B
       actions: [call]
       conditions:
         - type: sequenceBlock
           afterTools: [A]
   ```
2. Start the proxy with `--audit`.
3. Call tool A, then call tool B.
4. Observe: audit log shows B as "allow (observed)".
5. Start the proxy without `--audit` (enforcement mode).
6. Call tool A, then call tool B.
7. Observe: B is denied with CONDITION_FAILED / sequenceBlock.

The two modes produce different verdicts for the same tool call sequence; `--audit` never shows "would be denied".

### Suggested Fix

The cleanest fix is to separate the two use cases currently collapsed under `isDryRun`:

**Option A — Document the existing behavior accurately** (lowest effort):

Update `WithDryRun`'s godoc to state that `sequenceBlock` is also treated as always-passing, for the same reason as `maxCalls` (it depends on runtime call-order state that is not modelled during preflight). Add a note to `--audit` documentation that `sequenceBlock` observations are not accurate in this mode.

**Option B — Introduce a separate "quota-only dry-run" flag for audit mode** (correct fix):

Add a new unexported context key, e.g. `ctxSkipQuotaKey{}`, consumed by `handleMaxCalls` and `isDeferredCondition` but NOT by `handleSequenceBlock` or `recordSessionCall`. The `--audit` transport path sets this key instead of the full dry-run key; `WithDryRun` continues to set both, so preflight behaviour is unchanged. With this fix, `--audit` mode accurately observes `sequenceBlock` denials while still not consuming MaxCalls quota slots.

---

**Session:** rev-20260616T1342Z-9a3b6f
**HEAD SHA:** 62e6ba756dc3fbc043e00a4aa6ad33c70322cffe
**Files reviewed this cycle:** `pkg/enforcement/engine.go`, `pkg/capability/condition.go`, `cmd/eunox-mcp/stdio.go`

eunolabs:
## Bugs addressed (session rev-20260616T1342Z-9a3b6f)

- Closes #707 — **(Medium)** The proxy's `--audit` (observe) mode decorated the decision context with `enforcement.WithDryRun`, which is a *preflight* flag that neutralizes MaxCalls **and** `sequenceBlock` **and** session-history recording. In `--audit` mode that meant no session history was ever recorded and a `sequenceBlock`-gated call was always reported `allow (observed)` — so an operator using `--audit` to evaluate a `sequenceBlock` policy saw it as inert (a false-confidence observability gap; enforcement mode was always correct, only the observation was wrong), contradicting `WithDryRun`'s own (also inaccurate) godoc.

  **Fix (the issue's Option B, the correct fix):** add `enforcement.WithSkipQuota`, which skips **only** the quota-consuming MaxCalls side effect — `sequenceBlock` is still evaluated and `recordSessionCall` still records history. The eight audit-mode transport call sites (stdio + HTTP: `tools/call`, `resources/read`, `resources/subscribe`, `prompts/get`) now use `WithSkipQuota`, so a would-be `sequenceBlock` denial is logged as an observed deny while the observed call still consumes no quota. `WithDryRun` keeps its preflight semantics; its godoc is corrected to state it also skips `sequenceBlock`/history. MaxCalls observation in `--audit` remains intentionally inexact (skipped), since observing it would require consuming the quota it must leave untouched.

## Testing
- `go test -race -count=1 ./...` — all packages pass (no audit-mode behavior tests regressed; the existing `WithDryRun` preflight tests are unchanged since `WithDryRun`'s behavior is unchanged).
- New regression tests in `pkg/enforcement/sequence_block_test.go`:
  - `TestSequenceBlock_SkipQuotaObservesDenial_DryRunDoesNot` — `WithSkipQuota` observes the deny + records history; `WithDryRun` still skips it.
  - `TestMaxCalls_SkipQuotaDoesNotConsume` — `WithSkipQuota` does not consume MaxCalls quota.
- Threat-model changelog entry 1.42.

---


eunolabs (commented):
## Self-review

Reviewed the full diff: no substantive findings.

**Verified:**
- `WithDryRun`'s *behavior* is unchanged — I only added `WithSkipQuota` and switched the audit-mode call sites — so the preflight `/validate` paths and all existing `WithDryRun` tests are unaffected (confirmed: the full `-race` suite passes with no regressions).
- All eight switched call sites are audit-mode branches (`if rt.audit` / `if p.audit`); the only other `WithDryRun` mention in the transports is a comment at `http.go:829` referencing the context-flag *pattern*, left intact.
- `handleMaxCalls` skips on `isDryRun || skipQuota` (so audit mode consumes no quota); `handleSequenceBlock` and `recordSessionCall` still gate only on `isDryRun`, so audit mode evaluates `sequenceBlock` and records history. The flow is sound: `recordSessionCall` runs on the allow path with the same ctx, so the antecedent records history and the gated call's would-be-deny is observed.
- New tests assert both halves (sequenceBlock observed under `WithSkipQuota` but not `WithDryRun`; MaxCalls not consumed under `WithSkipQuota`).

`go test -race ./...`, `go vet`, `golangci-lint` (0 issues), `make check-license`, `gofmt` all clean. No further comments.

---

eunolabs:
✅ Ready to merge — session rev-20260616T1342Z-9a3b6f bug addressed (#707), self-review complete with no outstanding comments, tests passing (`go test -race ./...` locally + CI Test/Lint/Build and all demo/OPA/security checks green).

---

PR #713, opened 2026-06-16
Fix session rev-20260616T1350Z-1502c5 bugs

### LINKED ISSUE #708: [rev-20260616T1350Z-1502c5] HTTP session silently succeeds when upstream rejects initialize (missing msg.Error check in runInitHandshake)
## Bug Report

**Location:** `cmd/eunox-mcp/http.go:1636–1651` (`httpSession.runInitHandshake`)

**Reference:** https://github.com/eunolabs/eunox/blob/62e6ba756dc3fbc043e00a4aa6ad33c70322cffe/cmd/eunox-mcp/http.go#L1636-L1651

**Severity:** Medium — causes silent session creation with undefined upstream state when the upstream rejects initialization

---

### Description

`runInitHandshake` (the HTTP transport's upstream initialize handshake, used for local subprocess sessions) does not check whether the upstream returned a JSON-RPC error response. If the upstream rejects the `initialize` request with `msg.Error != nil`, the code falls through, silently skips populating `upstreamCaps`, and then sends `notifications/initialized` — signalling a successful handshake to an upstream that explicitly failed.

```go
// http.go – runInitHandshake (lines 1636–1651)
if msg.isResponse() && msgKey(msg.ID) == msgKey(initID) {
    var result mcpInitResult
    if err := json.Unmarshal(msg.Result, &result); err == nil {
        // msg.Result is nil when msg.Error is set, so Unmarshal fails silently.
        // upstreamCaps stays nil; no error is returned.
    }
    break  // ← exits the loop and sends notifications/initialized unconditionally
}
```

The stdio transport (`StdioProxy.initUpstream`) handles this correctly:

```go
// stdio.go – initUpstream (lines 340–342)
if msg.Error != nil {
    return fmt.Errorf("upstream initialize rejected: %s (code %d)", msg.Error.Message, msg.Error.Code)
}
```

The comment at `http.go:1644` even states that the block "Mirrors the identical block in http_remote.go and stdio.go" — but the error check is missing.

### Trigger / Reproduction

Any upstream MCP server that returns an error response to the `initialize` request (e.g., due to unsupported protocol version, authentication failure, or intentional rejection) will cause the HTTP transport to:

1. Silently continue past the error.
2. Return a session to the client with `upstreamCaps == nil` (defaulting to `{"tools":{}}` in `buildInitResponse`).
3. Forward subsequent tool/resource/prompt calls to an upstream that is in an undefined state (it never completed initialization).
4. Produce confusing downstream errors or silently dropping requests until the upstream terminates.

In contrast, the same upstream would cause `StdioProxy.initUpstream` to return an error immediately, aborting session creation and returning HTTP 500 to the client.

### Suggested Fix

Add the missing `msg.Error` guard to `runInitHandshake`, mirroring `StdioProxy.initUpstream`:

```go
if msg.isResponse() && msgKey(msg.ID) == msgKey(initID) {
    if msg.Error != nil {
        return fmt.Errorf("upstream initialize rejected: %s (code %d)",
            msg.Error.Message, msg.Error.Code)
    }
    var result mcpInitResult
    if err := json.Unmarshal(msg.Result, &result); err == nil {
        s.upstreamCaps = result.Capabilities
        if sv, ok := result.ServerInfo["version"].(string); ok {
            s.upstreamServerVersion = sv
        }
    }
    break
}
```

The same fix should be verified in `http_remote.go` if it has its own initialize handshake path.

---

**Session:** rev-20260616T1350Z-1502c5
**HEAD SHA:** 62e6ba756dc3fbc043e00a4aa6ad33c70322cffe
**Files reviewed this cycle:** `cmd/eunox-mcp/http.go`, `pkg/enforcement/glob.go`, `pkg/capability/token.go`

eunolabs:
## Bugs addressed (session rev-20260616T1350Z-1502c5)

- Closes #708 — **(Medium)** `runInitHandshake` (local-subprocess HTTP sessions) inspected only the `initialize` response's `Result`, never its `Error`. An upstream that returned a JSON-RPC error to `initialize` (unsupported protocol version, auth failure, deliberate refusal) left `Result` nil, so the capability-capture block silently no-oped and the handshake fell through to `notifications/initialized` — signalling a successful handshake to an upstream that explicitly failed, and returning the client a session backed by an upstream that never completed initialization (`upstreamCaps` defaulting to `{"tools":{}}`), with subsequent calls forwarded into undefined upstream state. The local path now rejects an error `initialize` response with a terminal error (aborting session creation), mirroring `StdioProxy.initUpstream`.

  The issue asked to verify `http_remote.go`; its `initRemoteUpstream` had the **same** gap (checked `respMsg.Result != nil` but not `respMsg.Error`), so this PR fixes both the local and remote HTTP handshake paths.

## Testing
- `go test -race -count=1 ./...` — all packages pass (the existing happy-path `TestRemoteUpstream_Initialize` still passes).
- New regression tests:
  - `TestRunInitHandshake_RejectsUpstreamErrorResponse` — local path returns an error and does **not** send `notifications/initialized` or populate `upstreamCaps`.
  - `TestRemoteUpstream_InitializeRejected` — remote path: an `initialize` rejection creates no session and sends no `notifications/initialized`.
- Threat-model changelog entry 1.43.

---


eunolabs (commented):
## Self-review

A fail-closed guard added to both HTTP init-handshake paths (`runInitHandshake` local, `initRemoteUpstream` remote), checking `Error` before the existing `Result` handling — mirroring `StdioProxy.initUpstream`. No behavior change for a successful `initialize` (the existing happy-path `TestRemoteUpstream_Initialize` still passes; the new guard only triggers when `Error != nil`, which previously fell through). The error message matches the stdio path's wording. Both new tests assert no session/notification leaks on rejection. `go test -race ./...`, `go vet`, `golangci-lint` (0 issues), `make check-license`, `gofmt` all clean. No further comments.

---

eunolabs:
✅ Ready to merge — session rev-20260616T1350Z-1502c5 bug addressed (#708), self-review complete with no outstanding comments, tests passing (`go test -race ./...` locally + CI Test/Lint/Build and all demo/OPA/security checks green).

---

Issue #686, opened 2026-06-16
Code review: 2 high, 2 medium, 1 low — sampling audit bypass, JWT missing RequestID, audit key fsync gap


eunolabs:
## Health Summary

The codebase is production-quality for a security-critical proxy: fail-closed posture is pervasive and consistently cited, the two-pass MaxCalls evaluation is architecturally sound, and the HMAC-SHA256 hash-chained audit log is correctly implemented. Two high-severity behavioral bugs exist: audit/wiretap mode silently hard-blocks server-initiated sampling (defeating a core guarantee of the mode), and JWT-only-allow responses omit `RequestID`/`DecidedAt` (breaking audit correlation for all JWT-gated calls). A medium-severity durability gap in audit key generation could render an entire log unverifiable after a power loss. The remaining findings are low-risk. Total fix cost for all five items is small — no structural changes needed.

---

## What's Done Well

- **Pervasive fail-closed posture** with issue-number citations (`// issue #215`, etc.) — every ambiguous path denies and logs.
- **Two-pass MaxCalls evaluation** correctly avoids quota consumption on calls that will be denied by other conditions.
- **`VerifyRecord` uses `json.Decoder.UseNumber()`** — the only correct way to round-trip large integers through HMAC without float64 precision loss.
- **`compositeCounterKey` length-prefix encoding** prevents colon-separator collisions that would allow one key to satisfy another's rate limit.
- **`copyDetails` deep clone** before the audit channel send prevents data races on the Details map.
- **`isKillSwitchDenial`** correctly bypasses the audit-mode observe gate — kill-switch blocks even in wiretap mode.
- **`boundAuditDetails`/`boundAuditObligations`/`boundEnvelopeField`** comprehensively defend against scanner buffer overflow from oversized upstream responses.

---

### [HIGH-1] `alwaysAllowPDP` doesn't implement `SamplingAuthorizer` — audit mode hard-denies sampling

**Location:** `cmd/eunox-mcp/pdp.go:99-107` (`decideSampling`); audit mode transport handler

**Priority/Effort/Risk:** High / S (3 lines) / Behavior change (fixes a bug)

**Problem:**

`decideSampling` type-asserts the PDP against `SamplingAuthorizer`. When the PDP is `alwaysAllowPDP` (audit/wiretap mode), the assertion fails and returns `denyResponse("SAMPLING_DENIED", ...)` with `AuditOnly: false`.

The transport observe gate:
```go
[...]

PR #714, opened 2026-06-16
fix: address code-review findings from issue #686

eunolabs:
Resolves all five findings from the issue #686 code review:

HIGH-1: alwaysAllowPDP now implements SamplingAuthorizer so audit/wiretap
mode observes server-initiated sampling instead of hard-blocking it. The
JWTPDP delegation is gated on innerEnforces() so an alwaysAllowPDP inner
(an unpoliced JWT route) is still not a sampling backstop — JWT mode keeps
failing closed without an explicit manifest opt-in.

HIGH-2: JWTPDP.Decide's JWT-only allow (no inner manifest) now stamps a
fresh RequestID and RFC3339Nano DecidedAt, matching every other allow path,
so audit records for JWT-gated calls keep their correlation fields.

MEDIUM-1: generateAndPersistAuditKey fsyncs the parent directory after the
link/rename publish so the new key's directory entry is durable across a
crash; without it a power loss could leave the whole log unverifiable.

MEDIUM-2: document that redactFields operates only on structural JSON object
keys and silently passes content embedded in string values/array elements;
clarified the same in the looksLikeUnparseableJSONContainer comment.

LOW-1: filterListResult takes a getField closure deriving the slice from the
list, making the "field must point into list" invariant type-enforced.

https://claude.ai/code/session_01Bx8s4QbDBtd9V1rmNAKiMP


cmd/eunox-mcp/pdp.go
```diff
@@ -197,6 +197,22 @@ func (p alwaysAllowPDP) DecidePromptGet(_ context.Context, _, _, _ string) capab
...
+// denied with SAMPLING_DENIED and AuditOnly=false, which the transport observe
+// gate (!isKillSwitchDenial && rt.audit) then hard-blocks even under --audit
+// because the denial is neither a kill-switch denial nor AuditOnly. That
+// silently broke any session using sampling in a mode whose whole purpose is
+// "observe everything, block nothing except kill-switch" (issue #686). Like the
+// other alwaysAllowPDP methods this carries no kill-switch check — a pure
+// wiretap route has no policy or kill manager wired to it; the kill switch is a
+// manifest/route concern (ManifestPDP.DecideSampling) that this passthrough mode
+// does not participate in.
+func (p alwaysAllowPDP) DecideSampling(_ context.Context, _, _ string) capability.EnforceResponse {
```
eunolabs, line 212: HIGH-1 looks correct and the gating against `innerEnforces()` in `JWTPDP.DecideSampling` properly prevents an `alwaysAllowPDP` inner from becoming a JWT-route sampling backstop. Worth flagging: this is a security-relevant behavior change — wiretap/audit mode now **forwards** `sampling/createMessage` (a server→host channel) where it previously failed closed. It correctly aligns the code with the documented promise in `architecture.md`/`threat-model-mcp.md` that `alwaysAllowPDP` forwards "everything", so this is the right fix.

However, `docs/threat-model-mcp.md` §6 has a version-history table that records security-relevant behavior changes (e.g. the 1.2/1.3 entries on fail-closed methods). Since this flips a previously-denied server-initiated channel to forwarded-in-observe-mode, consider adding a version-history line there so the security source-of-truth reflects it. No code change requested.

---

cmd/eunox-mcp/audit.go
```diff
@@ -2191,13 +2191,39 @@ func generateAndPersistAuditKey(keyPath string) ([][]byte, error) {
...
 		return nil, fmt.Errorf("publishing audit key file %q: %w", keyPath, err)
 	}
+	// fsync the parent directory so the new directory entry is durable, not just
+	// visible in the kernel page cache. tmp.Sync above made the key's DATA blocks
+	// durable, but a successful link(2)/rename(2) only updates the directory inode
+	// in cache; without this fsync a power loss between publish and the next
+	// OS-initiated flush can leave the key file absent on restart. A missing key
+	// renders every prior record in the log unverifiable — audit-verify fails on
+	// all of them — destroying the integrity guarantee entirely (issue #686).
+	syncDir(dir)
```
eunolabs, line 2210: MEDIUM-1 fix is correct: `tmp.Sync()` makes the key's data blocks durable, and this `syncDir(dir)` makes the directory entry durable after the `os.Link` publish — closing the crash window. The rename-fallback branch is also covered by its own `syncDir(dir)`.

One path intentionally not covered: the `os.IsExist` race-adopt branch (`readPublishedAuditKeys(..., "after create race")`) returns without an fsync, relying on the process that won the link race to have synced the directory itself. That's fine since the winner owns durability, but it's a non-obvious assumption — a one-line note there would save a future reader from thinking it's a missed `syncDir`. Treating `syncDir` as best-effort (log + continue) is the right call for a crash-recovery edge case.

---

eunolabs (commented):
Reviewed all five fixes against the surrounding code. The PR is clean and internally consistent — each finding is addressed correctly:

- **HIGH-1 (wiretap sampling):** `alwaysAllowPDP` now implements `SamplingAuthorizer` so `decideSampling`'s type assertion succeeds and observe mode forwards `sampling/createMessage` instead of hard-blocking it. The `JWTPDP.DecideSampling` delegation is correctly gated on `innerEnforces()`, so a JWT route wrapping an `alwaysAllowPDP` inner still fails closed without a manifest opt-in — matching the identity-only `Decide` path. Test coverage (`TestAlwaysAllowPDP_DecideSampling_...`, the rewritten gap test, and the `transport_unit_test` deny path now using a real no-opt-in `ManifestPDP`) is on point.
- **HIGH-2 (JWT-only allow correlation):** the JWT-only allow branch now stamps a fresh `RequestID` + RFC3339Nano `DecidedAt`, matching every other allow path; the `decideInner` branch correctly relies on the inner PDP to stamp. Verified this is the only JWTPDP allow return that didn't already inherit a downstream PDP's fields.
- **MEDIUM-1 (audit-key durability):** `syncDir` after the link/rename publish closes the directory-entry durability gap; `dir` is in scope and both publish paths are covered.
- **MEDIUM-2 / LOW-1:** the `redactFields` structural-only limitation is now documented in both the guide and the `looksLikeUnparseableJSONContainer` comment, and the `filterListResult` closure refactor genuinely makes the "field must point into list" invariant type-enforced.

Two non-blocking notes inline (a threat-model version-history entry for the HIGH-1 behavior change, and a clarifying note on the uncovered `os.IsExist` race path in `syncDir`). Neither requires a code change. LGTM.

---

cmd/eunox-mcp/pdp.go
eunolabs: Added a threat-model change-log entry (1.44) recording the HIGH-1 behavior change: audit/wiretap mode now forwards server-initiated `sampling/createMessage`, while noting the fail-closed posture is unchanged for policed and JWT-wrapped routes. Done in 70ce20c.

---

cmd/eunox-mcp/audit.go
eunolabs: Added a one-line note at the `os.IsExist` race-adopt branch explaining it intentionally skips `syncDir` (the link-race winner owns the directory fsync). Done in 70ce20c.

---

Issue #480, opened 2026-06-15
[rev-20260615T0110Z-758862] flushDropMarker bypasses boundAuditDetails, breaking the invariant asserted by writeRecord's comment


eunolabs:
**Location:** `cmd/eunox-mcp/audit.go:799–835` (`flushDropMarker`) and `cmd/eunox-mcp/audit.go:842–849` (`writeRecord` comment)

**Reference:** https://github.com/eunolabs/eunox/blob/affe5a66a16e39d9a6d0bd46391dac6ed24681be/cmd/eunox-mcp/audit.go#L799-L835 and https://github.com/eunolabs/eunox/blob/affe5a66a16e39d9a6d0bd46391dac6ed24681be/cmd/eunox-mcp/audit.go#L842-L849

**Severity:** Low (currently harmless; latent maintenance risk)

**Description:**

`writeRecord` begins with a comment that states an invariant that all callers are supposed to uphold:

```go
// Details is already bounded: record() applies boundAuditDetails before enqueue
// so the queued record (and thus its on-disk line) can never overflow the
// audit-verify scanner buffer or the chain-resume tail window ...
```

However, `flushDropMarker` constructs an `auditRecord` **directly** and calls `writeRecord` without going through `record()`:

```go
func (s *auditSink) flushDropMarker() {
    ...
    marker := auditRecord{
        ...
        Details: map[string]interface{}{
            "dropped":       newDrops,
            "dropped_total": cur,
        },
        ...
    }
    s.writeRecord(&marker)   // ← bypasses record() and boundAuditDetails
}
```

`boundAuditDetails` is never called on the drop-marker record's `Details`. The invariant asserted by the `writeRecord` comment does **not** hold for this code path.

**Why it is currently safe** — but fragile:

The marker's `Details` is hardcoded to two `int64` values (`dropped` and `dropped_total`). Their marshalled form is at most a few dozen bytes — far below the 1 MiB `auditDetailsTotalCap`. No scanner overflow can occur today.

**Why it is a maintenance risk:**

1. A developer extending the drop-marker `Details` (e.g., adding a `dropped_since_start` field or a list of approximate timestamps) would see the invariant comment in `writeRecord` and assume the bound is already enforced — it is not for this path. A large addition could push the record over the scanner limit without any compile-time or runtime warning.

[...]

PR #715, opened 2026-06-16
fix: address principal-review findings (JWT rotation, sampling hang, audit durability)

eunolabs:
Correctness:
- pdp_jwt: IdP-JWT validation now retries a forced JWKS refresh after a cached
  key fails signature verification for a kid-bearing token (refreshedForKid gate),
  matching JWKSClient.VerifyToken. Previously only kid-less tokens retried, so an
  IdP rotating key material under the same kid (rolling "current" slot) rejected
  valid tokens for up to a full cache TTL (#683). Adds a same-kid rotation test.
- http: a server-initiated request (sampling/createMessage, roots/list) that
  reaches no client SSE stream no longer hangs the upstream until session
  teardown. broadcast now reports delivery; broadcastServerRequest fails closed
  on non-delivery (untrack + JSON-RPC error to the upstream) and the audit record
  reflects the delivery failure (ENFORCEMENT_ERROR) instead of a forwarded allow
  (#648). Adds a fail-closed test and makes the round-trip test realistic.
- audit: the drainer now fsyncs once the record queue empties, bounding the window
  of records a power loss or hard kill can lose from the page cache after the
  in-memory chain head advanced — closing the steady-state gap that rotate()/Close()
  already guard.

Simplification / de-dup (no behavior change):
- Export capability.JWKSAlgorithms as the single asymmetric-only allowlist for both
  verification paths; delete the duplicated jwksAlgorithmsIDP.
- Extract resolveStringOrStringArray, shared by the allowedExtensions and
  recipientDomain conditions (identical string-or-array parsing kept in lock-step).
- Add capability.IsTargetType as the single source of truth for the namespace
  prefix set; splitEnginePrefix consults it instead of a second hardcoded list.
- Add auditSink.syntheticDenyMarker to collapse the four copy-pasted marker
  envelopes; inline the one-line forceRefreshReportingChange wrapper; remove a dead
  targetType assignment in recordSessionCall.

Polish:
- Rename actionMatchScore -> actionPermitted (it returns a bool, not a score).
- TrimSpace the --control-token flag value for parity with the env/file sources.

Tests: go test -race ./... green; go vet, gofmt, license checks clean.


eunolabs (commented):
## Review

**Verdict: looks good.** Build clean, `go vet` clean, and the affected tests pass locally (`SameKidKeyRotation`, `NoSubscriberFailsClosed`, `SamplingRoundTrip`, `MatchingEdgeCases`). The three correctness fixes are well-targeted and fail-closed; the de-dup refactors are behavior-preserving.

### Correctness — the three fixes

**JWT same-`kid` rotation (#683)** — Correct. The `refreshedForKid` gate is the right discriminator: the retry fires for `kid == "" || !refreshedForKid`, so a kid-bearing cache-hit token gets one rescue refresh while a token that already refreshed via the kid-miss branch stays terminal (no double-fetch). Genuinely brings `ValidateToken` in line with `JWKSClient.VerifyToken`, and amplification is bounded by `ForceRefreshForVerify`'s own rate-limit. New test exercises the real path (warm cache → rotate under same kid → must validate).

**HTTP fail-closed (#648)** — Correct and the most security-relevant change. `broadcast` now reports delivery; `broadcastServerRequest` untracks the ID and writes a JSON-RPC error to the upstream on non-delivery, and all three call sites audit `deny`/`ENFORCEMENT_ERROR` instead of a phantom `allow`. Verified `upWriter` is a mutex-guarded `msgWriter`, so the off-lock write is safe. No exploitable TOCTOU in the track→broadcast→take ordering — a subscriber that connects after `broadcast` returns `false` never received the message, so failing closed is correct.
- *By design:* a subscriber that exists but has a full buffer counts as non-delivery, turning a momentarily-slow client's server request into an `ENFORCEMENT_ERROR`. The comment calls this out; correct fail-closed tradeoff given there's no replay buffer.

**Audit `fsync`** — Sound. `len(s.records) == 0` gates the sync so bursts amortize to one fsync per drained batch while an idle proxy syncs promptly; the `s.f != nil` guard matches the existing lost-file path.
- *Only actionable item (non-blocking):* `_ = s.f.Sync()` swallows the error. A failing fsync (EIO) is **not** counted in `writeFailures`, unlike `writeRecord`'s failures — so a disk that accepts `Write` but fails `Sync` gives false durability confidence with no metric. Consider counting it, or a one-line note on why it's intentionally best-effort.

### De-duplication / cleanup — all behavior-preserving
- `resolveStringOrStringArray`: faithful extraction; the fail-closed taxonomy (non-string item / blank element → `CONDITION_FAILED`; wrong-type → `CONDITION_FAILED`; empty → `MISSING_CONTEXT`) is preserved verbatim with `condType` stamped per-caller. Keeps `allowedExtensions`/`recipientDomain` in genuine lock-step — highest-value refactor here.
- `capability.JWKSAlgorithms` exported as the single asymmetric-only allowlist; `jwksAlgorithmsIDP` deleted; both alg-confusion test comments updated.
- `IsTargetType` as single source of truth for the namespace prefix set; `splitEnginePrefix` consults it.
- `syntheticDenyMarker` collapses four copy-pasted envelopes and routes every marker through `boundAuditDetails`, so the writeRecord "Details already bounded" invariant now holds uniformly (#480).
- `recordSessionCall` dead-store removal: verified safe — the inner `targetType = req.Target.Type` was always overwritten by the unconditional authoritative block below it. Pure dead-store cleanup.
- `actionMatchScore` → `actionPermitted`, `forceRefreshReportingChange` inlined, `--control-token` TrimSpace — cosmetic/parity, correct.

### Conventions & coverage
- Each correctness fix ships a regression test; the existing round-trip test was made realistic (adds an SSE subscriber).
- CHANGELOG updated under `### Fixed`. No new top-level audit field, so no threat-model update required.
- Conventional Commit title, fail-closed invariant preserved throughout, no emojis.


---

eunolabs:
Addressed the one actionable item (the swallowed drain-time `fsync` error) in 1283302.

The drainer now counts a failed steady-state `fsync` in `writeFailures` and emits a one-shot stderr warning via a dedicated `syncErrWarned` flag — mirroring `writeRecord`'s existing Write-failure surfacing, so a disk that accepts `Write` but fails `Sync` is observable rather than silently trusted. Updated the `writeFailures` doc to list the fsync path as a contributing source.

---

PR #716, opened 2026-06-17
refactor: unify the two JWT verifiers behind one retry skeleton

eunolabs:
## What

Follow-up to #715. Extracts the duplicated JWT verification choreography shared by `JWKSClient.VerifyToken` (capability tokens) and `JWTPDP.ValidateToken` (IdP JWTs) into a single generic skeleton, `capability.VerifyWithKeyRotation[T]`.

## Why

Both verifiers hand-mirrored ~110 lines of identical logic: parse → select candidate keys for `kid` → force a JWKS refresh on a kid miss → try each key → and, when every key fails signature verification, force one rate-limited refresh and retry. That mirror had **already drifted once** — the IdP path was missing the same-kid rotation retry the native path had (#683, fixed in #715) — so the duplication was a live source of divergence bugs.

After this change each verifier contributes only its per-key claim validation as a closure; the `kid`-miss refresh, the `refreshedForKid` gate, and the single forced-refresh retry live in exactly one place and cannot drift again.

## Shape

`VerifyWithKeyRotation[T any](ctx, cache, kid, verify func(*jose.JSONWebKey) (*T, terminal, sig error)) (*T, error)` owns the loop; the closure keeps the existing fail-closed tri-state contract:
- `(result, nil, nil)` — full success
- `(nil, terminal, nil)` — signature verified but claims invalid; surfaced as-is, never retried (#408/#475)
- `(nil, nil, sig)` — signature failure; try the next candidate / one refreshed set

Net **−55 lines** (346 insertions, 401 deletions) — this deletes duplication rather than adding an abstraction layer.

## Behavior-preserving

Same control flow, same fail-closed semantics. The only observable change is that the IdP path now shares the capability path's error-wording set; verified every test assertion depends only on the stable `"refresh JWKS"` / `"verify signature"` substrings (#562), which are preserved.

## Verification

`go test -race ./...` green across all packages; `go vet`, `golangci-lint` (0 issues), `gofmt`, and license-header checks clean. The skeleton is exercised through the existing rotation / kid-miss / signature-failure / audience / issuer tests on both paths.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GgnvEvSxKTEJ5GBTvjhRUQ)_


pkg/capability/jwks.go
```diff
@@ -110,43 +110,57 @@ var JWKSAlgorithms = []jose.SignatureAlgorithm{
...
+// When every candidate (across at most one forced refresh) fails signature
+// verification, the last signature error is wrapped as "verify signature". A
+// forced-refresh failure is surfaced as "refresh JWKS ..." rather than the
+// cached-key signature error, so a transient JWKS outage during key rotation is
+// distinguishable from a forged token (#562).
+func VerifyWithKeyRotation[T any](
+	ctx context.Context,
+	cache *JWKSCache,
+	kid string,
+	verify func(key *jose.JSONWebKey) (result *T, terminal, sig error),
```
eunolabs: Contract ergonomics (non-blocking): `verify` returns two bare `error`s whose **position** carries all the meaning — `terminal` (signature OK, claims bad → surface as-is) vs `sig` (signature bad → try next key / one refresh). Swap them at a call site and a claims failure silently gets retried against other keys and a forced refresh (masking the real reason behind a later key's error or a refresh error), or a signature failure becomes un-retryable. The doc comment above documents this contract very well, but the type system gives a future closure author no help.

Worth considering a tiny typed signal — e.g. a distinct sentinel wrapper (`terminalErr{err}`) or a small result struct/enum — so the intent is encoded rather than only commented. Two callers today makes this borderline over-engineering, but since this is now an exported, shared helper, it's cheap insurance.

---

pkg/capability/jwks.go
```diff
@@ -157,196 +171,59 @@ func (c *JWKSClient) VerifyToken(ctx context.Context, tokenStr string) (*TokenPa
...
-			// RFC 7519 string-or-array form, so a multi-aud token parses (it used to be
-			// rejected at unmarshal) and every value is retained rather than truncated to
-			// the first (#571). Fall back to the validated standard claims' audience when
-			// the native payload did not carry one.
-			if len(p.Audience) == 0 && len(claims.Audience) > 0 {
-				p.Audience = Audience(claims.Audience)
-			}
-
-			return &p, nil, nil
+			lastSig = sig
```
eunolabs: Edge case in the tri-state handling (non-blocking): if `verify` ever returns `(nil, nil, nil)` — all three nil — `tryKeys` treats it as a signature failure with `lastSig = nil` and falls through. If that's the last/only candidate and the refresh path doesn't rescue it, the function ends at `return nil, fmt.Errorf("verify signature: %w", lastErr)` with `lastErr == nil`, i.e. a `verify signature: %!w(<nil>)` string rather than a meaningful fail-closed message.

Neither current closure can produce all-nil, so there's no live bug. But since this is now a shared helper future closures will plug into, a one-line guard — treat `r == nil && sig == nil` as a terminal "verifier returned no result/error" — would keep the fail-closed message intelligible if someone wires up a closure that slips through.

---

cmd/eunox-mcp/pdp_jwt.go
```diff
@@ -343,220 +343,141 @@ func (p *JWTPDP) ValidateToken(ctx context.Context, authHeader string) (context.
...
-	}
+	// The per-key verifier is the only IdP-JWT-specific part; the key-selection and
+	// rotation-retry choreography (kid-miss refresh, per-key trial, the
+	// same-kid/kid-less forced-refresh retry) is shared with the capability-token
+	// path via capability.VerifyWithKeyRotation so the two cannot drift (#683, #501).
+	// The closure returns the tri-state that helper expects: (claims, nil, nil) on
+	// success, (nil, terminal, nil) for a signature-verified-but-claims-invalid
+	// failure that must not be retried (#475), and (nil, nil, sig) for a signature
+	// failure (try the next candidate / a refreshed set).
+	validated, err := capability.VerifyWithKeyRotation[JWTClaims](ctx, p.cache, kid, func(key *jose.JSONWebKey) (*JWTClaims, error, error) {
```
eunolabs: Flagging for reviewers: the PR is behavior-preserving in **control flow and fail-closed semantics**, but the IdP path's user-visible error *strings* do change by adopting the capability path's wording:

- `fetching JWKS: …` → `fetch JWKS: …`
- `refreshing JWKS: …` → `refresh JWKS: …`
- `JWT signature verification failed: …` → `verify signature: …`
- `refresh JWKS for kid-less token: …` → `refresh JWKS after signature failure: …`

I checked the tests and this is safe: nothing asserts the old strings positively — the two `signature verification failed` checks in `pdp_jwt_test.go` (~L1046, ~L4077) are *negative* assertions, and L4074 keys on the stable `refresh JWKS` substring (preserved). Just noting that error *text* is not preserved on the IdP path (only the substrings the tests rely on, plus the structured fail-closed behavior), in case any external log-scrapers or runbooks key off the old IdP wording.

---

eunolabs (commented):
## Review: unify the two JWT verifiers behind one retry skeleton

**Overview.** Extracts the shared JWKS key-selection + rotation-retry choreography out of `JWKSClient.VerifyToken` (capability tokens) and `JWTPDP.ValidateToken` (IdP JWTs) into one generic helper, `capability.VerifyWithKeyRotation[T]`. Each verifier now contributes only its per-key claim validation as a closure; the kid-miss refresh, the `refreshedForKid` gate, and the single forced-refresh retry live in exactly one place. Net −55 lines, no new dependencies.

This is the right call — the duplication had already drifted once (#683), so collapsing it removes a real class of divergence bug rather than adding an abstraction layer.

**Correctness — no bugs found.** I traced both callers against the pre-PR code:
- the tri-state (`result` / `terminal` / `sig`) handling in the new generic `tryKeys` matches the old per-path loops (terminal returns immediately, success returns, signature failure records `lastSig` and continues);
- the `kid == "" || !refreshedForKid` forced-refresh guard is preserved verbatim on both paths;
- the per-key signature-error shape is preserved — capability wraps `key %q: …`, IdP returns the raw error — so the final `verify signature: …` strings keep their existing form;
- fail-closed posture is intact on every branch (fetch error, kid-miss exhausted, terminal claims failure, exhausted signatures, refresh failure).

**Findings (all non-blocking):**
1. **Contract ergonomics** — `verify`'s two bare `error` returns encode meaning purely by position; great doc comment, no type-level guardrail. (inline on `jwks.go`)
2. **All-nil edge** — a closure returning `(nil, nil, nil)` degrades to `verify signature: %!w(<nil>)`. No live path hits it; a one-line guard hardens the now-shared helper. (inline on `jwks.go`)
3. **IdP error-string changes** — semantics preserved, wording adopts the capability path's; verified no test breaks. (inline on `pdp_jwt.go`)
4. **Test coverage** — the helper is currently exercised only *transitively* through the two callers' existing rotation / kid-miss / signature-failure tests; no test was added/changed in this PR. Since `VerifyWithKeyRotation` is now an exported `pkg/` helper with a subtle contract, a small direct table-driven test with a fake `verify` closure — asserting (a) `terminal` is never retried, (b) `sig` is retried exactly once after a forced refresh, (c) a refresh error is surfaced over the cached-key sig error — would pin the contract independently of the two real callers and guard findings 1–2. CLAUDE.md gates `pkg/` at 80% and asks for tests on new `pkg` paths.

**Conventions.** SPDX/Apache headers intact, no emojis, Conventional-Commit title, no new runtime deps. No docs change required (no manifest-grammar / audit-shape / schema change). One process note: the combined status was still `pending` with 0 checks reported when I looked — worth confirming `make test` (`-race`) and `golangci-lint` are green on `491edf8` before merge, since the PR body's "all green" claim isn't yet reflected by CI.

Overall: a clean, well-reasoned deduplication that genuinely reduces future drift risk. The notes above are polish, not blockers.

---

eunolabs:
Addressed the review in `ed26151`.

1. **Contract ergonomics** — replaced the verifier's two positional `error` returns with a single error plus an exported `Terminal(err)` marker. A plain error is a retryable signature failure; a `Terminal`-wrapped one is surfaced as-is and never retried. The terminal-vs-sig distinction is now type-encoded, so a future verifier can't silently swap the two. Both `VerifyToken` and `ValidateToken` updated to the single-error form.
2. **All-nil edge** — `tryKeys` now treats a `(nil, nil)` verifier return as an explicit fail-closed error (`"per-key verifier returned neither a result nor an error"`) instead of letting it degrade into `verify signature: %!w(<nil>)`.
3. **IdP error-string changes** — confirmed, no code change; this was the intended consequence of sharing the capability path's wording, and the note for log-scrapers stands. (The structured fail-closed behavior and the test-relied-upon `refresh JWKS` / `verify signature` substrings are preserved.)
4. **Direct test** — added `TestVerifyWithKeyRotation_Contract` (table-driven, fake verifier) pinning: (a) a terminal failure is surfaced as-is and never retried, (b) a signature failure retries **exactly once** against a refreshed key set, (c) a forced-refresh failure is surfaced over the cached-key signature error, (d) the `(nil, nil)` contract violation fails closed.

`go test -race ./...` green; `go vet`, `golangci-lint` (0 issues), `gofmt`, license all clean.

---

PR #717, opened 2026-06-17
fix: harden transports and audit accuracy; unify enforced-forward

eunolabs:
Addresses the remaining deferred review items from the earlier code review.

## Changes

**A4 — unbounded goroutine spawn on the stdio transport**
`serveHost` now bounds host-request concurrency to 64 in-flight via a semaphore instead of `go handleHostRequest(...)` per request. A buggy/runaway local host that pipelines requests without reading responses hits pipe backpressure rather than growing goroutines without limit. The stdio transport is single/local/trusted, so this is robustness, not a remote-DoS defense; the HTTP transport was already bounded by `maxSessions`.

**A5 — JWKS cache handed out a live alias to the key set**
`FindKeys` now always returns a fresh copy, never an alias into the cached `*jose.JSONWebKeySet`. The cache serves the same set to every concurrent verification, so a caller that sorted/appended/rewrote the returned slice could corrupt the shared root-of-trust set. Defensive — no caller mutates it today; this keeps it true by construction. (`jwks.Key(kid)` already returned a new slice.)

**Audit accuracy — `allow` recorded on an upstream JSON-RPC error**
An allowed call whose upstream returns a JSON-RPC *error result* now records an `upstream_error_code` entry in `details`, so a forwarded upstream error is distinguishable from a clean success on the tape. The decision stays an honest `allow` — the proxy authorized and forwarded the call, and the audit decision field records the *proxy's* verdict, not the upstream's outcome. Only the numeric code is recorded, never the upstream message (which can carry sensitive content). Applies to the enforced methods and the `*/list` enumerations on both transports.

**B5 — `enforcedForward` mirrored across http/stdio**
Extracts the hand-mirrored deny/observe/forward/record logic into one shared `enforcedForwardCore` (`forward.go`) behind a small `auditRecorder` interface (`*auditSink` and `*routeSink` both satisfy it). The transports now differ only in delivery — HTTP returns the message, stdio writes it — so the load-bearing policy/audit logic lives in one place and can't drift.

**#9 — Origin port asymmetry (documented)**
Documents why `originAllowed` matches `allowedOrigins` exactly (including port) while the loopback/bind host-set path accepts any port: DNS rebinding is a host-based attack, not a port-based one, and any local port on a loopback host is equally local.

**#10 — `--trust-forwarded-for` footgun (startup warning)**
Logs a `SECURITY` warning at startup when `--trust-forwarded-for` is enabled on a non-loopback bind, where a directly-reachable client could forge `X-Forwarded-For` and spoof an `ipRange` condition's source IP. No change to the IP-parsing behavior itself — this flags the deployment constraint so ingress can be restricted to the trusted hop.

## Tests
New: `enforcedForwardCore` (allow / hard-deny-skips-upstream / upstream-error detail), `upstreamErrorDetail`, `bindIsLoopbackOnly`, and `FindKeys`-returns-a-copy. The B5 refactor is also covered by the existing enforced-forward tests on both transports.

`go test -race ./...` green; `go vet`, `golangci-lint` (0 issues), `gofmt`, and license-header checks all clean. CHANGELOG updated.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GgnvEvSxKTEJ5GBTvjhRUQ)_


cmd/eunox-mcp/stdio.go
```diff
@@ -379,9 +390,13 @@ func (p *StdioProxy) serveHost(ctx context.Context) {
 			continue
 		}
 		if msg.isRequest() {
+			// Acquire before spawning so in-flight requests stay bounded; a saturated
+			// semaphore blocks the read here, applying pipe backpressure to the host.
+			sem <- struct{}{}
```
eunolabs: **Potential deadlock / stall when the upstream issues server-initiated requests (`sampling/createMessage`, `roots/list`)**

Acquiring `sem` here blocks the `serveHost` read loop — but that same loop is the *only* path that routes the host's replies to server-initiated requests back to the upstream (the `isResponse()` branch just above, stdio.go:389-398). That creates a dependency cycle:

1. A pipelining host sends more than `maxStdioInflightRequests` (64) requests. `serveHost` dispatches 64, fills `sem`, and blocks here on the 65th.
2. Each in-flight `handleHostRequest` is parked in `callUpstream`'s `select { case resp := <-ch: ... }` (stdio.go:872), holding its slot, waiting for the upstream's reply.
3. To answer a `tools/call`, the upstream issues `sampling/createMessage` (or `roots/list`) back toward the host. The host's reply arrives on stdin **behind** the 65th request in the pipe.
4. `serveHost` can't read that reply — it's blocked on `sem <- struct{}{}` — so `ch` is never fulfilled, so no slot ever frees, so the loop never unblocks.

With the default `--upstream-timeout` (30s) this is a 30s stall that then fails all 64 parked calls with `-32603`; with `--upstream-timeout 0` (timeout disabled) it's a permanent hang. This is exactly the "runaway/pipelining host" scenario the change targets — for sampling/`roots`-capable upstreams the bound turns "many goroutines (but correct)" into "stall-or-deadlock," and the "pipe backpressure naturally throttles the host" reasoning in the doc comment doesn't hold when the throttled reader is also the in-band response demux.

Root cause: request-dispatch backpressure is gated on the same goroutine that must keep servicing the response stream. Possible directions:
- Decouple read from dispatch — a reader that always handles `isResponse()` inline and hands requests to a bounded worker pool, so backpressure never stalls the response path.
- Or scope the bound so server-initiated round-trips can't be starved (don't count requests that are themselves blocked on a host reply).

Upstreams that never issue server→client requests are unaffected, so this is conditional — but `sampling` and `roots/list` are first-class MCP and supported here (cf. #648), so it's worth resolving before merge.

---

cmd/eunox-mcp/http.go
```diff
@@ -434,6 +434,22 @@ func (p *HTTPProxy) Serve(ctx context.Context) error {
 		fmt.Fprintf(os.Stderr, "[eunox-mcp] HTTP proxy listening on http://%s%s\n", ln.Addr(), path)
 	}
 
+	// --trust-forwarded-for is only safe when EVERY client reaches the proxy through
+	// the trusted reverse proxy that sets X-Forwarded-For. On a non-loopback bind the
+	// proxy may also be reachable directly, and a direct client can forge
+	// X-Forwarded-For to spoof an ipRange condition's source IP (sourceIP reads the
+	// right-most XFF hop). Warn loudly at startup so this deployment footgun is not
+	// silent; the operator must restrict ingress to the trusted hop.
+	if p.trustFwdFor && !bindIsLoopbackOnly(p.bind) {
```
eunolabs, line 443: Nit (correct as written — `p.bind` is a bare host here, so `bindIsLoopbackOnly(p.bind)` is the right input, and this is a genuinely useful, actionable warning).

One thing: `main.go:620` already emits an unconditional `WARNING: --trust-forwarded-for is enabled ...` whenever the flag is set, so on a non-loopback bind the operator now gets two overlapping warnings for the same flag. Consider folding the generic one into this bind-aware check (or downgrading it) so this `SECURITY:` line is the single, specific signal.

---

eunolabs (commented):
## Review: transport + audit hardening (PR #717)

Overall this is high-quality work. The intent of each change is clearly documented, the typed-nil interface trap in the `enforcedForward` → `enforcedForwardCore` extraction is handled correctly (the concrete sink is assigned into `auditRecorder` only when non-nil, so the core's `fp.rec != nil` is a true "no sink" test), and `FindKeys` returning a fresh copy is a sound defensive fix with a good test. The new `upstream_error_code` audit detail also round-trips safely through sign/verify, because `VerifyRecord` already decodes `Details` with `UseNumber` — so the int re-marshals byte-for-byte and the HMAC still matches.

**One issue I'd resolve before merge** (inline on `stdio.go`): the new in-flight semaphore can deadlock/stall the stdio transport when the upstream issues server-initiated requests (`sampling/createMessage`, `roots/list`), because it blocks the same read loop that routes the host's replies for those requests. Full reasoning inline — short version: with the default `--upstream-timeout` it's a 30s stall + a burst of `-32603`s; with the timeout disabled (`0`) it's a permanent hang, and it triggers under the very runaway/pipelining-host scenario the change is meant to harden against.

Minor / optional:
- **B5 / convention:** CONTRIBUTING asks for a sign-and-verify round-trip test for new audit-record fields. `upstream_error_code` is a new *key* in `details` rather than a new top-level field, and the existing synthetic-marker tests already exercise int-in-`details` via `UseNumber`, so it's covered indirectly — but an explicit round-trip asserting an `allow` record carrying `upstream_error_code` verifies clean would lock the convention down for the live path.
- **Docs:** consider documenting the new `upstream_error_code` detail key where audit semantics live (e.g. `docs/threat-model-mcp.md`), so SIEM consumers understand it records the *upstream's* outcome on a proxy `allow`, not a proxy denial. (It isn't a top-level field, so the "new top-level field ⇒ update threat model" invariant doesn't strictly apply.)
- **trust-forwarded-for:** minor warning redundancy with `main.go:620` — noted inline.

Nice touches: the `#9` origin port-asymmetry comment and the `#10` non-loopback startup warning are well-reasoned documentation of deliberate decisions, and the B5 consolidation genuinely removes the drift risk between the two transports' enforced-forward paths.

---

eunolabs:
Addressed the review in `06604a3`.

**Blocking — stdio semaphore deadlock (A4):** You're right, and thanks for the precise trace. The read loop is also the response demux for server-initiated round-trips, so any bound that blocks dispatch starves the very replies the in-flight handlers wait on. I worked through whether a correct bound was salvageable: keeping the reader non-blocking is mandatory (it owns stdin), which forces either unbounded goroutines (the original) or an unbounded queue — the latter just trades goroutine growth for memory growth, not a clear win. For a single-tenant, trusted, local transport that isn't worth a reader/dispatcher redesign, so I **reverted A4** to the correct one-goroutine-per-request form and left a comment recording exactly why the read loop must not block.

**XFF warning redundancy:** Folded — dropped the unconditional `WARNING` in `serveHTTPGateway` so `HTTPProxy.Serve`'s bind-aware `SECURITY` line (non-loopback bind only) is the single, specific signal.

**Round-trip test (convention):** Added `TestEnforcedForwardCore_UpstreamErrorDetailSignsAndVerifies` — drives the real allow-with-upstream-error path through a live `*auditSink`, then asserts the persisted record carries `upstream_error_code` **and** verifies clean under the chain HMAC.

**Docs:** Documented the `upstream_error_code` detail in `docs/threat-model-mcp.md` §5.5, distinguishing a forwarded upstream JSON-RPC error (honest `allow` + code, no message) from a transport failure (deny + `UPSTREAM_ERROR`).

The other five items in the PR are unchanged. `go test -race ./...` green; `go vet`, `golangci-lint` (0 issues), `gofmt`, license all clean.

---

eunolabs (commented):
## Re-review @ 06604a3 — prior findings verified

Re-checked all four points from the last pass against the new commit; all are resolved.

1. **Stdio deadlock (the blocker) — fixed by reverting the semaphore.** `serveHost` is back to one-goroutine-per-request, with a comment recording *why* the read loop must never block on dispatch (it's also the demux for host replies to server-initiated `sampling`/`roots` requests). Confirmed no `maxStdioInflightRequests`/`sem` references remain — the net change vs `main` is just the comment. Trading the A4 bound away is the right call here: stdio is single-tenant/local/trusted, and a correct bound would mean decoupling response-demux from request-dispatch.
2. **XFF warning redundancy — fixed.** The unconditional warning in `serveHTTPGateway` is gone; the bind-aware `SECURITY` warning in `HTTPProxy.Serve` is the single signal. Verified the gateway path (`NewHTTPProxyGateway` → `TrustFwdFor: pf.trustFwdFor` → `Serve`) still reaches it, so no coverage gap.
3. **Sign-and-verify round-trip test — added.** `TestEnforcedForwardCore_UpstreamErrorDetailSignsAndVerifies` drives the real allow-with-upstream-error path through a live `*auditSink`, flushes the drainer, and asserts the persisted record both carries `upstream_error_code` and verifies clean under the chain HMAC. The helpers it uses all exist. This is exactly the round trip the convention asks for.
4. **Threat-model doc — added.** The new paragraph in `docs/threat-model-mcp.md` cleanly separates a completed round-trip returning a JSON-RPC `error` *result* (honest `allow` + `upstream_error_code`) from an infrastructure failure (deny), and restates the code-not-message rule.

**CI:** green except `Analyze (go)` (CodeQL) still in progress; CodeQL's other run is `neutral`, not a failure. No failing checks — the `unstable` mergeable state is just the pending analysis.

**One non-blocking nit:** the PR *description* is now stale — the **A4** section still says "`serveHost` now bounds host-request concurrency to 64 in-flight via a semaphore," which is the reverted approach. Worth updating the body (and the "harden transports" framing) so the description matches the merged code.

Otherwise this LGTM.

---

PR #733, opened 2026-06-17
Fix automated-review session bugs (64dd5e, be6961, 0b4ba9)

### LINKED ISSUE #722: [rev-20260617T0249Z-0b4ba9] JWTPDP JWT-only allow path stamps DecidedAt with time.Now() instead of p.clock
**Location:** `cmd/eunox-mcp/pdp_jwt.go:663-667`
**Reference:** https://github.com/eunolabs/eunox/blob/c28b88773c6bda9338cdb4562db22abc6d871116/cmd/eunox-mcp/pdp_jwt.go#L663-L667
**Severity:** Medium

**Description:**

The JWT-only allow path in `JWTPDP.Decide()` (the path taken when the JWT has capabilities and they pass, and there is no inner manifest PDP) hardcodes `time.Now()` when building the `DecidedAt` field of the `EnforceResponse`:

```go
    DecidedAt: time.Now().UTC().Format(time.RFC3339Nano),  // wrong
}
```

`JWTPDP` has a `clock enforcement.Clock` field injected via `JWTPDPOptions.Clock` precisely so the time source can be overridden in tests (and for internal consistency). Every other path that produces a timestamp in the same file and the sibling `pdp.go` respects `p.clock`:

- `ValidateToken()` (same file, line 411–416) uses `p.clock` if non-nil and falls back to `time.Now()`.
- `alwaysAllowPDP.wiretapAllow()` (`pdp.go`, lines 173–177) uses `p.clock` the same way.

The JWT-only allow path was added to fix issue #686 (blank `request_id`/`decided_at` in audit records), but the fix omitted the `p.clock` lookup, leaving this one path always on the real wall clock.

**Trigger / repro:**

Any test that:
1. Creates a `JWTPDP` via `NewJWTPDP` with a frozen `Clock` injected in `JWTPDPOptions.Clock`,
2. Issues a JWT with valid capabilities (so `HasCapabilities == true`),
3. Has no inner manifest PDP (`opts.Inner == nil`),
4. Calls `Decide()` and inspects `resp.DecidedAt`.

The `DecidedAt` timestamp will reflect real wall-clock time, not the frozen test time. Any assertion comparing it to the frozen instant will fail non-deterministically.

In production this causes no logical error (the right clock is used for JWT expiry validation, which is the security-critical part), but the audit log's `decided_at` field diverges from the time the enforcement engine reports for the same request epoch in frozen-clock integration tests, making test assertions on audit-record timing unreliable.

**Suggested fix:**

Mirror the pattern `ValidateToken()` and `alwaysAllowPDP.wiretapAllow()` use:

```go
now := time.Now
if p.clock != nil {
    now = p.clock.Now
}
    DecidedAt: now().UTC().Format(time.RFC3339Nano),
}
```

**Session:** rev-20260617T0249Z-0b4ba9

### LINKED ISSUE #723: [rev-20260617T0249Z-0b4ba9] circuitbreaker: comment overclaims that State() and Stats() are jointly consistent near the cooldown boundary
**Location:** `pkg/circuitbreaker/breaker.go:444-448` (comment on `projectedState`)
**Reference:** https://github.com/eunolabs/eunox/blob/c28b88773c6bda9338cdb4562db22abc6d871116/pkg/circuitbreaker/breaker.go#L444-L448
**Severity:** Low

**Description:**

The comment on `projectedState()` makes a guarantee about joint consistency across two *separate* public calls that the implementation cannot uphold:

> "Both State() and Stats() route their projection through this helper so a caller that reads both in rapid succession near the cooldown boundary cannot observe one reporting Open and the other HalfOpen: each public call samples now() once, and there is a single source of projection truth (#612)."

The guarantee *within* a single call is correct — `State()` calls `projectedState()` once, which calls `b.now()` once, so the state and any derived fields it returns are internally consistent. Same for `Stats()`.

The guarantee *across* two separate calls is not correct. `State()` and `Stats()` each acquire and release `b.mu.RLock()` independently:

```go
    b.mu.RLock()
    defer b.mu.RUnlock()
    state, _, _ := b.projectedState()  // calls b.now() — call A
    return state
}

    b.mu.RLock()
    defer b.mu.RUnlock()
    state, ... := b.projectedState()  // calls b.now() — call B
    ...
}
```

Between call A's `b.now()` and call B's `b.now()` the lock is released and re-acquired. If the cooldown elapses during that gap — small, but non-zero — `State()` can return `Open` while `Stats().State` returns `HalfOpen`. The comment says this *cannot* happen; it can.

**Trigger / repro:**

The scenario requires timing precision that makes it very hard to reproduce deterministically in production. In a test with a step-advancing `WithClock` clock:

1. Configure a breaker with a short `CooldownDuration` (e.g., 1 step).
2. Trip the breaker to `Open`.
3. Advance the fake clock past the cooldown boundary.
4. Call `b.State()` — it sees T_before_boundary from the fake clock (still Open).
5. Advance the clock one more step.
6. Call `b.Stats()` — it now sees T_after_boundary (projected HalfOpen).
7. `State()` returned Open; `Stats().State` returns HalfOpen. Both are logically stale, but they disagree, violating the guarantee in the comment.

In practice any code that reads `b.State() == StateHalfOpen` and then `b.Stats()` expecting `stats.State == StateHalfOpen` can observe the mismatch.

**Suggested fix:**

Two options:

Option A (safest, minimal change): Correct the comment to say the guarantee is per-call, not cross-call:

```go
// projectedState reports the logical state ... Must be called under b.mu.RLock (or b.mu.Lock).
//
// Both State() and Stats() route their projection through this helper so that WITHIN A SINGLE
// CALL each is internally self-consistent: one b.now() sample backs the entire snapshot (#612).
// A caller that needs both fields to agree atomically should call Stats() once and read
// Stats.State, rather than calling State() and Stats() as two separate calls.
```

Option B (stronger): Add a `StateAndStats()` method that acquires the lock once and returns both under a single `b.now()` call, then document that this is the way to get a jointly-consistent snapshot.

**Session:** rev-20260617T0249Z-0b4ba9

### LINKED ISSUE #725: [rev-20260617T0708Z-be6961] JWTPDP.filterList silently returns empty list when JWT claims are absent, while Decide returns ErrCodeNoJWTClaims
**Location:** `cmd/eunox-mcp/pdp_jwt.go` — `filterList` function, `!ok` branch (approximately lines 1025-1043)
**Reference:** https://github.com/eunolabs/eunox/blob/c7d09997ad7120d51570fd8fb441281912d4f449/cmd/eunox-mcp/pdp_jwt.go#L1025-L1043
**Severity:** Medium

**Description:**

There is a behavioral inconsistency between `JWTPDP.Decide()` and `JWTPDP.filterList()` when no JWT claims are present in the request context.

`Decide()` hard-fails with `ErrCodeNoJWTClaims` when the JWT claims key is absent from the context:

```go
claims, ok := jwtClaimsFromContext(ctx)
if !ok {
    return ..., ErrCodeNoJWTClaims  // structured error, upstream never called
}
```

`filterList()` takes a different path. When `!ok`, it falls through to `byClaims(result, nil)`. `byClaims` with a nil claims argument filters every entry out of the list, returning an empty slice. That empty slice is then passed to the inner PDP's `filterList` (if any). No error is ever returned or logged.

The result: a host that receives an allow from the inner PDP (when `inner != nil`) gets back an empty `*/list` response instead of a `ErrCodeNoJWTClaims` error. The caller has no indication that the empty list is caused by a missing JWT rather than a legitimately empty upstream. This also means the audit record for the list operation does not capture the "no JWT" condition, which can obscure misconfiguration in production.

**Trigger / repro:**

1. Configure a `JWTPDP` with an inner manifest PDP and a manifest that permits some tools.
2. Issue a `tools/list` request with no JWT in the context (e.g., misconfigured client, header stripped by middleware).
3. `Decide()` → returns `ErrCodeNoJWTClaims`.
4. `filterList()` (if the transport calls it separately, or on a different code path) → returns `[]any{}` with no error.

The gap widens if the transport calls `filterList` independently of `Decide`, which `HTTPProxy` does on certain paths.

**Suggested fix:**

Mirror the `Decide` guard at the top of `filterList`:

```go
func (p *JWTPDP) filterList(ctx context.Context, method string, result []any) ([]any, error) {
    claims, ok := jwtClaimsFromContext(ctx)
    if !ok {
        return nil, newRPCError(ErrCodeNoJWTClaims, "no JWT claims in context")
    }
    // ... rest of existing logic
}
```

Alternatively, if returning an error from `filterList` would break callers that treat it as best-effort, the `!ok` branch should at minimum log an audit record and pass the full unfiltered list to the inner PDP, preserving the behavior that `Decide` would have produced had it been called first (i.e., fail closed by denying rather than silently returning empty).

**Session:** rev-20260617T0708Z-be6961

### LINKED ISSUE #726: [rev-20260617T0708Z-be6961] handlers: parseTableArgument silently drops blank column strings instead of returning ConditionError
**Location:** `pkg/enforcement/handlers.go` — `parseTableArgument`, JSON array branch (approximately lines 1530-1532)
**Reference:** https://github.com/eunolabs/eunox/blob/c7d09997ad7120d51570fd8fb441281912d4f449/pkg/enforcement/handlers.go#L1530-L1532
**Severity:** Medium

**Description:**

`parseTableArgument` parses the `columns` argument for the `allowedTables` condition. When `columns` is a JSON array, blank elements are silently discarded:

```go
for _, col := range arr {
    if col != "" {           // blank element: silently skipped
        ta.Columns = append(ta.Columns, col)
    }
}
```

This is inconsistent with how blank values are treated elsewhere in the handler layer. `resolveStringOrStringArray` (the shared argument resolver used by `allowedValues`, `allowedExtensions`, `allowedOperations`, and `recipientDomain`) treats blank array elements as a `ConditionError`:

```go
if s == "" {
    return nil, condErr(cond, argName, "blank entry in string array")
}
```

The table-specific parser predates this shared resolver and was never updated to match the stricter validation.

**Impact:**

A manifest entry like:

```yaml
conditions:
  - type: allowedTables
    arguments:
      table: "users"
      columns: ["id", "", "secret"]
```

silently normalizes to `["id", "secret"]`. The manifest author who accidentally typed an empty string (or whose YAML generator emitted one) gets no error, no warning, and no audit trace. They likely intended `["id", "secret"]` but the silently-dropped blank masks the fact that `"secret"` was included. In a tighter case, `["id", "", ""]` becomes `["id"]` — a different effective policy than the author specified, with no indication of the discrepancy.

Manifest-load-time validation (`validateAllowedTables`) does not catch this because it validates the parsed `tableArgument` struct after `parseTableArgument` has already dropped the blanks.

**Trigger / repro:**

```yaml
conditions:
  - type: allowedTables
    arguments:
      table: "reports"
      columns: ["id", "", "name"]
```

Load the manifest. No error is returned. The effective constraint is `columns: ["id", "name"]`.

**Suggested fix:**

Replace the silent skip with a `ConditionError`:

```go
for _, col := range arr {
    if col == "" {
        return tableArgument{}, condErr(cond, "columns", "blank entry in columns array")
    }
    ta.Columns = append(ta.Columns, col)
}
```

This mirrors `resolveStringOrStringArray` and surfaces the malformed manifest at load time rather than silently altering the enforced policy.

**Session:** rev-20260617T0708Z-be6961

### LINKED ISSUE #727: [rev-20260617T0708Z-be6961] callcounter: PeekRetryAfter allocates valid []time.Time slice on every call, including the common allow path where it is immediately discarded
**Location:** `pkg/callcounter/memory.go` — `PeekRetryAfter`, window scan loop (approximately lines 471-479)
**Reference:** https://github.com/eunolabs/eunox/blob/c7d09997ad7120d51570fd8fb441281912d4f449/pkg/callcounter/memory.go#L471-L479
**Severity:** Low

**Description:**

`PeekRetryAfter` builds a `valid []time.Time` slice for every call to compute the retry-after duration. On the common allow path — where `cur < limit` — this slice is constructed in full and then immediately discarded because the function returns `(0, nil)` as soon as it detects the count is below the limit.

Compare with `Peek`, which solves the same count problem by simply incrementing a counter:

```go
// Peek — allocates nothing
var cur int64
for _, ts := range entry.timestamps {
    if now.Sub(ts) < window {
        cur++
    }
}
```

`PeekRetryAfter`:
```go
// PeekRetryAfter — always allocates a full slice
var valid []time.Time
for _, ts := range entry.timestamps {
    if now.Sub(ts) < window {
        valid = append(valid, ts)
    }
}
cur := int64(len(valid))
if cur < limit {
    return 0, nil   // valid is discarded here; allocation was wasted
}
// ... compute retry-after from valid
```

`PeekRetryAfter` is called on every `maxCalls`-gated request check (via `handleMaxCalls` with `ctxMaxCallsCheckOnlyKey`). Under sustained load this generates one slice allocation per check, all of which the GC must collect. The slice is only needed on the deny path (when `cur >= limit`), which is a minority of calls in well-behaved deployments.

**Trigger / repro:**

Any deployment with `maxCalls` conditions under moderate to high call volume, benchmarked with `go test -bench=BenchmarkPeekRetryAfter` or observable via `runtime.ReadMemStats` comparing allocations before and after requests.

**Suggested fix:**

Two-phase approach — count first, only build the slice on the deny path:

```go
var cur int64
for _, ts := range entry.timestamps {
    if now.Sub(ts) < window {
        cur++
    }
}
if cur < limit {
    return 0, nil
}
// Only on deny: rebuild to find the oldest timestamp for retry-after
var valid []time.Time
for _, ts := range entry.timestamps {
    if now.Sub(ts) < window {
        valid = append(valid, ts)
    }
}
// ... existing retry-after computation on valid
```

This matches the `Peek` pattern for the allow path and only pays the allocation cost on the deny path, which is already more expensive (it also computes a `retryAfterSeconds` duration).

**Session:** rev-20260617T0708Z-be6961

### LINKED ISSUE #728: [rev-20260617T0708Z-be6961] handlers: handleTimeWindow allows calls at exactly the notAfter boundary (inclusive upper bound)
**Location:** `pkg/enforcement/handlers.go` — `handleTimeWindow` (approximately lines 225-235)
**Reference:** https://github.com/eunolabs/eunox/blob/c7d09997ad7120d51570fd8fb441281912d4f449/pkg/enforcement/handlers.go#L225-L235
**Severity:** Low

**Description:**

`handleTimeWindow` checks whether the current time falls within the operator-specified `[notBefore, notAfter]` window. The upper-bound check uses `now.After(notAfter)`:

```go
if now.After(notAfter) {
    return deny(...)    // deny only when now is STRICTLY after notAfter
}
```

`time.Time.After(t)` returns `false` when `now == notAfter` (i.e., the times are equal), so a call whose timestamp is exactly equal to `notAfter` is **allowed**. The upper bound is inclusive.

The lower bound uses `now.Before(notBefore)`, which similarly allows a call at exactly `notBefore` (`now == notBefore` → not before → allowed). The lower bound is also inclusive. Window semantics are therefore `[notBefore, notAfter]` — a closed interval on both ends.

This behavior is not documented in `docs/capability-manifest-guide.md` or in any condition-level comment. Operators writing a manifest that says "allow until 17:00:00" likely expect the window to close at 17:00:00, not to allow exactly one more call at that instant.

The gap is only observable when a request timestamp lands on the exact nanosecond of the boundary — extremely unlikely in production but deterministically triggerable in tests with a frozen `Clock`. For a monitoring or security window (e.g., "maintenance window: no calls between T1 and T2") the boundary call is a quiet policy violation.

**Trigger / repro:**

With a frozen `Clock` set to exactly `notAfter`:

```go
now := notAfter                    // or notBefore
eng := NewEngine(WithClock(...))   // clock returns now
result := eng.Decide(ctx, ...)     // should deny; actually allows
```

**Suggested fix:**

Two options:

**Option A (behavior fix):** Change to a half-open `[notBefore, notAfter)` interval using `!now.Before(notAfter)` for the upper bound:

```go
if !now.Before(notAfter) {   // deny when now >= notAfter (exclusive upper bound)
    return deny(...)
}
```

And correspondingly document that the interval is `[notBefore, notAfter)`.

**Option B (documentation fix):** If closed-interval semantics are intentional (e.g., for `notAfter` expressed to second precision where a nanosecond-exact hit is negligible), document `[notBefore, notAfter]` explicitly in `docs/capability-manifest-guide.md` so operators know the boundary is included. Add a test that asserts this behavior for boundary values.

Option A is safer because it matches the common expectation that time windows have exclusive upper bounds (as in `[from, until)` scheduling conventions).

**Session:** rev-20260617T0708Z-be6961

### LINKED ISSUE #729: [rev-20260617T0807Z-64dd5e] handlers: resolveStringOrStringArray rejects native []string causing spurious CONDITION_FAILED in allowedExtensions and recipientDomain
**Location:** `pkg/enforcement/handlers.go` — `resolveStringOrStringArray`, lines 645–691
**Reference:** https://github.com/eunolabs/eunox/blob/c7d09997ad7120d51570fd8fb441281912d4f449/pkg/enforcement/handlers.go#L645-L691
**Severity:** Medium

**Description:**

`resolveStringOrStringArray` is the shared argument parser for the `allowedExtensions`, `recipientDomain`, `allowedOperations`, and `allowedValues` conditions. It handles two concrete types:

```go
switch t := v.(type) {
case string:
    if s := strings.TrimSpace(t); s != "" {
        out = []string{s}
    }
case []interface{}:
    for _, item := range t { ... }
default:
    return nil, &ConditionError{
        Message: fmt.Sprintf("argument %q must be a string or array of strings, got %T", argName, v),
    }
}
```

The `default` branch is hit whenever a library caller populates `EnforceRequest.Arguments` with a native `[]string` value:

```go
req := enforcement.EnforceRequest{
    Arguments: map[string]interface{}{
        "extension": []string{".go", ".ts"},
    },
}
```

`EnforceRequest.Arguments` is typed `map[string]interface{}` with no documented restriction on concrete value types. `json.Unmarshal` always produces `[]interface{}`, so the production proxy path is unaffected, but any library consumer — including tests, SDK wrappers, or the `policy` condition handler that constructs synthetic `EnforceRequest` objects internally — can naturally produce `[]string`. When they do, the condition always returns `CONDITION_FAILED` with a misleading error ("got []string") rather than evaluating the constraint.

This violates the fail-closed invariant in a subtle direction: the denial is correct but the *reason* is wrong. The audit record will attribute the denial to a type mismatch, not to the actual policy, making forensic analysis and debugging harder.

**Trigger / repro:**

```go
engine := enforcement.New(manifest)
result := engine.Decide(ctx, enforcement.EnforceRequest{
    ToolName: "read_file",
    Arguments: map[string]interface{}{
        "extension": []string{".go"},  // native []string
    },
})
// result.Decision == Deny, reason: "argument \"extension\" must be a string or array of strings, got []string"
// Even if .go is in allowedExtensions
```

**Suggested fix:**

Add a `[]string` case before `default`:

```go
case []string:
    for _, s := range t {
        s = strings.TrimSpace(s)
        if s == "" {
            return nil, condErr(cond, argName, "blank entry in string array")
        }
        out = append(out, s)
    }
```

This keeps parity with the `[]interface{}` branch (same blank-entry validation) and eliminates the spurious type error for library callers.

**Session:** rev-20260617T0807Z-64dd5e

### LINKED ISSUE #730: [rev-20260617T0807Z-64dd5e] handlers: parseTableArgument silently ignores native []string for columns, leaving column ACL unenforced
**Location:** `pkg/enforcement/handlers.go` — `parseTableArgument`, map branch, lines 1524–1535
**Reference:** https://github.com/eunolabs/eunox/blob/c7d09997ad7120d51570fd8fb441281912d4f449/pkg/enforcement/handlers.go#L1524-L1535
**Severity:** Medium

**Description:**

Note: this is distinct from issue #726 (blank string drop). This bug is about a native `[]string` type assertion failing silently.

`parseTableArgument` parses the `allowedTables` condition argument. When the argument is a `map[string]interface{}` (structured table spec), it extracts the `columns` field:

```go
case map[string]interface{}:
    ta.Table, _ = t["table"].(string)
    if cols, ok := t["columns"].([]interface{}); ok {  // type assertion
        for _, c := range cols {
            if s, ok := c.(string); ok && s != "" {
                ta.Columns = append(ta.Columns, s)
            }
        }
    }
    // if ok==false (e.g. columns is []string), ta.Columns stays nil — no error
    return []capability.TableAccess{ta}, nil
```

If a library caller (or an internal builder that assembles `map[string]interface{}` directly) sets `columns` to a native `[]string`, the type assertion `.([]interface{})` returns `ok=false`. The function does **not** return an error; it returns a `TableAccess` with a nil `Columns` slice.

A nil `Columns` means "allow all columns". The `allowedTables` handler interprets a nil column list as unrestricted:

```go
if len(ta.Columns) > 0 {
    // check requested columns against ta.Columns
} // else: any column is fine
```

The net effect: a manifest that intends to restrict column access (`columns: ["id", "name"]`) silently becomes a wildcard column allow when `columns` is supplied as `[]string`. The column-level ACL is quietly dropped with no error, no audit trace distinguishing this from an intentional unrestricted grant.

**Trigger / repro:**

```go
req := enforcement.EnforceRequest{
    Arguments: map[string]interface{}{
        "table": map[string]interface{}{
            "table":   "users",
            "columns": []string{"id", "secret_hash"},  // native []string
        },
    },
}
// allowedTables condition with columns: ["id"] — should deny "secret_hash"
// actual: column check silently skipped, request ALLOWED
```

**Suggested fix:**

Add a `[]string` arm to the column extraction:

```go
var cols []string
switch cv := t["columns"].(type) {
case []interface{}:
    for _, c := range cv {
        if s, ok := c.(string); ok && s != "" {
            cols = append(cols, s)
        }
    }
case []string:
    cols = cv
case nil:
    // no columns key: unrestricted (intentional)
default:
    return nil, condErr(cond, "columns", fmt.Sprintf("columns must be an array of strings, got %T", t["columns"]))
}
ta.Columns = cols
```

This closes the silent bypass and adds an explicit error for unexpected types, consistent with fail-closed behavior.

**Session:** rev-20260617T0807Z-64dd5e

### LINKED ISSUE #731: [rev-20260617T0807Z-64dd5e] handlers: asInt64/toFloat64 do not handle json.Number, causing allowedValues to always deny numeric arguments decoded with UseNumber()
**Location:** `pkg/enforcement/handlers.go` — `asInt64` and `toFloat64`, approximately lines 1131–1189; called by `numericEqual` (line ~1108) and `MatchAllowedValue` (line ~1103)
**Reference:** https://github.com/eunolabs/eunox/blob/c7d09997ad7120d51570fd8fb441281912d4f449/pkg/enforcement/handlers.go#L1131-L1189
**Severity:** Medium

**Description:**

`MatchAllowedValue` evaluates the `allowedValues` condition. For numeric arguments it delegates to `numericEqual`, which calls `asInt64` and `toFloat64`:

```go
func asInt64(v any) (int64, bool) {
    switch n := v.(type) {
    case int:     return int64(n), true
    case int64:   return n, true
    case float64: return floatToInt64(n)
    }
    return 0, false   // json.Number falls here
}

func toFloat64(v any) (float64, bool) {
    switch n := v.(type) {
    case float64: return n, true
    case int:     return float64(n), true
    case int64:   return float64(n), true
    }
    return 0, false   // json.Number falls here
}
```

Neither function handles `json.Number`. This matters because the audit log uses `json.NewDecoder` with `UseNumber()` to read records (preserving precision for large integers), and any code path that re-decodes a JSON blob with `UseNumber()` — including JWT claim parsing in `pkg/capability` and any caller that passes a `json.Decoder`-decoded `Arguments` map — will produce `json.Number` values instead of `float64`.

When `req.Arguments["count"]` is a `json.Number` (e.g. `"42"`):
- `asInt64(v)` returns `0, false`
- `toFloat64(v)` returns `0, false`
- `numericEqual` falls through to a string comparison between the `json.Number`'s string representation and the manifest's allowed value string
- The string comparison succeeds only for exact string identity (e.g. `"42"` == `"42"`) but fails for `"42"` vs. `"42.0"`, and completely fails for negative numbers or large integers compared against `float64`-serialized manifest values

More critically, the integer fast path in `numericEqual`:

```go
if iv, ok := asInt64(v); ok {
    // ...
}
```

never fires for `json.Number`, so even `42` compared against manifest value `42` returns false when the argument is `json.Number("42")` and the manifest value is an integer (stored as `int64`). The condition always returns `CONDITION_FAILED` regardless of whether the value is actually in the allowed list.

**Trigger / repro:**

Any pipeline that decodes request arguments via `json.NewDecoder(r).Decode(&args)` without calling `UseNumber()` is unaffected (produces `float64`). Any pipeline that *does* call `UseNumber()` — including tooling that re-parses saved audit records to replay requests, JWT-gated flows where claims are decoded with `UseNumber()`, or direct library usage with a `json.Number`-valued `Arguments` map — will see `allowedValues` conditions silently deny numeric arguments that should be allowed.

```go
var args map[string]interface{}
dec := json.NewDecoder(strings.NewReader(`{"count": 5}`))
dec.UseNumber()
_ = dec.Decode(&args)

req := enforcement.EnforceRequest{
    ToolName:  "read_rows",
    Arguments: args,  // args["count"] is json.Number("5")
}
// allowedValues condition: allowed: [5]
// result: CONDITION_FAILED — json.Number("5") not matched
```

**Suggested fix:**

Add `json.Number` handling to both helpers:

```go
import "encoding/json"

func asInt64(v any) (int64, bool) {
    switch n := v.(type) {
    case int:          return int64(n), true
    case int64:        return n, true
    case float64:      return floatToInt64(n)
    case json.Number:
        if i, err := n.Int64(); err == nil {
            return i, true
        }
        if f, err := n.Float64(); err == nil {
            return floatToInt64(f)
        }
    }
    return 0, false
}

func toFloat64(v any) (float64, bool) {
    switch n := v.(type) {
    case float64:      return n, true
    case int:          return float64(n), true
    case int64:        return float64(n), true
    case json.Number:
        if f, err := n.Float64(); err == nil {
            return f, true
        }
    }
    return 0, false
}
```

This ensures numeric comparisons work correctly regardless of whether the argument map was produced by `json.Unmarshal` (always `float64`) or `json.NewDecoder` with `UseNumber()`.

**Session:** rev-20260617T0807Z-64dd5e

eunolabs:
This batch resolves every open automated-review issue across the three review sessions still outstanding. All fixes land on one branch (the only branch this session may push to), one commit per issue.

### Session `rev-20260617T0807Z-64dd5e`
- Closes #731 — `asInt64`/`toFloat64` ignored `json.Number`, so numeric `allowedValues` comparisons silently denied any argument decoded with `json.Decoder.UseNumber()` (the float-only path fell through to a string compare). Added `json.Number` arms to both helpers.
- Closes #730 — `parseTableArgument`'s map branch asserted `columns` as `[]interface{}` only; a native `[]string` failed the assertion silently, leaving `Columns` nil (= allow all columns) and turning a column restriction into a wildcard. Replaced with a type switch handling `[]interface{}`, `[]string`, absent (`nil` = unrestricted), and an explicit error for other types.
- Closes #729 — `resolveStringOrStringArray` rejected a native `[]string` with a misleading `got []string` type error instead of evaluating the constraint. Added a `[]string` arm with the same blank-entry validation as `[]interface{}`.

### Session `rev-20260617T0708Z-be6961`
- Closes #728 — `handleTimeWindow` used `now.After(notAfter)`, making the upper bound inclusive (a call at exactly `notAfter` was allowed). Changed to `!now.Before(notAfter)` so the window is half-open `[notBefore, notAfter)`, matching the common `[from, until)` convention; documented in `docs/capability-manifest-guide.md`.
- Closes #727 — `PeekRetryAfter` allocated a `[]time.Time` slice on every call, including the common allow path where it was immediately discarded. Now counts first and only builds the slice on the deny path (mirrors `Peek`).
- Closes #726 — `parseTableArgument` silently dropped blank column strings, quietly altering the enforced policy. Now fails closed with `CONDITION_FAILED`, matching `resolveStringOrStringArray`.
- Closes #725 — `JWTPDP.filterList` returned an empty list when JWT claims were absent without distinguishing that from a legitimately empty upstream, and (with an inner backstop) ran the inner filter on an already-empty list. Added an explicit no-claims guard that fails closed (deny-all) without consulting the inner PDP, mirroring `Decide`'s `ErrCodeNoJWTClaims`.

### Session `rev-20260617T0249Z-0b4ba9`
- Closes #723 — corrected the `projectedState` comment that overclaimed `State()` and `Stats()` are jointly consistent across two separate calls; the guarantee is per-call only.
- Closes #722 — the JWT-only allow path stamped `DecidedAt` with `time.Now()` instead of the injected `p.clock`, making frozen-clock audit-timing assertions non-deterministic. Now resolves `now` from `p.clock` like `ValidateToken` and `alwaysAllowPDP.wiretapAllow`.

## Testing
- Added focused regression tests for each behavior fix: `json.Number` cases in `TestNumericEqual`; `TestEngine_AllowedExtensions_NativeStringSlice`; `TestEngine_AllowedTables_NativeStringColumns` and `TestEngine_AllowedTables_BlankColumnFailsClosed`; `TestEngine_TimeWindow_Boundaries`; `TestJWTPDP_Decide_JWTOnlyAllow_UsesInjectedClock`; `TestJWT_FilterToolsList_NoClaims_WithInner_FailsClosed`.
- `go build ./...`, `go vet ./...`, and `gofmt -l` clean.
- Full `go test ./...` passes; `go test -race -count=1` passes on the four touched packages (`pkg/enforcement`, `pkg/callcounter`, `pkg/circuitbreaker`, `cmd/eunox-mcp`).

https://claude.ai/code/session_01YXPM5X6xGT1wprjJoLt3Pu

---
_Generated by [Claude Code](https://claude.ai/code/session_01YXPM5X6xGT1wprjJoLt3Pu)_


eunolabs:
✅ Ready to merge — all outstanding automated-review bugs addressed across sessions 64dd5e (#731, #730, #729), be6961 (#728, #727, #726, #725), and 0b4ba9 (#723, #722). Self-review complete with no outstanding comments; all 43 CI checks (build, test, lint, gofmt, license, NOTICE, govulncheck, CodeQL, demo/OPA suites) pass.

---

PR #753, opened 2026-06-18
Fix session rev-20260617T1544Z-795679 bugs

### LINKED ISSUE #735: [rev-20260617T1544Z-795679] ForceRefresh singleflight race causes force=true caller to silently receive cached JWKS
## Location

`pkg/capability/jwkscache.go` — `ForceRefresh` / `ForceRefreshForKID`, approximately lines 354–436.

## Reference

Reviewed at commit `ad7b63c599c27e1ee346c2b47d12423206e7fb88`.

## Severity

**Medium-High** — can block JWT verification for up to 30 seconds during key rotation, causing denial of service for valid tokens.

## Description

`ForceRefresh(ctx, force bool)` deduplicates concurrent fetches via a single singleflight group (`sfGroup`). The singleflight key is the same regardless of the `force` argument, so a `force=true` caller that arrives while a `force=false` request is already in-flight joins the existing group and inherits its closure.

The TTL short-circuit (`if !force { ... return cached }`) is **inside** the singleflight closure, parameterised by the `force` value of the group leader. When the leader is `force=false` and the cache is fresh, the closure returns early without performing an HTTP round-trip. The `force=true` joiner receives the cached key set as if the forced refresh succeeded.

The method's own docstring is therefore incorrect:

> The joined fetch always performs a fresh HTTP round-trip (the TTL fast-path is outside the singleflight closure)

This is false when the group leader is non-forced.

## Trigger / Repro

1. A background TTL refresh (`force=false`) is in-flight.
2. A new-kid JWT arrives; `ForceRefreshForKID` calls `ForceRefresh(ctx, true)`.
3. The `force=true` call joins the `force=false` group.
4. The `force=false` closure's double-checked TTL finds the cache still fresh; returns without HTTP.
5. `ForceRefreshForKID` sees `changed=false`, calls `markKIDAbsent(kid)`.
6. The kid is now in `negKIDs` for `negativeKIDTTL` (30 s). All tokens carrying that kid are denied for 30 s even though the key exists at the JWKS endpoint.

The race is tight but reproducible under moderate load when key rotation coincides with a background refresh cycle.

## Suggested Fix

Use separate singleflight keys for forced and non-forced fetches, e.g. `"forced"` vs `"background"`. Alternatively, move the `if !force { ... }` TTL guard outside the `sfGroup.Do` closure so that a non-forced caller that finds the cache fresh returns early before entering the group, leaving forced callers a clean path to their own inflight slot.

Additionally, correct the docstring to match the actual behaviour.

## Session

`rev-20260617T1544Z-795679`

### LINKED ISSUE #736: [rev-20260617T1544Z-795679] JWT expiry leeway hardcoded at 60 s in JWTPDP, not operator-configurable
## Location

`cmd/eunox-mcp/pdp_jwt.go` — `ValidateToken`, call to `stdClaims.ValidateWithLeeway`.

## Reference

Reviewed at commit `ad7b63c599c27e1ee346c2b47d12423206e7fb88`.

## Severity

**Medium** — tokens expired up to 59 seconds ago are silently accepted with no way for operators to tighten or disable this.

## Description

`ValidateToken` validates standard JWT claims with a fixed 1-minute leeway:

```go
if err := stdClaims.ValidateWithLeeway(expected, time.Minute); err != nil {
    return nil, fmt.Errorf(...)
}
```

`ValidateWithLeeway` with `time.Minute` accepts a token whose `exp` is up to 60 seconds in the past (the leeway is added to `exp` before comparison). There is no `Leeway` field in `JWTPDPOptions` or any corresponding config key, so operators cannot reduce this window.

A 60-second grace period is large relative to the attack surface: a stolen short-lived token has a reuse window that is up to a minute longer than the IdP intended.

## Trigger / Repro

Issue a JWT with `exp = time.Now() - 30s`. Present it to a JWTPDP instance. The token passes `ValidateWithLeeway` and is forwarded to the inner PDP. The MCP host's user sees an allowed response; the IdP's intent (token expired) is violated.

## Suggested Fix

1. Add `Leeway time.Duration` to `JWTPDPOptions`; default to a conservative value (5–15 s is common practice).
2. Pass `opts.Leeway` (or `max(opts.Leeway, 0)`) to `ValidateWithLeeway`.
3. Document the field in `docs/capability-manifest-guide.md` and update the JSON schema if the leeway is expressible in the gateway config.
4. Use the injected `Clock` for all time arithmetic in `ValidateToken` to keep tests hermetic.

## Session

`rev-20260617T1544Z-795679`

### LINKED ISSUE #737: [rev-20260617T1544Z-795679] JWTPDP.Decide calls time.Now() unconditionally before checking injected clock
## Location

`cmd/eunox-mcp/pdp_jwt.go` — `Decide()`, first few lines of the function body.

## Reference

Reviewed at commit `ad7b63c599c27e1ee346c2b47d12423206e7fb88`.

## Severity

**Low** — no incorrect behaviour in production (the discarded value is never used); tests with an injected clock are correct. The issue is a wasteful syscall on every `Decide` call and a misleading code pattern.

## Description

`Decide()` always calls the real wall clock before checking whether a test-injected clock is set:

```go
now := time.Now()        // real syscall, always
if p.clock != nil {
    now = p.clock.Now()  // overwrites if clock injected
}
```

In production (`p.clock == nil`) this is a no-op in terms of correctness. In tests (`p.clock != nil`) the `time.Now()` result is immediately discarded — the injected clock is used — but the syscall still happens on every enforcement call.

The pattern also makes it easy for a future contributor to copy this idiom into a path where the overwrite is missing, silently bypassing the injected clock.

## Suggested Fix

Invert the check so the real clock is only called when no injection is present:

```go
var now time.Time
if p.clock != nil {
    now = p.clock.Now()
} else {
    now = time.Now()
}
```

Or introduce a small helper `(p *JWTPDP) now() time.Time` that centralises the pattern across the file.

## Session

`rev-20260617T1544Z-795679`

### LINKED ISSUE #738: [rev-20260617T1544Z-795679] boundFieldTo(s, 0) silently returns "" with no truncation marker in audit log
## Location

`cmd/eunox-mcp/audit.go` — `boundFieldTo(s string, limit int)`.

## Reference

Reviewed at commit `ad7b63c599c27e1ee346c2b47d12423206e7fb88`.

## Severity

**Low** — no current caller passes `limit=0`, but the silent empty-string return is a latent correctness hazard for audit integrity.

## Description

`boundFieldTo` truncates long strings to `limit` bytes and appends a truncation marker (e.g. `…<trunc>`). When `limit` is zero:

```
keep = 0 - len(marker)  // negative
end  = max(0, keep)     // 0
return s[:0] + marker[:0]  // ""
```

The function returns an empty string with no truncation marker. An audit consumer cannot distinguish:

- The original field was genuinely empty.
- The field had content that was silently dropped because `limit=0`.

This violates the "never put free-form text in a structured field" invariant in spirit: the structured field becomes a silent `/dev/null` for any `limit ≤ len(marker)` values, not just `limit=0`.

## Trigger / Repro

```go
boundFieldTo("sensitive content", 0)  // returns ""
boundFieldTo("sensitive content", 3)  // returns "" if marker is longer than 3 bytes
```

## Suggested Fix

Add a guard at the top of the function:

```go
if limit <= 0 {
    // Callers must provide a positive limit; treat as zero-length cap.
    // Return marker truncated to zero — or panic in debug builds.
    return ""
}
```

Then separately add an assertion (or documented invariant) that all callers must supply `limit > len(marker)` so at least the truncation marker is preserved. Alternatively, always return the marker untruncated when `len(s) > 0 && limit < len(marker)`:

```go
if len(s) <= limit {
    return s
}
if limit < len(marker) {
    return marker  // best-effort: signal truncation even if limit too small
}
```

## Session

`rev-20260617T1544Z-795679`

### LINKED ISSUE #739: [rev-20260617T1544Z-795679] ForceRefreshForVerify marks sentinel absent even when no HTTP fetch occurred
## Location

`pkg/capability/jwkscache.go` — `ForceRefreshForVerify`, approximately lines 235–266.

## Reference

Reviewed at commit `ad7b63c599c27e1ee346c2b47d12423206e7fb88`.

## Severity

**Low** — a sub-case of the ForceRefresh singleflight race (see related issue), specifically affecting kid-less token flows (signature-algorithm-only tokens or similar). Suppresses verify retries for up to 30 seconds.

## Description

`ForceRefreshForVerify` is the rate-limiting path for tokens without a `kid` header. After calling `ForceRefresh(ctx, true)` it inspects the `changed` flag:

```go
if !changed {
    c.markKIDAbsent(verifyRetrySentinel)
}
```

The intent is: if the JWKS didn't change, rate-limit further forced refreshes via the negative-kid cache. However, `changed=false` does not exclusively mean "HTTP fetch returned the same keys." It also occurs when:

1. A concurrent `force=false` singleflight was already in flight and returned a cached set (per the singleflight race described in the related issue) — no HTTP round-trip was performed at all.
2. A concurrent `ForceRefreshForVerify` from another goroutine completed first and already updated the cache.

In both cases the sentinel is marked absent, suppressing further `ForceRefreshForVerify` calls for `negativeKIDTTL` (30 s) — even though no actual HTTP fetch confirmed the keys are unchanged.

This is particularly problematic during a key rotation where the IdP has added a new key: the first forced-verify caller joins a non-forced singleflight, gets the old cache, marks the sentinel absent, and blocks all other kid-less verify callers from triggering a real refresh for 30 s.

## Trigger / Repro

Same race as the related ForceRefresh singleflight issue, restricted to code paths that go through `ForceRefreshForVerify` (kid-less tokens). Timing: any concurrent background TTL refresh in flight when a new key is needed.

## Suggested Fix

The root fix is to separate forced from non-forced singleflight keys so that a `force=true` call always runs its own HTTP fetch. Once that is in place, `changed=false` reliably means "HTTP confirmed no change" and `markKIDAbsent` is safe.

As an independent improvement: `ForceRefresh` could return an additional boolean `fetchOccurred` (distinct from `changed`) so callers like `ForceRefreshForVerify` can distinguish "cache hit via singleflight coalescing" from "HTTP confirmed unchanged."

## Related

This issue is a downstream consequence of the singleflight race filed in the companion issue `[rev-20260617T1544Z-795679] ForceRefresh singleflight race causes force=true caller to silently receive cached JWKS`.

## Session

`rev-20260617T1544Z-795679`

### LINKED ISSUE #740: [rev-20260617T1544Z-795679] ForceRefresh singleflight race: force=true caller silently receives cached JWKS
## Location

`pkg/capability/jwkscache.go` — `ForceRefresh` / `ForceRefreshForKID`, approximately lines 354-436.

## Reference

Reviewed at commit `ad7b63c599c27e1ee346c2b47d12423206e7fb88`.

## Severity

**Medium-High** — can block JWT verification for up to 30 seconds during key rotation, causing denial of service for valid tokens.

## Description

`ForceRefresh(ctx, force bool)` deduplicates concurrent fetches via a single singleflight group (`sfGroup`). The singleflight key is the same regardless of the `force` argument, so a `force=true` caller that arrives while a `force=false` request is already in-flight joins the existing group and inherits its closure.

The TTL short-circuit (`if !force { ... return cached }`) is **inside** the singleflight closure, parameterised by the `force` value of the group leader. When the leader is `force=false` and the cache is still fresh, the closure returns early without an HTTP round-trip. The `force=true` joiner receives the cached key set as if the forced refresh succeeded.

The method docstring is therefore incorrect:

> The joined fetch always performs a fresh HTTP round-trip (the TTL fast-path is outside the singleflight closure)

This is false when the group leader is non-forced.

## Trigger / Repro

1. A background TTL refresh (`force=false`) is in-flight.
2. A new-kid JWT arrives; `ForceRefreshForKID` calls `ForceRefresh(ctx, true)`.
3. The `force=true` call joins the `force=false` group.
4. The `force=false` closure's double-checked TTL finds the cache still fresh; returns without HTTP.
5. `ForceRefreshForKID` sees `changed=false`, calls `markKIDAbsent(kid)`.
6. The kid is now in `negKIDs` for `negativeKIDTTL` (30 s). All tokens carrying that kid are denied for 30 s even though the key exists at the JWKS endpoint.

## Suggested Fix

Use separate singleflight keys for forced and non-forced fetches (e.g. `"forced"` vs `"background"`). Alternatively, move the `if !force { ... }` TTL guard outside the `sfGroup.Do` closure so a non-forced caller that finds the cache fresh returns before entering the group, leaving forced callers a clean path to their own in-flight slot. Correct the docstring to match actual behaviour.

## Session

`rev-20260617T1544Z-795679`

eunolabs:
## Bugs addressed (session rev-20260617T1544Z-795679)

- Closes #735 — ForceRefresh shared one singleflight key for forced and non-forced fetches, so a `force=true` caller joining an in-flight `force=false` leader inherited its TTL fast-path and silently received the cached key set. Now keyed by `force` ("forced" vs "background"), so forced callers always run their own HTTP round-trip. Docstring corrected.
- Closes #740 — duplicate of #735; same root cause, fixed by the singleflight key separation.
- Closes #739 — downstream of the same race: `ForceRefreshForVerify` marked its sentinel absent on `changed=false` even when no HTTP fetch occurred (a coalesced cache hit). With separate keys a forced refresh always reflects a real fetch, so `changed=false` reliably means "HTTP confirmed unchanged".
- Closes #737 — `JWTPDP.Decide` called `time.Now()` unconditionally before checking the injected clock (wasteful syscall + an easy-to-miss idiom). Added a `now()` helper that consults the injected clock first; used in both `Decide` and `ValidateToken`.
- Closes #736 — JWT exp/nbf/iat leeway was hardcoded at one minute with no override (a stolen short-lived token's reuse window was up to 60s longer than the IdP intended). Added `JWTPDPOptions.Leeway` and a `--jwt-leeway` flag (default 10s; `0` disables), threaded through both constructors and the gateway route wrappers; documented in the capability-manifest guide.
- Closes #738 — `boundFieldTo` returned an empty string (or a confusing prefix of the long marker) when the limit was smaller than the marker, making a silently-dropped audit field indistinguishable from a genuinely empty one. Now emits the shortest recognizable marker (`...`) that fits while preserving the `<= limit` invariant; only a non-positive limit (a caller bug) returns empty.

## Testing
- `go vet ./...`, `make check-fmt`, `make check-license` — clean.
- New/updated regression tests:
  - `TestJWKSCache_ForceRefresh_DoesNotJoinInFlightNonForced` (rewrites the old test that pinned the buggy coalescing) asserts a forced refresh issues its own fetch.
  - `TestJWTPDP_ValidateToken_ConfigurableLeeway`, `TestEffectiveLeeway`, `TestJWTLeewayOption` pin the leeway window and resolution rules.
  - `TestBoundFieldToNeverSilentlyEmpties` asserts truncated content is never silently emptied for any usable limit.

---


eunolabs:
✅ Ready to merge — all session rev-20260617T1544Z-795679 bugs addressed (#735, #736, #737, #738, #739, #740), self-review complete with no outstanding comments, all 27 CI checks passing.

---

PR #754, opened 2026-06-18
Fix session rev-20260618T0010Z-563cce bugs

### LINKED ISSUE #741: [rev-20260618T0010Z-563cce] parseTableArgument does not handle native []string for top-level table argument
**Location:** `pkg/enforcement/handlers.go`, `parseTableArgument`, lines 1519–1613

**Severity:** Medium — incorrect denial with a misleading error message for any library caller that passes a pre-typed `[]string` as the `tables` argument value

**Description:**

The outer type-switch in `parseTableArgument` handles three shapes for the table-name argument:

```go
switch v := arg.(type) {
case string:          // single table name
case []interface{}:   // JSON-decoded array
case map[string]interface{}:  // single table object
default:
    return nil, fmt.Errorf("... got %T", arg)
}
```

`[]string` is missing from this switch.  A caller that constructs `Arguments` programmatically (rather than via `json.Unmarshal`) and writes:

```go
Arguments["tables"] = []string{"users", "orders"}
```

hits `default` and receives:

```
allowedTables: "tables" argument must be a string, object, or array of strings/objects; got []string
```

This causes the condition handler to return an error (which the engine treats as a deny), even though the intent is clearly a valid array of table names.

**Inconsistency:** the exact same gap was fixed for the `columns` sub-field inside the same function (issue #730, `case []string:` at lines 1587–1601) and for `resolveStringOrStringArray` (issue #729).  The top-level table-name argument is the only remaining site that lacks the `[]string` arm.

**Trigger / repro:**

```go
engine.Decide(ctx, capability.Arguments{
    "tables": []string{"users", "orders"},
}, constraint)
// → CONDITION_FAILED, "got []string"
```

**Suggested fix:**

Add a `case []string:` arm to the outer switch, immediately after `case []interface{}:`, that converts each element to a `tableEntry{Name: elem}`:

```go
case []string:
    for _, name := range v {
        if name == "" {
            return nil, fmt.Errorf("allowedTables: table name must not be empty")
        }
        tables = append(tables, tableEntry{Name: name})
    }
```

**Session:** rev-20260618T0010Z-563cce

### LINKED ISSUE #742: [rev-20260618T0010Z-563cce] verifyRetrySentinel cannot be inserted into negKIDs when cache is at capacity, disabling kid-less refresh rate-limiting
**Location:** `pkg/capability/jwkscache.go`, `markKIDAbsent`, lines 323–347; `ForceRefreshForVerify`, lines 234–267

**Severity:** Medium/High — targeted DoS: an attacker can force unbounded JWKS endpoint refreshes by exhausting the negative KID cache

**Description:**

`verifyRetrySentinel = ""` (line 220) is the empty-string key stored in `negKIDs` to rate-limit forced JWKS refreshes for tokens that carry no `kid` claim.  It is stored and looked up through exactly the same code paths as ordinary KID strings.

`markKIDAbsent` (lines 323–347) checks the cap before inserting:

```go
if len(c.negKIDs) >= negativeKIDMaxLen {   // negativeKIDMaxLen = 1024
    // sweep expired entries ...
        return   // still full — silently drop this KID
    }
}
c.negKIDs[kid] = time.Now()
```

The sentinel `""` gets no special treatment.  If an adversary floods the proxy with 1 024 JWT requests each carrying a distinct unknown `kid`, every slot is occupied by a real KID string.  After that:

1. `markKIDAbsent(verifyRetrySentinel)` silently returns without inserting.
2. `kidRecentlyAbsent(verifyRetrySentinel)` always returns `false`.
3. `ForceRefreshForVerify` no longer rate-limits: every subsequent kid-less bad-signature token triggers a live JWKS fetch.

Because `ForceRefreshForVerify` is called on every JWT verification failure for tokens without a `kid`, an attacker can then sustain a stream of such tokens to drive continuous JWKS endpoint polling — a potential DoS against the IdP or a resource exhaustion attack against the proxy.

The sweep in `markKIDAbsent` evicts only expired entries (TTL 30 s); if the attacker refreshes distinct unknown KIDs at a rate of ≥34/s the map stays full indefinitely.

**Trigger / repro:**

1. Send 1 024 JWT requests with unique unknown `kid` values to fill `negKIDs`.
2. Send a stream of JWT requests with no `kid` claim and an invalid signature.
3. Observe repeated JWKS endpoint fetches with no 30-second back-off between them.

**Suggested fix:**

Reserve the sentinel slot unconditionally.  In `markKIDAbsent`, skip the capacity check when `kid == verifyRetrySentinel`:

```go
    c.mu.Lock()
    defer c.mu.Unlock()
    if kid != verifyRetrySentinel && len(c.negKIDs) >= negativeKIDMaxLen {
        // sweep + capacity guard (existing logic)
        ...
            return
        }
    }
    c.negKIDs[kid] = time.Now()
}
```

Alternatively, use a dedicated `sentinelAbsentUntil time.Time` field on the cache struct, completely separate from `negKIDs`, so the sentinel is never subject to the KID cache cap at all.

**Session:** rev-20260618T0010Z-563cce

### LINKED ISSUE #743: [rev-20260618T0010Z-563cce] AgentID and TaskID stamped onto audit records without size bounding, can exceed 4 MiB scanner buffer
**Location:** `cmd/eunox-mcp/audit.go`, `record()`, lines 801–813

**Severity:** Medium — a malicious or misconfigured IdP can produce audit records that the built-in `audit-verify` subcommand cannot read, effectively breaking HMAC-chain verification

**Description:**

`record()` stamps JWT-derived claims directly onto the `auditRecord` struct without applying any size cap:

```go
AgentID: agentID,   // from JWT claim "mcp.agent_id"
TaskID:  taskID,    // from JWT claim "mcp.task_id"
```

Every other attacker-influenced field is bounded:

| Field | Cap |
|---|---|
| `SessionID` | `auditSessionIDCap` = 256 B |
| `Target` / `Method` | `auditEnvelopeFieldCap` = 8 KiB (via `boundEnvelopeField`) |
| `Details` map | `auditDetailsTotalCap` = 1 MiB |
| `Obligations` | `auditObligationsTotalCap` = 64 KiB |

`AgentID` and `TaskID` are the only two JWT-derived string fields with no corresponding guard.

`verifyAuditLog` (the scanner used by `audit-verify`) uses a 4 MiB buffer:

```go
scanner.Buffer(make([]byte, 4<<20), 4<<20)
```

A JWT with an `mcp.agent_id` value of, say, 5 MiB produces an audit record that is written successfully (the drainer goroutine does not cap it) but exceeds the scanner's line limit.  `bufio.Scanner.Scan()` returns `false` with `ErrTooLong`, the verify loop treats the first such record as a chain break, and every subsequent record appears to have a broken chain — the audit log is effectively unverifiable from that record onward.

Because JWTs come from the IdP and their claims are validated for structure (e.g. the `mcp.agent_id` field must be a string) but not for length, this can be triggered by a misconfigured or compromised IdP without any changes to the proxy configuration.

**Trigger / repro:**

1. Configure `eunox-mcp` with a `JWTPDP`.
2. Issue a JWT whose `mcp.agent_id` claim is a string of length > 4 MiB.
3. Send any request through the proxy; an audit record is written.
4. Run `eunox-mcp audit-verify`; it fails with a scanner error starting at that record.

**Suggested fix:**

Apply `boundFieldTo` in `record()`, mirroring the treatment of `SessionID`:

```go
AgentID: boundFieldTo(agentID, auditEnvelopeFieldCap),
TaskID:  boundFieldTo(taskID, auditEnvelopeFieldCap),
```

`auditEnvelopeFieldCap` (8 KiB) is a reasonable upper bound for an agent or task identifier; the truncation marker appended by `boundFieldTo` makes the truncation visible in the audit record.

**Session:** rev-20260618T0010Z-563cce

eunolabs:
## Bugs addressed (session rev-20260618T0010Z-563cce)

- Closes #741 — `parseTableArgument`'s outer type switch handled `string`, `[]interface{}`, and `map[string]interface{}` but not `[]string`, so a library caller passing a pre-typed `[]string` for the `tables` argument hit `default` and was denied with a misleading `got []string` error. Added a `[]string` arm mirroring the `[]interface{}` arm (and the existing `columns` `[]string` arm from #730/#729): each element is a table name; a blank element in a populated array fails closed.
- Closes #742 — the kid-less verify sentinel (`verifyRetrySentinel`) was subject to `negativeKIDMaxLen`, so an attacker flooding the proxy with 1024 distinct unknown-kid tokens could fill `negKIDs` and prevent the sentinel from ever being inserted. `ForceRefreshForVerify` then lost its rate-limit, letting a stream of kid-less bad-signature tokens drive unbounded JWKS endpoint fetches (DoS). The sentinel is now exempt from the cap (at most one extra slot, so the map stays bounded).
- Closes #743 — `AgentID`/`TaskID` were stamped onto audit records straight from IdP JWT claims with no length cap, unlike every other attacker-influenced field. A multi-megabyte `mcp.agent_id` could push a record past the 4 MiB `audit-verify` scanner buffer and break HMAC-chain verification from that record onward. Both are now bounded to `auditEnvelopeFieldCap` (8 KiB) via `boundFieldTo`, mirroring `SessionID`/`Target`/`Method`; the threat model envelope-bounding section is updated.

## Testing
- New regression tests:
  - `TestAllowedTables_ColumnsAndShapes` — added `native []string` accept / disallowed-table-on-the-merits / blank-element subtests.
  - `TestJWKSCache_MarkKIDAbsent_SentinelExemptFromCap` — fills the cache to capacity, confirms a real kid past the cap is dropped but the sentinel is still reserved.
  - `TestRecordOversizedAgentTaskIDBounded` — a ~5 MiB agent_id/task_id produces a bounded, HMAC-verifiable record.

---


eunolabs:
✅ Ready to merge — all session rev-20260618T0010Z-563cce bugs addressed (#741, #742, #743), self-review complete with no outstanding comments, all 27 CI checks passing.

---

PR #755, opened 2026-06-18
Fix session rev-20260618T0031Z-ae92e4 bugs

### LINKED ISSUE #745: [rev-20260618T0031Z-ae92e4] expectVersion with multiple policy files silently validates only the first file's version
**Location:** `cmd/eunox-mcp/gateway_config.go` `UpstreamConfig.ExpectVersion` (line ~183); `cmd/eunox-mcp/manifest.go` `MergeManifests` (lines 89–109); `cmd/eunox-mcp/gateway_config.go` `validate()` (lines ~266–372)

**Severity:** Medium — an operator's version pin is silently less effective than the config comment implies, potentially allowing an updated policy file to be loaded undetected

**Description:**

`UpstreamConfig.ExpectVersion` is documented with a clear constraint:

```go
// expectVersion pins the manifest's `version`. When set and the manifest
// version differs, BuildRoutes fails fast (the gateway refuses to start).
// Only supported with a single policy file — the pin is unambiguous only
// then. Empty ⟹ no pin.
ExpectVersion string `yaml:"expectVersion"`
```

Neither `validate()` nor `noPolicyStartupRejection` enforces the "single policy file" constraint. A config such as:

```yaml
upstreams:
  - name: stripe
    transport: http
    upstreamUrl: https://mcp.stripe.com
    policy:
      - ./policies/stripe-base.yaml   # version 1.0.0
      - ./policies/stripe-extra.yaml  # version 3.0.0  ← never pinned
    expectVersion: "1.0.0"
```

passes `LoadGatewayConfig` without error, and also passes `BuildRoutes` startup checks, because `MergeManifests` constructs the merged policy with the version from `ms[0]` only:

```go
merged := &LocalManifest{
    ...
    Version: ms[0].Version,   // ← always the first file's version
    ...
}
```

The version-pin check is then compared against `merged.Version`, which is always `ms[0].Version`. The second (and every subsequent) policy file's `version` field is never checked against `expectVersion`, even though their capabilities are fully merged and enforced.

**Impact:**

An operator who splits policy across two files and sets `expectVersion` to pin them both believes they will be alerted to any policy update. In practice:
- Only updates to the first policy file's `version` are caught.
- Updates to `stripe-extra.yaml` — including additions of new capabilities or changes to conditions — can be deployed without bumping the declared version, without `BuildRoutes` raising an error.

**Trigger / repro:**

1. Create two policy files both declaring valid manifests, with different `version` fields.
2. Configure `policy: ["file1.yaml", "file2.yaml"]` with `expectVersion` matching `file1.yaml`'s version.
3. Run `eunox-mcp proxy --config gateway.yaml`. Gateway starts successfully, silently ignoring `file2.yaml`'s version.
4. Update `file2.yaml` capabilities without bumping its version. Gateway continues to start without error.

**Suggested fix:**

Add a check to `validate()` or to the startup routine that errors when `ExpectVersion` is set alongside `len(Policy) != 1`:

```go
// In validate(), upstream loop:
if u.ExpectVersion != "" && len(u.Policy) != 1 {
    return fmt.Errorf("upstream %q: 'expectVersion' is only supported with a single policy file; "+
        "with %d files the pin is ambiguous (only the first file's version is compared) — "+
        "use a single merged policy file, or remove 'expectVersion'", u.Name, len(u.Policy))
}
```

**Session:** rev-20260618T0031Z-ae92e4

### LINKED ISSUE #746: [rev-20260618T0031Z-ae92e4] noPolicyStartupRejection suggests removing strictDrift as a standalone fix even when enforcement: enforce is also a barrier
**Location:** `cmd/eunox-mcp/gateway_config.go`, `noPolicyStartupRejection`, lines 415–432

**Severity:** Low — misleading error message in `validate`/`doctor` subcommands; no enforcement bypass

**Description:**

`noPolicyStartupRejection` evaluates guards in order and returns the message for the first one that fires:

```go
switch {
case resolveBool(u.StrictDrift, cfg.Defaults.StrictDrift):
    return `strictDrift requires a policy — add 'policy:' or remove 'strictDrift'`
case !cfg.auditModeFor(u):
    return `no policy and enforcement is not "audit" — add 'policy:' or set 'enforcement: audit'`
case u.ExpectVersion != "":
    return `expectVersion requires a policy — add 'policy:' or remove 'expectVersion'`
}
```

When a route has BOTH `strictDrift: true` AND `enforcement: enforce` (the default) with no policy, the first case fires and suggests two alternative fixes:

- `add 'policy:'` — always correct.
- `or remove 'strictDrift'` — **only correct when enforcement is also `audit`**. With `enforcement: enforce`, removing `strictDrift` leaves the second guard (`!auditModeFor`) active, and the gateway STILL refuses to start.

An operator who follows the "remove 'strictDrift'" suggestion expecting the gateway to then come up is immediately greeted by a second failure with a different error message, making the troubleshooting loop unnecessarily confusing.

**Example:**

```yaml
defaults:
  enforcement: enforce    # the default
upstreams:
  - name: fs
    transport: stdio
    command: npx -y @modelcontextprotocol/server-filesystem /data
    strictDrift: true
    # no policy:
```

Running `eunox-mcp validate --config gateway.yaml` (or `doctor`) reports:

```
strictDrift requires a policy — add 'policy:' or remove 'strictDrift'
```

If the operator removes `strictDrift` (the second suggested fix), the `validate` subcommand then reports:

```
no policy and enforcement is not "audit" — add 'policy:' or set 'enforcement: audit'
```

Two separate confusing failures instead of one accurate message.

**Suggested fix:**

Condition the `remove 'strictDrift'` suggestion on whether audit mode would be in effect once `strictDrift` is gone, so the alternative fix is only offered when it would actually succeed:

```go
case resolveBool(u.StrictDrift, cfg.Defaults.StrictDrift):
    if cfg.auditModeFor(u) {
        return `strictDrift requires a policy — add 'policy:' or remove 'strictDrift'`
    }
    return `strictDrift requires a policy, and enforcement is not "audit" — add 'policy:' (or both remove 'strictDrift' AND set 'enforcement: audit')`
```

Alternatively, surface all blocking reasons in a single pass rather than returning on the first.

**Session:** rev-20260618T0031Z-ae92e4

### LINKED ISSUE #747: [rev-20260618T0031Z-ae92e4] YAML manifest loading silently converts unquoted date values to RFC3339 strings, causing silent allowedValues policy mismatches
**Location:** `cmd/eunox-mcp/manifest.go`, `LoadManifest`, lines 52–60 (YAML→JSON conversion path)

**Severity:** Medium — an operator's `allowedValues` condition silently accepts different strings at runtime than what was written in the manifest YAML, causing calls that should be allowed to be denied

**Description:**

`LoadManifest` converts YAML to JSON via a generic round-trip:

```go
var raw interface{}
if err := yaml.Unmarshal(data, &raw); err != nil { ... }
if data, err = json.Marshal(raw); err != nil { ... }
```

`gopkg.in/yaml.v3` applies YAML 1.2 type inference when decoding into `interface{}`. An unquoted value that matches a date pattern (e.g., `2026-01-01`) is decoded as a Go `time.Time` value rather than a string. When `json.Marshal` serializes that `time.Time`, it produces the RFC3339 representation: `"2026-01-01T00:00:00Z"`.

The manifest file never reaches the JSON unmarshaler with the original `"2026-01-01"` string; it arrives as `"2026-01-01T00:00:00Z"`. The manifest loads without error. But at runtime, a tool call with argument `{"date": "2026-01-01"}` fails `allowedValues` enforcement with `CONDITION_FAILED`, because `"2026-01-01"` != `"2026-01-01T00:00:00Z"`.

**Minimal repro:**

YAML manifest:
```yaml
capabilities:
  - target: tool:send-email
    actions: [call]
    conditions:
      - type: allowedValues
        argument: scheduled_date
        values:
          - 2026-01-01       # unquoted — yaml.v3 decodes as time.Time
          - 2026-06-30       # same
```

Effective runtime allowedValues after YAML→JSON conversion:
```json
["2026-01-01T00:00:00Z", "2026-06-30T00:00:00Z"]
```

A tool call with `{"scheduled_date": "2026-01-01"}` is denied even though the operator intended to allow it.

**Why it's hard to spot:**

- `LoadManifest` succeeds with no warning.
- `validate` and `doctor` subcommands succeed.
- The audit log records the denial as a legitimate `CONDITION_FAILED` (the allowed value is `"2026-01-01T00:00:00Z"` — a perfectly valid string), giving no indication of a policy authoring error.
- Only operators who quote their date strings (`- "2026-01-01"`) avoid the problem; the YAML spec makes unquoted dates ambiguous, and most YAML authors are unaware of the auto-detection.

**Suggested fix (choose one):**

1. **Disable timestamp inference** for the YAML→JSON intermediate decode. Use a `yaml.Node` decode and walk the tree, converting `!!timestamp` nodes to their literal string representation instead of `time.Time`:

   ```go
   var node yaml.Node
   if err := yaml.Unmarshal(data, &node); err != nil { ... }
   // Walk node, replacing timestamp-tagged scalars with their string value
   data, err = marshalNodeToJSON(&node)
   ```

2. **Validate the condition values** after loading to detect `time.Time`-shaped strings that differ from what was written. This is harder since the original text is lost after the round-trip.

3. **Document and lint**: reject string values in `allowedValues` that match RFC3339 with a `Z` suffix but not the original manifest text. Hard in practice.

Option 1 is the only fully safe fix.

**Session:** rev-20260618T0031Z-ae92e4

eunolabs:
## Bugs addressed (session rev-20260618T0031Z-ae92e4)

- Closes #745 — `expectVersion` with multiple policy files silently pinned only the first file's version (`MergeManifests` takes `merged.Version` from `ms[0]`). `loadUpstreamPDP` already rejects this at real startup, but `cfg.validate()` (which backs the `validate`/`doctor` subcommands) did not, so those subcommands green-lit a config the proxy would refuse to start. Added a validate-time guard rejecting `expectVersion` with `len(policy) > 1` (the no-policy case stays handled by `noPolicyStartupRejection`).
- Closes #746 — `noPolicyStartupRejection` suggested "remove `strictDrift`" as a standalone fix even when `enforcement` was not `audit`, where removing it just trips the next guard and the gateway still won't start. The message now names both barriers when audit mode is not in effect, and keeps the single-fix suggestion when it is.
- Closes #747 — `LoadManifest` decoded YAML into `interface{}`, so yaml.v3's timestamp inference turned an unquoted date (`2026-01-01`) into a `time.Time` that the YAML→JSON round-trip rendered as `2026-01-01T00:00:00Z`, silently changing an `allowedValues` set and denying calls the operator meant to allow. Now decodes into a `yaml.Node` and retags `!!timestamp` scalars to `!!str` before the round-trip; other scalar types are untouched.

## Testing
- New regression tests:
  - `TestValidate_ExpectVersionMultiPolicyRejected` — multi-file pin rejected by `validate()`; single-file pin passes.
  - `TestNoPolicyStartupRejection_StrictDriftMessageDependsOnAuditMode` — message names both barriers under `enforce`, offers the standalone fix under `audit`.
  - `TestLoadManifest_UnquotedDatesStayStrings` — unquoted dates load as literal strings.
- Docs: capability-manifest guide (`allowedValues` date note).

---


eunolabs:
✅ Ready to merge — all session rev-20260618T0031Z-ae92e4 bugs addressed (#745, #746, #747), self-review complete with no outstanding comments. CI is green on re-run; the initial `Test` failure was an unrelated pre-existing flake in `TestRedis_MultiInstance_KillPropagatesAcrossInstances` (miniredis pub/sub "connection refused"), in a package this PR does not touch, and passed on re-run.

---

PR #756, opened 2026-06-18
Fix session rev-20260618T0124Z-567808 bugs

### LINKED ISSUE #750: [rev-20260618T0124Z-567808] Redis killswitch Stop() lacks WaitGroup — goroutines keep running after Stop() returns
## Location

`pkg/killswitch/redis.go` — `Stop()` (line ~244), `Start()` (line ~144), `reconcileLoop` goroutine, `listenPubSub` goroutine.

## Reference

Reviewed at commit `ad7b63c599c27e1ee346c2b47d12423206e7fb88`.

## Severity

**Medium** — calling `Stop()` and then immediately destroying shared state (closing the Redis client, dropping the logger) creates a data race: both background goroutines may still be touching `r.mu`, `r.lastRefreshErr`, `r.reconcileErrLogged`, and `r.logger` when `Stop()` has already returned. The window is small (context cancellation propagates quickly through Redis I/O) but real under load or test teardown.

## Description

`Stop()` calls `cancel()` and returns immediately:

```go
func (r *Redis) Stop() {
    r.mu.RLock()
    cancel := r.cancel
    r.mu.RUnlock()
    if cancel != nil {
        cancel()
    }
    // ← returns here; no wait for goroutines to exit
}
```

`Start()` launches two goroutines inside `startOnce.Do`:

```go
go r.listenPubSub(subCtx, pubsub)
go r.reconcileLoop(subCtx)
```

Neither goroutine is tracked with a `sync.WaitGroup`. After `cancel()` is called:

- `reconcileLoop` may be in the middle of `reconcileRefresh` → `refreshState` (Redis I/O) → `r.mu.Lock()` / `r.reconcileErrLogged` update → `r.logger.Warn/Info`. 
- `listenPubSub` may be in the middle of `handlePubSubMessage` → `reconcileRefresh` → same sequence.

These goroutines access `r.mu`, `r.lastRefreshErr`, `r.reconcileErrLogged`, and `r.logger` after `Stop()` has returned. A caller who closes or frees any of these after `Stop()` — including a test that replaces the Redis client or asserts no more log calls — has undefined behaviour.

## Contrast

`pkg/capability/token_cache.go` (sibling component) correctly handles this with `sync.WaitGroup`:

```go
// TokenCache.Start:
c.wg.Add(1)
go func() {
    defer c.wg.Done()
    c.cleanupLoop(ctx)
}()

// TokenCache.Stop:
close(c.stopCh)
c.mu.Unlock()
c.wg.Wait()  // blocks until goroutine exits
```

`Redis` has `startOnce sync.Once` but no `sync.WaitGroup`.

## Trigger / Repro

```go
r := NewRedis(redisClient)
r.WithLogger(slog.Default())
r.Start(ctx)
// ... use r ...
r.Stop()
// At this point a goroutine may still call r.logger.Warn() or access r.lastRefreshErr.
// Close or nil out redisClient here → possible crash or data race under -race.
```

In tests with `go test -race`, shutting down immediately after `Stop()` can surface a race on `r.reconcileErrLogged` or `r.lastRefreshErr` between the test goroutine and the still-running background goroutine.

## Suggested Fix

Add a `sync.WaitGroup` to the `Redis` struct:

```go
type Redis struct {
    ...
    startOnce sync.Once
    wg        sync.WaitGroup  // add this
}
```

In `Start()`, track each goroutine:

```go
r.wg.Add(1)
go func() {
    defer r.wg.Done()
    r.listenPubSub(subCtx, pubsub)
}()

r.wg.Add(1)
go func() {
    defer r.wg.Done()
    r.reconcileLoop(subCtx)
}()
```

In `Stop()`, wait after cancelling:

```go
func (r *Redis) Stop() {
    r.mu.RLock()
    cancel := r.cancel
    r.mu.RUnlock()
    if cancel != nil {
        cancel()
    }
    r.wg.Wait()  // block until both goroutines have exited
}
```

`r.wg.Wait()` is safe to call before `Start()` (counter is zero, returns immediately) and from multiple concurrent callers (each blocks until the counter reaches zero, matching `TokenCache.Stop()` behaviour documented in issue #447).

## Session

`rev-20260618T0124Z-567808`

eunolabs:
## Bugs addressed (session rev-20260618T0124Z-567808)

- Closes #750 — the Redis kill switch's `Stop()` cancelled the context and returned immediately without waiting for its two background goroutines (the pub/sub listener and the reconcile loop) to exit. Those goroutines keep touching `r.mu`, `r.lastRefreshErr`, `r.reconcileErrLogged`, and `r.logger` while finishing an in-flight refresh, so a caller that frees the Redis client or logger right after `Stop()` races them. Added a `sync.WaitGroup`, tracked both goroutines with `Add(1)`/`defer Done()`, and `wg.Wait()` in `Stop()` after cancelling — mirroring `capability.TokenCache`'s shutdown pattern (#447). `wg.Wait()` is safe before `Start` (zero counter) and from concurrent callers.

## Testing
- New regression test `TestRedis_Stop_WaitsForGoroutines` runs the reconcile loop at a 1ms interval, calls `Stop()`, then frees shared state — under `-race` this flags a lingering goroutine if the WaitGroup wiring regresses. Existing `TestRedis_StartStop_NoRace` continues to pass.

---


eunolabs:
✅ Ready to merge — session rev-20260618T0124Z-567808 bug addressed (#750), self-review complete with no outstanding comments, CI green (the 8 demo/OPA jobs are path-gated and skipped for a killswitch-only change).

---

PR #780, opened 2026-06-18
Fix session rev-20260618T0543Z-99bf89 bugs

### LINKED ISSUE #759: [rev-20260618T0543Z-99bf89] Redis Cluster scans only one master and misses kill-switch entries
**Location:** `pkg/killswitch/redis.go:522-555`
**Reference:** https://github.com/eunolabs/eunox/blob/c7d09997ad7120d51570fd8fb441281912d4f449/pkg/killswitch/redis.go#L522-L555
**Severity:** high
**Description:** `scanner` accepts `*redis.ClusterClient` and the comment explicitly says cluster clients are supported, but go-redis routes keyless `SCAN` commands to one random cluster master. `refreshState` therefore treats a partial one-node snapshot as healthy, swaps it into the local cache, and clears `lastRefreshErr`; kills stored on other masters disappear and `ShouldBlock` can allow those agents/sessions. `deleteByPrefix` has the same incomplete enumeration problem, and its multi-key `DEL` can additionally fail with `CROSSSLOT`.
**Trigger / repro:** Create a two-master Redis Cluster, place `killswitch:agent:agent-a` on one master and `killswitch:agent:agent-b` on the other, pass the `*redis.ClusterClient` to `NewRedis`, and run `scanPrefix`/`Start`. A regression probe returned only `map[agent-b:true]`; the other durable kill was omitted without an error.
**Suggested fix:** Detect cluster clients and iterate every master with `ForEachMaster`, merging each node-local scan into one snapshot. Perform reset deletion per node/key (or grouped by hash slot), and only mark refresh healthy after all masters succeed. Alternatively, reject cluster clients explicitly instead of claiming support.
**Session:** rev-20260618T0543Z-99bf89

### LINKED ISSUE #760: [rev-20260618T0543Z-99bf89] initialize notifications allocate unbounded mock MCP sessions
**Location:** `demo/mock-mcp-server/main.go:106-121`
**Reference:** https://github.com/eunolabs/eunox/blob/c7d09997ad7120d51570fd8fb441281912d4f449/demo/mock-mcp-server/main.go#L106-L121
**Severity:** low
**Description:** `handlePost` handles `method == "initialize"` before checking whether the message is a JSON-RPC request. An `initialize` notification with no `id` therefore creates and stores a session, returns a result/header as if it were a request, and leaves the session resident until a caller happens to delete it. Repeated invalid notifications grow the in-memory session map without bound.
**Trigger / repro:** POST `{"jsonrpc":"2.0","method":"initialize","params":{}}` repeatedly to `/mcp`. A regression probe observed HTTP 200, a new `Mcp-Session-Id`, and one additional entry in `server.sessions` for every notification.
**Suggested fix:** Require `msg.isRequest()` (non-nil ID plus method) before entering the initialize path and reject an initialize notification without mutating session state. Add session expiry/capacity as defense in depth for abandoned valid sessions.
**Session:** rev-20260618T0543Z-99bf89

### LINKED ISSUE #761: [rev-20260618T0543Z-99bf89] JWKS fetches can hang indefinitely with a zero-timeout custom HTTP client
**Location:** `pkg/capability/jwkscache.go:389-394`
**Reference:** https://github.com/eunolabs/eunox/blob/c7d09997ad7120d51570fd8fb441281912d4f449/pkg/capability/jwkscache.go#L389-L394
**Severity:** medium
**Description:** The shared JWKS fetch deliberately strips the initiating caller's cancellation and deadline with `context.WithoutCancel`, assuming the HTTP client timeout will bound the request. `JWKSCacheConfig` accepts an arbitrary custom `*http.Client`, and a normal zero-value/custom client has `Timeout == 0`. If its transport stalls, the fetch never returns; because refreshes share one singleflight key, every later token verification joins the same stuck call.
**Trigger / repro:** Construct `NewJWKSCache` with `Client: &http.Client{Transport: blockingRoundTripper}` and no client timeout, then call `GetKeys` with a 20 ms context deadline. A regression probe was still blocked after 100 ms and returned only after the transport was manually released.
**Suggested fix:** Guarantee an internal maximum fetch duration even for custom clients, for example by wrapping `context.WithoutCancel(ctx)` in `context.WithTimeout` when `Client.Timeout <= 0`, or reject/clone custom clients that do not define a finite timeout. Preserve singleflight decoupling while ensuring the shared operation is always bounded.
**Session:** rev-20260618T0543Z-99bf89

### LINKED ISSUE #762: [rev-20260618T0543Z-99bf89] OPA demo uses predictable sequential MCP session IDs
**Location:** `demo/opa-comparison/server/main.go:373-384`
**Reference:** https://github.com/eunolabs/eunox/blob/c7d09997ad7120d51570fd8fb441281912d4f449/demo/opa-comparison/server/main.go#L373-L384
**Severity:** low
**Description:** Session IDs are generated from a process-global counter (`opa-cmp-1`, `opa-cmp-2`, ...). The Streamable HTTP session header gates access to another client's server-side session, and the MCP 2025-11-25 transport specification says assigned session IDs should be globally unique and cryptographically secure. A peer that observes or guesses the counter can address another client's session.
**Trigger / repro:** Initialize once and observe `Mcp-Session-Id: opa-cmp-N`. After another client initializes, send `tools/list` or `tools/call` with `Mcp-Session-Id: opa-cmp-(N+1)`; the server accepts it as that client's live session. IDs also repeat after every process restart.
**Suggested fix:** Generate session IDs with `crypto/rand` (for example 128 random bits encoded as UUID/base64url), matching `demo/mock-mcp-server`, and keep the existing map lookup/delete behavior.
**Session:** rev-20260618T0543Z-99bf89

### LINKED ISSUE #763: [rev-20260618T0543Z-99bf89] Constraint unmarshal leaves stale conditions and directives after explicit null
**Location:** `pkg/capability/constraint.go:337-349`
**Reference:** https://github.com/eunolabs/eunox/blob/c7d09997ad7120d51570fd8fb441281912d4f449/pkg/capability/constraint.go#L337-L349
**Severity:** medium
**Description:** `Constraint.UnmarshalJSON` only assigns `Conditions` and `Directives` when the temporary slices are non-nil. JSON `null` decodes to nil, so unmarshalling an explicit `"conditions": null` or `"directives": null` into a reused `Constraint` leaves the old policy objects in place. This differs from normal JSON null semantics and makes policy reload/update code continue enforcing or redacting with stale configuration.
**Trigger / repro:** Unmarshal a constraint containing an `allowedValues` condition and `redactFields` directive, then unmarshal a second object into the same value with both fields explicitly set to `null`. A regression probe found both old slices still populated.
**Suggested fix:** Clear `c.Conditions` and `c.Directives` before rebuilding them, so absent/null input cannot retain prior values. If absent and explicit null must be distinguished, use pointer/raw-message fields in `constraintJSON`; explicit null must still clear the destination.
**Session:** rev-20260618T0543Z-99bf89

eunolabs:
## Bugs addressed (session rev-20260618T0543Z-99bf89)

- Closes #759 — Redis Cluster kill-switch scan/reset hit only one random master (keyless `SCAN` is routed to a single node), so `refreshState` loaded a partial snapshot, marked it healthy, and silently dropped kills on other masters; `deleteByPrefix` had the same gap plus a `CROSSSLOT` risk on multi-key `DEL`. Now detects `*redis.ClusterClient` and iterates every master with `ForEachMaster`, merging per-node scans, and deletes keys one at a time (slot-safe).
- Closes #760 — `demo/mock-mcp-server` handled `method == "initialize"` before checking `isRequest()`, so an `initialize` notification (no id) created/stored a session and returned a session header, growing the session map without bound. The initialize path is now gated on `isRequest()`; notifications are accepted without mutating state.
- Closes #761 — The shared JWKS fetch strips the caller's deadline (`context.WithoutCancel`) and relied on the HTTP client's `Timeout`, but a custom/zero-value client has `Timeout == 0`, so a stalled transport hung forever and every later verification joined the stuck singleflight call. An internal `maxFetch` ceiling now bounds the fetch when the client is itself unbounded.
- Closes #762 — `demo/opa-comparison` generated session IDs from a process-global counter (`opa-cmp-N`), which a peer could observe/guess to address another client's session and which repeated across restarts. IDs are now 128 cryptographically-random bits formatted as a UUID, matching `demo/mock-mcp-server`.
- Closes #763 — `Constraint.UnmarshalJSON` only assigned `Conditions`/`Directives` when the decoded slices were non-nil, so an explicit `null` (or omitted key) left stale policy objects on a reused `Constraint` during a policy reload. Both fields are now always reassigned (nil in => nil out).

## Testing
- `go test -race -count=1 ./pkg/capability/ ./pkg/killswitch/ ./demo/...` — all pass.
- New regression tests: constraint null/absent clearing (`constraint_unmarshal_clear_test.go`), bounded JWKS fetch with a zero-timeout client + blocking transport (`jwkscache_fetchtimeout_test.go`), initialize-notification session leak (`mock-mcp-server/initnotify_test.go`), and unpredictable/unique OPA session IDs (`opa-comparison/server/sessionid_test.go`).
- `make check-fmt`, `make check-license`, and `go vet` on touched packages all clean.
- #759 (cluster fan-out) has no unit test because miniredis does not emulate Redis Cluster; the standalone scan/reset path remains covered by existing `pkg/killswitch` tests, which exercise the new per-key delete.

---


eunolabs:
✅ Ready to merge — all session rev-20260618T0543Z-99bf89 bugs addressed (#759, #760, #761, #762, #763), self-review complete with no outstanding comments, tests passing.

CI: Test, Lint, all platform builds, every demo + OPA-comparison scenario, govulncheck, Trivy, enforcement compliance, gofmt, license, and NOTICE checks are green. (CodeQL "Analyze (go)" is an advisory, non-blocking scan still finishing; its summary check completed neutral.)

---

PR #784, opened 2026-06-18
Fix session rev-20260618T0622Z-e78a56 bugs

### LINKED ISSUE #776: [rev-20260618T0622Z-e78a56] serveHTTPGateway: --jwt-allow-any-audience warning uses wrong condition, suppressed when both --jwt-audience and --jwt-allow-any-audience are set
## Summary

`serveHTTPGateway` in `cmd/eunox-mcp/main.go` emits a security warning for the dangerous `--jwt-allow-any-audience` flag, but the condition used to trigger it is `pf.jwtAudience == ""` rather than `pf.jwtAllowAnyAudience`. When both `--jwt-audience=<value>` and `--jwt-allow-any-audience` are set simultaneously, `jwtAudience != ""` silences the warning even though the audience-bypass flag is active.

## Location

**File:** `cmd/eunox-mcp/main.go`  
**Line:** 671  
**Reference commit:** `96323a436c4528981bd8bfb5fa00a16a63150293`

```go
// cmd/eunox-mcp/main.go:671
if pf.jwtAudience == "" {
    fmt.Fprintf(os.Stderr, "[eunox-mcp] WARNING: --jwt-allow-any-audience is set; tokens minted for other audiences are accepted, disabling cross-service replay protection.\n")
}
```

## Why the condition is wrong

`validateJWTAudienceConfig` (line 663) rejects the combination `jwksURI != "" && jwtAudience == "" && !jwtAllowAnyAudience`, so after that check passes, `jwtAudience == ""` is a valid proxy for `jwtAllowAnyAudience == true` **only in the normal case** where the two flags are exclusive.

However, nothing prevents an operator from passing both flags together:

```
eunox-mcp proxy --jwks-uri https://idp.example.com/.well-known/jwks.json \
                --jwt-audience my-service \
                --jwt-allow-any-audience
```

`validateJWTAudienceConfig` returns nil (since `jwtAudience != ""`), and the warning block's condition `pf.jwtAudience == ""` evaluates to `false`. The warning is silently suppressed even though the dangerous `--jwt-allow-any-audience` flag is in effect.

## Security implication

`JWTPDPOptions.AllowAnyAudience: true` is still passed to `wrapRoutesWithJWT` (line 680):

```go
gwJWTPDP = wrapRoutesWithJWT(routes, JWTPDPOptions{
    ...
    AllowAnyAudience: pf.jwtAllowAnyAudience,
    ...
})
```

If `AllowAnyAudience` causes the audience check to be skipped entirely (regardless of whether `Audience` is also set), then this is a cross-service token replay opening: a token minted for another relying party of the same IdP is accepted with no operator warning. The operator who set both flags expecting belt-and-suspenders protection (audience pinned AND any-audience fallback) receives instead an audience-bypass with no diagnostic.

## Severity

**MEDIUM** — wrong condition for a security advisory warning; potential silent audience bypass when both flags are combined.

## Trigger / Reproduction

1. Launch the gateway with both `--jwt-audience=<value>` and `--jwt-allow-any-audience`.
2. **Expected (after fix):** stderr warning `"WARNING: --jwt-allow-any-audience is set; tokens minted for other audiences are accepted, disabling cross-service replay protection."`.
3. **Actual:** no warning printed; the dangerous flag is active but silent.

## Suggested Fix

Change line 671 to check the flag directly:

```go
// Before
if pf.jwtAudience == "" {

// After
if pf.jwtAllowAnyAudience {
```

This makes the warning fire whenever `--jwt-allow-any-audience` is active, regardless of whether `--jwt-audience` is also set. It is consistent with the flag description ("disabling audience pinning") and covers all permutations.

Optionally, also reject the contradictory combination (`--jwt-audience` + `--jwt-allow-any-audience`) in `validateJWTAudienceConfig` with an explicit error, since the two flags are documented as alternatives and setting both is almost certainly a mistake.

## Session

Session ID: `rev-20260618T0622Z-e78a56`  
Reviewed SHA: `96323a436c4528981bd8bfb5fa00a16a63150293`

eunolabs:
## Bugs addressed (session rev-20260618T0622Z-e78a56)

- Closes #776 — `serveHTTPGateway`'s `--jwt-allow-any-audience` advisory warning was gated on `pf.jwtAudience == ""`. When an operator passed **both** `--jwt-audience` and `--jwt-allow-any-audience`, `jwtAudience != ""` silenced the warning even though `JWTPDP` skips the audience check entirely (`pdp_jwt.go` only sets `expected.AnyAudience` when `!allowAnyAudience`), so the configured audience is silently ignored — a cross-service token-replay opening with no operator diagnostic. The warning is now keyed on the `--jwt-allow-any-audience` flag itself, and when an audience is also configured it explicitly states the pin is ignored.

## Changes
- `cmd/eunox-mcp/main.go`: extracted the decision into a pure, testable `jwtAudienceBypassWarning(allowAnyAudience, jwtAudience)` helper; the gateway emits its result. Behavior now: warning fires iff `--jwt-allow-any-audience` is set.
- `CHANGELOG.md`: Security entry.

## Testing
- New `TestJWTAudienceBypassWarning` (table-driven): no warning when pinning is active (with or without audience); warning fires for the bypass flag alone; **and crucially fires when both flags are set**, noting the configured `--jwt-audience` is ignored.
- `go test -race -count=1 ./cmd/eunox-mcp/` passes; `make check-fmt`/`check-license`/`go vet` clean.

## Note
The issue's optional suggestion to hard-reject the both-flags combination was deliberately not taken: `--jwt-allow-any-audience` is an explicit "accept the risk" opt-out and `AllowAnyAudience` wins deterministically (not ambiguous), so a clear warning that the pin is ignored is the proportionate fix without an over-aggressive startup failure.

---
_Generated by [Claude Code](https://claude.ai/code/session_019yY5D9YRVLgunXTZtxVVn5)_


eunolabs:
✅ Ready to merge — session rev-20260618T0622Z-e78a56 bug #776 addressed. Self-review (2 passes): confirmed `serveHTTPGateway` is the only JWT-wiring/audience-warning site (`wrapRoutesWithJWT` is gateway-only), so no other path carries the bug; the issuer warning is untouched. New table-driven test covers the both-flags regression. `go test -race ./cmd/eunox-mcp/` green; fmt/license/vet clean.

---

PR #793, opened 2026-06-18
Fix session rev-20260618T0559Z-6d8b6c bugs

### LINKED ISSUE #764: [rev-20260618T0559Z-6d8b6c] OPA comparison scenarios pass CI when Eunox returns the wrong verdict
**Location:** `demo/opa-comparison/scripts/credential-exfiltration.sh:40-44`, `demo/opa-comparison/scripts/path-gated-file-access.sh:48-61`
**Reference:** https://github.com/eunolabs/eunox/blob/c7d09997ad7120d51570fd8fb441281912d4f449/demo/opa-comparison/scripts/credential-exfiltration.sh#L40-L44 and https://github.com/eunolabs/eunox/blob/c7d09997ad7120d51570fd8fb441281912d4f449/demo/opa-comparison/scripts/path-gated-file-access.sh#L48-L61
**Severity:** medium
**Description:** Every Eunox call is followed by `|| true`, so the scenarios discard `mcp_call`'s allow/deny status for both positive and negative expectations. The scripts then print hard-coded success summaries and exit 0. These files are run as the OPA comparison integration jobs, so CI stays green if `sequenceBlock`, `allowedValues`, or `maxCalls` returns the opposite verdict. The expected OPA denial in the path scenario is also not asserted.
**Trigger / repro:** Replace the MCP endpoint with a stub that returns a valid ALLOW result for every tool call. `credential-exfiltration.sh` prints ALLOW for the write that is expected to be denied, then prints a summary claiming DENY and exits 0. `path-gated-file-access.sh` likewise allows `/etc/passwd` and call 6, prints a DENY summary, and exits 0.
**Suggested fix:** Add explicit `expect_allow` and `expect_deny` helpers that fail when `mcp_call` returns the opposite status. Remove blanket `|| true`; use an assertion wrapper for expected denials so the script can continue only after confirming the denial. Derive summaries from captured results instead of hard-coding them.
**Session:** rev-20260618T0559Z-6d8b6c

### LINKED ISSUE #770: [rev-20260618T0559Z-6d8b6c] OPA harness treats HTTP and malformed responses as ALLOW
**Location:** `demo/opa-comparison/scripts/common.sh:137-170`
**Reference:** https://github.com/eunolabs/eunox/blob/c7d09997ad7120d51570fd8fb441281912d4f449/demo/opa-comparison/scripts/common.sh#L137-L170
**Severity:** medium
**Description:** `mcp_call` uses `curl -s` without checking the HTTP status, then treats missing JSON fields as `error.message = empty`, `isError = false`, and result text = empty. Any HTTP error page, empty body, or malformed JSON is therefore printed as `ALLOW` and returns status 0. Because every scenario invokes `mcp_call` in an `|| true` context, Bash errexit does not reliably protect the curl assignment either, so transport failures can follow the same false-success path.
**Trigger / repro:** Make the MCP endpoint return HTTP 500 with a plain-text body such as `upstream returned HTTP 500`. Calling `mcp_call read_file {path:/reports/q3.pdf}` prints `ALLOW:` and returns 0. An isolated probe against this commit reproduced exactly that result.\n**Suggested fix:** Capture and validate the HTTP status separately, fail on curl transport errors, require the body to parse as a JSON-RPC response, and require either a well-formed `error` or `result` envelope before classifying the verdict. Return a distinct nonzero infrastructure-error status so scenario assertions cannot mistake it for a policy denial.\n**Session:** rev-20260618T0559Z-6d8b6c

eunolabs:
## Bugs addressed (session rev-20260618T0559Z-6d8b6c)

- Closes #770 — `demo/opa-comparison/scripts/common.sh` `mcp_call` used `curl -s` without checking the HTTP status and treated missing JSON fields as empty, so any HTTP error page, empty body, or malformed JSON printed `ALLOW` and returned 0. `mcp_call` now captures the HTTP status separately and fails on non-200, treats a curl transport failure (connection refused / timeout / DNS) as an infra error rather than an empty ALLOW, rejects empty bodies and anything that is not a well-formed JSON-RPC 2.0 envelope (object with `jsonrpc:"2.0"` and a non-null `error` or `result`), and returns a **distinct** infrastructure-error status (`2`) separate from a policy deny (`1`) so callers cannot mistake a broken harness for a verdict. The verdict contract (`0` ALLOW / `1` DENY / `2` INFRA_ERROR) is documented at the top of `common.sh`.

- Closes #764 — every Eunox `mcp_call` in `credential-exfiltration.sh` and `path-gated-file-access.sh` was followed by `|| true`, discarding the allow/deny status, after which the scripts printed hard-coded success summaries and exited 0 — so CI stayed green even when `sequenceBlock`, `allowedValues`, or `maxCalls` returned the opposite verdict (and the expected OPA denial of `/etc/passwd` was never asserted). Added `expect_allow` / `expect_deny` (and `expect_opa_allow` / `expect_opa_deny`) helpers in `common.sh` that **fail** the script when the captured verdict is the opposite of what is expected; an infrastructure error always fails and is never treated as the expected verdict. Both scenarios now drop the blanket `|| true` on Eunox calls, use the assertion helpers, assert the expected OPA denial, and only reach the summary block after every assertion has held (so the printed narrative reflects confirmed results, not a hard-coded claim).

## Testing

- `bash -n` clean on all three changed scripts.
- Unit-tested the rewritten `mcp_call` with a stubbed `curl` across 10 cases — ALLOW, JSON-RPC error deny, `isError:true` deny, HTTP 500, empty body, whitespace-only body, malformed JSON, curl transport failure, missing `jsonrpc` field, and `result:null`: each returns the expected `0`/`1`/`2` status (ALLOW=0, policy-deny=1, all infra failures=2). Confirmed the pre-fix false-ALLOW cases (HTTP 500 / empty / malformed) now return the distinct infra-error status.
- Unit-tested `expect_allow` / `expect_deny` under `set -euo pipefail`: each passes on the expected verdict (returning cleanly without tripping errexit) and exits non-zero on the opposite verdict and on an infra error.
- Could **not** run the full OPA-comparison stack end-to-end (needs Docker: live mock MCP server + OPA sidecar) — that is left to CI. `shellcheck` was not available in this environment, so it was not run; no new shellcheck-class issues were knowingly introduced.

https://claude.ai/code/session_01DQSJhGSAzf1cozmy7cHziu

---


eunolabs:
✅ Ready to merge — session rev-20260618T0559Z-6d8b6c bugs addressed (#764, #770). All 27 checks green, including all three live OPA-comparison jobs (token-reuse, path-gated-file-access, credential-exfiltration) which exercise the rewritten strict `mcp_call` + `expect_allow`/`expect_deny` assertions against the real proxy + OPA sidecar. Reviewed: an HTTP/transport/malformed response now returns the distinct infra-error status (never a silent ALLOW), and the scenario scripts fail-closed on the opposite verdict instead of `|| true`.

---

PR #794, opened 2026-06-18
Fix session rev-20260618T0557Z-f858a5 bugs

### LINKED ISSUE #765: [rev-20260618T0557Z-f858a5] Demo CI treats malformed responses as successful allows
**Location:** `demo/scripts/ci-test.sh:39-65`
**Reference:** https://github.com/eunolabs/eunox/blob/c7d09997ad7120d51570fd8fb441281912d4f449/demo/scripts/ci-test.sh#L39-L65
**Severity:** medium
**Description:** The HTTP demo test classifies a response by piping it through `jq` and falling back to the literal string `false` when parsing fails. `false` is exactly the value interpreted as an allowed call, so an HTML error page, truncated JSON, or other malformed response makes every expected-allow assertion pass. The same pattern is duplicated in `ci-test-gateway.sh` and `ci-test-jwt.sh`, allowing forwarding regressions to stay green when deny paths still return valid JSON-RPC errors.
**Trigger / repro:** Make the proxy or an intermediary return `<html>502 Bad Gateway</html>` for an allowed `read_file` call while denied calls continue returning JSON-RPC errors. The `jq ... || echo "false"` path sets `is_err=false`, and the allowed assertion prints `PASS`.
**Suggested fix:** Treat curl failures, non-success HTTP statuses, empty bodies, and JSON parse/schema failures as test failures. For an allow assertion, require a valid JSON-RPC response with the expected request ID, a non-null `result`, and no `error`; do not map parser failures to `false`. Apply the same strict parser to all three HTTP demo test scripts.
**Session:** rev-20260618T0557Z-f858a5

### LINKED ISSUE #766: [rev-20260618T0557Z-f858a5] Demo CI accepts unrelated errors as policy denials
**Location:** `demo/scripts/ci-test.sh:39-65`
**Reference:** https://github.com/eunolabs/eunox/blob/c7d09997ad7120d51570fd8fb441281912d4f449/demo/scripts/ci-test.sh#L39-L65
**Severity:** medium
**Description:** Expected-deny assertions only check whether any JSON-RPC `error` exists (or `result.isError` is true). Authentication failures, unknown-session errors, upstream crashes, internal errors, and unrelated tool errors are therefore all counted as successful policy enforcement. The gateway and JWT scripts repeat the same broad classification, so a broken denial path can still satisfy the security-focused integration tests.
**Trigger / repro:** Return a valid JSON-RPC internal error such as `{"jsonrpc":"2.0","id":2,"error":{"code":-32603,"message":"upstream crashed"}}` for a call expected to be denied by `allowedValues` or `allowedOperations`. `is_err` becomes `true`, and the assertion prints `PASS` even though no policy denial was demonstrated.
**Suggested fix:** Assert the documented policy-denial error code and structured reason/condition for deny cases, plus the expected response ID. Treat auth, session, transport, upstream, and internal errors as failures. Share one strict assertion helper across `ci-test.sh`, `ci-test-gateway.sh`, and `ci-test-jwt.sh`.
**Session:** rev-20260618T0557Z-f858a5

### LINKED ISSUE #767: [rev-20260618T0557Z-f858a5] Waitlist endpoint leaks fetch failures as runtime errors
**Location:** `site/functions/api/subscribe.js:37-50`
**Reference:** https://github.com/eunolabs/eunox/blob/c7d09997ad7120d51570fd8fb441281912d4f449/site/functions/api/subscribe.js#L37-L50
**Severity:** low
**Description:** The Buttondown `fetch` is awaited without an exception handler or timeout. Network, DNS, TLS, or connection failures reject `onRequestPost` outright, so Cloudflare returns a platform/runtime error instead of the clean JSON contract this endpoint promises. Only HTTP responses are converted to the intended 502 JSON response.
**Trigger / repro:** Stub `fetch` to reject (for example, `throw new Error("upstream down")`) and call `onRequestPost` with a valid email. A focused Node probe observed the handler promise reject with `upstream down` instead of returning a `Response`.
**Suggested fix:** Wrap the upstream request in `try/catch`, add a bounded timeout with `AbortSignal.timeout` or an `AbortController`, and return the existing 502 JSON body for transport failures.
**Session:** rev-20260618T0557Z-f858a5

### LINKED ISSUE #768: [rev-20260618T0557Z-f858a5] Adoption report breaks pagination checks after first page
**Location:** `scripts/signals/collect.sh:126-145`
**Reference:** https://github.com/eunolabs/eunox/blob/c7d09997ad7120d51570fd8fb441281912d4f449/scripts/signals/collect.sh#L126-L145
**Severity:** low
**Description:** `gh api --paginate` emits one JSON array per page, but the script stores those arrays as a concatenated stream and passes the stream to `jq length` inside one arithmetic `[[ ... -eq 0 ]]` comparison. Once releases exceed one API page, `jq` produces multiple lengths such as `30\n1`, which Bash rejects as a syntax error. The forks section repeats the same bug at lines 227-229.
**Trigger / repro:** Make the releases or forks API return two pages. Running the script with two one-item arrays produced `line 128: [[: 1\n1: syntax error in expression`; the forks check emitted the analogous error at line 229.
**Suggested fix:** Slurp and flatten paginated arrays before all downstream checks, for example with `gh api --paginate --slurp | jq add` (or `jq -s add`), then run `length` and row rendering against that single array.
**Session:** rev-20260618T0557Z-f858a5

### LINKED ISSUE #769: [rev-20260618T0557Z-f858a5] Stdio call helper invokes optional jq unconditionally
**Location:** `demo/scripts/mcp-call-stdio.sh:81-87`
**Reference:** https://github.com/eunolabs/eunox/blob/c7d09997ad7120d51570fd8fb441281912d4f449/demo/scripts/mcp-call-stdio.sh#L81-L87
**Severity:** low
**Description:** The helper advertises a raw-output fallback when `jq` is unavailable, but it invokes `jq` unconditionally one line earlier to extract each response ID. With `set -e` and `pipefail`, a machine without `jq` fails before the fallback branch can execute, so the documented no-`jq` behavior is unreachable.
**Trigger / repro:** Run the helper with valid `EUNOX_BIN` and `STDIO_SERVER_BIN` values while `jq` is absent from `PATH`. The ID extraction command fails with `jq: command not found`; the later `command -v jq` branch is never reached.
**Suggested fix:** Either declare and validate `jq` as a hard prerequisite at startup, or branch before ID filtering and provide a real raw-stream fallback when `jq` is missing.
**Session:** rev-20260618T0557Z-f858a5

eunolabs:
## Bugs addressed (session rev-20260618T0557Z-f858a5)

- Closes #765 — Demo CI treated malformed responses as successful allows. The HTTP demo tests classified responses with `jq ... || echo false`, and `false` is the "allowed" value, so HTML error pages, truncated JSON, empty bodies, and non-2xx statuses all made expect-ALLOW assertions pass.
- Closes #766 — Demo CI accepted unrelated errors as policy denials. Expect-DENY assertions only checked that *any* JSON-RPC `error` existed (or `result.isError`), so auth/session/transport/upstream/internal errors counted as successful policy enforcement.
- Closes #767 — Waitlist subscribe endpoint leaked fetch failures as runtime errors. The Buttondown `fetch` was awaited with no try/catch and no timeout, so network/DNS/TLS failures rejected `onRequestPost` (opaque Cloudflare platform error) instead of returning the clean 502 JSON contract.
- Closes #768 — Adoption-signals report broke pagination checks after the first page. `gh api --paginate` emits one JSON array per page; passing the concatenated stream to `jq length` inside `[[ ... -eq 0 ]]` produced multiple lengths (e.g. `30\n1`) and a Bash arithmetic syntax error, in both the releases and forks sections.
- Closes #769 — Stdio call helper invoked optional `jq` unconditionally. It advertised a no-`jq` raw-output fallback but extracted response ids with `jq` one line earlier, so under `set -e`/`pipefail` a machine without `jq` failed before the fallback could run.

### #765 / #766 — single shared strict helper

Both issues are the same three files and the same classification path, so they are fixed with one strict assertion helper, `demo/scripts/ci-test-common.sh`, sourced by `ci-test.sh`, `ci-test-gateway.sh`, and `ci-test-jwt.sh`:

- **ALLOW** requires curl success, a 2xx HTTP status, a non-empty body, a valid JSON-RPC envelope echoing the expected request id, a non-null `result`, no `error`, and `result.isError` not true. A parse failure is never mapped to "allowed".
- **DENY** requires a JSON-RPC error echoing the expected id whose structured `error.data.code` is one of the documented policy-denial codes (`AUTHORIZATION_FAILED`, `CAPABILITY_DENIED`, `CONDITION_FAILED`, `VALUE_NOT_PERMITTED`, `OPERATION_NOT_PERMITTED`, `RATE_LIMITED`, `MISSING_CONTEXT`, `KILL_SWITCH*`, `NO_JWT_CLAIMS`, `INVALID_PARAMS`). Auth/session/transport/upstream/internal JSON-RPC errors (e.g. a bare `-32603` from an upstream crash, which is what bug #766's repro produces) are treated as test failures, not as successful enforcement.

The legitimate demo ALLOW/DENY cases are unaffected: the proxy returns a clean `result` for allowed calls and routes every policy denial through `denialResult` (cmd/eunox-mcp/jsonrpc.go), which stamps `error.data.code` — exactly what the strict helper now requires.

## Testing

- `bash -n` clean on all touched shell scripts (`ci-test-common.sh`, `ci-test.sh`, `ci-test-gateway.sh`, `ci-test-jwt.sh`, `mcp-call-stdio.sh`, `collect.sh`).
- `shellcheck -x` clean on all touched shell scripts (added `# shellcheck shell=bash` to the sourced helper so it has no shebang but a declared shell).
- Unit-tested the strict helper with 12 canned responses (mocking `eunox_post`): clean allow passes; HTML 502, empty body, wrong id, truncated JSON, and `result.isError` all fail an ALLOW; documented policy denials (`VALUE_NOT_PERMITTED`, `AUTHORIZATION_FAILED`) pass a DENY; a bare `-32603` internal error, a non-policy `data.code`, and a leaked `result` all fail a DENY. All 12 behaved as intended.
- `node --check site/functions/api/subscribe.js` passes; `AbortSignal.timeout` confirmed available. Behavioral probe (stubbing `fetch`): transport-failure → 502 JSON, 3xx → 200 `{ok:true}`, 4xx → 502, invalid email → 400. Before the fix the transport-failure case rejected the handler.
- Validated the `--slurp | jq 'add // []'` flattening against multi-page (`[[1,2,3],[4]]` → length 4), single-empty-page, and no-page inputs.
- Could not run the full Docker demo stack (`make -C demo ci-test*`) locally — those integration jobs run in the `demo / *` CI jobs against a real proxy + mock upstream; relying on CI for that end-to-end verification.

https://claude.ai/code/session_01DQSJhGSAzf1cozmy7cHziu

---


eunolabs:
✅ Ready to merge — session rev-20260618T0557Z-f858a5 bugs addressed (#765, #766, #767, #768, #769). All 26 functional checks green, including the `demo / stdio`, `demo / gateway`, and `demo / JWT mode` jobs that run the rewritten strict `ci-test*.sh` scripts against the real proxy. Reviewed: the new shared `eunox_assert` helper requires a clean JSON-RPC result (with id echo) for ALLOW and a documented `error.data.code` policy-denial for DENY (so auth/session/transport/internal errors now fail the test); waitlist `fetch` is wrapped in try/catch + `AbortSignal.timeout`; paginated `gh api` output is slurped+flattened; the no-`jq` stdio fallback is now reachable.

---

PR #799, opened 2026-06-18
fix: concurrent session shutdown, pipe-FD cleanup, kill-timer stop, decideInner fail-closed (#779)

### LINKED ISSUE #779: Engineering review: 4 findings (serial shutdown, FD leak, stale kill timer, silent regression trap)
# Engineering review — 2026-06-18

## Health summary

The codebase is in excellent shape. The audit chain is tamper-evident with well-thought-out HMAC chaining, drop markers, and startup-resume integrity markers. The fail-closed posture is consistently applied — kill switch first, genesis-sentinel on chain breaks, hard denials on any ambiguity. Concurrency in the audit sink, HTTP session map, and call-counter backends is correct and carefully documented. JWT validation separates signature from claims (preventing error conflation), pins audience by default, rejects missing `exp`, and deterministically iterates conditions for reproducible audit records.

Four genuine findings below, none requiring architectural change.

---

### 1. `closeAllSessions` closes sessions serially — shutdown time is O(N × shutdownMs)

**Location:** `cmd/eunox-mcp/http.go:1641–1652`  
**Priority:** Medium | Effort: S | Risk: None (pure additive fix)

**Problem:** `closeAllSessions` is called during `Serve`'s shutdown path. It closes every active session one at a time: each `s.close(shutdownMs)` blocks up to `shutdownMs` (default 5 s) waiting for an unresponsive upstream. With N sessions this makes total shutdown time `N × 5 s`, violating any operator expectation about the shutdown window. Meanwhile `reapOnce` — which is the other place sessions are batch-closed — already does this correctly with a concurrent WaitGroup.

**Current code:**
```go
func (p *HTTPProxy) closeAllSessions() {
    p.mu.Lock()
    sessions := make([]*httpSession, 0, len(p.sessions))
    for _, s := range p.sessions {
        sessions = append(sessions, s)
    }
    p.sessions = make(map[string]*httpSession)
    p.mu.Unlock()
    for _, s := range sessions {   // ← serial: N * shutdownMs worst case
        s.close(p.shutdownMs)
    }
}
```

**Suggested fix:** mirror `reapOnce`'s concurrent pattern:
```go
func (p *HTTPProxy) closeAllSessions() {
    p.mu.Lock()
    sessions := make([]*httpSession, 0, len(p.sessions))
    for _, s := range p.sessions {
        sessions = append(sessions, s)
    }
    p.sessions = make(map[string]*httpSession)
    p.mu.Unlock()
    var wg sync.WaitGroup
    for _, s := range sessions {
        wg.Add(1)
        go func() {
            defer wg.Done()
            s.close(p.shutdownMs)
        }()
    }
    wg.Wait()
}
```

---

### 2. Pipe FD leak on subprocess startup error paths in both transports

**Location:** `cmd/eunox-mcp/http.go:1438–1452`, `cmd/eunox-mcp/stdio.go:229–242`  
**Priority:** Low | Effort: S | Risk: Error paths only

**Problem:** In `newSession` and `connectUpstream`, if `cmd.StdoutPipe()` returns an error after `cmd.StdinPipe()` has already succeeded, the stdin write-pipe is not closed. Similarly, if `cmd.Start()` fails, both already-created pipes are left unclosed. Go's `os.Cmd` does not close user-returned pipe ends on failure; they must be closed by the caller. Because these are error paths on program startup or initialization, the impact is minor — but the leak is real and could exhaust FDs under repeated failed restarts.

**`http.go` current:**
```go
upIn, err := cmd.StdinPipe()
if err != nil {
    return nil, fmt.Errorf("upstream stdin: %w", err)
}
upOut, err := cmd.StdoutPipe()
if err != nil {
    return nil, fmt.Errorf("upstream stdout: %w", err)  // upIn not closed
}
// ...
if err := cmd.Start(); err != nil {
    return nil, fmt.Errorf("starting upstream %q: %w", route.command, err)  // upIn, upOut not closed
}
```

**Suggested fix:**
```go
upIn, err := cmd.StdinPipe()
if err != nil {
    return nil, fmt.Errorf("upstream stdin: %w", err)
}
upOut, err := cmd.StdoutPipe()
if err != nil {
    _ = upIn.Close()
    return nil, fmt.Errorf("upstream stdout: %w", err)
}
if err := cmd.Start(); err != nil {
    _ = upIn.Close()
    _ = upOut.Close()
    return nil, fmt.Errorf("starting upstream %q: %w", route.command, err)
}
```

The same pattern applies in `connectUpstream` in `stdio.go` around lines 229–242.

---

### 3. `time.AfterFunc` SIGKILL timer in `signalUpstream` is never stopped — stale kill risk on PID reuse

**Location:** `cmd/eunox-mcp/stdio.go:272–289`  
**Priority:** Low | Effort: S | Risk: Theoretical

**Problem:** `signalUpstream` fires `time.AfterFunc(shutdownMs, func() { p.upCmd.Process.Kill() })` and returns. If the upstream exits cleanly within `shutdownMs`, `waitUpstream()` calls `cmd.Wait()`, which causes the OS to release and potentially reassign the PID. When the dangling timer fires, `Kill()` targets the recycled PID — a different, unrelated process. On Linux, PID reuse within the default 5 s is unlikely but possible on busy systems or with small `pid_max`.

Unlike `httpSession.close`, which uses `time.After` inside a `select` (and Go 1.23+ cleans those up properly), `time.AfterFunc` fires unconditionally regardless of whether the goroutine that set it is still running.

**Suggested fix:** stop the timer in the cleanup path:
```go
func (p *StdioProxy) signalUpstream(sig os.Signal) {
    if p.upHTTP != nil {
        p.upHTTP.close()
        return
    }
    if p.upCmd.Process != nil {
        _ = p.upCmd.Process.Signal(sig)
    }
    killMs := p.shutdownMs
    if killMs <= 0 {
        killMs = 5000
    }
    p.killTimer = time.AfterFunc(time.Duration(killMs)*time.Millisecond, func() {
        if p.upCmd.Process != nil {
            fmt.Fprintf(os.Stderr, "[eunox-mcp] Upstream did not exit; sending SIGKILL.\n")
            _ = p.upCmd.Process.Kill()
        }
    })
}
```
Then in the cleanup path (after `<-p.upstreamDone` in `Start`):
```go
if p.killTimer != nil {
    p.killTimer.Stop()
}
```

---

### 4. `decideInner`'s `default:` arm is a silent regression trap for future target types

**Location:** `cmd/eunox-mcp/pdp_jwt.go:723–737`  
**Priority:** Medium | Effort: S | Risk: None (defence in depth)

**Problem:** `decideInner` exists specifically because `resources/read` and `prompts/get` must synthesize an argument map from the target name before delegating to the inner PDP — falling through to `inner.Decide` with real args (or nil) causes MISSING_CONTEXT denials. The `default:` arm covers `TargetTypeTool` and `TargetTypeSystem` today and is safe for them. But a future developer who adds a new target type whose inner decision synthesizes arguments (analogous to resource/prompt) will silently fall through to `inner.Decide`, re-introducing the MISSING_CONTEXT regression this function exists to prevent. The comment warns about this, but a comment does not fail a test or block a PR.

**Current:**
```go
default:
    // NOTE: a new target type whose inner decision synthesizes
    // arguments from the target name MUST get its own case above…
    return p.inner.Decide(ctx, sessionID, target, args, sourceIP)
```

**Suggested fix:** enumerate the safe passthrough types explicitly so the `default:` arm is a hard denial (surfaced immediately in any integration test touching a new type):
```go
case capability.TargetTypeTool, capability.TargetTypeSystem:
    return p.inner.Decide(ctx, sessionID, target, args, sourceIP)
default:
    // Fail closed on any unrecognised target type. If a new type reaches here
    // it means decideInner was not updated — add a case above or a dedicated
    // Decide* method on the inner PDP.
    return denyResponse(capability.ErrCodeEnforcementError, "",
        fmt.Sprintf("decideInner: unhandled target type %q — update decideInner", target.Type))
```

This changes no observable behaviour for current code paths, converts a "please remember" comment into a runtime failure in tests, and removes the need for the warning comment.

---

## What's done well

- **Audit chain integrity**: three distinct startup markers (chain-resume-failed, tail-parse-failure, tail-integrity-failure), in-chain drop markers via `flushDropMarker`, and the genesis sentinel for fresh logs — all correct and well-tested.
- **Fail-closed everywhere**: kill switch before every decision, stray-argumentSchema guard positioned *before* the audit-mode defer, audience pinning ON by default with explicit opt-out, `ErrCodeNoJWTClaims` hard-denies even in list-filter paths.
- **JWT validation**: signature verified before claims decoded; terminal vs retryable error distinction prevents clock-skew retry on a bad payload; `json.Number` preservation for large integer claims.
- **Concurrency**: fine-grained locking in the HTTP session map, RWMutex for the audit send vs close race, atomic `lastActive` for idle reaping — no data races found.
- **Input validation**: JWKS URI scheme checked on redirects too, OAuth resource URI rejects control chars, JWT shorthand rejects percent-encoded keys and duplicate keys.
- **Circuit breaker**: generation-aware stale-probe dropping prevents split-brain close on concurrent probes.
- **Call counter**: length-prefixed composite keys are injective (fix for #273); `IncrementIfBelow` atomicity closes the TOCTOU gap that was letting denied retries self-extend the window (#204).

---

## Proposed action

Happy to implement any or all of these. Finding 1 (serial shutdown) and finding 4 (regression trap) are the highest-leverage changes. Finding 2 and 3 are small but complete the correctness story.

Which items would you like addressed first?

eunolabs:
Resolves the four findings from the 2026-06-18 engineering review in #779. All are localized correctness/defence-in-depth fixes — no architectural change, and no user-facing behavior change for current code paths.

### 1. `closeAllSessions` closed sessions serially — O(N × shutdownMs) shutdown
`cmd/eunox-mcp/http.go`. Each `s.close(shutdownMs)` blocks up to `shutdownMs` waiting for an unresponsive upstream, so closing N sessions one at a time made total HTTP shutdown time `N × shutdownMs`. Now fans the closes out with a `WaitGroup`, mirroring the existing `reapOnce` pattern, so the shutdown window stays bounded to a single `shutdownMs` regardless of session count.

### 2. Pipe FD leak on subprocess startup error paths (both transports)
`cmd/eunox-mcp/http.go` (`newSession`) and `cmd/eunox-mcp/stdio.go` (`connectUpstream`). `os/exec` does not close caller-held pipe ends on failure: if `StdoutPipe()` failed after `StdinPipe()` succeeded, the stdin pipe leaked; if `Start()` failed, both leaked. Both error paths now close the already-created pipes explicitly so repeated failed restarts cannot exhaust FDs.

### 3. Stale `time.AfterFunc` SIGKILL timer in `signalUpstream`
`cmd/eunox-mcp/stdio.go`. The graceful-shutdown SIGKILL fallback timer was armed but never stopped. If the upstream exited cleanly and `waitUpstream` reaped it, a later timer fire could `Kill()` a recycled PID. The timer is now captured on the proxy (mutex-guarded for the signal-goroutine-writes vs teardown-reads access) and stopped in `Start`'s teardown **before** `waitUpstream` reaps the subprocess and the OS can reuse the PID.

### 4. `decideInner`'s `default:` arm was a silent regression trap
`cmd/eunox-mcp/pdp_jwt.go`. The default arm forwarded unrecognised target types to `inner.Decide`. A future target type whose inner `Decide*` synthesizes an argument map from the target name (like resource/prompt) would silently fall through and re-introduce the empty-arg `MISSING_CONTEXT` denial this function exists to prevent. The safe passthrough types (`tool`, `system`) are now enumerated explicitly, and the `default:` arm fails closed with `ENFORCEMENT_ERROR` — converting a "please remember" comment into a test/runtime failure.

## Tests
- `TestJWT_decideInner_RoutesByTargetType` — pins the routing: resource/prompt reach the synthesizing inner methods, tool/system delegate verbatim, and an unknown target type fails closed with `ENFORCEMENT_ERROR` without consulting the inner PDP.
- `TestStdioProxy_SignalUpstream_KillTimerStoppable` — pins that `signalUpstream` captures the SIGKILL timer so teardown can `Stop()` it before PID reuse.
- Existing `-race` coverage (`TestHTTPCloseAllSessions`, `TestStdioProxy_ConnectUpstream_SubprocessLifecycle`, `TestStdioProxy_HTTPUpstreamWiring`) exercises the modified shutdown/teardown paths.

Full suite (`go test -race -count=1 ./cmd/eunox-mcp/ ./pkg/...`), `gofmt`, license check, and `golangci-lint` all pass.

Closes #779.

https://claude.ai/code/session_01XFWcx6j13bqbH3whRcWYnr

---
_Generated by [Claude Code](https://claude.ai/code/session_01XFWcx6j13bqbH3whRcWYnr)_


cmd/eunox-mcp/stdio.go
```diff
@@ -234,12 +249,17 @@ func (p *StdioProxy) connectUpstream(ctx context.Context) error {
...
+		// StdinPipe already succeeded; os/exec does not close caller-held pipe
+		// ends on failure, so close it here to avoid leaking the FD.
+		_ = upIn.Close()
 		return fmt.Errorf("upstream stdout: %w", err)
 	}
 	p.upWriter = newMsgWriter(upIn)
 	p.upReader = newMsgReader(upOut)
 
 	if err := p.upCmd.Start(); err != nil {
+		_ = upIn.Close()
```
eunolabs: These two closes are redundant: on a failed `Start()`, `os/exec` already closes the parent pipe ends. `Cmd.Start` defers `closeDescriptors(c.childIOFiles)` unconditionally and `closeDescriptors(c.parentIOPipes)` whenever it returns without `started` becoming true (`src/os/exec/exec.go`), and `parentIOPipes` is exactly `upIn`/`upOut`. So both the parent and child ends are released on `Start()` failure even without these lines — the explicit calls just double-close (harmless, error discarded).

The comment above (`os/exec does not close caller-held pipe ends on failure`) is accurate for the `StdoutPipe()`-failure branch — where `Start()` is never reached, so closing `upIn` there is genuinely necessary — but not for this `Start()` branch. Suggest either dropping these two lines, or, if keeping them defensively, giving them their own comment so the next reader isn't told something false about `Start()`.

Separate minor point on the `StdoutPipe()`-failure branch above: closing `upIn` releases the *parent* write end, but `StdinPipe` also created a *child* read end (held in `cmd.childIOFiles` / `cmd.Stdin`) that is only released by `Cmd.Start`'s defer — never reached on that path — so it lingers until the `*os.File` finalizer runs. Under the FD-exhaustion scenario this hardening targets, that's still ~1 FD per failed attempt not promptly reclaimed. The child end isn't exposed, so there's no clean fix without restructuring; worth a comment noting the guarantee is partial.

---

cmd/eunox-mcp/http.go
```diff
@@ -1462,6 +1465,8 @@ func (p *HTTPProxy) newSession(ctx context.Context, route *upstreamRoute, client
 	sess.upReader = newMsgReader(upOut)
 
 	if err := cmd.Start(); err != nil {
+		_ = upIn.Close()
```
eunolabs: Same as the note on `stdio.go`'s `connectUpstream`: on a failed `Start()`, `os/exec` already closes `parentIOPipes` (here `upIn`/`upOut`) and `childIOFiles` via `Cmd.Start`'s defer, so these two lines are redundant double-closes. The `StdoutPipe()`-failure `upIn.Close()` above is the one that's actually load-bearing (Start not reached there).

---

cmd/eunox-mcp/stdio.go
```diff
@@ -75,6 +75,13 @@ type StdioProxy struct {
 	upHTTP   *httpUpstream  // remote HTTP upstream only
 	upReader msgSource
 
+	// killTimer is the SIGKILL fallback armed by signalUpstream during a graceful
+	// shutdown; Start's teardown stops it once the upstream has exited so it
+	// cannot fire Kill() against a PID the OS has since reused (#779). killMu
```
eunolabs: The change is fine and the `killMu` guarding is correct (write in the signal goroutine, read in teardown). But the "cannot fire `Kill()` against a PID the OS has since reused" rationale doesn't hold on the supported runtime. The callback signals via `p.upCmd.Process.Kill()`, and after `waitUpstream()` → `Cmd.Wait()` the `os.Process` is marked done:

- On Linux/Go 1.24 a reaped `Process` uses a pidfd (`modeHandle`); `pidfdSendSignal` is bound to the original process and structurally cannot hit a reused PID.
- In the legacy PID path, `pidSignal` returns `ErrProcessDone` on `statusDone` *before* ever calling `syscall.Kill`, and `pidWait` deliberately serialises against in-flight signals via `sigMu`.

So a stale timer firing after `Wait()` is a no-op, not a wrong-process kill.

What stopping the timer *does* buy — and what I'd anchor the comment on — is suppressing the misleading `"[eunox-mcp] Upstream did not exit; sending SIGKILL."` line: the callback logs that unconditionally before the `Kill()`, so after a clean exit you'd otherwise get a bogus SIGKILL log line ~`shutdownMs` later. That, plus freeing the timer promptly, is real and worth doing — just consider retargeting the rationale so the threat model stays accurate.

---

eunolabs (commented):
Reviewed all four fixes against the base. Net: the changes are correct and safe, and the new tests are well-targeted. Two of the four are unambiguous improvements; the other two work and are harmless, but their stated rationale overstates the actual failure mode on the supported Go runtime. Details inline — none of this blocks merge.

- **#1 `closeAllSessions` concurrent fan-out** — correct. Mirrors `reapOnce` exactly; `s.close` is idempotent (`closeOnce.Do`) and per-session, the only shared read is the immutable `shutdownMs`, and the fan-out is bounded by `maxSessions`. (go.mod is 1.25, so the per-iteration loop variable captured by the goroutine is safe.)
- **#4 `decideInner` fail-closed default** — correct, and the strongest fix here. All four `TargetType`s (tool/resource/prompt/system) are now handled explicitly and `default` only catches a genuinely new type, matching the repo's fail-closed / no-silent-fall-through invariant. Test is sound: `staticPDP` returns the sentinel for all three inner methods, so the resource/prompt delegation assertions are meaningful and the unknown-type case proves the default arm doesn't consult `inner`.
- **#2 pipe-FD cleanup** — the `StdoutPipe()`-failure close is necessary; the `Start()`-failure closes are redundant and the comment is inaccurate for that branch (see inline on `stdio.go` / `http.go`).
- **#3 kill-timer stop** — `killMu` guarding is correct, but a stale timer firing post-`Wait()` can't actually signal a reused PID on this runtime; the real benefit is suppressing a misleading log line (see inline).

For transparency: I re-derived the `os/exec` / `os` behavior from the Go 1.24 sources rather than running the suite, so the #2/#3 notes are about rationale/precision, not observed failures.

---

cmd/eunox-mcp/stdio.go
eunolabs: Confirmed against the go1.25.11 source: `Cmd.Start`'s deferred cleanup closes `parentIOPipes` (upIn/upOut) when `!started` and `childIOFiles` unconditionally, so the `Start()`-failure closes were redundant double-closes. Dropped them in e0aa205 with a note so they don't get re-added.

Kept the `StdoutPipe()`-failure `upIn.Close()` (Start is never reached there) and reworded its comment to record exactly your second point: it only reclaims the parent write end — the child read end `StdinPipe` stashed in `childIOFiles` is unreachable from here and lingers until its `*os.File` finalizer, so the guarantee is partial (one of two FDs).

---

cmd/eunox-mcp/http.go
eunolabs: Same fix applied here in e0aa205 — dropped the redundant `Start()`-failure double-closes, kept the load-bearing `StdoutPipe()`-failure `upIn.Close()` and clarified that it only reclaims the parent end.

---

cmd/eunox-mcp/stdio.go
eunolabs: Agreed and verified against go1.25.11: after `Wait()`, `pidfdSendSignal` and `pidSignal` both return `ErrProcessDone` before any `syscall.Kill`, and the pidfd handle is bound to the original process — so a stale timer is a no-op, never a reused-PID kill. Retargeted the struct/teardown/test comments in e0aa205 onto the real benefit: suppressing the unconditional `"[eunox-mcp] Upstream did not exit; sending SIGKILL."` log line on clean shutdown and freeing the timer promptly. The `killMu` guarding and the fix itself are unchanged.

---

cmd/eunox-mcp/pdp_jwt.go
```diff
@@ -719,13 +719,21 @@ func (p *JWTPDP) decideInner(ctx context.Context, sessionID string, target Enfor
...
-	default:
-		// tools/call (and any target whose inner method does not synthesize an
-		// argument map). NOTE: a new target type whose inner decision synthesizes
-		// arguments from the target name MUST get its own case above — falling
-		// through to inner.Decide here forwards args verbatim and re-introduces the
-		// empty-arg MISSING_CONTEXT regression decideInner exists to prevent.
+	case capability.TargetTypeTool, capability.TargetTypeSystem:
+		// tools/call and system targets (e.g. sampling/createMessage) carry their
+		// real arguments, so the inner Decide evaluates conditions against them
+		// directly — no name-to-argument synthesis is needed.
```
eunolabs: Tracing the call graph, no production path reaches `decideInner` with `target.Type == TargetTypeSystem`, so this arm is currently unreachable and the `(e.g. sampling/createMessage)` example is inaccurate:

- `decideInner` is only called from `JWTPDP.Decide` (the no-capabilities fall-through and the post-JWT delegation), and `Decide` is only ever invoked with `TargetTypeTool` (transports' tools/call — `http.go:976` / `stdio.go:502`), `TargetTypeResource` (via `DecideResourceRead`), or `TargetTypePrompt` (via `DecidePromptGet`). Never `system`.
- `sampling/createMessage` is enforced through the separate `SamplingAuthorizer` path: `decideSampling(ctx, pdp, …)` (`http.go:1209` / `stdio.go:776`) → `JWTPDP.DecideSampling` → inner `sa.DecideSampling`. It never builds a `system` `EnforceTarget` nor calls `Decide`/`decideInner`. (`mcpMethodToTargetType`, which maps the method to `TargetTypeSystem`, is used only for audit-envelope derivation, not dispatch.)

So adding `system` here just preserves the pre-PR behavior (the old `default` also sent `system` → `inner.Decide`) for a path that can't currently be taken. Two things worth tightening:

1. **Comment accuracy** — don't cite sampling as the example; it points the next reader at the wrong enforcement entrypoint. Note too that `inner.Decide` is *not* where the sampling opt-in is checked (`inner.DecideSampling` is), so if a `system` target ever *did* reach this arm, delegating to `inner.Decide` likely wouldn't enforce sampling correctly.
2. **Fail-closed consistency** — given this function exists specifically to force explicit handling rather than silently delegate, consider dropping `system` from the passthrough and letting it fall to the fail-closed `default`. That keeps `tool` as the only intentional passthrough and turns any future "system target reaches `Decide`" into a loud `ENFORCEMENT_ERROR` instead of a silent (and probably-wrong) `inner.Decide`. The `system` subtest in `TestJWT_decideInner_RoutesByTargetType` is currently pinning a routing that production code can't otherwise exercise.

Not blocking — behavior is correct today either way; this is comment precision + a fail-closed judgement call.

---

eunolabs (commented):
Re-reviewed after the second commit (`e0aa205`). It cleanly resolves all three points from the first pass — I verified each against the current code:

- **Pipe FDs** — the redundant `Start()`-failure closes are gone from both `connectUpstream` and `newSession`, and the new comments correctly state that a failed `Cmd.Start` closes both the parent (`upIn`/`upOut`) and child ends via its own deferred cleanup. The retained `StdoutPipe()`-failure `upIn.Close()` now documents that it only reclaims the parent end (the child stdin end lingers until its `*os.File` finalizer) — accurate, and a fair statement of the partial guarantee.
- **Kill timer** — the rationale is retargeted to "the callback is already a no-op after reap (`Process.Signal` → `ErrProcessDone`, and the pidfd is bound to the original process so it can't reach a reused PID); stopping it suppresses the spurious `sending SIGKILL` log line on clean shutdown and frees the timer." That matches the Go runtime behavior. `killMu` guarding and the fix itself are unchanged. Test comment updated to match.

One **new** point I should have caught the first time, on the `decideInner` `system` arm — see inline. Short version: no production path routes a `system` target through `decideInner` (sampling is enforced via the separate `decideSampling`/`SamplingAuthorizer` path), so that case is currently unreachable and its `e.g. sampling/createMessage` comment is misleading. Behavior is still correct — it's a comment-accuracy / fail-closed-consistency nit, not a blocker.

Net: the four fixes are correct and well-tested; closeAllSessions and the decideInner fail-closed `default` remain the standouts. Nothing here blocks merge.

---

cmd/eunox-mcp/pdp_jwt.go
eunolabs: Good catch — confirmed the call graph: `decideInner` is only reached from `JWTPDP.Decide`, which the transports invoke solely with `TargetTypeTool` (`stdio.go:532`, `http.go:976`) plus `TargetTypeResource`/`TargetTypePrompt` internally; `sampling/createMessage` is enforced through the separate `decideSampling` → `SamplingAuthorizer.DecideSampling` path and never builds a `system` target nor calls `Decide`/`decideInner`. So the `system` passthrough was dead code and the `inner.Decide` example was doubly misleading (it isn't where the sampling opt-in is checked).

Went with your option 2 (f89f762): dropped `system` from the passthrough so `tool` is the only intentional one, and `system` now falls to the fail-closed `default` → `ENFORCEMENT_ERROR`. That turns any future "system target reaches `Decide`" into a loud failure instead of a silent, probably-wrong `inner.Decide`, matching the no-silent-fall-through invariant. Behavior is unchanged today (the arm was unreachable). The `system` subtest in `TestJWT_decideInner_RoutesByTargetType` now asserts the fail-closed deny alongside an unrecognised type, and the comments are rewritten to point at `DecideSampling` rather than `inner.Decide`.

---

Issue #732, opened 2026-06-17
Code review: 9 findings (1 High, 3 Med, 5 Low)


eunolabs:
# Principal engineer review — 2026-06-17

Findings from a rigorous read of `cmd/eunox-mcp/` and `pkg/`. Grouped by priority.

---

### #1 — `JWTPDP.Decide()` JWT-only allow path ignores injected clock · **High** · Effort S

**File:** `cmd/eunox-mcp/pdp_jwt.go:663-667`

`JWTPDP` carries a `clock enforcement.Clock` field for test-time freezing (used in `ValidateToken` lines 411–416). The JWT-only allow return (when `HasCapabilities == true`, conditions pass, `p.inner == nil`) stamps `DecidedAt` with `time.Now()` directly. Every other allow path uses the injected clock. Issue #686 added `RequestID`/`DecidedAt` to this path but omitted clock injection.

Tests that freeze time and exercise JWT-only mode get real wall-clock timestamps — `DecidedAt` assertions are non-deterministic.

```go
// Before (pdp_jwt.go:663-667)
return capability.EnforceResponse{
    Decision:  capability.DecisionAllow,
    RequestID: uuid.New().String(),
    DecidedAt: time.Now().UTC().Format(time.RFC3339Nano),
}

// After
now := time.Now()
if p.clock != nil {
    now = p.clock.Now()
}
return capability.EnforceResponse{
    Decision:  capability.DecisionAllow,
    RequestID: uuid.New().String(),
    DecidedAt: now.UTC().Format(time.RFC3339Nano),
}
```

A helper `func clockNow(c enforcement.Clock) time.Time` would also fix the three-pattern inconsistency (Finding #9 below).

---

### #2 — `ConditionTypeRedactFields` is a directive type constant placed among condition type constants, with no handler · **Med** · Effort S

**File:** `pkg/capability/condition.go:20`

```go
const (
    // ...
    ConditionTypeRedactFields = "redactFields"  // no struct, no ConditionType() method, no handler
    // ...
)
```

`registerBuiltins()` does not register a handler for `redactFields`. There is no `RedactFieldsCondition` struct. `redactFields` is a *directive/obligation* type. The constant is in the wrong file.

A manifest author who writes `type: redactFields` as a condition gets a hard-deny `CONDITION_FAILED: unknown condition type: redactFields` — correct outcome, confusing error, wrong file.

[...]

PR #803, opened 2026-06-18
Address code-review #732 findings and fix six MCP proxy defects

eunolabs:
This branch carries two independent commits.

## 1. `refactor:` triage of code review #732

Every one of the 9 findings in #732 was re-checked against current `main`; several were already fixed by commits that landed after the review.

**Implemented (behavior-preserving):**

| # | Finding | Change |
|---|---------|--------|
| 2 | `ConditionTypeRedactFields` misplaced among condition constants | Removed the duplicate constant (`redactFields` is a directive — `DirectiveTypeRedactFields` already exists with a struct + handler). The conditions-migration hint now compares against `DirectiveTypeRedactFields`. Same value, identical behavior. |
| 4 | Audit chain-state decode lacked `UseNumber` | Chain-state record now decodes with `json.NewDecoder` + `UseNumber`, matching `VerifyRecord`, so a future reader of `rec.Details` never sees rounded `float64` integers. Typed chain fields (`Seq`/`HMAC`/`PrevHMAC`) decode identically — defense-in-depth, no current bug. |
| 8 | `defaultAuditLog` stored unexpanded `~` (non-obvious contract) | `resolveAuditLogPath` is now the sole consumer of the constant; `openAuditSink` and `suggest` route through it. The "must `expandHome`" rule is now structural, not convention, and documented on the constant. |
| 9 | Clock nil-check pattern inconsistent across PDPs | Extracted a shared `clockNow(enforcement.Clock) time.Time`; `alwaysAllowPDP.wiretapAllow` and `JWTPDP.now` both use it, so the injected-clock fallback can't drift. Added a focused `TestClockNow`. |

**Already fixed (verified, no change needed):**

- **#1** (High) — `JWTPDP.Decide`'s JWT-only allow already stamps `DecidedAt` from the injected clock via `p.now()` (landed with #722/#737).
- **#3** (Med) — both HTTP drift paths (`http.go`, `http_remote.go`) now run `runHTTPDriftCheck` **synchronously** before the session is returned to the client, so the async FM-5 enforcement window the finding describes no longer exists.

**Deliberately left as-is (rationale):**

- **#5** — `findConstraint` is a correct single-pass selector in a fail-closed path. The suggested two-phase scan duplicates the filter and double-iterates; the alternative changes denial codes (`CAPABILITY_DENIED` → `AUTHORIZATION_FAILED`), which the finding itself flags as needing SIEM/runbook coordination. Not worth the regression risk for a readability tweak.
- **#6** — `truncatedObligations`' loop is intentional defense-in-depth (#619) that makes the cap invariant unconditional. Replacing it with a panic in the audit-write path would be a regression.
- **#7** — changing the genesis sentinel (`sha256:genesis`) would break HMAC-chain verification of existing audit logs (it's verified for `seq==1` records) for a cosmetic gain; the dual-accept the finding suggests conflicts with the repo's no-back-compat-shim rule.

## 2. `fix:` six MCP proxy defects

Unrelated correctness fixes that were already on this branch:

- **init** — carry `--upstream-tls-skip-verify` into the scaffolded config so the emitted `proxy --config` can reconnect to a self-signed upstream.
- **oauth** — build the RFC 9728 metadata URL by inserting the well-known segment between authority and path (matching the per-route served path), and require `https` + reject fragments in `--oauth-resource`.
- **kill** — bracket IPv6 literal hosts via `net.JoinHostPort` so `--host ::1` produces a parseable control-endpoint authority.
- **stdio http bridge** — surface a JSON-RPC error on a mismatched upstream response id instead of force-relabeling it with the request id.
- **jwt** — match the `Bearer` auth-scheme case-insensitively (RFC 7235 §2.1).

## Testing

`make test` (full `-race` suite), `make lint`, `make check-fmt`, and `make check-license` all pass. New regression tests accompany each fix.

Refs #732

https://claude.ai/code/session_01RPcveRmPmC31jje8EWafYX

---
_Generated by [Claude Code](https://claude.ai/code/session_01RPcveRmPmC31jje8EWafYX)_


cmd/eunox-mcp/audit.go
```diff
@@ -1668,7 +1671,15 @@ func verifyAuditLog(r io.Reader, verifier *auditSink, requestID string, since ti
...
-		unmarshalOK := json.Unmarshal(line, &rec) == nil
+		// Decode with UseNumber for parity with VerifyRecord: a plain json.Unmarshal
+		// coerces every number in the interface-typed Details map to float64, so any
+		// integer above 2^53 (e.g. a saturated AUDIT_RECORDS_DROPPED count) would be
+		// silently rounded for any future chain-state or reporting code that reads
+		// rec.Details. The typed chain fields (Seq, HMAC, PrevHMAC) decode identically
+		// either way; this is defense-in-depth, not a fix for a current bug.
+		dec := json.NewDecoder(bytes.NewReader(line))
+		dec.UseNumber()
+		unmarshalOK := dec.Decode(&rec) == nil
```
eunolabs: `json.Unmarshal` and `json.Decoder.Decode` are **not** equivalent here, and the difference is security-relevant: `Unmarshal` errors if there are non-whitespace bytes after the top-level value, whereas `Decoder.Decode` reads one value and silently ignores anything after it.

`VerifyRecord` already decodes with `Decoder.Decode` (≈line 1527) and re-computes the HMAC over the *re-marshaled fields*, so its HMAC check also ignores trailing bytes. That means the strict `json.Unmarshal` here was the **only** gate catching a tampered record line of the form `{…valid signed record…}GARBAGE`. After this change:

- `unmarshalOK` becomes `true` for such a line (trailing bytes dropped),
- `VerifyRecord` still matches (the trailing bytes aren't part of `rec`),
- so `verifyAuditLog` reports PASS where it previously emitted `INVALID  malformed record … not valid JSON` via the `if !unmarshalOK` gate (≈1788) — whose own comment assumes `VerifyRecord` would otherwise catch it, which it won't for this case.

So this isn't behavior-neutral: alongside the `UseNumber` precision fix (which is good), it also relaxes trailing-data strictness in the audit verifier — a narrow but real reduction in tamper-evidence in the one artifact the project treats as sacrosanct. The comment ("decode identically either way", "not a fix for a current bug") doesn't capture that.

Suggest restoring parity with `Unmarshal`:

```go
dec := json.NewDecoder(bytes.NewReader(line))
dec.UseNumber()
unmarshalOK := dec.Decode(&rec) == nil && !dec.More()
```

`Decoder.More()` skips trailing whitespace (so clean lines still pass) but reports any trailing token, matching `Unmarshal`'s rejection. Worth applying the same `!dec.More()` guard inside `VerifyRecord` too, so the HMAC path closes the same gap.

---

cmd/eunox-mcp/stdio_http_upstream.go
```diff
@@ -145,8 +145,16 @@ func (h *httpUpstream) post(ctx context.Context, msg rpcMsg) { //nolint:gocritic
...
+	case msgKey(resp.ID) != msgKey(msg.ID):
+		// JSON-RPC 2.0 §5: a response's id MUST equal the id of the request it
+		// answers. The bridge issues exactly one POST per request, so a mismatched
+		// (or absent) id means the upstream returned a response for a different
+		// request — a protocol violation. Force-relabeling it with msg.ID (the
+		// prior behavior) would bind an unrelated result to this request and mask
+		// the violation, so surface it as an internal error keyed to msg.ID
+		// instead. The matching case below needs no relabeling: the ids are already
+		// byte-identical.
+		resp = errorResponse(msg.ID, -32603, "upstream response id does not match request id for "+msg.Method)
```
eunolabs: Good change — relabeling an unverifiable response onto the caller's id was the worse default. Two small considerations:

1. `msgKey` is a raw-bytes compare (`string(*id)`), so this is byte-level, not semantic. For an id forwarded from the host and echoed verbatim by a well-behaved upstream this is exactly right, and it matches how `msgKey` is used elsewhere. The only edge is an upstream that re-encodes a semantically-equal id (number↔string, or a differently-formatted number) — that now hard-fails with `-32603` where the old code relabeled and succeeded. Almost certainly fine in practice; just flagging the strictness increase.

2. When the upstream returns a genuine *error* response whose id is null/absent (e.g. `{"jsonrpc":"2.0","id":null,"error":{"code":-32601,"message":"Method not found"}}`), this overwrites the upstream's real error code/message with the generic `-32603 "id does not match"`. Fail-closed is defensible, but the host loses the upstream's diagnostic. Worth deciding whether an already-`error` response should keep its payload (relabeled to `msg.ID`) rather than be replaced.

---

eunolabs (commented):
## Review: Address #732 findings and fix six MCP proxy defects

Thorough, well-tested PR. The #732 triage is exemplary — re-verifying each finding against current `main` and splitting it into implemented / already-fixed / deliberately-left with concrete rationale is exactly how a review batch should be closed out. Every fix ships a focused regression test that names the bug it pins, the refactors (`clockNow`, `resolveAuditLogPath`) are genuinely behavior-preserving, and the OAuth metadata-URL fix correctly aligns the advertised URL with the path-inserted variant that `docs/adr/0002` already specifies.

One finding worth addressing before merge, plus a few minor notes; details inline.

### Worth addressing
- **`audit.go` — the `json.Unmarshal` → `json.Decoder.Decode` switch in `verifyAuditLog` quietly relaxes trailing-data strictness** (inline). `Unmarshal` rejects non-whitespace bytes after the top-level value; `Decoder.Decode` ignores them. Since `VerifyRecord` already decodes with `Decoder.Decode`, the strict `unmarshalOK` was the *only* gate catching a record line with appended bytes — so after this change such a line now reports PASS instead of `INVALID malformed record`. Narrow vector (field-level edits are still caught by the HMAC chain), but it touches the audit-integrity invariant, and the new comment's "decode identically either way / not a fix for a current bug" understates it. One-line fix: `&& !dec.More()`.

### Minor / questions
- **`stdio_http_upstream.go`** (inline): two notes on the raw-byte `msgKey` comparison and on discarding the upstream's original error payload when its id is null/absent.
- **`oauth_metadata.go`** — `buildOAuthMetadataURL` uses only scheme/host/path, so any query string on the resource is dropped, while `validateOAuthResourceURI` rejects fragments but not queries. Resource identifiers with queries are unusual; for symmetry, consider rejecting a query in validation too (or documenting that it's intentionally ignored).
- The stricter `--oauth-resource` validation (now rejects `http://` and fragments) is a user-facing change. The metadata-URL fix needs no doc update (it matches the ADR), but a one-line changelog/flag-doc note for the new rejections would be courteous to anyone with an existing `http://` resource configured.

Nice work overall — the only substantive item is the audit-verifier strictness.

---

cmd/eunox-mcp/audit.go
eunolabs: Good catch — you're right, this was not behavior-neutral and my comment was wrong to claim so. `json.Unmarshal` rejected trailing bytes; `Decoder.Decode` drops them, and since `VerifyRecord` re-signs the decoded fields it ignores them too, so `{…record…}GARBAGE` would have flipped from `INVALID` to a false `PASS`.

Fixed in a7fec3e by adding the `!dec.More()` guard in **both** places, as suggested:
- `verifyAuditLog`: `unmarshalOK := dec.Decode(&rec) == nil && !dec.More()` — restores the `!unmarshalOK` → `INVALID` verdict.
- `VerifyRecord`: returns `(false, error)` on trailing data, closing the HMAC path independently.

Comment updated to describe the trailing-data guarantee accurately. Added regressions: a `VerifyRecord` subtest (record + appended `GARBAGE` must not verify) and `TestAuditChain_TrailingDataIsInvalid` (a tampered line makes the whole log verdict fail).

---

cmd/eunox-mcp/stdio_http_upstream.go
eunolabs: Thanks — addressed in a7fec3e.

**(2) error responses:** agreed, and implemented. On an id mismatch the bridge now splits by shape: an `error` response is preserved (its real code/message) and just stamped with the correct id, while a `result` with the wrong id is still refused with `-32603`. The reasoning is that this is a strict 1-POST/1-response-body bridge, so the reply *is* this request's — an error carries no result to mis-bind and its diagnostic pertains to our request, whereas a mismatched result might carry data computed for the wrong request (the masking the original fix was guarding against). Your `{"id":null,"error":{"code":-32601,...}}` example now reaches the host as `-32601 "Method not found"` keyed to the request id; added `TestHTTPUpstream_Post_MismatchedErrorIDPreserved` for exactly that.

**(1) byte-level vs semantic id compare:** left as-is intentionally, and called it out in the code comment. For the 1:1 bridge a verbatim echo is the norm; a semantically-equal re-encode (number↔string) is the rare edge, and it now only affects the *result* path (errors are preserved regardless).

---

eunolabs (commented):
## Re-review (commit `a7fec3e`)

Both findings from my previous review are addressed, and correctly.

**1. Audit trailing-data gap — resolved.** The `!dec.More()` guard is now applied in both places:
- `verifyAuditLog`: `unmarshalOK := dec.Decode(&rec) == nil && !dec.More()` — a `{…signed record…}GARBAGE` line now yields `unmarshalOK == false`, hits the `if !unmarshalOK` gate (audit.go:1788), is counted `INVALID` once, and `continue`s (so it never double-reports via the later `VerifyRecord` call).
- `VerifyRecord`: an explicit `if dec.More()` → `false, error`, closing the HMAC path for direct callers.

`Decoder.More()` skips trailing whitespace but flags any trailing token, so this matches `json.Unmarshal`'s old strictness exactly while leaving clean records passing. Good regression coverage in `TestAuditChain_TrailingDataIsInvalid` and the `security_test.go` `VerifyRecord` subtest. 

**2. stdio bridge id mismatch — resolved well.** Splitting by response shape is the right call: a genuine upstream *error* keeps its real code/message (relabeled to the request id, since the 1:1 POST guarantees it's this request's diagnostic), while a mismatched *result* is still refused with `-32603` (fail closed — don't bind data you can't attribute). The byte-level `msgKey` caveat I raised is now documented in-code, and `TestHTTPUpstream_Post_MismatchedErrorIDPreserved` pins the new path.

**Remaining (non-blocking):** the only open item is the optional nit from last time — `buildOAuthMetadataURL` drops any query string while `validateOAuthResourceURI` rejects fragments but not queries. Purely a symmetry/cosmetic point; fine to leave.

**CI:** all completed checks are green (demo e2e/gateway/stdio/JWT, OPA comparison, govulncheck, Trivy, NOTICE, Types Sync, builds, benchmark); Test/Lint/gofmt are still running from this push — nothing failing.

LGTM — no blocking items left. Nice, fast turnaround on the audit fix.

---

PR #811, opened 2026-06-18
sec: fix large-integer precision loss in policy enforcement

### LINKED ISSUE #687: [rev-20260616T1110Z-7bb934] JWT custom claims silently lose precision for integers > 2^53

eunolabs:
## Summary

Fix a critical precision-loss vulnerability where request arguments above 2^53 could be silently rounded to match a different authorized value, allowing an attacker to bypass policy constraints. Additionally, harden millisecond timeout settings against integer overflow and fix environment variable expansion to treat secrets as literal data instead of re-parsing them as YAML.

## Why

**Large-integer precision loss (CVE-worthy):** JSON numbers above 2^53 cannot be exactly represented as float64. When request parameters were decoded with the default `json.Unmarshal` path, a value like 9007199254740993 (2^53+1) would round to 9007199254740992 (2^53), matching a different enum/allowedValues constraint than the caller actually sent. The upstream would receive the original bytes verbatim, creating a gap where policy authorized one value while the upstream was handed another.

**Millisecond overflow:** Config settings like `listen.sessionIdleTimeoutMs` or `defaults.upstreamTimeoutMs` that exceed ~9.2e12 ms would overflow `time.Duration` (int64 nanoseconds), wrapping to negative values. A negative idle timeout would cause the reaper to treat every session as stale and close it immediately; a negative upstream timeout would expire calls instantly.

**YAML re-parsing of secrets:** Environment variable expansion happened on the raw config file text before YAML parsing. A secret like `#mytoken` would be read as a trailing comment and silently blank the field; `:` or newlines in a token could alter the parse structure. This is a high-severity auth bypass if a secret contains YAML metacharacters.

Closes #687 (large-integer precision), #699 (millisecond overflow), and the YAML re-parsing issue.

## Type of change

- Security hardening

## How it was tested

```
make test    # All new tests pass; existing enforcement tests still pass
make lint    # No new findings
```

New tests added:
- `TestStdioHandleToolsCall_AllowedValuesLargeIntPrecision` — end-to-end: 2^53 is allowed, 2^53+1 is denied
- `TestDecodeParams_PreservesLargeIntegerPrecision` — unit: json.Number survives decode without float64 rounding
- `TestValidateArgumentSchema_EnumLargeIntPrecision` — unit: enum comparison at full int64 precision
- `TestMsToDuration_SaturatesInsteadOfOverflowing` — unit: millisecond overflow saturates to MaxInt64
- `TestLoadGatewayConfig_EnvValueIsLiteralNotYAML` — unit: env values are not re-parsed as YAML

## Security considerations

**Fail-closed preserved:** Policy now correctly denies large integers that don't exactly match an enum/allowedValues constraint, even if they round to the same float64. Millisecond overflow is caught at config validation time (startup refusal). Environment secrets are no longer re-interpreted as YAML syntax.

**Authority:** This fix *narrows* what a manifest permits (large integers now require exact matches), closing a fail-open gap. No new authority is granted.

**Secrets:** Environment variable expansion now happens on decoded string values, never on raw file text. A secret containing `#`, `:`, or other YAML metacharacters is substituted as literal data and used verbatim, not re-parsed.

**Audit tape:** No changes to audit record structure or signing.

## Changes

1. **`cmd/eunox-mcp/jsonrpc.go`** — New `decodeParams()` function that unmarshals request params with `UseNumber()` to preserve json.Number for large integers instead of collapsing to float64.

2. **`cmd/eunox-mcp/http.go`, `stdio.go`** — Use `decodeParams()` for all tools/call and prompts/get parameter decoding. Add `msToDuration()` helper that saturates at `math.MaxInt64` instead of overflowing. Use it for all millisecond-to-duration conversions (shutdown, upstream timeout, idle reaper).

3. **`cmd/eunox-mcp/gateway

https://claude.ai/code/session_01JY9iWnMVHqLsEdrd6v4x5P


cmd/eunox-mcp/gateway_config.go
```diff
@@ -79,6 +80,40 @@ func expandEnvRefs(s string) string {
...
+		for i := 0; i < v.NumField(); i++ {
+			if f := v.Field(i); f.CanSet() { // CanSet skips unexported fields
+				expandEnvInStrings(f)
+			}
+		}
+	case reflect.Slice, reflect.Array:
+		for i := 0; i < v.Len(); i++ {
+			expandEnvInStrings(v.Index(i))
+		}
+	}
```
eunolabs, line 120: This walk silently skips `reflect.Map` (and `reflect.Interface`), so its correctness rests entirely on the invariant in the doc comment — "the tree contains only structs, slices, strings…". That's true today (I checked: no map fields), but it's a quiet, security-relevant landmine. If anyone later adds a `map[string]string` field — e.g. per-upstream custom/auth headers, which this PR's own description lists as a secrets use case — its values would silently *not* be env-expanded, and a `${TOKEN}` reference would be handed to the upstream as the literal string `${TOKEN}`. (Map values also aren't addressable via `CanSet`/`SetString`, so they'd need `SetMapIndex` — i.e. real work, not a free fall-through.)

Since the failure mode is silent and auth-related, consider hardening the invariant rather than relying on the comment: a `default:` arm that panics on `reflect.Map`/`reflect.Interface` (caught at startup and in tests), or a small reflective test asserting the config tree contains no map/interface fields. Non-blocking.

---

schemas/eunox-gateway-config.schema.json
```diff
@@ -68,6 +68,7 @@
           "type": "integer",
           "description": "Reap a session whose host has sent no request for this many milliseconds, closing its upstream so idle sessions cannot pin resources. 0 (default) disables idle reaping. Only valid for transport: http.",
           "minimum": 0,
+          "maximum": 9223372036854,
```
eunolabs, line 71: `9223372036854` is hard-coded here (and again at the `upstreamTimeoutMs` `maximum` below), while the loader derives the same bound as `maxDurationMs = int64(math.MaxInt64) / int64(time.Millisecond)` in `http.go`. These are independent literals with nothing pinning them together — `TestGatewaySchema_MatchesStructs` only checks field-name parity, not value bounds — so if `maxDurationMs` ever changes, this schema silently drifts out of lockstep with the Go validation. Consider a one-line test asserting the schema `maximum` equals `maxDurationMs`, or at least a comment cross-referencing the constant. Minor.

---

eunolabs (commented):
## Review: `sec: fix large-integer precision loss in policy enforcement`

Tightly-scoped, well-tested security fix. I traced all three issues through the consuming code and the fixes hold up. Two non-blocking notes inline; everything else below is confirmation.

### What it does
- **Large-int precision (#687):** new `decodeParams` decodes `tools/call` / `prompts/get` params with `json.Decoder.UseNumber()`, so arguments arrive as `json.Number` instead of `float64`. The engine already compares `json.Number` at full `int64` precision (`asInt64` / `numericEqual`), and `schema.go` is reordered so the enum check runs on the un-coerced `json.Number` *before* the float64 coercion the keyword checks need. Net effect: `2^53+1` no longer collides with `2^53`.
- **ms→Duration overflow (#699):** `msToDuration` saturates at `MaxInt64`; `validate()` additionally rejects `sessionIdleTimeoutMs` / `upstreamTimeoutMs` above `maxDurationMs` at startup; the JSON schema gets matching `maximum`s.
- **YAML re-parse of secrets:** env expansion moved to *after* the strict decode, walking decoded string fields via reflection — so a secret like `#token` can no longer be re-interpreted as YAML (comment/anchor/mapping) and silently blank `authToken`.

### Verified correct
- **Scope is complete.** The only enforced methods that evaluate caller-supplied numeric arguments are `tools/call` and `prompts/get` — both switched to `decodeParams`. `sampling/createMessage` goes through `decideSampling`, which passes only session/IP (binary opt-in + kill-switch + ipRange), and `resources/read|subscribe` key on a URI string. Neither was ever on the lossy numeric path, so leaving them untouched is right.
- **`schema.go` reorder is sound** — enum compares at full precision, then the coercion guarantees `val` is `float64` before the `val.(float64)` assertions in the type/keyword checks.
- **`msToDuration` callers are guarded** — `<= 0` is handled before each call (drift/upstream-timeout guards, the `killMs` default), and the idle reaper is only launched when `sessionIdleMs > 0` (`http.go:472`). The `validate()` upper bound also makes the saturation branch unreachable for the two config-driven values — fine as defense-in-depth.
- **Reflection walk covers the current tree** — confirmed there are no `map`/`interface` fields today, so every secret-bearing field (`authToken`, `upstreamAuthHeader`, `command`, `args`, `policy`, URLs) is reached. Fail-closed preserved.
- Tests are strong: end-to-end allow/deny at `2^53` / `2^53+1`, the unit decode, the enum-precision unit, ms saturation, and the literal-not-YAML cases. Docs + schema updated in the same PR per CONTRIBUTING.

### Minor / non-blocking
- `decodeParams` uses `Decoder.Decode`, which (unlike `json.Unmarshal`) tolerates trailing bytes after the first JSON value. Not reachable here, since `msg.Params` is a `json.RawMessage` already extracted by a successful outer unmarshal (exactly one well-formed value, no trailing) — just worth knowing if params are ever sourced differently.
- **Behavior change worth a changelog line:** expanding *after* parse means `${VAR}` no longer works in non-string fields (e.g. `port: ${PORT}`) — the strict decoder now rejects `"${PORT}"` as a non-integer before expansion runs. Consistent with the updated docs ("resolves only in string-typed fields") and no shipped example relies on it, so this is intentional — just flag it for operators migrating configs.
- Two inline notes: hardening the reflection-walk invariant, and the duplicated `maxDurationMs` literal in the schema.

Overall: correct, fail-closed preserved, no audit-shape changes. LGTM modulo the inline nits.

---

cmd/eunox-mcp/gateway_config.go
eunolabs: Hardened the invariant in 5ba7cdc with the reflective-test option: `TestExpandEnvInStrings_ConfigTreeHasNoUnhandledKinds` walks the `GatewayConfig` type tree and fails the moment a `map` or `interface` field is added, so the silent-skip can't sneak in — whoever adds such a field has to extend `expandEnvInStrings` (maps need `SetMapIndex`, as you noted) first. The walk's doc comment now points at the test. Went with the test over a runtime `default:` panic to keep config loading panic-free while still being CI-enforced.

---

schemas/eunox-gateway-config.schema.json
eunolabs: Pinned in 5ba7cdc: `TestGatewaySchema_DurationMaxMatchesConstant` reads both schema `maximum`s (`listen.sessionIdleTimeoutMs` and the `$defs/routeDefaults` `upstreamTimeoutMs`) and asserts each equals `maxDurationMs`, so a change to the constant now fails the test instead of silently drifting the schema.

---

eunolabs (commented):
## Re-review (commit `5ba7cdc`)

The follow-up cleanly resolves both items from the previous review, and CI is fully green (43/43 checks; `mergeable_state: clean`). Verified the new work:

- **Reflection-walk invariant — resolved.** `TestExpandEnvInStrings_ConfigTreeHasNoUnhandledKinds` walks the `GatewayConfig` type tree and fails the moment a `map`/`interface` field is added, with a message pointing at the fix (`SetMapIndex` for maps). The walk is correct — it descends pointer/slice/array/struct and guards against recursive types via a `seen` set — so the "silent skip leaks a `${TOKEN}` to the upstream un-expanded" failure mode is now caught at test time before it can ship. The `expandEnvInStrings` doc comment also now spells out the limitation and references the test. Exactly what I was after.

- **Schema/loader drift — resolved.** `TestGatewaySchema_DurationMaxMatchesConstant` pins both schema `maximum` literals to `maxDurationMs`. Nice catch that `defaults` is a `$ref` to `#/$defs/routeDefaults` — the test navigates `$defs.routeDefaults.properties.upstreamTimeoutMs.maximum` rather than the (nonexistent) `properties.defaults...` path, which is the correct resolution. The bound is exactly representable as float64, so the `int64(got) == maxDurationMs` comparison is exact. Combined with the existing "exceeds the maximum" loader error-cases, the schema bound, the `validate()` bound, and the error message are now all chained to one constant.

- **CHANGELOG — added and accurate**, including the env-refs-string-only migration note (`port: ${PORT}` is now rejected at load) that I flagged as worth surfacing for operators.

No new findings. Both prior nits were non-blocking and are now addressed with regression tests; the fix itself was already sound. From my side this is good to merge — flagging only that I can't stamp a formal approval here since the review is on my own PR (GitHub restriction), so treat this comment as the sign-off.

---

PR #812, opened 2026-06-18
fix: bound JWKS refreshes, fix co-signature model, tighten token-cache expiry

### LINKED ISSUE #742: [rev-20260618T0010Z-563cce] verifyRetrySentinel cannot be inserted into negKIDs when cache is at capacity, disabling kid-less refresh rate-limiting

eunolabs:
## Summary

Addresses three findings from security review `rev-20260618T1232Z-b8fe95`. They are bundled because they come from one review pass over `pkg/capability/`; each is independent and individually reviewable below. (Happy to split into separate PRs if preferred.)

1. **JWKS shared refresh budget** (high) — `jwkscache.go`, `jwks.go`
2. **Co-signature detached model + self-referential rejection** (medium) — `proofs.go`, `token.go`, docs
3. **Token-cache exact-expiry fix** (low) — `token_cache.go`

---

### 1. JWKS shared refresh budget (high)

**Problem.** The per-kid negative cache only suppresses a repeat of the *same* unknown kid. A flood of *distinct* unknown kids forced one upstream JWKS round-trip each (N kids → N fetches), and a successful unchanged `200` is not a circuit-breaker failure, so the breaker never engaged.

**Fix.** `ForceRefreshForKID` now charges a shared refresh budget (`sharedRefreshSentinel`, the renamed empty-kid sentinel, now shared with `ForceRefreshForVerify`) whenever a forced fetch returns a key set identical to the cached one. While charged, distinct unknown kids are served from cache (fail closed) instead of fetching, bounding the flood to one fetch per `negativeKIDTTL`. A *changed* set does not charge the budget, so distinct unknown kids keep fetching while rotations are actively landing.

**Accepted tradeoff (per review).** While the budget is charged, forced fetches are suppressed, so a kid the issuer rotates *in* during that window is not observed until the budget expires — a rotated-in kid can be delayed up to `negativeKIDTTL` (30s). The sentinel is chargeable by a single unknown-kid token, so this window is attacker-triggerable, but it is bounded, self-healing, and far below `CacheTTL` (5m): the first probe fetch after the budget expires observes the rotation, resolves the new kid, and (because the set changed) does not re-charge. The code comments now state this explicitly instead of claiming "resolved promptly".

**Tests.**
- `TestJWKSCache_ForceRefreshForKID_SharedBudgetAcrossDistinctKIDs` — 50 distinct unknown kids trigger 1 fetch, not 50.
- `TestJWKSCache_ForceRefreshForKID_ChangedSetLeavesBudgetOpen` — rotation-before-charge: a changed set does not charge the budget, so a rotated-in kid resolves immediately.
- `TestJWKSCache_ForceRefreshForKID_ChargedBudgetDelaysRotatedKidUntilTTL` — charged-then-rotation: a kid rotated in while the budget is charged is suppressed for ≤`negativeKIDTTL`, then resolves on the first probe fetch (pins the accepted tradeoff).
- `TestJWKSCache_MarkKIDAbsent_SentinelExemptFromCap` updated for the renamed, dual-path sentinel.

Closes #742.

---

### 2. Co-signature detached model + self-referential rejection (medium)

**Problem.** The documented co-signature model had each co-issuer sign the exact compact token that *embeds* the co-signature — self-referential. The natural "sign → co-sign → embed → re-sign" flow can never verify, because embedding the proof and re-signing changes the signed bytes.

**Fix.** Documented the **detached** model in `proofs.go`, `token.go`, and the guide: co-issuers sign the proof-less *primary* token, and the proofs travel detached (passed to `VerifyCoSignatures` separately), never folded back into the signed payload. `VerifyCoSignatures` now **fails closed** (a new rejection path) when `tokenStr` embeds its own `proofs.signatures`, before any crypto. The guard sits after the empty-proofs early return, so a token embedding proofs but verified with empty detached proofs is left to `JWKSClient.VerifyToken` to reject, as before. The double payload parse the review flagged is folded into a single `parseTokenPayloadMeta`.

**Tests.**
- `TestVerifyCoSignatures_DetachedOverPrimaryToken` — end-to-end happy path over a proof-less primary token.
- `TestVerifyCoSignatures_EmbeddedProofsRejected` — signs the embedded bytes to prove the guard fires on structure, before crypto.

---

### 3. Token-cache exact-expiry fix (low)

**Problem.** `TokenCache.Get` used `now.After(expiresAt)`, serving a token at the exact expiry instant; a second-rounded clock made this last the whole expiry second.

**Fix.** `Get` now treats `now == expiresAt` as expired (`!now.Before`), consistent with RFC 7519 §4.1.4 (`exp` must be strictly after the current time) and with `Put`'s `tokenRemaining <= 0` admission check. The clock is still sampled under the read lock, so #535/#554 stay closed.

**Test.** `TestTokenCache_Get_ExactExpiryInstantIsMiss` pins the one-second-before hit and the exact-instant miss.

---

## Security considerations

- **Fail-closed preserved throughout.** Suppressed JWKS paths return the cached set lacking the key → key selection misses → reject. The co-signature guard fires before any crypto. The expiry change is the fail-closed direction (exact instant = expired).
- **No audit-shape change**, no new top-level fields, no widening of authority.
- **DoS mitigation:** bounds distinct-unknown-kid JWKS amplification from O(N) to O(1) per `negativeKIDTTL`.

## Checklist

- [x] `make test` — race detector clean
- [x] `make lint` — `go vet` + golangci-lint (0 issues)
- [x] `make check-fmt`, `make check-license`
- [x] Docs updated — `docs/capability-manifest-guide.md` (detached co-signature model)

https://claude.ai/code/session_017myfV4xAQRRPmih1SRi7qD


pkg/capability/jwkscache.go
```diff
@@ -197,38 +197,71 @@ func (c *JWKSCache) ForceRefresh(ctx context.Context) (*jose.JSONWebKeySet, erro
...
+//     N distinct unknown kids would otherwise cost N fetches, and a successful
+//     unchanged 200 never trips the circuit breaker.
+//
+// An empty kid is never suppressed (it selects the whole set, so it is not an
+// unknown-kid lookup). A found kid never marks either signal, so a known kid keeps
+// fetching exactly like ForceRefresh. When the cache holds no set at all, the fetch
+// is not suppressed so a cold start never denies purely for lack of a cached copy.
 func (c *JWKSCache) ForceRefreshForKID(ctx context.Context, kid string) (*jose.JSONWebKeySet, error) {
-	if kid != "" && c.kidRecentlyAbsent(kid) {
+	if kid != "" && (c.kidRecentlyAbsent(kid) || c.kidRecentlyAbsent(sharedRefreshSentinel)) {
```
eunolabs, line 233: **Gating `ForceRefreshForKID` on the *shared* sentinel delays resolution of legitimately rotated-in kids by up to `negativeKIDTTL` (30s).**

Before this change the guard was `kid != "" && c.kidRecentlyAbsent(kid)`, so a never-before-seen rotated-in kid always fetched and resolved on first presentation (the whole point of #410). Now, once the sentinel is charged, *any* non-empty kid — including a brand-new one — short-circuits to the cached set and is rejected (fail-closed). Concretely:

1. one unknown-kid token charges the sentinel via an unchanged fetch (line 244-245);
2. the IdP rotates and publishes `kidNew` (<30s later);
3. a valid token carrying `kidNew` arrives → this guard sees the sentinel charged → returns the cached set without `kidNew` → rejected until the sentinel expires.

The sentinel is chargeable by a *single* unknown-kid (or kid-less bad-signature) token and kept charged by a low-rate flood, so this window is attacker-triggerable.

The PR summary says *"A CHANGED key set clears the sentinel, so genuinely rotated-in kids are still resolved promptly,"* and the comment at L241-242 says the same. That only holds when a forced fetch actually runs to observe the change — but while the sentinel is charged, forced fetches are suppressed, so nothing observes the rotation until the sentinel expires (≤30s later, when one probe fetch is allowed through). `TestJWKSCache_ForceRefreshForKID_ChangedSetLeavesBudgetOpen` only exercises the ordering where the rotation lands *before* the sentinel is charged; it doesn't pin the charged → rotation → new-kid case.

This is likely an acceptable, bounded (≤30s), self-healing tradeoff — `negativeKIDTTL` (30s) ≪ `CacheTTL` (5m), so it's still far faster than TTL-only. But please:
- add a test for the charged-sentinel-then-rotation ordering so the accepted behavior is pinned (the property the summary claims is exactly the one not covered); and
- soften the "resolved promptly" wording (summary + L241-242) to acknowledge the up-to-`negativeKIDTTL` delay while the shared budget is engaged.

Minor, same doc block (L218): *"a known kid keeps fetching exactly like ForceRefresh"* isn't strictly true when the sentinel is charged — a call for a kid that *is* in the cache also short-circuits to cached. Harmless in practice (`VerifyWithKeyRotation` only calls this for absent kids), but the comment slightly overstates.

---

pkg/capability/proofs.go
```diff
@@ -77,6 +88,15 @@ func VerifyCoSignatures(ctx context.Context, tokenStr string, proofs *IssuancePr
 		return errors.New("capability: co-signatures present but no key resolver provided; failing closed")
 	}
 
+	// Co-signatures are detached: tokenStr must be the PRIMARY token, not a token
+	// that has the co-signatures embedded in its payload. An embedded-proofs token is
+	// self-referential — re-signing the token after embedding a co-signature changes
+	// the very bytes that co-signature signed, so it could never verify. Fail closed
+	// rather than run a verification that can only ever fail in a confusing way.
+	if tokenEmbedsCoSignatures(tokenStr) {
```
eunolabs: Good guard — failing closed on the self-referential embed-then-re-sign shape *before* any crypto is the right call, and `TestVerifyCoSignatures_EmbeddedProofsRejected` signing the embedded bytes to prove the guard fires on structure (not signature) is a nice touch. Placement is correct too: it sits after the empty-proofs early return (L73-75), so a token that embeds proofs but is verified with empty detached proofs is left to `JWKSClient.VerifyToken` to reject, as documented.

Two notes:
- **Scope/description.** This is a behavior change to the `VerifyCoSignatures` contract — a new rejection path — but the PR summary is entirely about the JWKS refresh budget and doesn't mention it (its two new tests aren't in the "How it was tested" list either). The commit message does cover it; the PR body should too. See the top-level review.
- **Optional.** `tokenEmbedsCoSignatures` base64-decodes + JSON-unmarshals the payload, which `parseTokenIssuedAt` (L122) then does again a few lines later — two passes over the same payload. Negligible here (library-only, not on a wired request path), but worth folding into one parse if this ever gets wired in.

---

pkg/capability/token_cache.go
```diff
@@ -167,13 +167,21 @@ func (c *TokenCache) Get(tokenStr string) (*TokenPayload, bool) {
...
-	// expired during that wait would still pass now.After and be served past its
-	// TTL (issue #554). Now() does not touch c.mu, so calling it under the read lock
+	// expired during that wait would still pass the expiry check and be served past
+	// its TTL (issue #554). Now() does not touch c.mu, so calling it under the read lock
 	// is deadlock-free, and the lock stays held across the lookup+check so the #535
 	// revocation race remains closed (sampling after RUnlock would reopen it).
 	now := c.cfg.Now()
-	if now.After(entry.expiresAt) {
-		// Expired — treat as miss; background cleanup will remove it.
+	if !now.Before(entry.expiresAt) {
```
eunolabs, line 175: Correct: `!now.Before(expiresAt)` is `now >= expiresAt`, so the exact expiry instant is now a miss — consistent with RFC 7519 §4.1.4 (`exp` must be strictly after the current time) and with `Put()`'s `tokenRemaining <= 0` not-cached handling. The clock is still sampled under the read lock, so the #535/#554 races stay closed. Verified the inversion is faithful and the targeted test (`TestTokenCache_Get_ExactExpiryInstantIsMiss`) pins both the one-second-before hit and the exact-instant miss. 

(Filed under "unrelated token-cache precision fix" in the summary — see the top-level note on bundling.)

---

eunolabs (commented):
## Review

Thorough, well-commented work — the engineering quality here is high. The code is correct and preserves fail-closed everywhere I checked. My notes are about one overstated guarantee + a test gap, and about PR scope; none are "this is broken."

### What this PR does
Three logically distinct changes:
1. **JWKS shared refresh budget** (`jwks.go`, `jwkscache.go`, `+ tests`) — the titled change. A `sharedRefreshSentinel` (the renamed empty-kid sentinel) is now charged by *both* `ForceRefreshForKID` and `ForceRefreshForVerify` when a forced fetch returns an unchanged key set, and consulted by both, so a flood of *distinct* unknown kids can no longer amplify into one upstream JWKS fetch each.
2. **Co-signature detached-model rejection** (`proofs.go`, `token.go`, docs, `+ tests`) — `VerifyCoSignatures` now fails closed when `tokenStr` embeds its own `proofs.signatures` (the self-referential embed-then-re-sign shape).
3. **Token-cache exact-expiry fix** (`token_cache.go`, `+ test`) — `now.After` → `!now.Before`, so the exact expiry instant is a miss.

### Main concern — JWKS budget (inline at `jwkscache.go:222`)
Coupling the unknown-kid path to the shared sentinel means that, once the sentinel is charged (doable with a single unknown-kid token, sustainable with a low-rate flood), a genuinely **rotated-in kid is suppressed and rejected for up to `negativeKIDTTL` (30s)** — previously it resolved on first presentation (#410). The summary's claim that *"a CHANGED key set clears the sentinel, so genuinely rotated-in kids are still resolved promptly"* only holds when a forced fetch runs to observe the change; while the sentinel suppresses fetches, nothing observes the rotation until the sentinel expires. `TestJWKSCache_ForceRefreshForKID_ChangedSetLeavesBudgetOpen` covers only the rotation-before-charge ordering, not the charged → rotation → new-kid case. The tradeoff is probably fine (bounded ≤30s, self-healing, still ≪ the 5m TTL), but please pin it with a test and soften the wording. Details inline.

### Scope / description
The PR title (`fix:`) and summary describe **only** the JWKS budget. Changes (2) and (3) are bundled in: (3) is flagged as "unrelated," and (2) is a real change to the `VerifyCoSignatures` contract (a new rejection path) that the summary doesn't mention at all — its two new tests aren't in "How it was tested" either. The commit message does cover all three. Per the repo's one-logical-change/document-behavior-changes conventions, consider splitting (2) and (3) into their own PRs, or at minimum expand the description so the co-signature contract change gets reviewed as its own unit. A reviewer scanning the title would not expect the co-signature path to move here.

### Looks good
- **Fail-closed preserved throughout.** Suppressed paths return the cached set lacking the key → key selection misses → reject. The self-referential guard fires before any crypto. The expiry change is the fail-closed direction (exact instant = expired).
- `refresh(ctx, true)` substituted for `ForceRefresh(ctx)` is behavior-identical apart from capturing `changed`; the `changed` flag is computed atomically inside the singleflight closure (#633), so the shared-budget decision can't race a concurrent rotation. Sentinel auto-expires (anchored to first observation, #680) so it can't wedge.
- Tests are strong: `SharedBudgetAcrossDistinctKIDs` exercises the 50-distinct-kid core; `EmbeddedProofsRejected` signs the embedded bytes to prove the guard precedes crypto; the token-cache test pins both the before-hit and exact-instant-miss.
- Docs and the `IssuerSignature`/`IssuanceProofs` godoc updates are clear and match the code.

### Recommendation
Not blocking on correctness. Before merge I'd like to see: (a) a test for the charged-sentinel-then-rotation ordering, (b) the "resolved promptly" wording softened to note the ≤`negativeKIDTTL` delay, and (c) the PR description expanded (or the PR split) to cover the co-signature and token-cache changes.

---

pkg/capability/jwkscache.go
eunolabs: Addressed in 210d6fc:

- Added `TestJWKSCache_ForceRefreshForKID_ChargedBudgetDelaysRotatedKidUntilTTL`, which pins exactly the charged → rotation → new-kid ordering: the rotated-in kid is suppressed (fails closed, no fetch) within the window, then resolves on the first probe fetch after the sentinel expires.
- Softened the doc + inline wording: dropped "resolved promptly", and now document the up-to-`negativeKIDTTL` delay while the budget is charged (attacker-triggerable, bounded, self-healing, ≪ `CacheTTL`).
- Fixed the L218 overstatement — a charged budget short-circuits even a cached kid (harmless, since that set already contains it).
- Folded the double payload parse into a single `parseTokenPayloadMeta`.

Also expanded the PR description so the co-signature contract change and the token-cache fix are documented as their own units. Left all three bundled (one review pass over `pkg/capability/`), but happy to split if you'd prefer.

---

eunolabs (commented):
## Re-review (head `210d6fc`)

The follow-up commit addresses every point from my first pass. Verified the new state rather than taking the description at face value:

**1. Charged-budget → rotation ordering — now pinned.** `TestJWKSCache_ForceRefreshForKID_ChargedBudgetDelaysRotatedKidUntilTTL` exercises exactly the gap I flagged: charge the sentinel via an unchanged fetch → rotate `knew` in *after* (clock not advanced) → the lookup for `knew` is suppressed and fails closed with no extra fetch (hits stays 2) → after `negativeKIDTTL` one probe fetch runs (hits 3) and resolves `knew`. That's the bounded, self-healing behavior, asserted on both the fetch count and the key presence. Good test.

**2. Wording — corrected.** The `ForceRefreshForKID` doc/inline comments drop the "resolved promptly" claim and now state the up-to-`negativeKIDTTL` delay, that it's attacker-triggerable, bounded, and self-healing, and why a CHANGED set doesn't re-charge. The "known kid keeps fetching exactly like ForceRefresh" line is fixed to note the charged-budget short-circuit (harmless, since that set already contains the kid). No logic changed here — comment-only.

**3. Double parse — folded.** `parseTokenIssuedAt` + `tokenEmbedsCoSignatures` are now one `parseTokenPayloadMeta` returning `(iat, embedsCoSigs)` from a single payload decode. Equivalent: still best-effort `(0, false)` on non-JWT input, guard still sits after the empty-proofs early return and before any crypto, and the iat bound is unchanged. Confirmed no dangling references to the removed functions.

**4. Scope/description — resolved.** Title and body now cover all three changes with severities, each independently reviewable, and the accepted tradeoff is written up explicitly. The bundle-vs-split call is the maintainer's; the author offered to split.

Locally: `go test -race` on the affected `pkg/capability` tests passes. CI is green (35/35) and the PR is mergeable.

### One trivial, non-blocking nit
`pkg/capability/proofs_test.go:697` — the `makeJWTWithIAT` doc comment still says *"so parseTokenIssuedAt has a real iat to read"*; that function no longer exists (it's `parseTokenPayloadMeta` now). It's outside this PR's diff hunks so I couldn't anchor an inline comment, and `makeJWTWithIAT` wasn't otherwise touched here — fold it in whenever convenient.

LGTM from my side — all review feedback addressed and CI green. (Submitting as a comment rather than a formal approval since this PR is under the same account I'm reviewing from; aside from the one cosmetic nit above I see nothing blocking a merge.)

---

PR #813, opened 2026-06-18
Fix review-session bugs (07e43b, 37e71f, f73824, 9d6c2a, 174fa2)

### LINKED ISSUE #786: [rev-20260618T0852Z-174fa2] Redis killswitch Stop can return before concurrent Start finishes
**Location:** `pkg/killswitch/redis.go:152-198` and `pkg/killswitch/redis.go:263-278`

**Reference:** https://github.com/eunolabs/eunox/blob/33641cfe0f668d61e4f188eda43a1f0f14ded5f3/pkg/killswitch/redis.go#L152-L198 and https://github.com/eunolabs/eunox/blob/33641cfe0f668d61e4f188eda43a1f0f14ded5f3/pkg/killswitch/redis.go#L263-L278

**Severity:** medium

**Description:** `Start` publishes `r.cancel` before its synchronous initial `refreshState` finishes and before either `wg.Add` call. A concurrent `Stop` can therefore observe the cancel function, cancel the context, call `wg.Wait` while the counter is still zero, and return. `Start` remains inside Redis I/O and later registers and launches the two background goroutines after `Stop` has already promised that Redis/client/logger teardown is safe. This also violates the `sync.WaitGroup` rule that positive `Add` calls from a zero counter must happen before a concurrent `Wait`. The new WaitGroup closes the post-cancel goroutine race from #750 only when `Start` has already reached both `Add` calls; it does not make the documented concurrent `Start`/`Stop` path safe.

**Trigger / repro:** Use a Redis client whose initial `Get` blocks. Launch `Start` in one goroutine and wait until `refreshState` enters `Get`; call `Stop` in another goroutine. `Stop` returns immediately because the WaitGroup is still zero, even though `Start` is still using the Redis client. Closing the client after `Stop` races that in-flight refresh. Releasing `Get` then lets `Start` execute both `wg.Add` calls and launch goroutines after `Stop` returned. A focused regression probe at this commit failed deterministically with `Stop returned while concurrent Start was still using the Redis client`.

**Suggested fix:** Track the entire synchronous `Start` body as lifecycle work before exposing `r.cancel` to `Stop` (for example, increment the WaitGroup before publishing `cancel`, defer its `Done` until all background goroutines are registered, and keep the counter non-zero while those later `Add` calls occur). Alternatively serialize `Start`/`Stop` with an explicit lifecycle mutex/state and a `startDone` channel. Add a blocking-client regression test that requires `Stop` to wait for the initial refresh and all goroutine registration to finish.

**Session:** rev-20260618T0852Z-174fa2

### LINKED ISSUE #787: [rev-20260618T0852Z-174fa2] E2E HTTP client accepts malformed initialize responses
**Location:** `demo/e2e/mock-host/main.go:637-671`

**Reference:** https://github.com/eunolabs/eunox/blob/33641cfe0f668d61e4f188eda43a1f0f14ded5f3/demo/e2e/mock-host/main.go#L637-L671

**Severity:** medium

**Description:** The black-box HTTP client discards both `io.ReadAll` and `json.Unmarshal` errors. `initialize` then treats any HTTP 200 response carrying `Mcp-Session-Id` and no decoded `error` field as successful; it does not require a valid JSON-RPC envelope or an initialize result. A gateway regression that returns malformed or truncated JSON can therefore pass initialization. The same unchecked decode path affects later calls: malformed JSON that partially populates `Result` or `Error` can be classified by the assertion helpers as a valid policy outcome. This undermines the e2e suite's stated transport-framing coverage.

**Trigger / repro:** Point `httpConn` at a test server that responds to initialize with HTTP 200, header `Mcp-Session-Id: fake-session`, and body `not-json`. `httpConn.initialize()` returns nil. A focused regression probe at this commit failed with `initialize accepted a malformed JSON response`. The subsequent `notifications/initialized` result is also discarded unconditionally, so the probe needs no valid JSON-RPC response at any point.

**Suggested fix:** Propagate `io.ReadAll` errors and require every non-empty response body to decode successfully as a JSON-RPC envelope. For request responses, require exactly one of `result` or `error`, preserve and validate the response ID, and make `initialize` require a non-empty valid result before accepting the session. Check the status/error from `notifications/initialized` instead of discarding all three return values. Add an `httptest` regression covering malformed and truncated initialize responses.

**Session:** rev-20260618T0852Z-174fa2

### LINKED ISSUE #788: [rev-20260618T0852Z-9d6c2a] JSON-RPC id:null requests are misclassified as notifications
**Location:** `cmd/eunox-mcp/jsonrpc.go:18-45`
**Reference:** https://github.com/eunolabs/eunox/blob/33641cfe0f668d61e4f188eda43a1f0f14ded5f3/cmd/eunox-mcp/jsonrpc.go#L18-L45
**Severity:** medium
**Description:** `rpcMsg.ID` is a `*json.RawMessage`. Go unmarshalling sets that pointer to nil both when `id` is absent and when the JSON value is explicitly `null`. The classification helpers therefore treat a valid JSON-RPC request carrying `"id": null` as a notification. HTTP requests are acknowledged with 202 and no JSON-RPC response; stdio requests are forwarded as notifications. An initialize request can also allocate a session while returning a response with the id omitted.
**Trigger / repro:** Unmarshal `{"jsonrpc":"2.0","id":null,"method":"tools/list"}` into `rpcMsg`. `isNotification()` returns true and `isRequest()` returns false, even though JSON-RPC permits a null request identifier.
**Suggested fix:** Track field presence independently from its value, for example with a custom optional raw-ID type or custom `MarshalJSON` and `UnmarshalJSON`. Classify present-null as a request/response ID and preserve `"id": null` in the response.
**Session:** rev-20260618T0852Z-9d6c2a

### LINKED ISSUE #789: [rev-20260618T0852Z-9d6c2a] HTTP resource query delimiters act as authorization wildcards
**Location:** `cmd/eunox-mcp/pdp_jwt.go:980-995`
**Reference:** https://github.com/eunolabs/eunox/blob/33641cfe0f668d61e4f188eda43a1f0f14ded5f3/cmd/eunox-mcp/pdp_jwt.go#L980-L995
**Severity:** high
**Description:** HTTP resource claims deliberately keep the URI query string attached to the resource name, but the resulting bare name is passed to `matchBare`, which uses `path.Match`. The literal URI query delimiter `?` is therefore interpreted as a one-character glob. An exact-looking capability for one URL authorizes a different resource path.
**Trigger / repro:** A signed claim `resource:https://api.example.com/search?q=widget` also matches and allows `resources/read` for `https://api.example.com/searchXq=widget`. A focused probe showed `buildConstraintsFromClaims` returns a granting constraint for that different target.
**Suggested fix:** Match the structural `?` and query component of HTTP resource URIs literally. Parse HTTP resource claims separately and either exact-match the query string while applying glob semantics only to the intended path portion, or escape URL metacharacters before calling the glob matcher.
**Session:** rev-20260618T0852Z-9d6c2a

### LINKED ISSUE #790: [rev-20260618T0852Z-9d6c2a] mcp.capabilities:null bypasses the JWT exhaustive allowlist
**Location:** `cmd/eunox-mcp/pdp_jwt.go:147-155`
**Reference:** https://github.com/eunolabs/eunox/blob/33641cfe0f668d61e4f188eda43a1f0f14ded5f3/cmd/eunox-mcp/pdp_jwt.go#L147-L155
**Severity:** high
**Description:** The pointer representation used to distinguish an absent `mcp.capabilities` field from an empty array also collapses an explicitly present JSON null to nil. `ValidateToken` accepts a correctly signed token containing `"capabilities": null`, sets `HasCapabilities=false`, and treats it as identity-only. On a route with a manifest backstop, the JWT no longer narrows access and the broader manifest alone governs.
**Trigger / repro:** Sign a valid token with `{"mcp":{"v":"0.2","capabilities":null}}`. `ValidateToken` succeeds and stores `HasCapabilities:false`; a permissive route manifest can then allow targets that a malformed capability claim should not authorize.
**Suggested fix:** Use presence-aware decoding that distinguishes absent, null, and array values. Reject null and all non-array values during token validation with HTTP 401; alternatively treat present-null as an empty exhaustive allowlist, but rejection gives the clearest IdP configuration error.
**Session:** rev-20260618T0852Z-9d6c2a

### LINKED ISSUE #791: [rev-20260618T0852Z-9d6c2a] Slower background JWKS refresh can overwrite a forced rotation
**Location:** `pkg/capability/jwkscache.go:369-448`
**Reference:** https://github.com/eunolabs/eunox/blob/33641cfe0f668d61e4f188eda43a1f0f14ded5f3/pkg/capability/jwkscache.go#L369-L448
**Severity:** high
**Description:** Forced and non-forced refreshes use different singleflight keys, so both HTTP fetches can run concurrently. Each fetch writes the shared cache when it completes, with no generation or ordering guard. A non-forced refresh that started first but finishes last can overwrite keys installed by a later forced rotation with its older response. This can reject newly rotated tokens and can reintroduce a key that the newer JWKS response removed for the full cache TTL.
**Trigger / repro:** Seed key set A, expire the TTL, start a background `Refresh` whose response is delayed, then run `ForceRefresh` and return set B. After B is installed, release the background response containing A. The final `GetKeys` returns A; the focused concurrency probe reproduced this deterministically.
**Suggested fix:** Prevent older-started fetches from overwriting newer results. Serialize cache installation across both refresh classes or assign fetch generations and commit only when the generation is not older than the currently installed generation. Preserve the forced-refresh guarantee without allowing cross-key completion order to decide cache freshness.
**Session:** rev-20260618T0852Z-9d6c2a

### LINKED ISSUE #795: [rev-20260618T1230Z-f73824] TokenCache.Get() returns a shared *TokenPayload pointer — callers can corrupt the cache through the returned reference
**Location:** `pkg/capability/token_cache.go:150-179`
**Reference:** https://github.com/eunolabs/eunox/blob/af8aab0983e10c5f8f0724ed16d739c986068b95/pkg/capability/token_cache.go#L150-L179
**Severity:** medium

**Description:** `TokenCache.Get()` returns `entry.payload` directly as `*TokenPayload`, giving the caller a live pointer into the struct stored inside the cache map. The doc-comment says the returned value "must be treated as read-only," but this constraint is not enforced at the type level or at runtime. Any caller that mutates a field through the returned pointer — even inadvertently, via a future refactor or a code path that updates the struct after a cache hit — silently corrupts the cached entry for every subsequent `Get()` on the same token hash, because the cache and the caller share the same memory. The bug is latent today and can be invisible in testing (the mutation is not detectable without a second `Get()` under a read lock).

**Trigger / repro:** Call `Get()` for a valid token, receive the `*TokenPayload`, then write through the pointer (e.g., `payload.ExpiresAt = 0`). A subsequent `Get()` for the same raw token string returns the mutated payload with the zeroed expiry. The `now.After(entry.expiresAt)` expiry check in `Get()` would then pass unconditionally (time is always after epoch zero), serving the corrupted payload indefinitely until the entry is evicted.

**Suggested fix:** Return a shallow copy of the payload struct so the caller owns the returned value:
```go
copy := *entry.payload
return &copy, true
```
`TokenPayload` contains only value types and a slice (which a shallow copy leaves shared but whose backing array is not mutated by the cache), so a value copy is sufficient. Alternatively enforce immutability by making `TokenPayload` fields unexported and accessible only through getter methods, but that is a larger API change.

**Session:** rev-20260618T1230Z-f73824

### LINKED ISSUE #796: [rev-20260618T1230Z-f73824] ValidateToken re-samples p.now() inside the per-key closure — non-deterministic accept/reject at token expiry boundary across JWKS retries
**Location:** `cmd/eunox-mcp/pdp_jwt.go:380-437`
**Reference:** https://github.com/eunolabs/eunox/blob/af8aab0983e10c5f8f0724ed16d739c986068b95/cmd/eunox-mcp/pdp_jwt.go#L380-L437
**Severity:** low

**Description:** The key-verification closure passed to `capability.VerifyWithKeyRotation` executes `now := p.now()` on line 437 inside the closure body, so it samples a fresh timestamp on every invocation. `VerifyWithKeyRotation` can call the closure multiple times: once per matching key candidate (unusual but possible when a JWKS endpoint publishes multiple keys for the same `kid`), and a second round after a forced JWKS refresh triggered by a signature failure on the first round. A forced refresh involves a network round-trip that can take from tens to hundreds of milliseconds. If the clock crosses the token's `exp + leeway` boundary between the first call (token accepted) and the forced-refresh retry (token rejected — or vice versa), the acceptance decision is non-deterministic and depends on network latency. A token near expiry presented during a key-rotation event can be accepted on the original cached-key attempt then rejected after the forced refresh in the same `ValidateToken` call, surfacing an erratic 401 to the caller.

**Trigger / repro:** Configure `leeway=0`. Construct a JWT that expires in the time it takes `ForceRefreshForKID` to complete (inject a delayed JWKS server). Submit the token. The first closure call (against the cached, unmatched key set) returns a signature error; `VerifyWithKeyRotation` forces a refresh. By the time the retry closure runs, `p.now()` is past `exp`, so the retry returns `Terminal(ErrExpired)`. The request is denied despite the token being valid at the time it was first received.

**Suggested fix:** Sample the clock exactly once before the closure is created and capture the fixed instant in the closure:
```go
now := p.now()
validated, err := capability.VerifyWithKeyRotation[JWTClaims](ctx, p.cache, kid, func(key *jose.JSONWebKey) (*JWTClaims, error) {
    // use captured now, not p.now()
    expected := jwt.Expected{Time: now}
    ...
})
```
This matches the pattern used by the `Engine`, `TokenCache`, and `alwaysAllowPDP.wiretapAllow`, all of which sample the clock once per operation.

**Session:** rev-20260618T1230Z-f73824

### LINKED ISSUE #797: [rev-20260618T1230Z-f73824] Breaker.Allow() in HalfOpen state is not panic-safe — an unhandled panic permanently wedges the breaker when HalfOpenMaxProbes=1
**Location:** `pkg/circuitbreaker/breaker.go:190-193`, `pkg/circuitbreaker/do.go:53-58`
**Reference:** https://github.com/eunolabs/eunox/blob/af8aab0983e10c5f8f0724ed16d739c986068b95/pkg/circuitbreaker/breaker.go#L190-L193

**Severity:** medium

**Description:** `Breaker.Allow()` increments `halfOpenProbes` inside `allow()` when the breaker is in HalfOpen state, consuming one of the limited probe slots. The documented usage pattern expects the caller to call `RecordSuccess()` or `RecordFailure()` after running the protected function; these are the only mechanisms that can advance the breaker out of HalfOpen (to Closed or back to Open). `Allow()` itself provides no safety net. `Do()` — the recommended call path — wraps the protected function in a deferred closure that unconditionally calls `probe.Failure()` on panic (do.go:53–58), so panics never leave the probe slot unreported. Callers using `Allow()`+`RecordFailure()` directly do not get this guarantee. If the protected function panics and the caller's goroutine has no recover-and-RecordFailure logic, the probe slot is consumed but no outcome is reported. With `HalfOpenMaxProbes=1` (the default), the single probe slot is permanently occupied: every subsequent `Allow()` or `AllowProbe()` call returns `false`, and the breaker is stuck in HalfOpen indefinitely with no way to recover short of an explicit `Reset()`.

**Trigger / repro:**
```go
    FailureThreshold:  1,
    CooldownDuration:  10*time.Millisecond,
    HalfOpenMaxProbes: 1,
})
b.Allow(); b.RecordFailure()        // trip to Open
time.Sleep(15 * time.Millisecond)   // wait for cooldown

admitted := b.Allow()               // admitted=true; halfOpenProbes=1
if admitted {
    panic("simulated panic")        // RecordFailure never reached
}

// breaker is now permanently stuck in HalfOpen:
admitted2, _ := b.AllowProbe()      // admitted2=false (halfOpenProbes==1 >= MaxProbes==1)
```

**Suggested fix:** Document explicitly in `Allow()`'s doc-comment that callers MUST use `defer b.RecordFailure()` as the very first statement after the admission check (not only on the success/failure branches). A stronger fix is to provide a panic-safe helper:
```go
func (b *Breaker) AllowFunc(fn func() error) error {
    admitted, probe := b.AllowProbe()
    if !admitted {
        return ErrOpen
    }
    completed := false
    defer func() {
        if !completed {
            probe.Failure()
        }
    }()
    err := fn()
    completed = true
    if err != nil {
        probe.Failure()
        return err
    }
    probe.Success()
    return nil
}
```
Deprecating `Allow()` in favour of `Do()` / `AllowProbe()` / `AllowFunc()` would make the panic-safe path the path of least resistance for all callers.

**Session:** rev-20260618T1230Z-f73824

### LINKED ISSUE #798: [rev-20260618T1230Z-f73824] handleAllowedExtensions uses filepath.Ext for "no extension" gate but strings.HasSuffix for allowlist — allows broader compound-extension matches than the early-return implies
**Location:** `pkg/enforcement/handlers.go:792-832`
**Reference:** https://github.com/eunolabs/eunox/blob/af8aab0983e10c5f8f0724ed16d739c986068b95/pkg/enforcement/handlers.go#L792-L832
**Severity:** low

**Description:** `handleAllowedExtensions` uses two different extension mechanisms on the same file path, and their semantics diverge for compound extensions in a way that can mislead operators:

1. **"No extension" gate** (line 792): `ext := strings.ToLower(filepath.Ext(filePath))`. For `backup.tar.gz`, `filepath.Ext` returns `.gz` — non-empty, so the file passes the gate.
2. **Allowlist matching** (line 812): `strings.HasSuffix(base, suffix)` where `base = filepath.Base(filePath)`. For the same `backup.tar.gz`, `strings.HasSuffix("backup.tar.gz", ".gz")` is `true`, so if `.gz` is in the allowlist, the file is admitted.
3. **Audit denial detail** (line 828): `fileSuffix(base)` = `.tar.gz` is written to `details.extension`.

The net effect: the "no extension" early-return uses `filepath.Ext` (single-segment, `.gz`), while the allowlist check uses `HasSuffix` (any suffix including compound, `.tar.gz`). An operator who configures `extensions: [".gz"]` may intend to allow only single-extension gzip files, not compound `.tar.gz` archives. The code's own comment at line 762–769 documents this: *"a SIMPLE entry like '.gz' also matches a COMPOUND file like 'archive.tar.gz'"* and states there is no way to allow `.gz` while denying `.tar.gz` using this condition alone. Despite the inline warning, this design makes it impossible to express a precise single-extension allowlist without a separate `allowedValues` constraint, and the audit denial message compounds the confusion: when a `.tar.gz` file is denied, `details.extension` reports `.tar.gz` (from `fileSuffix`), while the code that decided the outcome used a different segment (`.gz` from `filepath.Ext`) for the "no extension" determination — two different representations of the same file's extension appear across the two code paths.

**Trigger / repro:** Manifest allowlist: `extensions: [".gz"]`. Submit a `tools/call` for `read_file` with `path=backup.tar.gz`. The call is permitted (HasSuffix matches). Now submit `path=backup.tar.bz2`. Denied. Audit log shows `details.extension: ".tar.bz2"`. An operator reviewing the log sees `.tar.bz2` denied and might assume `.tar.gz` would also be denied — but it is not, because `.gz` is a suffix of `.tar.gz`.

**Suggested fix:** Either (a) unify the extension computation: replace `filepath.Ext` with `fileSuffix(filepath.Base(filePath))` for the "no extension" gate so both branches use the same representation; or (b) add a separate `allowedExtensions` validation warning at manifest-load time when a simple-extension entry (one without a compound prefix) is present alongside files that are expected to have compound extensions, prompting operators to also pin via `allowedValues`. At minimum, update the manifest guide to make the compound-match asymmetry prominent, with a worked example of `.gz` matching `.tar.gz`.

**Session:** rev-20260618T1230Z-f73824

### LINKED ISSUE #800: [rev-20260618T1310Z-37e71f] readLastAuditLine returns ("", nil) on file shrink, silently breaking the tamper-evident chain
**Location:** `cmd/eunox-mcp/audit.go:729-750`

**Reference:** https://github.com/eunolabs/eunox/blob/509758192e647c0fbd53e2490759741815bc8293/cmd/eunox-mcp/audit.go#L729-L750

**Severity:** Medium

**Description:**

`readLastAuditLine` calls `f.Stat()` to get the file size, then issues `f.ReadAt(buf, offset)` to read the tail. If the file is truncated or replaced (log rotation) between `Stat` and `ReadAt`, `ReadAt` returns `(0, io.EOF)`.

The function interprets `(0, io.EOF)` as "empty file" and returns `("", nil)`. The caller in `openAuditSink` treats `""` the same as a genuinely empty log — it starts a fresh chain without emitting a `chain_resume_failed` audit marker. The result is an unmarked gap in the HMAC-SHA256 chain: an attacker who can rotate the log file at the right moment causes the proxy to silently start a new chain, with no forensic evidence of the interruption.

**Trigger / repro:**

1. Start the proxy writing to an audit log with many records.
2. Atomically replace (or truncate to 0) the log file while the proxy is mid-`openAuditSink` (e.g. during restart after a crash).
3. Observe: `openAuditSink` resumes normally; no `chain_resume_failed` record is written; the new chain's `seq` starts at 1 with `prev_hmac` empty.

The race window is small but real on any system with log-rotation daemons running concurrently with proxy restarts.

**Suggested fix:**

Distinguish "true empty file" from "file shrank under us":

```go
n, err := f.ReadAt(buf, offset)
if n == 0 && errors.Is(err, io.EOF) {
    if size > 0 {
        // File shrank between Stat and ReadAt — treat as a read error,
        // not an empty file, so the caller emits chain_resume_failed.
        return "", fmt.Errorf("audit tail read: file shrank (stat=%d): %w", size, ErrAuditFileShrunk)
    }
    return "", nil // genuinely empty
}
```

Define `ErrAuditFileShrunk` as a sentinel the caller can distinguish from transient I/O errors if finer-grained handling is desired.

**Session:** rev-20260618T1310Z-37e71f

### LINKED ISSUE #801: [rev-20260618T1310Z-37e71f] Bare Allow()+RecordSuccess/Failure bypasses half-open generation guard, can prematurely close breaker
**Location:** `pkg/circuitbreaker/breaker.go:281-286, 362-366`

**Reference:** https://github.com/eunolabs/eunox/blob/509758192e647c0fbd53e2490759741815bc8293/pkg/circuitbreaker/breaker.go#L281-L286

**Severity:** Medium

**Description:**

`Breaker` exposes two admission paths:
- `AllowProbe()` — returns a generation-bound `Probe`; `Probe.Success()` / `Probe.Failure()` call `recordSuccess(gen, guard=true)` / `recordFailure(gen, guard=true)`, which verify that the stored `halfOpenGen` still matches before crediting the outcome.
- `Allow()` — returns a bare bool, and callers are expected to follow up with `RecordSuccess()` / `RecordFailure()`, which call `recordSuccess(gen=0, guard=false)` / `recordFailure(gen=0, guard=false)`.

The `guard=false` path skips the generation check entirely. A slow probe admitted in the **closed** state that calls `RecordSuccess()` after the breaker has transitioned `Closed → Open → HalfOpen` will unconditionally increment `halfOpenSuccess` and decrement the outstanding probe count in the new half-open window — even though that outcome belongs to a completely different request from a prior generation. With `HalfOpenMaxProbes = 1`, a single such stale `RecordSuccess()` closes the breaker immediately without any real success in the new window.

`Do[T]()` avoids this by using `AllowProbe()` exclusively, but any caller using the lower-level `Allow()` + `RecordSuccess()`/`RecordFailure()` API is exposed.

**Trigger / repro:**

```go
b := circuitbreaker.New(circuitbreaker.Options{
    Threshold:         1,
    HalfOpenMaxProbes: 1,
    HalfOpenTimeout:   100 * time.Millisecond,
    ResetTimeout:      50 * time.Millisecond,
})

// 1. Trip the breaker (closed → open).
b.Allow(); b.RecordFailure()

// 2. Wait for half-open window to open.
time.Sleep(60 * time.Millisecond)
admitted := b.Allow() // admitted=true; probe is now "in flight"

// 3. Before the probe reports, let the half-open window expire and
//    re-open the breaker (simulate by another failure from a concurrent path).
//    halfOpenGen is now bumped.

// 4. The slow probe from step 2 finishes successfully.
b.RecordSuccess() // guard=false → credits new half-open window → breaker closes prematurely
```

**Suggested fix:**

Capture the generation at admission time in `Allow()` and carry it in the returned token, or deprecate `Allow()`/`RecordSuccess()`/`RecordFailure()` in favor of the generation-safe `AllowProbe()` API. The simplest approach: change `Allow()` to return a `Probe` (a breaking API change, but pre-1.0) and remove the unguarded `RecordSuccess`/`RecordFailure` methods entirely. This would make the guard-bypass path impossible by construction.

**Session:** rev-20260618T1310Z-37e71f

### LINKED ISSUE #802: [rev-20260618T1310Z-37e71f] TokenCache Stop() before Start() violates "no-op" contract; subsequent Start() launches a goroutine that exits immediately
**Location:** `pkg/capability/token_cache.go:131-145`

**Reference:** https://github.com/eunolabs/eunox/blob/509758192e647c0fbd53e2490759741815bc8293/pkg/capability/token_cache.go#L131-L145

**Severity:** Low

**Description:**

`Stop()` closes `c.stopCh` (a `chan struct{}`) and returns. The code comment at `Start()` documents: *"Calling Start after Stop is a no-op."*

This is incorrect. `stopCh` is closed in `Stop()`, but `startOnce` (`sync.Once`) has not yet been triggered. A subsequent call to `Start(ctx)` will fire `startOnce.Do(...)`, launch the cleanup goroutine, and the goroutine will read from the already-closed `stopCh` via the `select` and exit immediately. The `wg.Add(1)` has already been called before the goroutine exits, so `wg.Wait()` in any subsequent `Stop()` call will complete normally — but the cleanup loop never ran.

The practical consequence is that a caller who:
1. Creates a `TokenCache`,
2. Calls `Stop()` (perhaps in an error-path teardown before `Start()` was called),
3. Later calls `Start(ctx)` (perhaps after a retry),

…ends up with a `TokenCache` that silently never performs background cleanup. Expired tokens accumulate and are only purged on the next `Put()` call's two-phase sweep.

**Trigger / repro:**

```go
tc := capability.NewTokenCache(10, time.Minute)
tc.Stop()                         // closes stopCh; startOnce not yet fired
tc.Start(context.Background())    // fires startOnce, launches goroutine → exits on closed stopCh
// tc cleanup loop is permanently dead; no background expiry occurs
```

**Suggested fix:**

Track whether `Stop()` was called before `Start()` with an atomic flag, and have `Start()` check it:

```go
func (c *TokenCache) Start(ctx context.Context) {
    if atomic.LoadUint32(&c.stopped) == 1 {
        return // genuinely a no-op after Stop
    }
    c.startOnce.Do(func() { ... })
}
```

Alternatively, replace `startOnce` with a mutex-guarded state machine (not started / running / stopped) so that `Start`-after-`Stop` can be documented and enforced with precision.

**Session:** rev-20260618T1310Z-37e71f

### LINKED ISSUE #804: [rev-20260618T1310Z-37e71f] findMatchingCapability tie-break condition is unreachable for second equal-score principal constraint
**Location:** `pkg/enforcement/engine.go:800-808`

**Reference:** https://github.com/eunolabs/eunox/blob/509758192e647c0fbd53e2490759741815bc8293/pkg/enforcement/engine.go#L800-L808

**Severity:** Low

**Description:**

`findMatchingCapability` iterates over constraints and selects the best match via:

```go
if score > bestScore || (score == bestScore && hasPrincipal && !bestPrincipal) {
    bestScore = score
    bestPrincipal = hasPrincipal
    best = &constraints[i]
}
```

The intent of the tie-break `(score == bestScore && hasPrincipal && !bestPrincipal)` is to prefer a principal-scoped constraint over a wildcard constraint at the same specificity score. However, once `best` has been set to a principal-scoped constraint (`bestPrincipal = true`), the condition `hasPrincipal && !bestPrincipal` evaluates to `true && false = false` for any subsequent principal-scoped constraint at the same score. The second constraint is silently discarded.

This means that when two principal-scoped constraints have identical specificity scores — a realistic configuration for two principals who both match on the same tool — the winner is determined entirely by iteration order (i.e., declaration order in the manifest YAML). An operator who lists a more-specific-in-intent principal constraint second will see the wrong constraint applied, with no error or warning.

**Trigger / repro:**

```yaml
capabilities:
  - tools:
      - name: "query_db"
    principals: ["alice"]
    conditions:
      - type: maxCalls
        args: { limit: 10, window: 60s }
  - tools:
      - name: "query_db"
    principals: ["alice"]
    conditions:
      - type: maxCalls
        args: { limit: 100, window: 3600s }
```

A request from `alice` for `query_db` will always match the first constraint (limit 10/60s) regardless of which was intended to be more authoritative.

**Suggested fix:**

Clarify the intended semantics and enforce them explicitly:

- If two equal-score principal constraints for the same principal are an operator error, add a validation step in manifest loading (`Validate()`) that rejects ambiguous duplicates and surfaces a clear error.
- If the intended behavior is "last wins" or "most-recently-defined wins", replace `!bestPrincipal` with `true` in the tie-break to allow later principal constraints to override earlier ones.
- Document the current first-wins behavior in `docs/capability-manifest-guide.md` so operators are not surprised.

**Session:** rev-20260618T1310Z-37e71f

### LINKED ISSUE #805: [rev-20260618T1310Z-37e71f] Redis incrIfBelowScript omits EXPIRE on denied path; key can expire and silently reset the quota window
**Location:** `pkg/callcounter/redis.go` — `incrIfBelowScript` Lua source (approximately line 210)

**Reference:** https://github.com/eunolabs/eunox/blob/509758192e647c0fbd53e2490759741815bc8293/pkg/callcounter/redis.go#L210

**Severity:** Low

**Description:**

`incrIfBelowScript` is the atomic Lua script that backs `IncrementIfBelow`. Its logic is:

1. `ZREMRANGEBYSCORE` — prune entries older than `windowSec`.
2. `ZCARD` — count remaining entries.
3. If `count < limit`: `ZADD` the new entry + `EXPIRE` the key for `windowSec * cleanupMarginFactor`.
4. If `count >= limit` (denied): return the retry-after time. **No `EXPIRE` is issued.**

The TTL on the Redis key is only refreshed when a call is **admitted**. If every call is denied (the key is constantly at or above limit), the key's TTL counts down from the last admitted call. After `windowSec * cleanupMarginFactor` seconds of continuous denial with no admitted calls, the Redis key expires and is deleted.

On the next call, `ZCARD` returns 0 (key doesn't exist yet), `count < limit` is satisfied, and the call is admitted — even if the real sliding-window count (had the key survived) would still be at or above limit.

In practice this edge case requires a sustained period of all-denied traffic with zero admitted calls and no other key activity for `2 * windowSec`. For short windows (seconds) this is reachable; for hour-scale windows it is theoretical. The behavior is safe in the sense that the window has legitimately slid past the old entries, but it violates the semantic contract that a key at limit stays at limit until the oldest admitted entries slide out of the window.

**Trigger / repro:**

```
windowSec = 5, limit = 1, cleanupMarginFactor = 2 (TTL = 10s)
t=0: call admitted → key TTL = 10s
t=1..9: 9 calls denied → no EXPIRE issued; TTL counting down
t=10: Redis key expires (TTL hit 0)
t=11: call admitted (count=0 < 1) → new key created; quota effectively reset
```

**Suggested fix:**

Issue `EXPIRE` unconditionally on every script execution, not only on admission:

```lua
-- after pruning, always refresh TTL regardless of admit/deny decision
redis.call('EXPIRE', KEYS[1], tonumber(ARGV[3]) * 2)
```

This costs one extra Redis command per denied call (absorbed into the same pipeline/script execution), ensures the key survives as long as there are entries to track, and matches the behavior callers would expect from a sliding-window counter.

**Session:** rev-20260618T1310Z-37e71f

### LINKED ISSUE #806: [rev-20260618T1410Z-07e43b] pkg/callcounter: StartCleanup second TOCTOU permanently disables cleanup goroutine
**Location:** `pkg/callcounter/memory.go` lines 542–550
**Reference:** https://github.com/eunolabs/eunox/blob/509758192e647c0fbd53e2490759741815bc8293/pkg/callcounter/memory.go#L542-L550
**Severity:** Medium

## Description

`StartCleanup` uses an `atomic.Bool` (`cleanupStarted`) as a run-once guard. The implementation contains a second context liveness check (line 542) to reject an already-cancelled context before launching the goroutine. However, if the context is cancelled **between** the `ctx.Err() != nil` check on line 542 and the `go func()` launch on line 550, the goroutine starts and immediately exits via `<-ctx.Done()` — but `cleanupStarted` is never reset to `false`.

All future callers see `CompareAndSwap(false, true)` fail (the bool is already `true`) and return `false` without starting a new goroutine. Cleanup is permanently disabled for the lifetime of the `InMemory` instance, causing unbounded growth of `m.entries` as sliding-window buckets accumulate without expiry.

## Trigger

Pass a context that is cancelled concurrently while `StartCleanup` is executing — e.g. a test that uses `context.WithTimeout` with a very short deadline, or a service that shuts down quickly during startup.

## Suggested Fix

Reset `cleanupStarted` to `false` inside the goroutine when it exits via `ctx.Done()`, so a future caller can restart cleanup with a fresh context:

```go
go func() {
    t := time.NewTicker(interval)
    defer func() {
        t.Stop()
        m.cleanupStarted.Store(false) // allow restart with a new context
    }()
    for {
        select {
        case <-t.C:
            m.Cleanup()
        case <-ctx.Done():
            return
        }
    }
}()
```

Alternatively, drop the second `ctx.Err()` check entirely and let the goroutine's `<-ctx.Done()` case handle an already-cancelled context atomically.

### LINKED ISSUE #807: [rev-20260618T1410Z-07e43b] pkg/capability: fetchKeys does not drain non-200 response body, breaking HTTP keep-alive
**Location:** `pkg/capability/jwkscache.go` lines 504–505
**Reference:** https://github.com/eunolabs/eunox/blob/509758192e647c0fbd53e2490759741815bc8293/pkg/capability/jwkscache.go#L504-L505
**Severity:** Low

## Description

`fetchKeys` correctly closes the HTTP response body via `defer resp.Body.Close()`, but when the server returns a non-200 status code it returns an error immediately without draining the body first:

```go
if resp.StatusCode != http.StatusOK {
    return nil, fmt.Errorf("JWKS endpoint returned %d", resp.StatusCode)
    // body closed by defer but not drained
}
```

Go's `net/http` transport can only reuse a keep-alive connection if the previous response body is fully consumed before `Close()`. Closing without draining forces the transport to discard the TCP connection. Under the circuit-breaker-guarded JWKS refresh path this means every non-200 response (e.g. a transient 503 during JWKS rotation) tears down a pooled connection and opens a new TCP+TLS handshake on the next probe — adding latency on the hot JWT verification path.

## Trigger

A JWKS endpoint that returns any non-200 status (503, 429, 404, etc.) repeatedly.

## Suggested Fix

Drain the body before returning the error:

```go
if resp.StatusCode != http.StatusOK {
    _, _ = io.Copy(io.Discard, resp.Body)
    return nil, fmt.Errorf("JWKS endpoint returned %d", resp.StatusCode)
}
```

eunolabs:
## Bugs addressed

This PR clears every open issue tagged with one of the current review-session IDs, across five sessions, in one branch.

### Session rev-20260618T1410Z-07e43b
- Closes #807 — `fetchKeys` closed a JWKS non-200 response body without draining it, forcing `net/http` to discard the keep-alive connection on every transient 5xx/429 during rotation. Now drains the body (bounded to 1 MiB) before returning the error.
- Closes #806 — A cancel racing the window between `StartCleanup`'s post-CAS re-check and the goroutine launch left `cleanupStarted` stuck at `true`, permanently disabling cleanup. The goroutine now releases the start guard on exit (deferred), so a later live context can restart cleanup.

### Session rev-20260618T1310Z-37e71f
- Closes #805 — `incrIfBelowScript` issued `EXPIRE` only on admission, so a key held continuously at limit could expire mid-window and silently reset the quota. The denied path now refreshes the TTL while in-window entries remain.
- Closes #804 — Documented and test-locked the first-wins resolution for two equal-specificity principal constraints (the tie-break clause is reachable only for principal-beats-general); added a manifest-guide warning against positional ambiguity.
- Closes #802 — `TokenCache.Stop()` before `Start()` left a later `Start()` launching a goroutine that died on the closed `stopCh`. `Start` now checks the stopped flag under the lock, making Start-after-Stop a true no-op.
- Closes #795 — `TokenCache.Get()` returned a live pointer into the cached entry; a caller mutating a scalar (e.g. `ExpiresAt`) corrupted the cache. Now returns a shallow copy.
- Closes #801 — The bare `Allow()` + `RecordSuccess/Failure` API bypasses the half-open generation guard and can prematurely close the breaker on a stale outcome. Added the generation-safe, panic-safe `AllowFunc` helper and hardened the bare-API docs to steer callers to `Do`/`AllowFunc`/`AllowProbe`.
- Closes #800 — `readLastAuditLine` returned `("", nil)` when the log shrank between `Stat` and `ReadAt`, so the caller started a fresh chain with no `chain_resume_failed` marker. Now returns `errAuditFileShrunk` so a signed resume-failure marker is written.

### Session rev-20260618T1230Z-f73824
- Closes #798 — Clarified the intentional single-vs-compound extension asymmetry (`.gz` matching `.tar.gz`) in the manifest guide and code, including how to read the denial log's `extension` field. The `filepath.Ext` gate is intentional (#243) and must not be unified naively.
- Closes #797 — `Allow()` in HalfOpen consumes a probe slot with no panic safety, wedging the breaker if the guarded call panics. Added the panic-safe `AllowFunc` (shared with #801) and documented the requirement.
- Closes #796 — `ValidateToken` re-sampled `p.now()` inside the per-key verifier closure, making the exp/leeway decision non-deterministic across JWKS retries. The clock is now sampled once before the closure.
- Closes #795 — (see above; shipped with #802 in the same file).

### Session rev-20260618T0852Z-9d6c2a
- Closes #791 — Forced and background JWKS refreshes use separate singleflight keys and run concurrently; an older-started background fetch could finish last and clobber a newer forced rotation. Added a fetch-generation guard so the cache only installs a result newer than the one already installed.
- Closes #790 — A JWT with `mcp.capabilities: null` collapsed to "absent" and bypassed the exhaustive allowlist (the manifest backstop governed). Now rejected as a terminal validation error.
- Closes #789 — The literal `?` query delimiter in an http(s) resource claim was interpreted by `path.Match` as a single-char glob, so one URL's claim authorized a different path. The query is now matched literally; globbing applies only to the path portion.
- Closes #788 — `id: null` JSON-RPC requests were misclassified as notifications (the `*json.RawMessage` collapses absent and null to nil). A custom `rpcMsg.UnmarshalJSON` now preserves present-null as a non-nil `null` id, so it is classified as a request and echoed back as `null`.

### Session rev-20260618T0852Z-174fa2
- Closes #787 — The e2e HTTP client discarded `io.ReadAll`/`json.Unmarshal` errors, so a 200 with a malformed/truncated body passed initialization. It now propagates read errors, requires a valid JSON-RPC envelope, and requires a non-empty initialize result.
- Closes #786 — Redis killswitch `Stop` could return while a concurrent `Start` was still using the client (Start published `cancel` and ran `refreshState` before its `wg.Add` calls). A lifecycle mutex now serializes `Start`/`Stop`, ordering the `wg.Add` calls before `Stop`'s `Wait`; a Stop that wins the race marks the switch stopped so Start becomes a no-op.

## Testing
- `make check-fmt`, `make check-license`, `make check-notice`, `make lint` — all clean.
- `make coverage` — pkg/ at 95.6% (gate is 80%).
- New regression tests accompany every fix (table-driven where applicable; `httptest`/miniredis/blocking-client harnesses for the concurrency and transport fixes), each referencing its issue number.

https://claude.ai/code/session_01KjZAV8brNYqjjZ6NfDE8p7

---
_Generated by [Claude Code](https://claude.ai/code/session_01KjZAV8brNYqjjZ6NfDE8p7)_


eunolabs:
✅ Ready to merge — all session bugs addressed (#807, #806, #805, #804, #802, #801, #800, #798, #797, #796, #795, #791, #790, #789, #788, #787, #786), self-review complete with no outstanding comments, tests passing.

Self-review found and fixed one issue before this: the initial #787 hardening errored on any non-JSON body, which broke the e2e unknown-route case (a 404 with a plain-text body). It's now scoped to 2xx responses (the transport-framing concern #787 targets), with a regression test pinning both directions. CI is green on the fixed head (Test, Lint, demo/e2e + all demo modes, OPA comparisons, builds, and every format/license/NOTICE/enforcement gate).

---

PR #815, opened 2026-06-18
Fix session rev-20260618T1410Z-07e43b bugs

### LINKED ISSUE #808: [rev-20260618T1410Z-07e43b] cmd/eunox-mcp: jwtShorthandValues omits JSON null, so null-typed tool arguments can never match JWT claims
**Location:** `cmd/eunox-mcp/pdp_jwt.go` lines 944–954
**Reference:** https://github.com/eunolabs/eunox/blob/509758192e647c0fbd53e2490759741815bc8293/cmd/eunox-mcp/pdp_jwt.go#L944-L954
**Severity:** Medium

## Description

`jwtShorthandValues` expands a raw claim value string into a slice of candidates for `enforcement.MatchAllowedValue`. It handles `float64` and `bool` JSON scalars but has no case for `null`:

```go
func jwtShorthandValues(raw string) []interface{} {
    values := []interface{}{raw}
    var scalar interface{}
    if err := json.Unmarshal([]byte(raw), &scalar); err == nil {
        switch scalar.(type) {
        case float64, bool:
            values = append(values, scalar)
        // nil / JSON null: not handled
        }
    }
    return values
}
```

When a JWT claim encodes `tool:foo?enabled=null`, `raw` is the string `"null"`. `json.Unmarshal` decodes that to Go `nil`, which falls through the type switch unhandled. The function returns `[]interface{}{"null"}` — a one-element slice containing only the string `"null"`.

If the incoming tool call has an `enabled` argument whose JSON value is `null`, that argument arrives as Go `nil`. `enforcement.MatchAllowedValue(nil, ["null"])` attempts to glob-match `nil` against the string pattern `"null"`, which fails — the match never succeeds. A JWT claim intended to restrict calls where an argument is explicitly `null` is silently inoperative.

This means the intended policy `tool:sensitive?secret=null` (permit only calls where `secret` is JSON null) cannot be expressed using the v0.2 shorthand syntax: calls with any value of `secret` will all fail the `allowedValues` check equally, whether or not `secret` is null.

## Suggested Fix

Add a `nil` case to the type switch:

```go
switch v := scalar.(type) {
case float64, bool:
    values = append(values, v)
case nil:
    values = append(values, nil)
}
```

### LINKED ISSUE #810: [rev-20260618T1410Z-07e43b] cmd/eunox-mcp: observedDescHash in ManifestPDP grows without bound under high-cardinality upstream
**Location:** `cmd/eunox-mcp/pdp.go` lines 252–268 (`recordObservedDescription`) and 822–827 (`filterToolsListResult`)
**Reference:** https://github.com/eunolabs/eunox/blob/509758192e647c0fbd53e2490759741815bc8293/cmd/eunox-mcp/pdp.go#L252-L268
**Severity:** Medium

## Description

`filterToolsListResult` calls `mdp.recordObservedDescription(t.Name, t.Description)` for **every tool** the upstream advertises, including tools that are not in the manifest and will never be permitted:

```go
// pdp.go ~line 827
for _, t := range upstream.Tools {
    mdp.recordObservedDescription(t.Name, t.Description)
    // ...filter to permitted entries...
}
```

`recordObservedDescription` stores a SHA-256 hash of the description keyed by tool name in `p.observedDescHash`, a plain `map[string]string` with no size cap or eviction policy:

```go
func (p *ManifestPDP) recordObservedDescription(name, description string) {
    h := computeDescriptionHash(description)
    p.descMu.Lock()
    if p.observedDescHash == nil {
        p.observedDescHash = make(map[string]string)
    }
    p.observedDescHash[name] = h // grows without bound
    p.descMu.Unlock()
}
```

A malicious or misconfigured upstream that advertises a large number of unique tool names (e.g. by including a UUID or timestamp in each name across successive `tools/list` responses) will cause `observedDescHash` to grow without limit. Each entry is small (~100 bytes of map overhead + key + value), but at 100 k unique tool names that is ~10 MB of heap — never reclaimed because the map is rooted in the long-lived `ManifestPDP` struct. In gateway mode, each route owns its own `ManifestPDP`, so the attack surface multiplies by the number of routes.

## Trigger

An upstream MCP server that rotates tool names on each `tools/list` response — for example, by appending a timestamp or nonce. The host will call `tools/list` periodically; each call adds new entries that are never evicted.

## Suggested Fix

Cap insertion before storing, discarding entries beyond a reasonable bound:

```go
func (p *ManifestPDP) recordObservedDescription(name, description string) {
    h := computeDescriptionHash(description)
    p.descMu.Lock()
    defer p.descMu.Unlock()
    if p.observedDescHash == nil {
        p.observedDescHash = make(map[string]string)
    }
    const maxObserved = 4096
    if _, exists := p.observedDescHash[name]; !exists && len(p.observedDescHash) >= maxObserved {
        return // discard; description-drift detection is best-effort
    }
    p.observedDescHash[name] = h
}
```

Alternatively, only record descriptions for tools that appear in the manifest — tools not in the manifest are filtered out anyway, so tracking their descriptions provides no security value.

### LINKED ISSUE #814: [rev-20260618T1410Z-07e43b] cmd/eunox-mcp: NewJWTPDP does not wire opts.Clock into JWKSCacheConfig.Now, so JWKS cache TTL ignores the injected clock
**Location:** `cmd/eunox-mcp/pdp_jwt.go` lines 270–276 (`NewJWTPDP`)
**Reference:** https://github.com/eunolabs/eunox/blob/509758192e647c0fbd53e2490759741815bc8293/cmd/eunox-mcp/pdp_jwt.go#L270-L276
**Severity:** Low

## Description

`NewJWTPDP` stores `opts.Clock` in the `JWTPDP.clock` field (used for JWT `exp`/`nbf`/`iat` validation in `ValidateToken`), but never passes `opts.Clock.Now` to `JWKSCacheConfig.Now` when constructing the `JWKSCache`:

```go
// pdp_jwt.go line 270-276
return &JWTPDP{
    cache: capability.NewJWKSCache(capability.JWKSCacheConfig{
        JWKSURL:  opts.JWKSURI,
        CacheTTL: opts.CacheTTL,
        Client:   opts.Client,
        Breaker:  breaker,
        Logger:   logger,
        // Now: opts.Clock.Now  <-- missing
    }),
    ...
    clock: opts.Clock,   // stored for JWT validation only
```

`JWKSCacheConfig.Now` defaults to `time.Now` when unset (see `pkg/capability/jwkscache.go:148-150`), so the `JWKSCache` always uses the real wall clock regardless of what clock is injected into the `JWTPDP`.

The code comment at lines 183–186 explicitly documents the intent:

```
// a test that freezes time injects one to exercise the expiry path
// deterministically (#581). enforcement.Clock is the same abstraction
// the engine and TokenCache use, so a frozen clock stays consistent
// across all three in integration tests.
```

The "all three" refers to the engine, `TokenCache`, and `JWTPDP`. The `JWKSCache` should also be in that set, but it is silently excluded, breaking the stated design goal.

## Impact

- **Test correctness:** Integration tests that inject a `testutil.FakeClock` to drive JWT expiry, JWKS cache TTL, or key-rotation scenarios see two different "nows": the fake clock for JWT validation, and `time.Now` for JWKS cache expiry. Tests that advance the fake clock to trigger a JWKS cache miss and re-fetch will never actually expire the cache — the cache TTL runs on real wall time and is never reached during a fast test. This can mask bugs in the JWKS cache's expiry/refresh paths when accessed through `JWTPDP`.

- **Subtle inconsistency in boundary conditions:** In production both clocks are `time.Now`, so there is no observable behavioral difference. The risk is purely in test fidelity: the code promises deterministic time injection but silently fails to deliver it for the cache component.

## Trigger

Any integration test that:
1. Constructs a `JWTPDP` via `NewJWTPDP` with a `FakeClock`,
2. Issues a token validation request (populating the JWKS cache),
3. Advances the fake clock past the JWKS cache TTL, and
4. Expects the cache to have expired and re-fetch.

The cache will not expire regardless of how far the fake clock is advanced.

## Suggested Fix

Wire the clock into the cache config when `opts.Clock` is non-nil:

```go
cacheConfig := capability.JWKSCacheConfig{
    JWKSURL:  opts.JWKSURI,
    CacheTTL: opts.CacheTTL,
    Client:   opts.Client,
    Breaker:  breaker,
    Logger:   logger,
}
if opts.Clock != nil {
    cacheConfig.Now = opts.Clock.Now
}
return &JWTPDP{
    cache: capability.NewJWKSCache(cacheConfig),
    ...
}
```

This change is backward-compatible: callers that do not inject a clock continue to get `time.Now` (the existing default), and tests that inject a `FakeClock` now get a fully consistent time view across JWT validation and JWKS cache TTL management.

eunolabs:
## Bugs addressed (session rev-20260618T1410Z-07e43b)

- Closes #810 — `ManifestPDP.recordObservedDescription` recorded a hash for **every** tool an upstream advertised, keyed by name in an uncapped `map[string]string` rooted in the long-lived PDP (once per route in gateway mode). An upstream rotating tool names (nonce/timestamp per `tools/list`) grew it without bound. Now capped at `maxObservedDescriptions` (4096): new names past the cap are discarded, updates to already-tracked names still apply. Drift detection is best-effort, so a never-seen name has no pinned `descriptionHash` to enforce and is safe to drop.
- Closes #808 — `jwtShorthandValues` expanded `float64`/`bool` JSON scalars but had no `null` case, so a claim like `tool:foo?secret=null` yielded only the string `"null"`. A tool argument set to JSON null arrives as Go `nil`, which `MatchAllowedValue` glob-matches against `"null"` and never matches — making the v0.2 shorthand for null-valued arguments silently inoperative. Added a `nil` case so null is offered as a match candidate.
- Closes #814 — `NewJWTPDP` stored `opts.Clock` for JWT `exp`/`nbf` validation but never passed `opts.Clock.Now` to `JWKSCacheConfig.Now`, so the JWKS cache TTL always ran on the real wall clock. Tests freezing time to drive JWKS cache expiry/key-rotation through `JWTPDP` saw two different "nows" and could never expire the cache. Now wires `opts.Clock.Now` into the cache config when non-nil; a nil clock leaves it unset (cache defaults to `time.Now`), so production wiring is unchanged.

## Testing
- New regression tests, each verified to fail before its fix and pass after:
  - `TestRecordObservedDescriptionCap` (#810) — fills to the cap, asserts overflow names are discarded and tracked-name updates still apply with no growth.
  - `TestJWTShorthandValuesMatchTypedArgs` (#808) — added cases: null shorthand matches a nil arg; null shorthand denies a non-null arg.
  - `TestJWT_JWKSCacheTTLUsesInjectedClock` (#814) — advancing the injected `FakeClock` past `CacheTTL` forces a JWKS re-fetch (confirmed to fail with `hits=1` before the fix).
- `go build ./...`, `go vet`, and `make check-fmt` clean.
- Full suite green: `go test -race -count=1 ./cmd/eunox-mcp/ ./pkg/...`

---
_Generated by [Claude Code](https://claude.ai/code/session_018VThV6SUzKJdsDKbUUYiFm)_


cmd/eunox-mcp/pdp.go
```diff
@@ -249,14 +249,25 @@ func NewManifestPDP(manifest *LocalManifest, engine *enforcement.Engine, ks kill
...
+const maxObservedDescriptions = 4096
+
 func (p *ManifestPDP) recordObservedDescription(name, description string) {
 	h := computeDescriptionHash(description)
 	p.descMu.Lock()
+	defer p.descMu.Unlock()
 	if p.observedDescHash == nil {
 		p.observedDescHash = make(map[string]string)
 	}
+	if _, exists := p.observedDescHash[name]; !exists && len(p.observedDescHash) >= maxObservedDescriptions {
```
eunolabs: The cap stops the unbounded growth, but the **discard-new** eviction can be turned into a bypass of the very #668 call-leg check this map exists for — by the same name-rotating upstream this fix targets.

`observedDescHash` is read in exactly one place — `decideTarget`, gated to pinned exact-tool constraints:

```go
if schema == validateSchema && isPinnedExactToolConstraint(matched) {
    if observed, ok := p.observedDescriptionHash(target.Name); ok && observed != matched.DescriptionHash {
        // deny (fail closed)
    }
}
```

An *unobserved* name (`ok == false`) is **allowed** — the comment there trusts start-time drift verification. Combined with discard-new, a malicious upstream can suppress a pinned tool's record: on the first `tools/list` of the process, return 4096 junk names (filling the cap via `recordObservedDescription`, which runs per-entry *before* the constraint check) followed by a mid-session-poisoned pinned tool. That tool's changed description is discarded → `ok == false` → the `tools/call` leg forwards the poisoned call. Narrow (needs the flood to land before that tool is ever observed, and a host that calls the cached name), but it's a fail-open under the stated threat model.

Since the map is only ever consulted for pinned tools, consider **recording only for pinned exact-tool constraints** rather than capping: compute the constraint first in `filterToolsListResult` and gate `recordObservedDescription` on `isPinnedExactToolConstraint(c)`. That bounds the map by the manifest's pinned set (operator-controlled, small), closes the rotation/DoS vector completely, and removes the eviction fail-open — no arbitrary cap needed. The "record even the about-to-be-hidden tool" property (#668) is preserved because `findConstraint` matches by name regardless of the hash mismatch.

Minor, related: the struct-field comment above (`observedDescHash` … *"bounded by the upstream's advertised tool set, so it cannot grow without bound"*) is now contradicted by this fix and should be updated.

---

cmd/eunox-mcp/pdp_jwt.go
```diff
@@ -1009,6 +1017,12 @@ func jwtShorthandValues(raw string) []interface{} {
 		switch scalar.(type) {
 		case float64, bool:
 			values = append(values, scalar)
+		case nil:
```
eunolabs, line 1021: Fix is correct: `ResolveArgument` returns `present=false` for an absent key (so absent args still deny via `MISSING_CONTEXT` — only an *explicitly* null arg reaches the matcher), and `MatchAllowedValue` matches the nil candidate via `reflect.DeepEqual`. The added "denies non-null arg" case nails down the deny side.

One cleanup: the function doc comment just above is now stale — it still reads *"Values that are not a JSON number or boolean (plain strings, globs, `"null"`) add no scalar and keep the string-only behaviour."* After this change, `"null"` **does** add a scalar (nil). Worth updating that sentence so the doc matches the new `case nil`.

---

eunolabs (commented):
Reviewed all three fixes against the surrounding enforcement code. Two are correct as-is; the #810 fix bounds growth but the chosen eviction policy has a narrow fail-open I'd like to flag before merge. Not blocking — flagging for a decision.

## Overview
Three independent, well-scoped bug fixes, each with a fail-before/pass-after regression test. PR description is excellent and the test design (isolating cache-TTL from JWT-exp, asserting no-growth + tracked-name updates) is solid.

## #814 — wire injected clock into JWKS cache — LGTM
Correct and clean. `opts.Clock` is `enforcement.Clock`, `JWKSCacheConfig.Now` is `func() time.Time`, and `NewJWKSCache` defaults `Now` to `time.Now` when nil (`jwkscache.go:161`), so the nil-clock production path is genuinely unchanged. The `#581` reference in the new comment is consistent with the file's existing convention for the clock-injection lineage (not a typo for #814).

## #808 — null shorthand value — correct, one stale comment
The `case nil` is right, and importantly it does **not** make absent args match: `ResolveArgument` returns `present=false` for a missing key, which denies via `MISSING_CONTEXT` before the matcher runs — so only an explicitly-null arg matches. Left an inline nit: the function doc comment still lists `"null"` as adding no scalar, which is now wrong.

## #810 — observed-description cap — fixes growth, but eviction can fail open
The cap does stop unbounded growth. My concern (inline on `pdp.go`) is the **discard-new** eviction: `observedDescHash` is read only for pinned exact-tool constraints, and an unobserved name is *allowed* on the call leg, so a name-flooding upstream (the same actor #810 cites) can push a poisoned pinned tool past the cap and bypass the #668 call-leg drift check. Since the map is only useful for pinned tools, recording **only** for pinned exact-tool constraints (instead of capping) bounds it by the manifest's pinned set, closes the DoS vector entirely, and avoids the fail-open. Details + suggested change inline.

## Conventions
Tests land in the right files (`pdp_test.go`, `pdp_jwt_*_test.go`); no manifest-grammar / audit-shape / schema changes, so no docs or threat-model updates are required for these fixes. `fmt`/`vet`/suite reported green.

---

cmd/eunox-mcp/pdp.go
eunolabs: Good catch — agreed this is a real fail-open under the stated threat model. Fixed in 7b0d29d by recording only for pinned exact-tool constraints (gated on `isPinnedExactToolConstraint(c)` after computing the constraint in `filterToolsListResult`), which bounds the map by the operator-controlled manifest pinned set and removes the eviction fail-open entirely — the cap and `maxObservedDescriptions` are gone, and `recordObservedDescription` reverts to its simple form. The #668 "record the about-to-be-hidden tool" property is preserved (`findConstraint` matches by name regardless of hash). Also updated the stale struct-field comment. Added `TestFilterToolsList_RecordsOnlyPinnedDescriptions` (verified it reports 5002 entries under the old record-everything behavior, 1 after the fix) and `TestFilterToolsList_RecordsHiddenPinnedTool` for the #668 case.

---

cmd/eunox-mcp/pdp_jwt.go
eunolabs: Done in 7b0d29d — updated the doc comment to note that a JSON `"null"` now adds the nil scalar, and that only values parsing as none of number/boolean/null keep the string-only behaviour.

---

PR #822, opened 2026-06-18
Fix session rev-20260618T1410Z-07e43b bugs

### LINKED ISSUE #808: [rev-20260618T1410Z-07e43b] cmd/eunox-mcp: jwtShorthandValues omits JSON null, so null-typed tool arguments can never match JWT claims

### LINKED ISSUE #810: [rev-20260618T1410Z-07e43b] cmd/eunox-mcp: observedDescHash in ManifestPDP grows without bound under high-cardinality upstream

### LINKED ISSUE #814: [rev-20260618T1410Z-07e43b] cmd/eunox-mcp: NewJWTPDP does not wire opts.Clock into JWKSCacheConfig.Now, so JWKS cache TTL ignores the injected clock

eunolabs:
## Bugs addressed (session rev-20260618T1410Z-07e43b)

- Closes #808 — `jwtShorthandValues` only appended a typed scalar for JSON numbers and booleans, so a v0.2 shorthand claim like `tool:foo?secret=null` expanded to just `["null"]`. A tool argument whose JSON value is `null` arrives as Go `nil`, and the string `"null"` can never glob-match `nil`, so the policy was silently inoperative. Added `nil` to the type switch so the parsed null scalar is appended (matching a nil argument by value) while keeping the string form for glob matching.
- Closes #810 — `filterToolsListResult` recorded a description hash for every advertised tool (including unpermitted ones) into the long-lived `ManifestPDP.observedDescHash`, which had no size cap. A name-rotating upstream (nonce/timestamp in each tool name) could grow it without bound, multiplied per route in gateway mode. Capped the map at 4096 entries; updates to already-tracked names still proceed (so the call-leg descriptionHash pin keeps following live drift), only brand-new names beyond the cap are discarded.
- Closes #814 — `NewJWTPDP` stored `opts.Clock` for JWT exp/nbf/iat validation but never passed it to `JWKSCacheConfig.Now`, so the JWKS cache TTL always ran on `time.Now`. A test that freezes time with a `FakeClock` saw two different "nows", making the cache expiry/refresh path unreachable in fast tests. Now sets `cacheConfig.Now = opts.Clock.Now` when a clock is injected; unset, `JWKSCacheConfig` still defaults `Now` to `time.Now`, so production behaviour is unchanged.

## Testing
- Added regression tests: `TestJWTShorthandValuesMatchTypedArgs` (new null cases, #808), `TestRecordObservedDescriptionBounded` (#810), `TestJWT_JWKSCacheUsesInjectedClock` (#814). Confirmed the #808 and #814 tests fail without their respective fixes.
- `make check-fmt`, `make check-license`, `go vet ./...` — all clean.
- `go test -race -count=1 ./cmd/eunox-mcp/ ./pkg/...` — all packages pass.

## Related
- Spec/docs for the #808 typed-value (incl. `null`) shorthand matching are in `eunolabs/mcp-capability-manifest` branch `claude/affectionate-goldberg-7y6b1l` (separate repo, separate PR).

---
_Generated by [Claude Code](https://claude.ai/code/session_018tF6kRfeZ8xzUeJfWX7PcG)_


eunolabs:
## Review — three session bug fixes (#808, #810, #814)

Traced each fix against the surrounding enforcement code at the head commit (`096799c`), checked out the branch locally (`go vet ./cmd/eunox-mcp/` clean, the three new regression tests pass under `-race`), and all 43 CI checks are green / the PR is mergeable.

Overall a tight, well-scoped set: each bug is real, each fix is minimal and matches the repo's fail-closed conventions, and each ships a regression test. One substantive point on #810, otherwise non-blocking.

### #808 — null shorthand value — correct, and correctly fail-closed
Adding `nil` to the `jwtShorthandValues` switch is right: `json.Unmarshal("null")` yields Go `nil`, so the set becomes `["null", nil]` and `MatchAllowedValue`'s exact path matches a `nil` argument via `reflect.DeepEqual(nil, nil)`.

The subtle risk here is whether it lets a *missing* argument slip through as "null → allow". It doesn't, and I verified both legs:
- **engine** — `handleAllowedValues` → `ResolveArgument` uses the comma-ok idiom, so a missing key returns `(nil, false)` → `MISSING_CONTEXT` deny *before* matching; an explicit null returns `(nil, true)` → reaches the matcher.
- **JWT** — `evaluateJWTConditions`' `AllowedValuesCondition` arm gates on `argExists` identically.

So `secret=null` only matches an argument that is *present and explicitly null*. Good — and nice to see the deny case in the table too.

### #814 — JWKS cache clock injection — correct, production-safe
`cacheConfig.Now = opts.Clock.Now` is set only when `opts.Clock != nil`; `NewJWKSCache` defaults `Now` to `time.Now` (`jwkscache.go:161`), so production behaviour is unchanged. `enforcement.Clock.Now` is `func() time.Time`, matching the field. The test correctly exercises cached-hit (no fetch) vs. TTL-expiry (re-fetch) under an advancing fake clock.

### #810 — bounded `observedDescHash` — fixes the leak, but interacts with the call-leg pin
The cap logic is correct (updates to tracked names always proceed; only brand-new names past 4096 are dropped) and the `defer Unlock` cleanup is a nice touch. It does fix the unbounded growth.

One interaction worth flagging, because it touches the exact threat `descriptionHash` defends against — a hostile upstream (#668):

`recordObservedDescription` is called for **every advertised tool, before the constraint check** (`pdp.go:847`), so junk/unpermitted tools consume cap capacity. The call-leg pin (`pdp.go:439`) only denies when a description has been **observed *and* mismatches** — an *unobserved* description is deliberately not denied. So a malicious upstream can flood the (cross-session, persistent) map to 4096 distinct junk names such that a genuinely pinned tool, first advertised later, is never recorded → its mid-session drift is no longer caught on a `tools/call`-by-name. The list leg still hides it (it recomputes the hash inline, independent of the map) and the start-time drift probe still runs, so the residual gap is narrow — but it does make the #668 call-leg pin best-effort under flooding.

**Suggestion:** record only tools that match a name+`call` constraint — move the `recordObservedDescription` call *below* the `findConstraint`/`containsAction` check but keep it *above* the hash comparison (so a drifted-but-named tool is still recorded and the call leg still denies it). That bounds the map by the manifest's matched-tool set for everything except a `tool:*` grant and removes the junk-flooding eviction vector; keep the 4096 cap as the backstop for the `tool:*` case.

Either way, this narrows a security control from "enforced" to "best-effort under a >cap rotating upstream", so a line in `docs/threat-model-mcp.md` seems warranted per the repo's "behaviour changes need a docs update in the same PR" rule (the PR currently has no doc changes).

### Minor
- `maxObservedDescriptions` is a function-local const — consider hoisting it next to the `observedDescHash` field doc, and emitting a debug log / one-shot warning when the cap is first hit so a flooding upstream is observable in ops.
- PR title ("Fix session rev-…") is opaque; the commit subjects (`fix: match JSON null in JWT shorthand allowed values (#808)`, …) are much clearer — consider promoting one.

Nice work overall, especially the care on the #808 fail-closed boundary.

---

eunolabs:
Closing this as superseded.

While resolving the merge conflicts here, I found that **#815 has already landed on `main`** (commit `135f5f8`) and closes the same three issues this PR targets:

- **#808** — `jwtShorthandValues` offers Go `nil` for a JSON-null shorthand (identical code to this PR).
- **#814** — `NewJWTPDP` threads `opts.Clock.Now` into `JWKSCacheConfig.Now` (identical code to this PR).
- **#810** — bounds `observedDescHash` by **recording only pinned exact-tool constraint matches**, so the map is bounded by the operator-controlled manifest set "without a cap and without an eviction fail-open."

That last point is the key reason to prefer `main`'s version: it's the cleaner approach (recording is gated by the pinned set rather than capped), and it specifically avoids the cap's eviction fail-open — the exact weakness raised in the review above, where a name-rotating upstream could flood the map past the cap and slip a pinned tool's drift past the call-leg check.

Mechanically rebasing this branch onto `main` would *revert* `main`'s better `#810` fix back to the 4096-cap approach and also revert the `#823` `findConstraint` refactor; the net diff against `main` is 115 insertions / 154 deletions (it removes more than it adds). With all three bugs already fixed on `main`, there's nothing left for this PR to contribute, so closing rather than merging.

---

PR #832, opened 2026-06-18
Fix batch of automated-review bugs (sessions rev-20260618T16/17/18Z)

### LINKED ISSUE #816: [rev-20260618T1610Z-ac1ef7] handleAllowedExtensions: filepath.Base/Ext use OS path separators, breaking extension enforcement for Windows-style paths on Linux
**Location:** `pkg/enforcement/handlers.go`, lines 802 and 819 (`handleAllowedExtensions`)

**Reference:** commit `51bf11221a216defe5b33f777eb9251b124eebd4`

**Severity:** High — security enforcement bypass

**Description:**

`handleAllowedExtensions` extracts the file's base name and extension using `filepath.Base(filePath)` and `filepath.Ext(filePath)`:

```go
// line 802
// ...
// line 819
ext := strings.ToLower(filepath.Ext(base))
```

`filepath.Base` uses the **OS path separator**. On Linux the separator is `/`, so `\` is treated as a regular character — not a directory separator. A Windows-style path like `evil.exe\report.pdf` passed as `filePath` on a Linux proxy produces:

```
filepath.Base("evil.exe\\report.pdf") → "evil.exe\\report.pdf"   // whole string, no split
filepath.Ext("evil.exe\\report.pdf")  → ".pdf"                    // last dot segment
```

The comment on line 816 explicitly states the intent: _"Match the file name — not the directory path, so a dotted directory cannot smuggle an extension."_ This safety property is broken on Linux for paths containing `\`. An attacker can supply `malicious.exe\anything` and the extension check sees the full path as the "base name", but `filepath.Ext` still extracts `.exe` — or, conversely, supply `anything\report.pdf` where `report.pdf` is the permitted extension and the check passes even though the caller intended to deny paths containing directory traversal via `\`.

More critically, path components that look like extensions on the directory portion are now included in the comparison rather than stripped. For example, if only `.pdf` is allowed, `attack.exe\document.pdf` passes because `filepath.Ext` of the whole string is `.pdf`.

**Trigger:**

Any `allowedExtensions` constraint on a tool that accepts a file-path argument, where the upstream MCP client is Windows-origin or where an adversarial client constructs a path with backslash separators.

**Fix:**

Normalize path separators before calling `filepath.Base`/`filepath.Ext`, or use a cross-platform split that strips both `/` and `\` directory components:

```go
// Replace backslashes with forward slashes before splitting,
// so filepath.Base behaves consistently regardless of OS.
normalized := strings.ReplaceAll(filePath, "\\", "/")
base := path.Base(normalized)   // use "path" (Unix), not "filepath" (OS-specific)
ext  := strings.ToLower(path.Ext(base))
```

Alternatively, explicitly reject any path containing `\` as a structural validation step before extension extraction, consistent with the fail-closed invariant.

**Session:** rev-20260618T1610Z-ac1ef7

### LINKED ISSUE #817: [rev-20260618T1610Z-ac1ef7] interpretAuditTail: partial file-shrink (0 &lt; n &lt; buf, io.EOF) silently bypasses integrity check
**Location:** `cmd/eunox-mcp/audit.go`, lines 764–779 (`interpretAuditTail`)

**Reference:** commit `51bf11221a216defe5b33f777eb9251b124eebd4`

**Severity:** Medium — audit integrity gap

**Description:**

`interpretAuditTail` seeks to `size - len(buf)` bytes from the start of the file and reads into a fixed-size buffer to recover the last audit record. The read result is handled as:

```go
n, err := f.Read(buf)
if n == 0 && errors.Is(err, io.EOF) {
    return nil, 0, errAuditFileShrunk
}
// ... processes buf[:n] as though it is a valid tail
```

The `n == 0` check catches only the case where the file is completely empty (or has shrunk to fewer than `start` bytes, where `start = size - len(buf)`). It does **not** catch the case where:

```
n > 0  &&  n < len(buf)  &&  err == io.EOF
```

This occurs when the file has been truncated to between `start` bytes and `size` bytes — i.e., the file still has *some* content in the tail window, but less than expected. In this scenario:

1. `n > 0`, so `errAuditFileShrunk` is never returned.
2. `err == io.EOF` after a partial read is not treated as an error by the current code (standard Go `Read` behavior: partial read + io.EOF is normal for the last block of a file).
3. The truncated buffer is parsed for a newline-delimited JSONL record. If the truncation happened mid-record, no newline is found and the function returns `(nil, 0, nil)` — a silent non-error. If truncation happened after a record boundary, an earlier (now-incorrect) record is validated as "the tail".

In both sub-cases, the function does not surface `errAuditFileShrunk`, so the caller (`readLastAuditLine` and the audit-verify path) silently treats a shrunk file as intact, undermining the tamper-evidence guarantee.

**Trigger:**

An audit log file that has been truncated (e.g., by a log-rotation tool or an adversary with write access) to a size between `size - tailBufSize` and `size` bytes passes integrity check without raising `errAuditFileShrunk`.

**Fix:**

After the `Read` call, also check for a short read paired with `io.EOF`:

```go
n, err := f.Read(buf)
if n == 0 && errors.Is(err, io.EOF) {
    return nil, 0, errAuditFileShrunk
}
// A partial read (n < len(buf)) with io.EOF means the file is shorter than
// expected — it was truncated between start and size bytes.
if n < len(buf) && errors.Is(err, io.EOF) {
    return nil, 0, errAuditFileShrunk
}
if err != nil && !errors.Is(err, io.EOF) {
    return nil, 0, fmt.Errorf("reading audit tail: %w", err)
}
```

**Session:** rev-20260618T1610Z-ac1ef7

### LINKED ISSUE #818: [rev-20260618T1610Z-ac1ef7] TokenCache.Put: O(n) expired-entry sweep held under write lock stalls concurrent Get calls
**Location:** `pkg/capability/token_cache.go`, lines 286–295 (`TokenCache.Put`)

**Reference:** commit `51bf11221a216defe5b33f777eb9251b124eebd4`

**Severity:** Medium — latency/throughput regression under load

**Description:**

`Put` includes a fast O(1) path that reclaims expired entries from the front of the insertion-order list. When the cache is still at capacity after that front reclaim, it falls back to an O(n) full sweep:

```go
    for el := c.insertOrder.Front(); el != nil; {
        next := el.Next()
        ek := el.Value.(string)
        if e, ok := c.entries[ek]; !ok || now.After(e.expiresAt) {
            c.insertOrder.Remove(el)
            delete(c.entries, ek)
        }
        el = next
    }
}
```

This entire loop runs **under `c.mu.Lock()`** (the write lock). `Get` acquires `c.mu.RLock()`, which is blocked for the full duration of the write lock. For a cache at `MaxSize` (default 4096) with uniformly-distributed TTLs (common when all tokens share the same issuer-configured lifetime), most entries expire in the middle or rear of the list rather than at the front — so the front-reclaim pass finds nothing and the O(n) sweep runs on **every** `Put` for the duration of the busy period.

In a gateway handling many concurrent sessions, `Put` is called once per token validation. The O(n) sweep under the write lock serializes all concurrent `Get` calls — which sit on the enforcement hot path — for the duration of the scan over 4096 list elements. Under load this can cause measurable latency spikes on the JWT validation path.

**Trigger:**

A deployment where:
1. The token cache is at or near `MaxSize` (default 4096), AND
2. Tokens have a uniform (or near-uniform) TTL — so expired entries are spread throughout the list, not concentrated at the front.

Under these conditions every `Put` call traverses the full list under the write lock.

**Fix:**

Move the O(n) sweep out of the write lock. One approach: run the full sweep under the read lock (collecting expired keys into a slice), then re-acquire the write lock only to delete the collected keys and check capacity. This mirrors the two-phase approach already used in `purgeExpired`:

```go
// Phase 1 (under RLock): collect expired keys from the full list
// Phase 2 (under Lock): delete collected keys, then evict oldest if still at cap
```

Alternatively, reduce the default `MaxSize` so the sweep is bounded by a smaller constant, or schedule the capacity-triggered sweep to the background cleanup goroutine rather than the hot `Put` path.

**Session:** rev-20260618T1610Z-ac1ef7

### LINKED ISSUE #819: [rev-20260618T1610Z-ac1ef7] JWTPDP.Decide: buildConstraintsFromClaims re-parses all capability claims on every request
**Location:** `cmd/eunox-mcp/pdp_jwt.go`, lines 1022–1062 (`buildConstraintsFromClaims`), called from `Decide` / `DecideResourceRead` / `DecidePromptGet` / `DecideSampling`

**Reference:** commit `51bf11221a216defe5b33f777eb9251b124eebd4`

**Severity:** Medium — CPU amplification under load

**Description:**

Every call to `JWTPDP.Decide` (and the resource/prompt/sampling variants) invokes `buildConstraintsFromClaims` unconditionally:

```go
func (p *JWTPDP) Decide(ctx context.Context, ...) (Decision, error) {
    payload, err := p.ValidateToken(ctx, token)
    // ...
    constraints, err := buildConstraintsFromClaims(payload.Capabilities, target)
    // ...
}
```

`buildConstraintsFromClaims` iterates over every claim key in `payload.Capabilities`, calls `splitV2Claim` (string split + field extraction) and `ParseTarget` (string parsing of `<type>:<name>`) on each one, and collects the matching constraints:

```go
for claimKey, claimVal := range caps {
    toolName, condSuffix, ok := splitV2Claim(claimKey)   // string ops
    // ...
    t, err := ParseTarget(toolName)                        // string parse
    // ...
}
```

For a JWT with 50 capability claims (not unusual for a broad-scope service account token), this is 50 × (string split + target parse) operations **per enforced MCP call**. The results are discarded at the end of each `Decide` invocation and recomputed from scratch on the next call.

By contrast, the list-filtering path (`FilterToolsList`, `FilterResourcesList`) uses `parseCapClaims(payload.Capabilities)` which parses all claims once per list response — not per entry in the filtered list.

There is no equivalent pre-parse caching on the per-request decision path. The `TokenCache` caches the `TokenPayload` (raw claims map), but not the parsed constraints derived from it, so the parsing cost is paid again on every `Decide` call even for a cache-hit token.

**Trigger:**

A gateway handling high-throughput `tools/call` traffic with JWT tokens carrying large capability sets (many `mcp.tool.*` claims). The CPU cost scales as O(C × R) where C is the number of claims per token and R is the request rate.

**Fix:**

Cache the parsed `[]capability.Constraint` slice keyed by `(tokenHash, targetKey)` or keyed by `tokenHash` (the full parsed-claims map). The `TokenPayload` cached in `TokenCache` could carry a pre-parsed constraints map as an unexported field populated on first `ValidateToken` and reused on subsequent cache hits. Alternatively, add a second in-process cache in `JWTPDP` mapping `tokenHash → parsedClaimsMap` with the same TTL as `TokenCache`.

A simpler incremental improvement: parse all claims once in `ValidateToken` (or on first `Decide`) and attach the result to `TokenPayload.parsedClaims`, so repeated `Decide` calls on the same token do not re-parse.

**Session:** rev-20260618T1610Z-ac1ef7

### LINKED ISSUE #820: [rev-20260618T1610Z-ac1ef7] openAuditSink: stat failure on newly-created empty file forces unnecessary rotation and may prune a valid older log
**Location:** `cmd/eunox-mcp/audit.go`, lines 420–426 (`openAuditSink`)

**Reference:** commit `51bf11221a216defe5b33f777eb9251b124eebd4`

**Severity:** Low — correctness / log retention

**Description:**

In `openAuditSink`, after opening or creating the audit log file, the code stats the file to determine how many bytes have already been written:

```go
fi, err := f.Stat()
if err != nil {
    // Cannot determine current size; force rotation now to get a clean file.
    written = rotateSizeBytes
}
```

Setting `written = rotateSizeBytes` is intended as a safe fallback that triggers immediate rotation when the current file size is unknowable. However, the stat is called immediately after `os.OpenFile` — including on the code path where the file was **just created** (O_CREATE) and is therefore empty. If `f.Stat()` fails transiently (e.g., due to a filesystem inconsistency, NFS delay, or a bug in a test harness), the newly-created empty file is immediately rotated:

1. The empty file is renamed to the rotated name (e.g., `audit-20260618T162000.jsonl`).
2. `pruneRotated` is called, which counts rotated files and deletes the oldest ones down to the `retain` limit.
3. The rotated **empty** file counts against the `retain` quota, and the oldest **valid** rotated log file is deleted to make room.

The net effect: a transient `Stat` failure on an empty new file causes permanent, silent deletion of a valid historical audit log. This violates the append-only audit invariant by destroying records.

**Trigger:**

A transient `f.Stat()` error during proxy startup when the audit log file is first created, on systems where the configured `retain` limit is already reached by existing rotated files.

**Fix:**

Distinguish the newly-created case. If the file was just created (size is known to be 0 from the O_CREATE path), set `written = 0` rather than `rotateSizeBytes`. The stat failure only matters for pre-existing files whose size is genuinely unknown:

```go
fi, err := f.Stat()
switch {
case err == nil:
    written = fi.Size()
case isNewFile: // flag set on the O_CREATE|O_EXCL path
    written = 0  // new file is empty; no rotation needed
default:
    written = rotateSizeBytes  // existing file, size unknown: rotate to be safe
}
```

Alternatively, always call `os.Stat(path)` (not `f.Stat()`) after `OpenFile` so the size is obtained via the path rather than the file descriptor, which is more robust to certain filesystem implementations.

**Session:** rev-20260618T1610Z-ac1ef7

### LINKED ISSUE #821: [rev-20260618T1610Z-ac1ef7] handleRecipientDomain: missing domain-syntax validation yields misleading denial reason for structurally invalid email addresses
**Location:** `pkg/enforcement/handlers.go`, lines 1035–1075 (`handleRecipientDomain`)

**Reference:** commit `51bf11221a216defe5b33f777eb9251b124eebd4`

**Severity:** Low — operator UX / defensive correctness

**Description:**

`handleRecipientDomain` validates that an email recipient's domain is in the configured allowlist. The check extracts the domain by splitting on `@`:

```go
if len(parts) != 2 {
    return deny("recipient missing @")
}
if !domainSet[domain] {
    return deny("domain not allowed: " + domain)
}
```

There is no validation that `parts[1]` (the domain portion) is a structurally valid domain name. Values like `"user@[192.168.1.1]"` (IP-literal syntax, RFC 5321 §4.1.3), `"user@.example.com"` (leading dot), `"user@exam ple.com"` (space), or `"user@exam*ple.com"` (wildcard character) all pass the `@`-split check and reach the `domainSet` lookup.

For most of these, `domainSet[domain]` returns `false` and the request is correctly denied — but the denial reason is `"domain not allowed: [192.168.1.1]"` rather than `"invalid email address"`. This creates ambiguity for operators reading audit logs: was the request denied because the domain is not in the allowlist, or because the address is syntactically malformed? Distinguishing these cases matters for incident response.

More subtly, an operator who adds `"[192.168.1.1]"` to the `allowedDomains` list to permit a specific relay would inadvertently allow IP-literal syntax while intending to permit only the literal string — the current code never rejects it as invalid, so it would silently match.

**Trigger:**

Any `recipientDomain` constraint evaluated against a tool that accepts email addresses where the upstream client sends malformed addresses (IP literals, addresses with spaces, wildcard characters, etc.).

**Fix:**

Add a lightweight domain-syntax check after the `@` split. A minimal check — no leading/trailing dot, no spaces, no `[` characters — is sufficient to catch the common structural cases without introducing a full RFC 5321 parser:

```go
if domain == "" || strings.ContainsAny(domain, " \t[]") ||
    strings.HasPrefix(domain, ".") || strings.HasSuffix(domain, ".") {
    return deny("invalid email address: malformed domain")
}
if !domainSet[domain] {
    return deny("domain not allowed: " + domain)
}
```

This preserves the fail-closed behavior (invalid → deny) while giving operators a correct and actionable denial reason.

**Session:** rev-20260618T1610Z-ac1ef7

### LINKED ISSUE #824: [rev-20260618T1610Z-ac1ef7] peekRetryAfterScript: O(count − limit) ZRANGEBYSCORE LIMIT pivot blocks Redis while incrIfBelowScript uses O(log N) ZRANGE by rank
**Location:** `pkg/callcounter/redis.go`, lines 360–376 (`peekRetryAfterScript`)

**Reference:** https://github.com/eunolabs/eunox/blob/51bf11221a216defe5b33f777eb9251b124eebd4/pkg/callcounter/redis.go#L360-L376

**Severity:** Medium — Redis-server-wide latency amplification under high-rate denied traffic

**Description:**

`peekRetryAfterScript` locates the retry-after pivot entry using `ZRANGEBYSCORE ... LIMIT idx 1`:

```lua
local count = redis.call('ZCOUNT', KEYS[1], '(' .. cutoff, '+inf')
local retry = 0
if limit >= 1 and count >= limit then
  local idx = count - limit
  local pivot = redis.call('ZRANGEBYSCORE', KEYS[1], '(' .. cutoff, '+inf', 'WITHSCORES', 'LIMIT', idx, 1)
```

`ZRANGEBYSCORE ... LIMIT offset count` has complexity O(log N + offset), because Redis must traverse `offset` in-window entries after finding the score boundary before returning the requested element. Here `offset = idx = count − limit`.

By contrast, `incrIfBelowScript` (the commit path, which prunes first) uses an O(log N) rank-based lookup:

```lua
redis.call('ZREMRANGEBYSCORE', KEYS[1], '-inf', ARGV[1])  -- prune; after this, all entries are in-window
local count = redis.call('ZCARD', KEYS[1])
...
local idx = count - limit
local pivot = redis.call('ZRANGE', KEYS[1], idx, idx, 'WITHSCORES')  -- O(log N), no offset scan
```

`ZRANGE` with explicit rank indices is O(log N + M) where M is the number of elements returned; for a single-element range `[idx, idx]` M=1, so the total cost is O(log N) regardless of `idx`.

**Practical impact:**

The `PeekRetryAfter` call sits *before* `IncrementIfBelow` on the check-only maxCalls path — it is called on every incoming request that will be denied. Under burst traffic against a tight quota (e.g. `maxCalls: count: 1, windowSeconds: 86400`), `count` grows to the number of calls already in the window and `limit = 1`, yielding `idx = count − 1`. Each `PeekRetryAfter` Redis call must traverse `count − 1` in-window entries to find the pivot. For `count = 10 000`, that is 9 999 skip-list node traversals inside the Lua script.

Because Redis Lua scripts run atomically (blocking all other Redis operations for their duration), a slow `peekRetryAfterScript` call directly adds latency to every other client sharing the Redis instance — including unrelated maxCalls and sequenceBlock operations across all routes in gateway mode. Under sustained denial traffic the amplification is O(count × denied-RPS) Redis-server-time per second.

**Trigger:**

Any `maxCalls` constraint with a tight limit against a high-request-rate client. The check-only pass is the `PeekRetryAfter` call in `handleMaxCalls`; it is invoked before the commit (`IncrementIfBelow`) on every evaluation.

**Suggested fix:**

Decompose the in-window entry count into (total − expired), then use `ZRANGE` with an explicit absolute rank — eliminating the offset traversal:

```lua
local count = redis.call('ZCOUNT', KEYS[1], '(' .. cutoff, '+inf')
local retry = 0
if limit >= 1 and count >= limit then
  local total   = redis.call('ZCARD', KEYS[1])          -- O(1)
  local expired = total - count                          -- entries below cutoff
  local idx     = count - limit
  local rank    = expired + idx                          -- absolute rank in full set
  local pivot   = redis.call('ZRANGE', KEYS[1], rank, rank, 'WITHSCORES')  -- O(log N)
  if pivot[2] then
    retry = (tonumber(pivot[2]) + window) - now
    if retry < 0 then retry = 0 end
  end
end
return {count, retry}
```

`ZCARD` is O(1). `ZRANGE [rank, rank]` is O(log N) regardless of `rank`. The three Redis calls (ZCOUNT + ZCARD + ZRANGE) execute atomically inside the script and collapse to a single round-trip, so network overhead is unchanged. The retry-after estimate is identical to the current implementation.

This brings `peekRetryAfterScript`'s pivot cost in line with `incrIfBelowScript`, eliminating the O(count − limit) traversal that currently blocks the Redis server under high denial load.

**Session:** rev-20260618T1610Z-ac1ef7

### LINKED ISSUE #825: [rev-20260618T1710Z-28c9ad] JWT segment count guard checks < 2 instead of < 3, allowing unsigned 2-part tokens through defensive check
**Session:** rev-20260618T1710Z-28c9ad
**Reviewed at commit:** 135f5f83904f9dc8c4a66d5b53cc9b6167d09265

## Location

`cmd/eunox-mcp/pdp_jwt.go`, function `decodeJWTClaimsPreservingNumbers`, line 344

## Severity

Medium — not currently exploitable (go-jose validates the token before this function is called), but wrong defensive check in a security-critical code path with a wrong error message.

## Description

`decodeJWTClaimsPreservingNumbers` re-decodes the JWT payload segment with `json.Number` precision to preserve large integer claims for OPA/Cedar policy evaluators. It splits the token string on `.` and guards against malformed input:

```go
parts := strings.Split(tokenStr, ".")
if len(parts) < 2 {
    return nil, fmt.Errorf("malformed JWT: expected at least 2 segments, got %d", len(parts))
}
payloadBytes, err := base64.RawURLEncoding.DecodeString(parts[1])
```

A valid JWS compact serialization (JWT) has exactly **3** segments: `header.payload.signature`. The guard `len(parts) < 2` passes any string with 2 or more parts, including a 2-segment `"header.payload"` string that has no signature component. The error message compounds the error by saying "expected at least 2 segments."

### Why this matters

1. **Wrong semantics:** A 2-segment string (no signature) passes the guard and its second component is decoded as claims without any indication that the signature is missing.
2. **Wrong error message:** A 1-segment string produces `"malformed JWT: expected at least 2 segments, got 1"`, which is confusing — valid JWTs need 3 segments, not 2.
3. **Latent vulnerability:** The function is currently called only after `jwt.ParseSigned` + `tok.Claims(key, ...)` fully validate the signature, so the wrong guard is never the first line of defence in production. However, `decodeJWTClaimsPreservingNumbers` has no package-private visibility fence; if a future refactoring calls it before signature verification (e.g., for logging, metrics, or a new code path), the check `< 2` would silently accept unsigned or tampered tokens.

## Trigger

```go
claims, err := decodeJWTClaimsPreservingNumbers("eyJhbGciOiJub25lIn0.eyJzdWIiOiJhdHRhY2tlciJ9")
// len(parts) == 2 — passes the < 2 guard; claims decoded without a signature segment
```

In production this string would be rejected by `jwt.ParseSigned` first, but the guard in `decodeJWTClaimsPreservingNumbers` itself offers no protection against it.

## Fix

```go
if len(parts) < 3 {
    return nil, fmt.Errorf("malformed JWT: expected 3 segments (header.payload.signature), got %d", len(parts))
}
```

No other change needed; `parts[1]` is still the payload segment.

### LINKED ISSUE #826: [rev-20260618T1710Z-28c9ad] kidRecentlyAbsent acquires write lock for a read-only hot path, serializing all concurrent JWT validations under the anti-flood sentinel
**Session:** rev-20260618T1710Z-28c9ad
**Reviewed at commit:** 135f5f83904f9dc8c4a66d5b53cc9b6167d09265

## Location

`pkg/capability/jwkscache.go`, function `kidRecentlyAbsent`, lines 376-388

## Severity

Medium — correctness is unaffected, but the anti-flood protection mechanism serializes concurrent JWT validations under the exact attack conditions it is designed to mitigate, creating a secondary DoS amplification vector.

## Description

`negMu` is declared as `sync.RWMutex`, signalling intent to permit concurrent reads of `negKIDs`. However, `kidRecentlyAbsent` always acquires the exclusive write lock:

```go
func (c *JWKSCache) kidRecentlyAbsent(kid string) bool {
    c.negMu.Lock()          // exclusive write lock — blocks all concurrent readers
    defer c.negMu.Unlock()
    at, ok := c.negKIDs[kid]
    if !ok {
        return false
    }
    if c.now().Sub(at) >= negativeKIDTTL {
        delete(c.negKIDs, kid)  // write only on this rare expiry branch
        return false
    }
    return true             // hot path: pure read, no mutation
}
```

The **hot path** (entry present and not expired — `return true`) performs only a map read, yet holds an exclusive lock for its entire duration. The only mutation (`delete`) happens on the rare cold path when an entry is found but expired.

### Impact under a KID-flood attack

The shared sentinel (`sharedRefreshSentinel = ""`) is the primary rate-limit for distinct-unknown-kid floods: once a forced JWKS fetch returns an unchanged key set, the sentinel is marked, and every subsequent `ForceRefreshForKID` call checks `kidRecentlyAbsent(sharedRefreshSentinel)` before performing an expensive HTTP round-trip.

Under the attack this sentinel is designed to absorb:

1. N concurrent goroutines each call `ForceRefreshForKID` with a distinct unknown KID.
2. After the first forced fetch, the sentinel is set.
3. Every subsequent goroutine calls `kidRecentlyAbsent("")` — which acquires `negMu.Lock()`.
4. All N goroutines queue on the exclusive write lock even though the sentinel check is a pure read.

The protection that was supposed to make flood absorption *cheap* (return cached set immediately) instead serializes all N concurrent callers. The serialization adds O(N) queuing latency at exactly the point where latency matters most.

`markKIDAbsent` also holds `negMu.Lock()` while sweeping up to `negativeKIDMaxLen` (1024) entries for expiry. A sweep running in parallel with many `kidRecentlyAbsent` calls extends the serialisation window further.

## Fix

Use `negMu.RLock()` for the common case (entry present and valid) and upgrade to `negMu.Lock()` only when a deletion is needed. Go's `sync.RWMutex` does not support atomic lock upgrade, so re-check after acquiring the write lock:

```go
func (c *JWKSCache) kidRecentlyAbsent(kid string) bool {
    c.negMu.RLock()
    at, ok := c.negKIDs[kid]
    if !ok {
        c.negMu.RUnlock()
        return false
    }
    if c.now().Sub(at) < negativeKIDTTL {
        c.negMu.RUnlock()
        return true // hot path: concurrent reads allowed
    }
    c.negMu.RUnlock()

    // Entry is expired: acquire write lock, re-check, then delete.
    c.negMu.Lock()
    defer c.negMu.Unlock()
    if at2, ok2 := c.negKIDs[kid]; ok2 && c.now().Sub(at2) >= negativeKIDTTL {
        delete(c.negKIDs, kid)
    }
    return false
}
```

This allows N concurrent callers to read the sentinel in parallel with zero write-lock contention, which is the intended behaviour of `negMu` being a `sync.RWMutex`.

### LINKED ISSUE #827: [rev-20260618T1810Z-60f7cb] token_cache: Get() expires at boundary (now >= expiresAt) but eviction uses strictly-after (now > expiresAt), causing live-entry eviction when cache is full
**Session:** rev-20260618T1810Z-60f7cb
**Reviewed at commit:** 135f5f83904f9dc8c4a66d5b53cc9b6167d09265

## Location

`pkg/capability/token_cache.go`

- `Get()` — line 193
- `Put()` front-scan eviction — line 279
- `Put()` O(n) sweep eviction — line 298
- `purgeExpired()` Phase 1 — line 385

## Reference

https://github.com/eunolabs/eunox/blob/135f5f83904f9dc8c4a66d5b53cc9b6167d09265/pkg/capability/token_cache.go#L193

## Severity

Medium — no security bypass (expired tokens are never served), but at the exact expiry nanosecond the cache underestimates available capacity and prematurely evicts a live entry when at `MaxSize`.

## Description

`Get()` treats the exact expiry instant as expired:

```go
// line 193 — expired when now >= expiresAt (inclusive boundary)
if !now.Before(entry.expiresAt) {
    // "The check is inclusive of the expiry instant (now >= expiresAt, not now > expiresAt)"
    return nil, false
}
```

`Put()`'s front-scan, `Put()`'s O(n) sweep, and `purgeExpired()` all treat the exact expiry instant as **live** (strictly-after semantics):

```go
// Put() front-scan, line 279 — stops/treats as live when now <= expiresAt
if !ok || !now.After(e.expiresAt) {
    break // "front is live (or already gone): stop reclaiming"
}

// Put() O(n) sweep, line 298 — evicts only when now > expiresAt
if e, ok := c.entries[ek]; !ok || now.After(e.expiresAt) {
    ...delete...
}

// purgeExpired() Phase 1, line 385 — collects only when now > expiresAt
    expired = append(expired, key)
}
```

At the exact nanosecond `now == expiresAt`:

- `Get()` returns a miss (correctly — the token is expired per the inclusive rule and the RFC 7519 §4.1.4 rationale documented in the comment).
- The eviction scans in `Put()` and `purgeExpired()` skip the entry because `now.After(entry.expiresAt)` is `false`.

The `Get()` comment says "background cleanup will remove it" — this is inaccurate for boundary entries: `purgeExpired()` will also skip them until `now` advances past `expiresAt` into strictly-after territory.

## Trigger / Repro

When the cache is at `MaxSize` and one or more entries have `expiresAt == now`:

1. `Put()` runs the front-scan and O(n) sweep to reclaim expired entries.
2. Both passes skip the boundary-expired entries (`now.After(entry.expiresAt)` is `false`).
3. The "still at capacity" fallback at line 307 evicts the oldest **live** entry to make room.
4. A valid, serviceable token is dropped from the cache, forcing an unnecessary re-verification on its next use.

In practice this requires sub-nanosecond clock precision colliding with an exact expiry; it is rare in production but reproducible with a stubbed `Now()` clock in tests.

## Suggested Fix

Make all four expiry checks agree on inclusive-boundary semantics by replacing the strictly-after comparisons in the eviction paths with `!now.Before(...)`:

```go
// Put() front-scan (line 279): stop when entry is live (now < expiresAt)
if !ok || !now.Before(e.expiresAt) {  // was: !now.After(e.expiresAt)
    break
}

// Put() O(n) sweep (line 298): evict when now >= expiresAt
if e, ok := c.entries[ek]; !ok || !now.Before(e.expiresAt) {  // was: now.After(e.expiresAt)
    ...delete...
}

// purgeExpired() Phase 1 (line 385): collect when now >= expiresAt
if !now.Before(entry.expiresAt) {  // was: now.After(entry.expiresAt)
    expired = append(expired, key)
}
```

This makes the eviction paths consistent with `Get()`'s existing `!now.Before(entry.expiresAt)` check and the comment that explains the inclusive-boundary choice (RFC 7519 §4.1.4 / fail-closed with coarse clocks).

The `purgeExpired()` Phase 2 re-check (not shown above) should receive the same fix for the same reason.

## Session

rev-20260618T1810Z-60f7cb

### LINKED ISSUE #828: [rev-20260618T1806Z-9b543c] PeekRetryAfter: int64 capacity arg to make() panics on 32-bit when limit > MaxInt32
**Session:** rev-20260618T1806Z-9b543c
**Reviewed at commit:** 135f5f83904f9dc8c4a66d5b53cc9b6167d09265

## Location

`pkg/callcounter/memory.go`, function `PeekRetryAfter`, line 485

## Reference

https://github.com/eunolabs/eunox/blob/135f5f83904f9dc8c4a66d5b53cc9b6167d09265/pkg/callcounter/memory.go#L485

## Severity

Low — triggers only on 32-bit target platforms (GOARCH=386, arm) with `maxCalls` limits above `math.MaxInt32` (~2.1 × 10⁹). 64-bit production deployments are unaffected.

## Description

`PeekRetryAfter` collects the current window's timestamps into a slice to find the earliest slot that would free under the limit. The slice is pre-allocated with `cur` (type `int64`) as the capacity:

```go
// memory.go line 485
valid := make([]time.Time, 0, cur)  // cur is int64
```

In Go, the capacity argument to `make` is of type `int`. On a 32-bit platform `int` is 32 bits wide. An `int64` value above `math.MaxInt32` (~2.1 × 10⁹) is silently truncated on the implicit conversion, potentially producing a **negative** capacity, which causes a runtime panic:

```
runtime: makeslice: cap out of range
goroutine N [running]:
panic(interface {}(0x...))
```

### Why a large `cur` is reachable

`MaxLimit` is defined as `1 << 53` (~9 × 10¹⁵) — far above `math.MaxInt32`. Unlike `MaxWindowSeconds`, which is capped at `365 * 24 * 3600` with an explicit `math.MaxInt` overflow note in the adjacent comment, `checkLimit` has no corresponding 32-bit guard:

```go
// callcounter.go
const MaxLimit = 1 << 53  // no math.MaxInt cap

func checkLimit(limit int64) error {
    if limit < 1 || limit > MaxLimit {
        return fmt.Errorf(...)
    }
    return nil  // passes limits up to 1<<53 — far above MaxInt32 on 32-bit
}
```

A manifest with `maxCalls: 3000000000` (3 × 10⁹, which is > MaxInt32 but ≤ MaxLimit) passes validation successfully. On a 32-bit host, once the in-memory window accumulates enough entries and `cur` is computed above `math.MaxInt32`, the `make` call panics.

## Trigger / Repro

1. Build for a 32-bit target: `GOARCH=386 go build ./...`
2. Configure a capability constraint with `maxCalls: 3000000000` (any value > 2 147 483 647).
3. Drive enough calls within the window to fill `cur` beyond `math.MaxInt32`.
4. Any subsequent call that invokes `PeekRetryAfter` panics with `makeslice: cap out of range`.

## Suggested Fix

**Option A** — cap `cur` immediately before `make` (surgical fix, localised to the panic site):

```go
capVal := cur
if capVal > math.MaxInt {
    capVal = math.MaxInt
}
valid := make([]time.Time, 0, int(capVal))
```

**Option B** — add a platform-aware guard to `checkLimit` (analogous to `MaxWindowSeconds`), so manifests with oversized limits are rejected at validation time rather than panicking at runtime:

```go
func checkLimit(limit int64) error {
    if limit < 1 || limit > MaxLimit || int64(int(limit)) != limit {
        return fmt.Errorf("limit %d is out of range [1, %d]", limit, MaxLimit)
    }
    return nil
}
```

Option A is the safer surgical fix; Option B closes the gap at the manifest-validation boundary for all future callers and is consistent with how `MaxWindowSeconds` protects against the same class of overflow.

## Session

rev-20260618T1806Z-9b543c

### LINKED ISSUE #829: [rev-20260618T1821Z-af4249] WithMaxKeys godoc and threat model claim sequenceBlock recording fails OPEN at the key ceiling, but recordSessionCall fails CLOSED — threat model self-contradicts
**Location:**
- `pkg/callcounter/memory.go:79-98` (`WithMaxKeys` godoc, the "best-effort sequenceBlock recorder swallows it" claim)
- `docs/threat-model-mcp.md:287` ("A refused `seq:` key fails *open*: `recordSessionCall` swallows the error by design")

contradicted by:
- `pkg/enforcement/engine.go:384-399` and `:727-729` / `:925-927` (actual fail-closed behavior)
- `docs/threat-model-mcp.md:204` (correctly says recording errors are "not swallowed")

**Reference:** https://github.com/eunolabs/eunox/blob/135f5f83904f9dc8c4a66d5b53cc9b6167d09265/pkg/callcounter/memory.go#L79-L98

**Severity:** Medium — the threat model (a load-bearing artifact per CONTRIBUTING) is internally self-contradictory and documents a `sequenceBlock` bypass that does not actually exist in the shipped code; it also hides the real availability consequence of `--max-call-counter-keys`.

**Description:**

There are two mutually exclusive descriptions of what happens when an InMemory counter with `WithMaxKeys` set hits its key ceiling during `sequenceBlock` antecedent recording.

The `WithMaxKeys` godoc and threat-model §287 say it fails **OPEN** (best-effort swallow):

> `pkg/callcounter/memory.go` — "...the best-effort sequenceBlock recorder swallows it and skips the antecedent marker — so a later sequenceBlock check cannot see the prior call and does not block the dependent tool, i.e. fail-open for sequence ordering at saturation."

> `docs/threat-model-mcp.md:287` — "A refused `seq:` key fails *open*: `recordSessionCall` swallows the error by design (best-effort, `engine.go`)... a `sequenceBlock(afterTools: [A])` guard on B is bypassable under sustained saturation..."

But the actual code does **not** swallow. `recordSessionCall` returns the counter error (`engine.go:450-451`), and both `ValidateAction` (`engine.go:727-729`) and `EvaluateConditions` (`engine.go:925-927`) turn that into a hard `recordFailureDenial` — a fail-**closed** DENY. The `recordSessionCall` contract comment itself (`engine.go:384-399`) states this is deliberate:

> "...callers MUST treat that as a fail-closed deny... A write fault here therefore denies every allowed call in a session carrying a sessionId while the backend is faulting... the deliberate cost of a single, consistent fail-closed rule."

The same threat-model document even states the correct (fail-closed) behavior one section earlier:

> `docs/threat-model-mcp.md:204` — "an error while *recording* an allowed antecedent is **not swallowed**... the engine surfaces the write error and denies the call that triggered it with `CONDITION_FAILED`... a counter-backend fault denies every allowed call in the session for its duration."

So §204 and §287 of the same file directly contradict each other, and §287 + the `WithMaxKeys` godoc contradict the code.

The only place that actually swallows is the **audit-mode** path (`cmd/eunox-mcp/pdp.go:528`), which logs a warning and continues because audit mode forwards regardless. The stale docs appear to generalize that audit-mode behavior (or a pre-#213 design) to the enforce path.

**Why it matters:**
1. §287 claims a concrete, exploitable `sequenceBlock` bypass ("flood the map to its ceiling and the victim session's A call records no marker, so B is admitted"). In the shipped code the antecedent call A is **denied** at saturation, so A never runs — the described bypass does not exist. A security reader either wastes effort chasing a non-bug or wrongly concludes the guard is unsafe.
2. The godoc points operators in the wrong direction about the real trade-off: enabling `--max-call-counter-keys` (default 1,000,000) does **not** silently skip markers at the ceiling — it **denies every allowed call** in any session that needs a new history key while the map is full (an availability footgun under session churn, exactly what `engine.go:384-399` documents).

**Trigger / repro:** Configure an InMemory counter with `WithMaxKeys(n)`, fill the map to `n` distinct keys, then issue an allowed call in a new session (needs a new `seq:` key). Observe: the call is DENIED with `CONDITION_FAILED` / `phase: record` — not allowed-with-skipped-marker as the godoc and threat-model §287 describe. A unit test driving `recordSessionCall` against a saturated `WithMaxKeys` counter confirms the deny.

**Suggested fix:** The code (fail-closed) is intentional per `engine.go`; fix the stale documentation:
- Rewrite the `WithMaxKeys` godoc (`memory.go:79-98`) so the seq: ceiling path is described as fail-**closed** (the antecedent call is denied, consistent with `recordSessionCall`), not "best-effort swallow / fail-open".
- Rewrite `docs/threat-model-mcp.md:287` to match §204 and the code: at the key ceiling the antecedent call is denied (fail-closed), so the §287 "sequenceBlock bypassable under saturation" claim should be removed/corrected. Note the real consequence is an availability impact (denials under high session churn), and reconcile §204 and §287 so the document no longer contradicts itself.
- Keep the audit-mode swallow (`pdp.go:528`) as the one explicitly best-effort path and call it out as such.

**Session:** rev-20260618T1821Z-af4249

### LINKED ISSUE #830: [rev-20260618T1821Z-af4249] drift.go file-header comment misstates which drift findings abort startup: omits FM-4 (strict-fatal) and FM-5 (unconditionally fatal)
**Location:** `cmd/eunox-mcp/drift.go:14-16` (package/file header comment)

**Reference:** https://github.com/eunolabs/eunox/blob/135f5f83904f9dc8c4a66d5b53cc9b6167d09265/cmd/eunox-mcp/drift.go#L1-L17

**Severity:** Low — documentation, but the inaccurate text concerns the startup-abort safety mechanism for a security feature.

**Description:**

The file header summarizes the abort behavior of the drift check:

```go
// In non-strict mode all findings are emitted as structured log lines to
// stderr and the session continues.  With --strict-drift, FM-1 and FM-2
// findings abort session establishment.
```

Both sentences are inaccurate relative to the actual code in the same file:

1. **"In non-strict mode ... the session continues"** is false for FM-5. `hasCriticalDrift` (`drift.go:293-300`) returns true on any FM-5 finding, and `evaluateDrift` (`drift.go:336-338`) aborts on FM-5 **regardless of `--strict-drift`**:
   ```go
   if hasCriticalDrift(warnings) {
       return fmt.Errorf("startup aborted: description hash mismatch detected ...")
   }
   ```
   The `DriftFM5` doc (`drift.go:73-76`) confirms it is "always fatal (aborts startup regardless of --strict-drift)". So in non-strict mode an FM-5 finding does **not** let the session continue.

2. **"With --strict-drift, FM-1 and FM-2 findings abort"** omits FM-4. `IsFatal()` (`drift.go:96-98`) returns true for FM-1, FM-2, **and FM-4**, and `evaluateDrift` (`drift.go:339-341`) aborts under `--strict-drift` whenever `hasFatalDrift` is true. The `IsFatal` doc itself says "FM-1, FM-2, and FM-4 are fatal". So FM-4 also aborts under `--strict-drift`, contrary to the header.

The header comment was evidently written when only FM-1/FM-2/FM-3 existed and was not updated when FM-4 and FM-5 were added (their `DriftKind` constants and the `IsFatal`/`hasCriticalDrift` logic all post-date it).

**Trigger / repro:** Read the file header and compare against `IsFatal`, `hasFatalDrift`, `hasCriticalDrift`, and `evaluateDrift`. A maintainer relying on the header would wrongly believe (a) nothing aborts in non-strict mode and (b) only FM-1/FM-2 abort under `--strict-drift` — both wrong for the security-relevant abort path.

**Suggested fix:** Update the header to match the implemented policy, e.g.:

```go
// In non-strict mode, findings are emitted to stderr and the session
// continues — EXCEPT FM-5 (description-hash mismatch), which aborts startup
// unconditionally. With --strict-drift, FM-1, FM-2, and FM-4 findings also
// abort session establishment. FM-3 and "uncovered" are always advisory.
```

(No code change needed; the behavior is correct — only the comment is stale.)

**Session:** rev-20260618T1821Z-af4249

### LINKED ISSUE #831: [rev-20260618T1821Z-b22270] `suggest` mines the `_eunox_truncated` audit marker as a real tool argument, emitting an allowedValues condition that denies every call
**Location:** `cmd/eunox-mcp/suggest.go:207-221` (`mineArgs`), interacting with `cmd/eunox-mcp/audit.go:984-1021` (`boundAuditDetails` / `auditTruncatedKey`)

**Reference:** https://github.com/eunolabs/eunox/blob/135f5f83904f9dc8c4a66d5b53cc9b6167d09265/cmd/eunox-mcp/suggest.go#L207-L221

**Severity:** Medium

**Description:**

`suggest` builds draft `allowedValues` conditions by mining the `details` map of every allowed `tools/call` audit record (`mineArgs`), treating each key as a tool argument name and each string value as an observed value. Its core design invariant (stated repeatedly in `writeTargetEntry`) is that it must never emit a condition that would reject a call the tape recorded as allowed.

That invariant is violated when the audit log truncated the record. When an allowed `tools/call`'s argument map serializes to more than the 1 MiB detail cap, `boundAuditDetails` (audit.go) replaces the **entire** `details` map with a single sentinel:

```go
return map[string]interface{}{
    auditTruncatedKey: fmt.Sprintf("details omitted: %d-byte map exceeds the %d-byte audit record detail cap", len(encoded), auditDetailsTotalCap),
}
```

where `auditTruncatedKey == "_eunox_truncated"`. The audit code's own comment says the underscore prefix is meant to "keep it clear of real tool-argument names" — but the `suggest` consumer has no awareness of this marker.

`mineArgs` therefore records `_eunox_truncated` as a regular argument with a single string value (the truncation message). Walk it through `writeTargetEntry`:

- `a.calls == t.allow` (present on every observed call, if it's the affected tool's only allow record): passes the "optional argument" guard.
- `a.nonString == false`, `len(vals) == 1`: passes the type/size guards.
- `valuesSelfMatch` is **true** — the message contains no glob metacharacters, so `MatchValueGlob(msg, msg)` matches literally.

So it falls through to the `default` arm and emits an **active** condition:

```yaml
- target: tool:<name>
  actions: [call]
  conditions:
    - type: allowedValues
      argument: "_eunox_truncated"
      values:
        - "details omitted: 1500000-byte map exceeds the 1048576-byte audit record detail cap"
```

`_eunox_truncated` is a phantom argument the real tool never carries. Once this draft is enforced, `ResolveArgument(req.Arguments, "_eunox_truncated")` finds nothing, the condition fails `MISSING_CONTEXT`, and **every** call to that tool is denied — the exact "deny a call the tape showed as allowed" outcome the module promises to avoid.

(Note: the per-value placeholder for an over-512KiB single string — `[eunox: omitted N-byte value ...]` — is incidentally caught by the `valuesSelfMatch` glob-metacharacter guard because it begins with `[`, so it degrades to a review comment. The total-cap `_eunox_truncated` marker is the one that slips through as an active condition.)

**Trigger / repro:**
1. Run the proxy in wiretap mode: `eunox-mcp proxy --audit ...`.
2. Have the agent invoke a tool with a total argument map larger than the 1 MiB audit detail cap (e.g. a large batch payload or a big structured argument). The allow record's `details` becomes `{"_eunox_truncated": "details omitted: ..."}`.
3. Run `eunox-mcp suggest` over that tape. The draft for that tool contains an active `allowedValues` condition on argument `_eunox_truncated`.
4. Enforcing the draft denies all calls to that tool.

**Suggested fix:**
In `mineArgs` (or `computeSuggestions`), recognize the audit truncation marker and never treat it as a tool argument. Concretely, skip a `details` map whose sole key is `auditTruncatedKey` (and skip any individual `auditTruncatedKey` key), and instead record a flag on the `observedTarget` so `writeTargetEntry` emits a review note such as `# arguments were truncated in the audit log for some calls — values could not be mined; constrain manually`. Reusing the exported `auditTruncatedKey` constant keeps the two sides in lock-step if the marker name ever changes.

**Session:** rev-20260618T1821Z-b22270

eunolabs:
This PR fixes every open automated-review issue from the 2026-06-18 review sessions, one commit per bug. All were reviewed against the current HEAD commit `135f5f8`.

### Session rev-20260618T1610Z-ac1ef7
- Closes #816 — `allowedExtensions` matched via OS-specific `filepath.Base/Ext`, so on Linux a Windows-style `dir\file` path was matched as one element. Now normalizes `\`→`/` and matches via the OS-independent `path` package, so the directory component can't be smuggled into the matched file name.
- Closes #817 — `interpretAuditTail` processed a partial `ReadAt` (`0 < n < len(buf)`, `io.EOF`) as a valid tail, silently masking a file shrink. Since the buffer is sized to exactly the tail window, a short read can only be a truncation, so it now fails closed with `errAuditFileShrunk` (the chain-resume path then writes a signed marker). Updates the superseded #449 test.
- Closes #818 — token-cache `Put` ran its O(n) expired-entry sweep under the write lock, stalling concurrent `Get`. The expensive identification now happens under a read lock (gated to the full-cache/live-front case the old sweep ran in); the write lock only deletes via O(1) list pointers, with the O(n) sweep kept as a correctness fallback for the rare inter-phase fill race.
- Closes #819 — `JWTPDP.Decide` re-parsed every capability claim head on every request. The cheap-phase parse is now memoized on the validated `JWTClaims` (cached across requests), so repeated decisions skip it; condition parsing stays lazy (#575 preserved).
- Closes #820 — `openAuditSink` forced rotation when `f.Stat` failed, which on a newly-created empty log could prune a valid older rotated log. Now stats the path before `O_CREATE` to distinguish a brand-new (empty → counter 0) file from a pre-existing one of unknown size.
- Closes #821 — `recipientDomain` gave a misleading "domain not allowed" reason for structurally invalid addresses. Now rejects IP-literals (`[...]`) and leading/trailing-dot domains with a distinct "invalid recipient email" reason (fail closed), so an IP-literal can never silently match an allowlist entry.
- Closes #824 — `peekRetryAfterScript` located the retry-after pivot with `ZRANGEBYSCORE ... LIMIT offset`, an O(log N + offset) scan that blocked Redis under denial bursts. Now selects by absolute rank with `ZCARD` + `ZRANGE rank rank` (O(log N)), matching the commit path.

### Session rev-20260618T1710Z-28c9ad
- Closes #825 — `decodeJWTClaimsPreservingNumbers` guarded `len(parts) < 2`, accepting an unsigned 2-segment token. Now requires the 3 JWS segments (fail closed) with a corrected message.
- Closes #826 — `kidRecentlyAbsent` took the exclusive lock on its read-only hot path, serializing JWKS anti-flood sentinel checks under the very flood they mitigate. `negMu` is now an `RWMutex`; the hot path read-locks and only upgrades (with a re-check) on the rare expiry-prune branch.

### Session rev-20260618T1806Z-9b543c
- Closes #828 — `PeekRetryAfter` passed an `int64` count as `make`'s capacity, which truncates to a negative `int` and panics on 32-bit targets when `maxCalls > MaxInt32`. The capacity hint is now capped at `math.MaxInt` (count is unaffected; `append` still grows).

### Session rev-20260618T1810Z-60f7cb
- Closes #827 — token-cache eviction used strictly-after (`now > expiresAt`) while `Get` expires inclusively (`now >= expiresAt`), so a boundary-expired entry was skipped by eviction and a live entry was evicted instead. All four eviction checks now use the inclusive boundary.

### Session rev-20260618T1821Z-af4249
- Closes #829 — `WithMaxKeys` godoc and threat-model §287 claimed the `seq:` key ceiling fails *open* (a documented `sequenceBlock` bypass), contradicting the code (`recordSessionCall` surfaces the error → fail-closed deny) and §204. Docs now describe the fail-closed behavior and the real availability cost, and reconcile §287 with §204.
- Closes #830 — `drift.go` file header omitted FM-4 (strict-fatal) and FM-5 (unconditionally fatal). Header now matches `IsFatal`/`hasCriticalDrift`. (Comment-only; behavior was already correct.)

### Session rev-20260618T1821Z-b22270
- Closes #831 — `suggest` mined the `_eunox_truncated` audit marker as a real argument, emitting an `allowedValues` condition on a phantom argument that would deny every call to the tool. `mineArgs` now skips the marker (reusing `auditTruncatedKey`) and flags the target so the draft carries a review note instead.

## Testing
- `go vet ./...`, `make lint` (0 issues), `make check-license`, `gofmt` — all clean.
- Added targeted regression tests: `allowedExtensions` backslash handling (#816), `recipientDomain` malformed-domain reasons (#821), audit-tail partial-shrink (#817), `suggest` truncation marker (#831), token-cache boundary-expiry and multi-expired-behind-live-front eviction (#827, #818). Existing JWKS/JWT/token-cache/redis/audit suites pass under `-race`.

https://claude.ai/code/session_018gcnvNkPFXQ8zqAskiVbdL

---
_Generated by [Claude Code](https://claude.ai/code/session_018gcnvNkPFXQ8zqAskiVbdL)_


pkg/callcounter/redis.go
```diff
@@ -365,8 +365,21 @@ local window = tonumber(ARGV[4])
...
+  -- blocks the whole single-threaded Redis server for that scan on every denied
+  -- request. The expired entries (score <= cutoff) occupy the lowest ranks, so
+  -- the in-window set starts at absolute rank (total - count); the (count-limit)th
+  -- in-window entry, oldest-first, is at rank (total - count) + (count - limit).
+  -- ZCARD is O(1) and ZRANGE on a single explicit rank is O(log N) regardless of
+  -- the rank, matching incrIfBelowScript's commit-path cost (#824).
+  local total   = redis.call('ZCARD', KEYS[1])
+  local expired = total - count
+  local idx     = count - limit
+  local rank    = expired + idx
```
eunolabs, line 381: **#824 — rank math is correct, but the change isn't pinned by a test.**

The reasoning here is sound: the commit path (`incrIfBelowScript`) runs `ZREMRANGEBYSCORE` first, so its `idx = count - limit` indexes an all-in-window set, whereas this read-only script never prunes — so expired members linger at ranks `[0, expired-1]` and the `expired = total - count` offset is exactly what skips past them. (`rank` here simplifies to `total - limit`, for what it's worth.)

The gap is coverage: `TestRedis_PeekRetryAfter_MatchesIncrementIfBelow` inserts 6 entries that are all still in-window when `PeekRetryAfter` runs (at T=55 the cutoff is -5s), so `expired == 0` and `rank == idx`. That test would still pass if the offset were dropped — i.e. it doesn't lock in the one behavior this commit introduces.

Consider a regression test where stale (pre-cutoff) members are still present in the set at peek time: insert entries, advance the clock past the window **without** calling `IncrementIfBelow` (which would prune them), then `PeekRetryAfter` and assert the retry tracks the correct in-window pivot despite the low-rank expired entries. That's the case that distinguishes this fix from the old `ZRANGEBYSCORE ... LIMIT` form.

---

cmd/eunox-mcp/pdp_jwt.go
```diff
@@ -341,8 +350,14 @@ func (p *JWTPDP) innerEnforces() bool {
...
 	parts := strings.Split(tokenStr, ".")
-	if len(parts) < 2 {
-		return nil, fmt.Errorf("malformed JWT: expected at least 2 segments, got %d", len(parts))
+	// A JWS compact serialization has exactly three segments
+	// (header.payload.signature). A 2-segment "header.payload" string carries no
+	// signature, so it must be rejected here too: the caller verifies the
+	// signature before reaching this function today, but this defensive guard
+	// must not silently accept an unsigned token if a future caller invokes it
+	// before verification (fail closed).
+	if len(parts) < 3 {
```
eunolabs: Nit / defense-in-depth: this correctly rejects the unsigned 2-segment token, but `< 3` still admits a 4+ segment string while the message says `expected 3 segments (header.payload.signature)`. Since the guard exists precisely for a hypothetical future caller that reaches here *before* signature verification, `!= 3` would match the message and also reject malformed multi-segment input. Minor — the signing check is the real gate today.

Separately: I don't see a unit test pinning this rejection (the PR adds targeted tests for #816/#817/#821/#827/#818/#831 but not #825). A two-line table test (`"a.b"` → error, `"a.b.c"` → ok) would guard against a future regression back to `< 2`.

---

cmd/eunox-mcp/audit.go
```diff
@@ -409,23 +409,38 @@ func openAuditSink(logPath, keyPath string, rotateSizeBytes int64, retainRotated
...
-	// rotation by a full rotateSizeBytes (the file could reach 2x the bound).
-	// Fall back to the conservative rotateSizeBytes so the next write rotates
-	// immediately, and warn.
+	// existing log tail.
 	var written int64
-	if info, statErr := f.Stat(); statErr != nil {
+	switch info, statErr := f.Stat(); {
+	case statErr == nil:
+		written = info.Size()
+	case preexisted:
```
eunolabs: #820 reads cleanly and fixes the common case. One residual edge in the same class as the bug being fixed: a *pre-existing but empty* log whose `f.Stat()` fails lands in `case preexisted:` and forces rotation (`written = rotateSizeBytes`), which can still rename the empty file and let `pruneRotated` drop a valid older rotated log — the exact outcome this commit set out to avoid, just for the (rare) empty-preexisting-file + fstat-failure combination. There's also a narrow TOCTOU between the `os.Stat` probe and `O_CREATE` (a create/remove in between flips `preexisted`).

Both are nearly unreachable given how rarely fstat-on-an-open-fd fails, so likely fine to leave as-is — flagging for awareness, not as a blocker.

---

eunolabs (commented):
## Review: batch automated-review bug fixes

Reviewed all 14 commits against base `135f5f8`. This is a high-quality, tightly-scoped batch — one commit per issue, each carrying a clear rationale comment, and the fail-closed invariant is preserved throughout. I checked the load-bearing claims against the surrounding code rather than taking them on faith:

**Verified correct**
- **#817** (partial audit-tail read → shrink): sound. `readLastAuditLine` sizes `buf := make([]byte, size-start)` to exactly the tail window from the same `size`, so `n < len(buf) && io.EOF` can only be a truncation, never a legitimate short read.
- **#819** (cache parsed cap heads): the hot path `buildConstraintsFromParsed(claims.parsedCaps, …)` and the fallback `buildConstraintsFromClaims(claims.Capabilities, …)` are provably equivalent — both reduce to `buildConstraintsFromParsed(parseCapHeads(capsList), …)`, since `parsedCaps = parseCapHeads(capsList)` and `Capabilities = capsList`. `parseCapHeads` always returns non-nil, so the `parsedCaps != nil` gate correctly separates validated tokens from test-constructed claims.
- **#827** (inclusive eviction boundary): all four eviction checks now use `!now.Before(expiresAt)`, matching `Get`'s `>=`. Consistent.
- **#828** (32-bit makeslice cap): `cur` is `int64`; capping at `math.MaxInt` before the `int(...)` conversion is correct and a no-op on 64-bit.
- **#829 / #830** (docs): both accurate against the code — `recordSessionCall` returns the counter error and `engine.go:727/925` turn it into `recordFailureDenial` (fail-closed); `IsFatal` = FM-1/2/4 and `hasCriticalDrift` = FM-5 match the new `drift.go` header verbatim.
- **#818, #826, #831**: read correctly — read-lock scan + write-lock re-check; RWMutex hot path with a re-check on the expiry upgrade so a concurrent refresh isn't clobbered; truncation marker skipped + review note emitted while the observed call is still allowed.

**Worth addressing (none blocking)**
1. **#824 — test gap on the crux of the fix.** The `expired = total - count` offset is the whole point of the change, but the existing regression test only exercises `expired == 0`. Detail + suggested test inline on `pkg/callcounter/redis.go`.
2. **#825 — `!= 3` vs `< 3`, and no unit test.** Minor; inline on `pdp_jwt.go`.
3. **#820 — residual force-rotation edge for a pre-existing *empty* log on fstat failure.** Corner case; inline on `audit.go`.

**One observation, no action needed:** #816 normalizes `\`→`/` before extension matching. Backslash is a legal byte in a POSIX filename, so on a genuinely-Linux upstream this reinterprets a `\`-containing name as having directory components — but because the allowlist match is suffix-only (`strings.HasSuffix`), the allow/deny outcome is preserved in every case I could construct; only the *reported* extension and the empty-extension gate get cleaner. Good hardening for Windows upstreams.

Nice work overall — the rationale comments make this batch easy to audit.

---

pkg/callcounter/redis.go
eunolabs: Good catch — added `TestRedis_PeekRetryAfter_PivotSkipsLingeringExpiredMembers`. It inserts six members at T=0,5,…,25, then peeks at T=70 without another `IncrementIfBelow` (so scores 0/5/10 linger out-of-window at the low ranks), and asserts retry == 10s. Dropping the `expired` offset (`rank == idx`) would pick the stale score-5 member and clamp retry to 0, so the offset is now pinned. (8b546f3)

---

cmd/eunox-mcp/pdp_jwt.go
eunolabs: Done — changed the guard to `!= 3` so it matches its message and also rejects 4+-segment non-JWS input, and added `TestDecodeJWTClaimsPreservingNumbers_SegmentGuard` (1/2/4 segments → error, 3 → ok) to guard against a regression back to `< 2`/`< 3`. (committed in the latest push)

---

cmd/eunox-mcp/audit.go
eunolabs: Closed the residual edge: the path is now probed with `os.Stat` before `O_CREATE`, and that pre-open size is reused as the fallback when `f.Stat()` fails. So a pre-existing **empty** log (preSize == 0) no longer force-rotates; a large existing log still continues the counter; only the genuinely-unknown case (both stats fail) keeps the conservative `rotateSizeBytes` (#363). The `os.Stat`→`O_CREATE` TOCTOU remains inherent but is unchanged in scope.

---

PR #844, opened 2026-06-18
refactor: complexity guardrails, dedup quick wins, and transport unification (audit phases 0-2)

eunolabs:
## What this is

An incremental code-quality pass from a complexity/design audit, in three phases. All behavior-preserving; net code is negative once the linter config is excluded. Every commit validated under `go test -race ./...` with lint/fmt/license/coverage gates green.

## Phase 0 — complexity guardrails (`ci`)

Enables `funlen`, `gocyclo`, `gocognit`, `nestif`, `dupl` as **ratchet ceilings** — each threshold sits just above today's worst offender, so the tree is green now while new code worse than today's worst is rejected (CI lints a shallow checkout, so a diff-based baseline isn't viable). Genuine outliers are grandfathered with named exclusions that double as a refactor worklist.

## Phase 1 — quick-win dedups (`refactor`)

| Area | Change |
|---|---|
| audit | `resolveAuditKeyPath` reuse; shared `newAuditLineScanner`; fused `cloneAndBound`/`cloneAndBoundDetails`; collapsed three integrity-marker writers into `writeIntegrityMarker` |
| pdp | `newAllowResponse` for the wiretap + JWT-only allow paths |
| callcounter | `retryAfterFromPivot` + `redisWindowKey` |
| killswitch | `cutKillID` over `strings.CutPrefix` for the four prefixed pub/sub events |

(plus the two review nits: `writeIntegrityMarker` no longer mutates its caller's map; the kill-switch dispatch is a single-eval tagged switch.)

## Phase 2 — transport unification (`refactor`)

- **Typed audit API:** `RecordAllow`/`RecordDeny` replace the 10-arg positional `Record`, so the decision is encoded by which method you call (an allow can't carry a denial code; a deny can't carry obligations). All 100 call sites migrated AST-correctly with `gofmt -r`.
- **`tools/call` folded into the shared `enforcedForwardCore`:** the highest-risk hand-mirrored path — both transports each carried a ~50-line copy of the deny/observe/forward/redact/record flow. The core now applies `redactFields` to the response (gated on `dec.Obligations`, fail-closed) and takes an `allowDetails` func so `tools/call` keeps logging its audit-mode argument map while resources/prompts keep `upstreamErrorDetail`. Both `handleToolsCall` handlers drop to ~29 lines.

Remaining transport-unification steps (not in this PR): fold `sampling/createMessage` into a shared core, extract `upstreamConn` for the triplicated correlation/handshake/timeout, share the method-dispatch switch, then split `http.go`.

## Validation


cmd/eunox-mcp/audit.go
```diff
@@ -620,59 +611,35 @@ func (s *auditSink) syntheticDenyMarker(code string, details map[string]interfac
...
+// The three are mutually exclusive (a read failure precludes a parse, which
+// precedes the HMAC check), so at most one is written per start. Every variant
+// reuses the AUDIT_INTEGRITY_FAILURE deny shape via syntheticDenyMarker — adding no
+// new top-level record field — and is chained and signed like any other record, so
+// audit-verify and any external sink surface it. Without this in-band marker the
+// event would survive only as a transient stderr warning that container log
+// rotation can lose, leaving the tamper-evident trail with no evidence the prior
+// session ended on a forged, corrupt, or unreadable tail.
+func (s *auditSink) writeIntegrityMarker(kind string, details map[string]interface{}) {
+	details["kind"] = kind
```
eunolabs: Nit (non-blocking): `writeIntegrityMarker` mutates the caller's `details` map in place. It's safe today — all three call sites pass fresh map literals, and `syntheticDenyMarker` clones via `cloneAndBoundDetails` immediately after — but the old three writers each built their map literal inline with no shared input, so this is new implicit contract: `details` must be non-nil (a `nil` map would panic on the `details["kind"] = kind` assignment) and the caller must not reuse the map afterward.

Cheap to make robust, e.g. start from a fresh map so the caller's argument is never touched:

```go
func (s *auditSink) writeIntegrityMarker(kind string, details map[string]interface{}) {
	d := map[string]interface{}{"kind": kind}
	for k, v := range details {
		d[k] = v
	}
	marker := s.syntheticDenyMarker("AUDIT_INTEGRITY_FAILURE", d)
	s.writeRecord(&marker)
}
```

or just a one-line doc note that `details` is mutated and must be non-nil. Either is fine; flagging since it's the one spot in the dedup that took ownership of a caller-supplied map.

---

pkg/killswitch/redis.go
```diff
@@ -775,14 +776,14 @@ func (r *Redis) handlePubSubMessage(payload string) {
...
 		shouldRefresh = true
-	case len(payload) > len("agent:kill:") && payload[:len("agent:kill:")] == "agent:kill:":
-		r.killedAgents[payload[len("agent:kill:"):]] = true
-	case len(payload) > len("agent:revive:") && payload[:len("agent:revive:")] == "agent:revive:":
-		delete(r.killedAgents, payload[len("agent:revive:"):])
-	case len(payload) > len("session:kill:") && payload[:len("session:kill:")] == "session:kill:":
-		r.killedSessions[payload[len("session:kill:"):]] = true
-	case len(payload) > len("session:revive:") && payload[:len("session:revive:")] == "session:revive:":
-		delete(r.killedSessions, payload[len("session:revive:"):])
+	case cutKillID(payload, "agent:kill:") != "":
```
eunolabs: Trivial nit (optional): each matching branch evaluates `cutKillID(payload, prefix)` twice — once in the `case` guard and again in the body — so a match does two `strings.CutPrefix` scans instead of one. The cost is negligible (rare control-plane pub/sub messages, short strings) and the equivalence with the old `len()>... && payload[:...]==...` form is correct, including treating a bare `"agent:kill:"` (empty id) as a non-match that falls through to a full refresh. Only mentioning because the switch-with-guard shape is what forces the double call; an if/else chain binding `id` once would avoid it:

```go
if id, ok := strings.CutPrefix(payload, "agent:kill:"); ok && id != "" {
	r.killedAgents[id] = true
} else if id, ok := strings.CutPrefix(payload, "agent:revive:"); ok && id != "" {
	delete(r.killedAgents, id)
} // ...
```

Not worth churning the switch over if you prefer its symmetry — leaving to your taste.

---

eunolabs (commented):
Reviewed as a pure refactor — the "no behavior changes" claim is the thing to verify, so I checked each extraction against the code it replaced rather than just reading the diff. It holds up; this is clean work. CI is green (Test, Lint, all demo/* jobs), and test coverage for the renamed paths is intact.

**Behavior-preservation, verified extraction by extraction:**

- **`openAuditSink` → `resolveAuditKeyPath`** — `resolveAuditKeyPath` is byte-for-byte the old inline precedence (`flag → EUNOX_AUDIT_KEY_PATH → defaultAuditKeyPath`, then `expandHome`). Equivalent, and it now genuinely shares one source of truth with audit-verify/doctor. ✓
- **`copyDetails`+`boundAuditDetails` → `cloneAndBound`/`cloneAndBoundDetails`** — the fused single pass produces identical output: strings bounded the same way, `[]string` cloned element-wise *without* per-element bounding (matching the old `boundDetailValue`, which didn't recurse into `[]string`), `nil`/scalars/`json.Number` passed through, and the total-cap marshal check + marker message unchanged. Cloning before the bound (so the bounded queue can't retain un-truncated multi-MiB clones, #461) is preserved. ✓
- **`wiretapAllow` / JWT-only allow → `newAllowResponse`** — this was the most likely place to drift, because the JWT path used `p.now()` while wiretap used `clockNow(p.clock)`. Confirmed `JWTPDP.now()` is literally `return clockNow(p.clock)`, so collapsing both onto `newAllowResponse(p.clock)` is exactly equivalent; the frozen-clock test path is honored identically. `p.now()` is still used in `ValidateToken`, so no dead code. ✓
- **`retryAfterFromPivot`** — both call sites (`IncrementIfBelow`, `PeekRetryAfter`) pass `cur`/`limit` as `int64`, so the #532 32-bit overflow fix is preserved at both, and the `return 0` fallback matches the old named-return-starts-at-zero behavior. ✓
- **`redisWindowKey` / `cutKillID`** — `redisWindowKey` is a verbatim format extraction across all four Redis methods. `cutKillID` preserves the "strictly longer than prefix" semantics (bare `agent:kill:` → empty id → no match → full refresh). ✓
- **Integrity markers** — collapsing the three writers into `writeIntegrityMarker(kind, details)` keeps coverage for all three kinds: `tail_hmac_mismatch` and `tail_parse_failure` exercised end-to-end through `openAuditSink`, `chain_resume_failed` through the (updated) direct test. ✓

**One observation on the guardrails themselves (not blocking):** the grandfathered complexity hotspots use message-text exclusions (`func \`validateLocalManifest\` is high`), which match regardless of the actual cyclo/cognit number. So while the ceilings stop *new* functions from regressing, they won't catch a *grandfathered* function getting worse — e.g. `validateLocalManifest` could drift from cyclo 70 to 90 and stay silently excluded. That's an inherent limit of text-based exclusions (golangci-lint has no per-function numeric ceiling), and the same applies to the whole-file `dupl` exclusions for `http.go`/`stdio.go`/`killswitch/`, which are broader than the function-level grandfathers elsewhere. Worth keeping in mind as the worklist gets burned down — an inline `//nolint:gocyclo // grandfathered at N, do not increase` per function would at least document the baseline even if it can't enforce it. Fully reasonable to defer to the planned refactors.

Two minor inline nits below (the `writeIntegrityMarker` in-place map mutation is the more substantive of the two). Neither blocks.

---

cmd/eunox-mcp/audit.go
eunolabs: Fixed in 76ea7a6 — `writeIntegrityMarker` now copies into a fresh map and sets `"kind"` last, so the caller's map is never mutated and a nil `details` no longer panics on the assignment. Added `TestWriteIntegrityMarker_CallerMapSafety` to pin both the no-mutation and nil-safe properties.

---

pkg/killswitch/redis.go
eunolabs: Done in 76ea7a6 — switched to a tagged `switch payload` for the exact-match events and bind the id once via `cutKillID` in the `default` arm, so each prefix is scanned a single time. Semantics are unchanged, including a bare `"agent:kill:"` (empty id) falling through to a full refresh.

---

PR #851, opened 2026-06-18
refactor(transport): unify tools/call and sampling onto shared cores + typed audit API (audit phase 2)

eunolabs:
## What this is

Phase 2 of the code-quality audit: **transport unification**, building on the phase 0–1 work merged in #844. The two transports (`StdioProxy`, `HTTPProxy`) hand-mirrored their most security-critical paths line-for-line; this collapses them onto shared cores so they can no longer drift. All behavior-preserving; every commit validated under `go test -race ./...`.

## Commits

**1. Typed audit API (`refactor(audit)`)** — `RecordAllow`/`RecordDeny` replace the 10-arg positional `Record`, so the decision is encoded by which method you call (an allow can't carry a denial code; a deny can't carry obligations). All 100 call sites migrated AST-correctly with `gofmt -r`.

**2. `tools/call` → shared `enforcedForwardCore` (`refactor(transport)`)** — both transports each carried a ~50-line copy of the deny/observe/forward/redact/record flow. The core now applies `redactFields` to the response (gated on `dec.Obligations`, fail-closed — a no-op for non-tool methods, which the manifest forbids from carrying directives) and takes an `allowDetails` func so `tools/call` keeps logging its audit-mode argument map while resources/prompts keep `upstreamErrorDetail`. Both `handleToolsCall` drop from ~80 to **29 lines**.

**3. `sampling/createMessage` → shared `forwardServerRequest` (`refactor(transport)`)** — the second hand-mirrored pair (upstream-initiated requests). Parameterized on the only real differences: delivery (`forward func(rpcMsg) bool` — stdio always delivers, HTTP's SSE broadcast may not, audited as `ENFORCEMENT_ERROR` per #648), the upstream-deny writer, and the JWT `claims` attached only on the sampling path. Both handlers drop to ~18 lines.

**4. Shared initialize handshake + timeout (`refactor(transport)`)** — the three handshakes (`initUpstream`, `runInitHandshake`, `initRemoteUpstream`) now share `buildInitializeRequest` + `applyInitializeResult`, so the fail-closed-on-error (#708) and `serverInfo.version` extraction (#545) logic that previously had to land in all three lives once; each keeps only its transport-specific send/receive. `withUpstreamTimeout`'s two real impls unify behind `boundUpstreamCall`, making the `withoutUpstreamTimeout` opt-out mean the same thing on both transports (a no-op for stdio today).

**5. Dead-code cleanup (`test(http)`)** — removed the orphaned `captureSink` flagged in review.

## Remaining transport-unification steps (follow-ups, not in this PR)

Extract an `upstreamConn` type for the `callUpstream` pending-map correlation and `readUpstream` router (the concurrency-critical core, where `done`-channel ownership differs by transport mode); then share the method-dispatch switch and split `http.go`.

## Validation


eunolabs:
Verdict: clean, behavior-preserving refactor. No blocking findings. This is the deferred work [#844](https://github.com/eunolabs/eunox/pull/844) flagged, and it's executed carefully. Build + go vet clean, and go test -race ./cmd/eunox-mcp/ passes in full (~70s) on the head commit.

What I verified (the parts that could have hidden a behavior change)
1. Record → typed RecordAllow/RecordDeny split. Both delegate to the unchanged internal record(...); RecordDeny hard-wires obligs=nil so a deny structurally can't carry obligations, and the allow/deny string is now chosen by method rather than passed positionally. routeSink and the auditRecorder interface were updated in lockstep. Confirmed zero remaining callers of the old positional Record in production or test code, and the method is fully removed.

2. Redaction moved into the shared enforcedForwardCore — the one change that's more than a refactor. Previously only the bespoke tools/call path applied redactFields; now the fail-closed redaction block lives in the shared core, gated on len(dec.Obligations) > 0. Its safety rests entirely on directives being impossible on non-tool targets — and that holds: [pkg/capability/directive.go:30-32](https://github.com/eunolabs/eunox-internal/blob/claude/awesome-newton-1s1jgr/pkg/capability/directive.go#L30-L32) rejects any directive on a non-tool: target at manifest load. So dec.Obligations is always empty for resources/read, resources/subscribe, and prompts/get, making the relocated block a true no-op for them. The fail-closed semantics, the -32603 error, and the stderr SECURITY: line are all preserved (the line now interpolates kind/denialTarget, yielding identical text for tools).

3. tools/call allow details. The allowDetails closure returns the precomputed audit-mode argument map and ignores the response arg — matching the old args-only behavior (tools/call deliberately does not use upstreamErrorDetail, unlike resources/prompts). The aliasing of params.Arguments is safe because record() deep-clones via cloneAndBoundDetails before enqueue.

4. sampling fold into forwardServerRequest. Checked every branch against both old transports: hard-deny writes AUTHORIZATION_FAILED to the upstream then records (same order); the kill-switch hard-block survives audit mode (#358); the two-record observe pattern is deny-then-forward (#678); the no-subscriber ENFORCEMENT_ERROR deny is preserved on HTTP (#648) and correctly unreachable on stdio (its forward always returns true); JWT claims are attached only on the sampling path, and only on HTTP — matching the prior code, where non-sampling server requests recorded without claims.

Minor / optional (non-blocking)
captureSink in http_test.go is orphaned dead code. Its Record method still has the old positional signature and now matches no interface. This is pre-existing — http_test.go isn't in this PR's changeset, and TestAudit_RecordsObserveFlag already discards the sink (_ = sink // we'll wire it differently below) and captures via a recordingDecisionPoint instead. It compiles only because the struct is never assigned to an auditRecorder. Since this PR is precisely a recorder-signature migration, it'd be a natural time to delete the now-doubly-dead helper — but entirely optional and out of the stated scope.
Notes
No docs/threat-model update needed: the audit record shape is unchanged (same fields, same "allow"/"deny" values emitted by record()), consistent with the "no behavior changes" claim.

eunolabs:
Addressed the one optional item in `22cc21c`: removed the orphaned `captureSink` (its `Record` method carried the old positional signature and matched no interface after the migration) along with the dead setup block it anchored in `TestAudit_RecordsObserveFlag` (which also orphaned a throwaway proxy). Kept `capturedRecord` — the live `recordingDecisionPoint` path still uses it. Full `-race` suite still green.

Also pushed two further transport-unification steps since your review (`0365d18`): the three initialize handshakes now share `buildInitializeRequest`/`applyInitializeResult` (so the #708 fail-closed and #545 version-extraction logic lives once), and `withUpstreamTimeout` is unified behind `boundUpstreamCall`. PR description updated to match.

---

PR #858, opened 2026-06-19
refactor(transport): share the upstream request/response correlation

eunolabs:
## What this is

Continues the transport-unification work from #851. The two transports (`StdioProxy`, `httpSession`) hand-mirrored the concurrency-critical JSON-RPC request/response correlation; this collapses it onto shared helpers. The extraction is behavior-preserving; one folded-in hardening (below, from review) deliberately fixes a pre-existing corner case. The full `-race` suite passes.

## The change

`callUpstream` (stdio) and `callSubprocessUpstream` (HTTP session) each carried the identical pending-map correlation: register a response channel under a **dup-ID guard** (refusing to orphan an in-flight caller's channel), send, then wait on a **three-way `select`** (response / `ctx` / upstream-exited). `readUpstream`'s response delivery (locked map lookup + buffered send) was duplicated too.

Extracted `awaitUpstreamResponse` and `deliverUpstreamResponse` as free functions over the existing `pendingMu`/`pending` fields, so the dup-ID guard, cleanup, and three-way wait live once. The genuine per-transport differences stay at the call site:
- the **send step** (the stdio HTTP bridge uses `postWithCtx`, everything else `upWriter.Write`) is a passed closure;
- the **done channel** (`p.upstreamDone` / `s.done`) is a parameter.

### Why free functions, not a `pendingCalls` type

The audit suggested an `upstreamConn` type embedding the pending map. I deliberately **didn't** change the field representation: ~65 test sites construct these structs with `pending: make(map[string]chan rpcMsg)` or poke `pendingMu`/`pending` directly (injecting entries to exercise the dup-ID guard). Changing the field type would force a large, risky test migration for **no behavior gain**. The free-function form keeps the field layout — and every test — untouched, while still removing the duplication.

The two `"upstream exited"` / `"session upstream exited"` messages (operator-log only; no test or client contract depends on the text) are unified to `errUpstreamExited`.

This completes the `upstreamConn` correlation dedup; the initialize-handshake and `withUpstreamTimeout` pieces of that finding landed in #851.

## Folded-in hardening (review follow-up)

Review flagged that the now-centralized `deliverUpstreamResponse` did a **blocking** send on the buffered(1) pending channel — faithfully carried over from both transports. A misbehaving upstream emitting two responses for the same JSON-RPC ID could wedge the reader goroutine forever if the awaiting caller had already left via `ctx`/`done` without draining (buffer full, second send blocks with no receiver).

Since consolidation gave us exactly one delivery path, this fixes it for both transports at once: the send is now **non-blocking** (`select { case ch <- msg: default }`), dropping a duplicate/late response for an already-served or abandoned ID. The awaiting caller's first response always has buffer room, so legitimate traffic is unaffected. Added `TestDeliverUpstreamResponse_DuplicateDoesNotWedgeReader` (`-race`) and corrected the doc comment that previously over-claimed the send "never blocks the reader goroutine."

## Remaining transport-unification follow-ups (not in this PR)

Share the per-method dispatch switch between the two transports; then split the 2k-line `http.go` into server/security/session/handlers files.

## Validation

---


cmd/eunox-mcp/jsonrpc.go
```diff
@@ -169,6 +171,60 @@ func (t *serverReqTracker) take(key string) bool {
...
+// deliverUpstreamResponse routes an upstream response to the caller in pending
+// (guarded by mu) awaiting its ID, if any. Each pending channel is buffered (1), so
+// the send never blocks the reader goroutine; an untracked or already-served ID is
+// dropped. Both transports' readUpstream loops deliver responses through here.
+func deliverUpstreamResponse(mu *sync.Mutex, pending map[string]chan rpcMsg, msg rpcMsg) { //nolint:gocritic // hugeParam: rpcMsg by value, matching the reader loop
+	mu.Lock()
+	ch, ok := pending[msgKey(msg.ID)]
+	mu.Unlock()
+	if ok {
+		ch <- msg
```
eunolabs: Optional hardening (pre-existing, not a regression — flagging only because this PR centralizes the delivery into one spot): the send happens outside the lock on a buffered(1) channel, faithfully preserved from both transports. In the normal single-response case this is fine (the buffer always has room). But if a misbehaving upstream ever emits two responses for the same JSON-RPC ID, the second `ch <- msg` can wedge the reader goroutine in the corner case where the awaiting caller already left via `ctx.Done()`/`done` without draining the first send — the buffer stays full and the key isn't deleted yet, so this blocks forever.

Now that there's exactly one delivery path, it's a one-line fix if you want it — a non-blocking send drops the stray duplicate instead of risking a stuck reader:

```go
select {
case ch <- msg:
default: // ID already served or caller gone; drop the duplicate/late response
}
```

Legitimate traffic is unaffected. Entirely out of scope for a behavior-preserving PR — just noting that consolidating this code is the natural moment to consider it.

---

eunolabs (commented):
Clean, faithful extraction. I checked the two free functions byte-for-byte against the inline logic they replace (this is concurrency-critical, so it's the thing worth verifying), and re-ran the full `cmd/eunox-mcp` suite under `-race` locally on top of green CI.

**Behavior-preservation, verified:**

- **`awaitUpstreamResponse`** is exactly the prior `callUpstream` body: dup-ID guard → register channel → `defer` cleanup → send → three-way `select`. The genuine per-transport differences are correctly lifted out: the **send step** as a `send func() error` closure (stdio's `upHTTP.postWithCtx(ctx, msg)` returning nil vs `upWriter.Write(msg)` returning its error; HTTP's plain `upWriter.Write`) and the **done channel** as a parameter (`upstreamDone` / `done`). The cleanup `defer` is registered before `send()` in both old and new, so a send failure still deletes the key. ✓
- **`deliverUpstreamResponse`** is identical to both `readUpstream` delivery blocks (locked lookup, unlock, buffered send). ✓
- **Field-layout decision** — keeping the `pending`/`pendingMu` fields and going with free functions rather than a `pendingCalls`/`upstreamConn` type is the right pragmatic call given ~65 test sites construct/poke these directly; the diff stays tiny and no test had to move. Agreed with the reasoning in the description.

**The one deliberate behavior change is safe:** unifying `"session upstream exited"` / `"upstream exited"` into the `errUpstreamExited` sentinel. Confirmed nothing matches on the text — no production caller inspects the error, the two test references to "upstream exited" are unrelated (one is an arbitrary non-timeout error in a `DeadlineExceeded` test; the others assert only that the exit path returns *non-nil*). The HTTP-side text shift is operator-log-only, and the change is a net improvement since the error is now a comparable sentinel.

**Coverage note (non-blocking):** no test file is in the changeset, so the new helpers have no dedicated unit test — but they're transitively exercised through the real call sites by the existing dup-ID-guard and upstream-exit-while-pending tests in `transport_unit_test.go`, which is adequate for a pure extraction.

One optional, explicitly pre-existing hardening idea left inline on `deliverUpstreamResponse`. Nothing blocking.

---

cmd/eunox-mcp/jsonrpc.go
eunolabs: Done in b558bd1. Switched the delivery to a non-blocking send (`select { case ch <- msg: default }`), so a duplicate/late response for an already-served or abandoned ID is dropped rather than wedging the reader on a full buffer. Legitimate single-response traffic is unaffected — the awaiting caller's first response always has buffer room.

Added `TestDeliverUpstreamResponse_DuplicateDoesNotWedgeReader`, a `-race` regression test that fills the buffer then asserts a second delivery returns instead of blocking (without the fix it deadlocks → 2s timeout). Also corrected the doc comment, which previously over-claimed the send "never blocks the reader goroutine."

---

eunolabs (commented):
Re-reviewed after the follow-up commit (`b558bd17`). My one prior note is fully resolved and the fix is correct — no further findings, LGTM. (Would approve, but GitHub won't let this account approve its own PR.)

**The hardening (`deliverUpstreamResponse` non-blocking send):** correct. `default` is only reachable when the buffered(1) channel is already full, i.e. a second/duplicate response for an ID whose single legitimate response is still undrained. A legitimate first response always finds the empty buffer and lands, so nothing real is ever dropped — the change is strictly safer and now covers both transports through the one shared delivery path. Promoting it to `sec(transport)` is the right call: the old blocking send let a misbehaving upstream wedge the reader goroutine.

**Test quality — verified, not assumed:** I ran `TestDeliverUpstreamResponse_DuplicateDoesNotWedgeReader` under `-race` (passes), then reverted the send back to the blocking `ch <- msg` locally to confirm the test actually *catches* the wedge — it fails, hanging the full 2s and tripping the regression assertion. So it's a genuine guard, not a tautology, and it models the exact trigger (buffer full + caller gone without draining).

**Doc comment:** good catch fixing the prior comment that over-claimed the send "never blocks the reader goroutine" — it now accurately describes the non-blocking drop and its cases.

**Rest of the PR** is unchanged from my first pass: the `awaitUpstreamResponse`/`deliverUpstreamResponse` extraction is a byte-for-byte-faithful consolidation of the per-transport correlation, the unified `errUpstreamExited` sentinel is safe (nothing matches on the text), and the free-functions-over-a-type decision is well justified.

Verification: build + `go vet` clean, full `cmd/eunox-mcp` suite green under `-race` locally, on top of green CI (only the CodeQL/Analyze-go job was still completing at review time; all functional checks passed).

---

PR #860, opened 2026-06-19
refactor(transport): split http.go into server/security/session/handlers files

eunolabs:
## What this is

The last of the transport-unification follow-ups from #858: `http.go` had grown to ~1810 lines spanning five distinct concerns, making it hard to navigate. This splits it by concern. **Pure relocation — no logic changed.**

## The split

| File | Lines | Responsibility |
|------|------:|----------------|
| `http.go` | 438 | server core: constants, config resolvers, `HTTPProxy` + options + constructors, `Serve`, `serveCtx` |
| `http_routing.go` | 360 | HTTP endpoint routing: `/mcp` POST/GET/DELETE + loopback-only `/control/kill` |
| `http_security.go` | 203 | bearer/control-token auth, Origin (DNS-rebinding) validation, source-IP extraction, loopback gating |
| `http_session.go` | 584 | `httpSession` lifecycle, session registry + idle reaper, upstream I/O, SSE broadcast |
| `http_handlers.go` | 292 | per-method MCP enforcement handlers (`tools/call`, `resources/*`, `prompts/*`, server-initiated requests) + upstream-timeout/denial helpers |

Every declaration moved verbatim; all cross-references stay within `package main`, so nothing needed re-wiring. Each file's imports are the minimal set its code actually uses.

## Why this is safe (verified, not assumed)

This touches security-critical transport code, so I verified the move is lossless three ways:

1. **Declaration inventory** — the set of 65 top-level `func`/`type`/`var`/`const` declarations is byte-identical before and after (`diff` of the sorted inventories is empty).
2. **Code-line multiset** — every non-scaffolding code line in the prior `http.go` is present exactly once across the five files (zero lost, the only additions are the new per-file license headers and one-line purpose comments).
3. **Behavior** — the full `cmd/eunox-mcp` suite passes under `-race`.

## Validation

`go build`, `go test -race ./cmd/eunox-mcp/` (green, ~70s), `go vet` + `golangci-lint` (0 issues), `gofmt`, license headers — all pass. Updated the architecture section in `CLAUDE.md` to describe the new layout.

## Remaining follow-up (not in this PR)

The per-method dispatch switch is still hand-mirrored between the two transports. Sharing it is the larger remaining piece (stdio handlers write to the host and return void, while the HTTP handlers return `rpcMsg`, so a full share needs handler-signature unification) — best done as its own focused PR now that the handlers have a dedicated home.

---


eunolabs (commented):
Reviewed as a pure relocation — for a "no logic changed" file split the whole question is whether the move is truly lossless, so I verified that mechanically rather than reading the 2.8k-line diff. It checks out cleanly; no findings. (Would approve, but GitHub won't let this account approve its own PR.)

**Lossless-move proof (independent of the PR's own checks):** I parsed the pre-split `http.go` and the 5 new files with Go's AST, stripped all comments so only *code* is compared, and diffed the multiset of top-level declarations. Result: **65 declarations on each side, every one byte-identical at the code level, same multiplicity, zero mismatches.** That's stronger than a name-inventory diff — it confirms each declaration's *body* is identical, not just that the names survived. Combined with `go build` succeeding (which independently proves each file's imports are the exact minimal set — Go rejects both missing and unused imports), the relocation is provably behavior-preserving.

**Things a file move can silently break, explicitly checked:**
- **`//nolint` / `//go:` directives** — 21 in the old file, 21 across the new files (lint stays green, so each still suppresses at its new location). A dropped `//nolint` is the classic file-split regression; none lost here.
- **License headers** — present on all 5 files (CI License Headers job green).
- **No collateral edits** — only the 5 `http*` files + a one-line `CLAUDE.md` architecture update changed; nothing else in the tree.

**Behavior:** full `cmd/eunox-mcp` suite green under `-race` locally (~74s), on top of fully green CI (Test, Lint, gofmt, CodeQL, Analyze-go, all builds/demos/OPA jobs).

The split itself is sensible — server core / routing / security / session / handlers are genuinely distinct concerns, and giving the per-method handlers a dedicated home sets up the noted follow-up (sharing the dispatch switch across transports) well. LGTM.

---

PR #864, opened 2026-06-19
refactor(transport): extract shared dispatchRequest (finishes audit phase 2)

eunolabs:
## What this is

The final Phase 2 (transport unification) item from the code-quality audit: **extract the shared `dispatchRequest`**. The per-method MCP dispatch switch — the method→handler mapping plus the fail-closed default for unmapped methods — was hand-mirrored between the stdio (`handleHostRequest`) and HTTP (`handleMCPPost`) transports. That switch was the last of the "highest-risk hand-mirrored security paths" Phase 2 set out to eliminate.

## The change

New transport-neutral `dispatch.go`:
- **`dispatchParams`** carries each transport's policy/audit/upstream wiring (PDP, audit sink, session id, source IP, `callUpstream`). stdio builds it from the proxy (empty source IP); HTTP from `sess.route` + the request.
- **`dispatchRequest`** is the single switch — the one source of truth for the method→handler map and the fail-closed `default` deny.
- The per-method handlers (`tools/call`, `resources/read|list|subscribe`, `prompts/get|list`, and the unmapped deny) are now shared free functions returning the `rpcMsg` to deliver.

Both transports build `dispatchParams` and call `dispatchRequest`; the per-transport handler bodies (`handleHTTPToolsCall`, `handleToolsCall`, … and both `enforcedForward` wrappers) and the duplicated switch are deleted. `initialize` stays per-transport (stdio logs a host-initialized line; HTTP ties it to session creation) and never reaches the shared switch — the transports differ only in delivery (HTTP returns the message, stdio writes it to the host).

This builds on #851, which already shared the deny/forward/record core (`enforcedForwardCore`); this PR consolidates the remaining dispatch + param-parsing on top of it.

## Impact

**Net −146 production lines** (`dispatch.go` +204, but `stdio.go` −238, `http_handlers.go` −162, `http_routing.go` −22).

## Why this is safe (behavior-preserving)

- The merged handlers pass the **identical** arguments to `enforcedForwardCore` as the two originals (same `method`/`auditID`/`denialTarget`/`kind`/`recordObligations`/`allowDetails`), so audit records are unchanged. The audit-mode quota-skip is preserved via `dispatchParams.decideCtx`. The fail-closed default logs and records exactly as before (`AUTHORIZATION_FAILED`, upstream never called).
- The ~40 stdio unit tests that drove the per-method methods directly now drive `handleHostRequest` — each test's `rpcMsg` already sets `Method`, so routing is identical and the assertions are unchanged.
- Full `cmd/eunox-mcp` suite passes under **`-race`** (~69s); `go vet` + golangci-lint **0 issues**; gofmt + license clean.

## Plan status

This **completes Phase 2 (transport unification)**. Updated the `CLAUDE.md` architecture note to describe the shared dispatch. Remaining audit work is Phase 3's audit-side decomposition (`auditSink` → writer/rotator/verifier; `verifyAuditLog`/`cmdProxy`) and Phases 4–5.

---


eunolabs (commented):
Reviewed as a behavior-preserving consolidation of a security-critical path (the dispatch switch + fail-closed default). The real risk in merging two hand-mirrored switches is a *pre-existing divergence* the merge silently picks one side of, so I diffed the old per-transport code against each other and against the new shared handlers. It holds up — faithful merge, no findings. (Would approve, but GitHub blocks self-approval for this account.)

**Method map — identical pre-merge.** Both old switches (`handleHostRequest`, `handleMCPPost`) mapped the same 8 methods + the same fail-closed default; only `initialize` differed, and it's correctly kept per-transport *before* the shared switch — stdio routes it to `handleInitialize` and returns; HTTP keeps `buildInitResponse`, so re-initialize on an existing session still answers locally rather than falling into the `AUTHORIZATION_FAILED` default. That was the main divergence trap and it's handled.

**Merged handlers — argument-identical to both originals.** For all four param-parsing handlers (`tools/call`, `resources/read`, `resources/subscribe`, `prompts/get`) and `dispatchList`, the old stdio and old HTTP bodies were already byte-identical except for the pdp/audit/sessionID/sourceIP that `dispatchParams` now abstracts. The new handlers pass the same `method`/`auditID`/`denialTarget`/`kind`/`recordObligations`/`allowDetails` into `enforcedForwardCore`, the same `-32602` param-error strings, the same empty-URI/empty-name guards, and the same audit-mode quota-skip (now `dispatchParams.decideCtx`, applied only to the decision, not the forward — matching the old `decCtx`).

**Two subtle things specifically checked:**
- **Typed-nil sink trap** (the one #851 was careful about): both `dispatchParams` builders do `var rec auditRecorder; if sink != nil { rec = sink }`, so `d.rec != nil` in `dispatchList`/`dispatchUnmapped`/the core is a true "no sink" test. `*auditSink.RecordAllow` has no nil-receiver guard, so getting this wrong would panic — it's correct.
- **The `decodeParams` (UseNumber) vs `json.Unmarshal` split** now sits side-by-side in `dispatch.go` and looks like an inconsistency, but it's intentional: `decodeParams` preserves `json.Number` for argument-carrying methods to avoid the float64-rounding policy-bypass (#687); resources only carry a URI string. Worth keeping that asymmetry documented so nobody "unifies" it later — but the merge preserves it exactly, which is the correct call.

**Verification:** no dangling references to the deleted per-method handlers / both `enforcedForward` wrappers; the ~39 stdio test migrations are pure call-site swaps to `handleHostRequest` with the same `Method`-bearing `rpcMsg` (zero assertion changes — and they now exercise the real dispatch path); `go build`/`go vet` clean; full `cmd/eunox-mcp` suite green under `-race` locally, on top of fully green CI. LGTM — nice clean finish to Phase 2.

---

PR #865, opened 2026-06-19
refactor(audit): split audit.go into writer/rotator/verifier files

eunolabs:
## What this is

Phase 3 (file/concern splits) of the code-quality audit, audit-side. `audit.go` had grown to ~2410 lines holding the entire `auditSink`. This splits it by concern. **Pure relocation — no logic changed.**

## The split

| File | Lines | Responsibility |
|------|------:|----------------|
| `audit.go` | 1688 | writer core: `auditSink` construction, the record/drain/write/close path, record bounding, key management, tail resume, path resolution |
| `audit_rotate.go` | 314 | size-triggered rotation, retention pruning, rotated-sibling naming |
| `audit_verify.go` | 436 | per-record HMAC verification, the tamper-evident chain check, `verifyAuditLog` |

`auditSink` **stays one type** — its fields are shared drainer-goroutine state (the rotator mutates the same `f`/`written`/`activePath` the writer owns), so the concerns split cleanly by *file*, not by type. All files are `package main`, so cross-file references resolve without rewiring.

## Why this is safe (verified, not assumed)

This is the most integrity-critical file in the tree (the tamper-evident HMAC chain), so I verified the move is lossless four ways:

1. **Declaration inventory** — the set of 68 top-level `func`/`type`/`var`/`const` declarations is byte-identical before/after (sorted-inventory `diff` empty).
2. **Code-line multiset** — every code line in the prior `audit.go` is present exactly once across the three files (**0 lost**; the only additions are the two new files' license headers and one-line purpose comments).
3. **Behavior** — the full `cmd/eunox-mcp` suite (≈15 audit-specific test files) passes under **`-race`**.
4. **Static** — `go vet` + golangci-lint **0 issues**; gofmt + license clean.

## Plan status

This is the **first** of Phase 3's audit-side decompositions. Updated the `CLAUDE.md` architecture note. Remaining Phase 3 work (separate follow-up PRs, kept apart so a pure file-move isn't mixed with logic-restructuring decompositions):
- decompose `verifyAuditLog` (~283 lines) into focused helpers
- decompose `cmdProxy` (~257 lines) into focused helpers

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01BGW2TZerGmSkP6BTfNY3bB

---
_Generated by [Claude Code](https://claude.ai/code/session_01BGW2TZerGmSkP6BTfNY3bB)_

PR #869, opened 2026-06-19
refactor: decompose verifyAuditLog and cmdProxy (finishes Phase 3)

eunolabs:
## What this is

The final Phase 3 (code-quality audit) work: decompose the two remaining long functions, **`verifyAuditLog`** and **`cmdProxy`**. Together with the `audit.go` split (#865), this **completes Phase 3**.

Two commits, each behavior-preserving and independently reviewable.

### 1. `refactor(audit): decompose verifyAuditLog into an auditChainVerifier`

`verifyAuditLog` was a single ~280-line scan loop carrying the tamper-evident chain state (prev hmac/seq, signed-era flag) and the running tallies as locals. Lift that state into an `auditChainVerifier` struct and split the loop body into focused methods, **moving the load-bearing comments verbatim**:

- `decodeAuditRecord` — one line → (record, well-formed?)
- `processLine` — driver: decode, derive the forged-decoy flags, maintain the chain, classify
- `updateChain` — prev_hmac linkage + seq-gap detection and anchor advance
- `classify` — malformed / forged-decoy / legacy / HMAC-verified counting

`verifyAuditLog` is now a thin scanner wrapper. Behavior-preserving: `updateChain` runs before `classify` exactly as the inline order did, and `classify` reads only chain state that is unchanged for the record category it inspects. (Also merges a pre-existing `inReportWindow` conditional — staticcheck QF1007 — that the Phase 0 lint baseline had suppressed and the move resurfaced.)

### 2. `refactor(cmd): decompose cmdProxy into flag/backend/sink helpers`

`cmdProxy` bundled ~30 flag declarations, the usage text, backend construction, audit-sink opening, signal handling, and transport dispatch in one ~250-line function. Extracted into focused helpers so `cmdProxy` is now a **~134-line orchestration shell**:

- `proxyCLIFlags` + `registerProxyFlags` — the ~30 flag declarations as one struct
- `printProxyUsage` — the subcommand help text
- `buildCallCounterAndKillSwitch` — in-memory vs Redis-backed counter + kill switch
- `openConfiguredAuditSink` — audit-sink config-over-flag precedence + open

Flag declarations and operator-facing log strings move verbatim; flag pointers are read through the returned struct (`*f.x`), and the backend/sink helpers take parsed values as parameters.

## Verification

For each commit: `go build`, the full `cmd/eunox-mcp` suite under **`-race`** (incl. the dedicated forged-seq0 / forged-legacy / seq0-wrap / chain-break verifier tests), `go vet` + golangci-lint **0 issues**, gofmt + license clean.

## Plan status

This **completes Phase 3**. Remaining audit work: Phase 4 (interface cleanups — PDP interface fold, counter-mirror collapse, engine registry/mutex, circuitbreaker trim) and the optional Phase 5 (package restructure).

---


eunolabs (commented):
Reviewed both commits independently. Both are behavior-preserving; no findings. (Would approve, but GitHub blocks self-approval for this account.)

**Commit 1 — `verifyAuditLog` → `auditChainVerifier` (the security-critical one).** I traced the new state machine against the old inline loop branch by branch, because a tamper-detection decomposition is exactly where a reordering can silently weaken a check:
- `decodeAuditRecord` is the old decode verbatim (`UseNumber` + `!dec.More()` trailing-data guard).
- The forged-decoy derivation (`postWrap`, `forgedSeq0`, `forgedLegacySeq0`) and the chain-update gate (`unmarshalOK && !forgedSeq0 && !forgedLegacySeq0`) are identical.
- `updateChain` reproduces the old chain block exactly: the seq-1 genesis-sentinel check, the `prev_hmac` mismatch, the seq-gap gated on `signedSeen` (not `prevSeq > 0`, preserving the post-wrap fix #629), and the anchor advance.
- `classify` reproduces the malformed → forged-seq0 → legacy/forged-legacy → HMAC-verify order; `continue`→`return` is equivalent.
- The load-bearing part: `updateChain` runs before `classify`, and `classify` reads chain state (`havePrev`) **only** in the `forgedSeq0` branch — which is precisely the record category for which `updateChain` is skipped, so no field it reads was mutated for that record. The split is clean.
- The one real change, collapsing `inReportWindow` (staticcheck QF1007), is logically identical by De Morgan: `!(reqID != "" && rec != reqID)` ≡ `reqID == "" || rec == reqID`. The dedicated forged-seq0 / forged-legacy / seq0-wrap / chain-break tests all pass under `-race`.

**Commit 2 — `cmdProxy` → flag/backend/sink helpers (mechanical).**
- All **65** flag `name/default/usage` triples across `main.go` are byte-identical before/after (empty diff), so no flag changed name, default, or help text.
- `openConfiguredAuditSink` and `buildCallCounterAndKillSwitch` are byte-identical to the old inline blocks, the only change being flag-pointer deref → parameter.
- The orchestration shell keeps every lifecycle detail: the `exitCode`/outermost-defer pattern, `sink.Close()` deferred **in the orchestrator** (not the helper — correct, or it'd close immediately), `ksRedis.Start(ctx)`, in-memory `StartCleanup`, SIGINT/SIGTERM handling, and the `proxyFlags` mapping (field-name-consistent, no swaps). The `--audit`/`--config` mutual-exclusion and config-precedence ordering are unchanged.

**Verification:** `go build`/`go vet` clean; full `cmd/eunox-mcp` suite green under `-race` locally; all 43 CI checks green. Clean finish to Phase 3 — LGTM.

---

PR #871, opened 2026-06-19
refactor(pdp): fold ListFilterer and SamplingAuthorizer into the PDP contract

eunolabs:
## What this is

The first of Phase 4's interface cleanups: **fold the optional PDP interfaces into one contract.**

`ListFilterer` (list filtering) and `SamplingAuthorizer` (sampling authorization) were optional PDP facets the transports detected by **type assertion**, with a silent fallback when a PDP didn't implement them (forward lists verbatim / deny-by-default sampling). This embeds both into `PolicyDecisionPoint`, so every PDP decides list filtering and sampling **explicitly** — the same "no silent fall-through" rule the core `Decide*` methods already enforce — and replaces the assertions with direct calls.

## The change

- **`PolicyDecisionPoint`** now embeds `ListFilterer` + `SamplingAuthorizer` (kept as named facets for their docs and method expressions).
- **`decideSampling`** and **`dispatchList`** call `pdp.DecideSampling` / the filter directly — the type assertions and the deny/verbatim fallbacks are gone.
- **`JWTPDP.DecideSampling`** keeps its `innerEnforces()` gate (an `alwaysAllowPDP` inner is still *not* a sampling backstop) but drops the inner `SamplingAuthorizer` assertion.
- **`JWTPDP.innerFilter`** calls the inner filter directly, guarding a **nil inner** (JWT-only mode) for the verbatim pass-through the assertion used to provide.
- **`alwaysAllowPDP`** gains verbatim `Filter*` methods, preserving wiretap-mode list pass-through.
- Four test doubles (`denyAllPDP`, `condPDP`, `staticPDP`, `recordingDecisionPoint`) implement the folded methods — reproducing their prior fallback (deny-by-default sampling, verbatim lists; the recording wrapper delegates).

## Why it's safe (behavior-preserving)

Every former fallback is now reproduced by a concrete method: `alwaysAllowPDP`, a JWT-only PDP (nil inner), and an `alwaysAllow` inner all pass lists through unchanged; a PDP with no sampling opt-in denies. One nil-pointer panic surfaced in testing (the JWT-only `innerFilter` path) and is fixed with the nil guard. The full `cmd/eunox-mcp` suite under **`-race`** — including the #686 wiretap-sampling and ADR-0001 JWT list-filter tests — passes; `go vet` + golangci-lint **0 issues**; gofmt + license clean. Updated the `CLAUDE.md` PDP note.

## Plan status

Phase 4 (interface cleanups), **item 1 of 4**. Remaining: collapse the call-counter mirror set, freeze the engine registry / drop its mutex, trim the circuit breaker.

---


cmd/eunox-mcp/pdp.go
```diff
@@ -96,14 +104,11 @@ type SamplingAuthorizer interface {
...
+// transport handler.  DecideSampling is part of the PDP contract, so every PDP
+// decides sampling explicitly — each is deny-by-default unless its policy carries
+// an explicit system:sampling/createMessage opt-in.
 func decideSampling(ctx context.Context, pdp PolicyDecisionPoint, sessionID, sourceIP string) capability.EnforceResponse {
-	if sa, ok := pdp.(SamplingAuthorizer); ok {
-		return sa.DecideSampling(ctx, sessionID, sourceIP)
-	}
-	return denyResponse("SAMPLING_DENIED", "",
-		"server-initiated sampling requires a PDP with an explicit system:sampling/createMessage opt-in")
+	return pdp.DecideSampling(ctx, sessionID, sourceIP)
```
eunolabs, line 111: The implementation now calls through the mandatory contract, but the `alwaysAllowPDP.DecideSampling` comment and `TestAlwaysAllowPDP_DecideSampling_AllowsAndIsSamplingAuthorizer` still say `decideSampling` performs an optional type assertion. That assertion is now guaranteed by assigning to `PolicyDecisionPoint`, so the test is tautological and the comments contradict this code. Please update them to call/assert `DecideSampling` through the PDP contract directly and describe the explicit wiretap allow behavior.

cmd/eunox-mcp/pdp_jwt.go
```diff
@@ -1235,19 +1233,21 @@ func (p *JWTPDP) FilterPromptsList(ctx context.Context, result json.RawMessage)
...
 	result json.RawMessage,
 	sel func(ListFilterer, context.Context, json.RawMessage) json.RawMessage,
 ) json.RawMessage {
-	if lf, ok := p.inner.(ListFilterer); ok {
-		return sel(lf, ctx, result)
+	if p.inner == nil {
+		return result
 	}
-	return result
+	return sel(p.inner, ctx, result)
```
eunolabs, line 1250: `TestInnerFilterList_NoListFilterer` and `TestInnerFilterToolsList_NoListFilterer` still construct `inner: alwaysAllowPDP{}`, but this PR makes `alwaysAllowPDP` implement `ListFilterer`. Those tests now exercise this delegate path, not a no-filterer/nil fallback, while their names and comments claim the opposite. Rename them as always-allow passthrough tests, or use a nil inner if the intent is to pin the new nil guard.

eunolabs (commented):
The runtime refactor looks behavior-preserving; `go test -race ./cmd/eunox-mcp`, `go vet ./cmd/eunox-mcp`, Go CI, and Demo Integration Tests pass. I left two non-runtime cleanup findings where the tests/comments still describe the optional-interface design this PR removes.

eunolabs (commented):
Reviewed as a behavior-sensitive interface fold — the risk here is that the old code had four *silent* type-assertion fallbacks (verbatim lists / deny-by-default sampling), so the question is whether each concrete PDP's now-mandatory method reproduces exactly what its fallback produced. It holds up; behavior-preserving, and a genuine fail-explicit improvement. (Would approve, but GitHub blocks self-approval for this account.)

**Production fallbacks were already dead code.** `ManifestPDP` already implements all four facets (`pdp.go:729/792/797/802`), and `JWTPDP`/`alwaysAllowPDP` already had `DecideSampling`. So every production route PDP satisfied both assertions before this PR — the `else deny` / `else verbatim` branches were only ever reachable by test doubles. Removing them and calling the methods directly is a no-op for production.

**The four fallback sites each reproduce their prior result:**
- `dispatchList` → `filter(d.pdp, …)`: `alwaysAllowPDP`'s new verbatim `Filter*` methods reproduce the old "no `ListFilterer` ⇒ skip filtering" pass-through; `ManifestPDP`/`JWTPDP` filter as before.
- `decideSampling` → direct call: prod PDPs all implemented it; the test doubles now deny explicitly (same decision/`SAMPLING_DENIED` code).
- `JWTPDP.DecideSampling`: the `innerEnforces()` gate is preserved, and I confirmed it still returns false for both a nil inner and an `alwaysAllowPDP` inner (`_, allowAll := p.inner.(alwaysAllowPDP); return !allowAll`) — so an unpoliced inner is still **not** a sampling backstop, which is the load-bearing correctness point here.
- `JWTPDP.innerFilter`: the new `if p.inner == nil` guard + `alwaysAllowPDP`'s verbatim methods together reproduce the old assertion's behavior for both nil-inner (JWT-only) and alwaysAllow-inner routes. The nil guard is a real necessity, not cosmetic: calling `sel(p.inner, …)` on a nil interface would panic, whereas the old `p.inner.(ListFilterer)` returned `ok=false` and fell through — good catch fixing it.

**Verification:** build + `go vet` clean; the #686 wiretap-sampling and JWT list-filter tests pass under `-race`; full `cmd/eunox-mcp` suite green under `-race`; all 43 CI checks green. Build passing also confirms no PDP implementer was missed (the now-required methods would fail to compile otherwise).

**One minor, optional follow-up (pre-existing file, not in this diff):** `TestAlwaysAllowPDP_DecideSampling_AllowsAndIsSamplingAuthorizer` in `pdp_test.go` still does `sa, ok := pdp.(SamplingAuthorizer); require.True(t, ok, …)`. Now that `PolicyDecisionPoint` embeds `SamplingAuthorizer`, that assertion is a tautology — it can never be false, and the #686 property it was written to guard (that `alwaysAllowPDP` implements the facet) is now compile-time enforced. The behavioral assertions just below it (`DecideSampling` → allow, `decideSampling` → allow) still carry the real coverage, so nothing is lost — but the type-assertion line is now dead and could be dropped or turned into a comment noting the contract is compile-enforced, so a future reader isn't puzzled by a runtime check of a structural guarantee. Non-blocking. LGTM.

---

cmd/eunox-mcp/pdp.go
eunolabs: Done in a8d3311. Rewrote the `alwaysAllowPDP.DecideSampling` comment to describe the explicit wiretap allow through the mandatory contract (the type assertion is gone), and renamed the test to `TestAlwaysAllowPDP_DecideSampling_WiretapAllows` — it drops the now-tautological `pdp.(SamplingAuthorizer)` assertion and calls `DecideSampling` directly through `PolicyDecisionPoint`.

---

cmd/eunox-mcp/pdp_jwt.go
eunolabs: Done in a8d3311. Renamed both to `TestInnerFilter{List,ToolsList}_AlwaysAllowInner_Passthrough` (they now exercise the verbatim-delegate path, since `alwaysAllowPDP` implements `ListFilterer`), and added `TestInnerFilter_NilInner_Passthrough` to pin the new nil-inner guard directly.

---

PR #877, opened 2026-06-19
chore: expand VS Code debug configs and add local gateway/HTTP configs

eunolabs:
## Summary

Reorganized and expanded `.vscode/launch.json` with 18 focused debug targets covering all CLI subcommands and transport modes, added three new local debug config files (`eunox.local.stdio.yaml`, `eunox.local.http.yaml`, `eunox.local.gateway.yaml`) for self-contained testing, and added VS Code extension recommendations.

## Why

The original launch.json had only 5 generic targets with hardcoded flags that didn't reflect the actual CLI surface or config-driven architecture. Developers had to manually construct commands or edit configs to test different scenarios (stdio vs. HTTP, enforce vs. audit, single upstream vs. gateway). The new configs and targets:

- Enable one-click debugging of each subcommand (`proxy`, `validate`, `init`, `suggest`, `stats`, `doctor`, `audit-verify`, `kill`, `version`)
- Provide self-contained local configs that require no external services (mock upstreams are spawned as subprocesses)
- Support the gateway multiplexing mode with two independent stdio upstreams
- Include a compound target to launch proxy + mock upstream together
- Document the intended usage pattern (paths, ports, prerequisites) in YAML comments

## Type of change

- Docs / site (debug configuration and developer experience)
- Refactor or chore (no behavior change)

## How it was tested

N/A — configuration and documentation only; no runtime behavior change.

## Security considerations

No security impact. These are local debug targets (loopback bind, no external network exposure by default).

## Checklist

- [x] `make check-license` passes — Apache-2.0 header added to new `.vscode/extensions.json`
- [x] No new source code or behavior changes — configuration and docs only
- [x] PR description explains the motivation (developer experience improvement)

https://claude.ai/code/session_01HTKCZ5KXRizrnUqesJo5ME


.vscode/launch.json
```diff
@@ -2,108 +2,231 @@
...
+      "console": "integratedTerminal",
       "args": [
         "proxy",
-        "--transport", "http",
-        "--port", "3000",
-        "--policy", "${workspaceFolder}/demo/manifest.yaml",
-        "--upstream-url", "http://localhost:8080",
-        "--audit-log", "${workspaceFolder}/demo/audit/audit.jsonl",
-        "--audit-key-path", "${workspaceFolder}/demo/audit/audit.key"
+        "--audit",
```
eunolabs, line 27: [P2] Point this wiretap at the demo audit tape. With only `--audit`, the proxy writes to the default `~/.eunox/audit.jsonl` and `~/.eunox/audit.key`, while the new `suggest`, `stats`, and `audit-verify` targets below all read `${workspaceFolder}/demo/audit/audit.*`. Running this capture target and then any of those readers therefore reports that the demo log is missing (or reads a different tape). Add matching `--audit-log` and `--audit-key-path` arguments here, or make all of the readers use the default location.

.vscode/launch.json
```diff
@@ -2,108 +2,231 @@
...
+    {
+      "name": "doctor: support bundle",
+      "type": "go",
+      "request": "launch",
+      "mode": "debug",
+      "program": "${workspaceFolder}/cmd/eunox-mcp",
+      "cwd": "${workspaceFolder}",
+      "args": [
+        "doctor",
+        "--config", "${workspaceFolder}/demo/eunox.local.stdio.yaml"
```
eunolabs: [P2] Pass the demo audit paths to `doctor` as well. `doctor --config` includes the config and manifests, but it does not derive its audit log/key flags from that config; without explicit `--audit-log` and `--audit-key-path`, this target reads `~/.eunox/audit.*`. I reproduced the target showing unrelated existing home-directory audit records instead of `demo/audit/audit.jsonl`, so the generated support bundle is misleading and can expose metadata from the wrong tape.

eunolabs (commented):
Reviewed the current head and exercised the new CLI/config paths. The strict YAML configs validate and the live stdio validation succeeds, but the audit-related launch targets are not wired to the same tape.

PR #878, opened 2026-06-19
refactor: finish Phase 4 interface cleanup (counter fold, registry freeze, circuitbreaker trim)

eunolabs:
## What & why

Finishes Phase 4 (interface cleanup) of the complexity-reduction plan with the three remaining items, each as its own commit. All three are behavior-preserving refactors that shrink interface surface and remove runtime optionality in favor of compile-time contracts — the same move as the PDP fold in #871. No manifest grammar, audit shape, config, or `go.mod` changes; the fail-closed posture is preserved throughout.

### 1. Fold the optional counter capabilities into `capability.CallCounter` (`a909263`)

`pkg/enforcement` declared three consumer-side mirror interfaces — `CallHistory`, `RateLimiter`, `RateLimitPeeker` — that a counter backend *might* additionally implement, detected at runtime by type assertion and failing closed when absent. Both built-in backends (`callcounter.InMemory`, `callcounter.Redis`) always implemented all three, so the optionality bought nothing but three extra interfaces and a fail-closed branch per condition.

`Peek`, `IncrementIfBelow`, and `PeekRetryAfter` now live directly on `capability.CallCounter`. Every counter satisfies the whole contract at compile time; the `maxCalls` / `sequenceBlock` handlers drop their type assertions for a single `if e.counter == nil` fail-closed guard. A backend that omits a method now breaks the build at `counter_contract.go` instead of failing closed at runtime with an opaque deny.

### 2. Freeze the condition-handler registry, drop the mutex (`05ad8eb`)

The `Engine` guarded its `ConditionHandler` map with a `sync.RWMutex` to allow `RegisterCondition` after construction, but handlers are only ever registered at construction time. Built-ins now register before the option loop runs, and custom handlers register through a new `WithConditionHandler` option, so the map is fully built by the time `New` returns and never mutated again. The mutex (and its per-evaluation `RLock`) is gone.

### 3. Trim the circuitbreaker public surface (`c3b6d6c`)

The breaker exposed its full internal mechanics — `Allow`, `AllowProbe`, `AllowFunc`, `RecordSuccess`, `RecordFailure`, `Reset`, and the `Probe` handle — but the only production caller (the JWKS fetch guard in `pkg/capability`) drives it exclusively through the generic `Do` helper. Those entry points invited callers to reimplement `Do`'s logic incorrectly — a non-atomic `Allow`-then-`Record` races the state machine.

Mechanics are now un-exported (`allowReq`, `allowProbe`, `markSuccess`, `markFailure`, `reset`, `probe`) and the unused `AllowFunc` is deleted, leaving `New`/`DefaultConfig`/`Config`/`WithClock`/`Do`/`ErrOpen`/`Breaker` as the construction-and-use surface and `State`/`Stats` as read-only observability. Tests that drove the mechanics moved into the package (white-box).

## Verification

- `go build ./...` — clean
- `make lint` (go vet + golangci-lint) — 0 issues
- `make check-fmt`, `make check-license`, `make check-notice` — clean (NOTICE in sync, 7 modules; `go.mod`/`go.sum` untouched)

---


pkg/circuitbreaker/breaker.go
```diff
@@ -180,81 +183,34 @@ func New(cfg Config, opts ...Option) *Breaker {
...
-// bare Allow, place `defer b.RecordFailure()` as the first statement after a true
-// admission and overwrite it with RecordSuccess only on the success path.
-func (b *Breaker) Allow() bool {
+// It returns only a bool, so the outcome must be reported through the separate
+// bare markSuccess / markFailure methods, which carry NO generation. That makes
+// this bare pattern safe ONLY for strictly synchronous, single-probe use where no
+// state transition can occur between admission and outcome. [Do] (via allowProbe)
+// is the generation- and panic-safe path; this wrapper exists for the package's
+// own synchronous tests.
+func (b *Breaker) allowReq() bool {
```
eunolabs, line 199: [P1] Mark these exported API removals as breaking. This PR removes/renames `Breaker.Allow`, `AllowProbe`, `AllowFunc`, `RecordSuccess`, `RecordFailure`, `Reset`, and `Probe`; it also removes `Engine.RegisterCondition` and expands `capability.CallCounter`, so downstream consumers will stop compiling. `CONTRIBUTING.md` requires a `!` commit prefix plus a `BREAKING CHANGE:` migration footer, and the PR description must flag the break; the current `refactor:` commits can be omitted from generated release notes while the PR calls the work behavior-preserving. Pre-1.0 means no compatibility shim is required, but the break still needs explicit release metadata and migration guidance.

eunolabs (commented):
Full runtime verification passes (`go test -race -count=1 ./...`, `go vet ./...`, Go CI, and Demo Integration Tests). I found one blocking release/API issue and three documentation inconsistencies introduced or exposed by the public-surface changes.

pkg/enforcement/engine.go
```diff
@@ -509,12 +454,14 @@ func recordFailureDenial(requestID, now string, auditOnly bool, obligations []ca
...
-// for the same condition type.
-func (e *Engine) RegisterCondition(name string, handler ConditionHandler) {
-	e.mu.Lock()
-	defer e.mu.Unlock()
-	e.handlers[name] = handler
+// WithConditionHandler registers a custom condition handler under name. Applied
+// after the built-ins (see New), it overwrites a built-in handler of the same
+// type. Registration happens during construction, so the handler map is frozen
+// by the time New returns and the engine reads it lock-free on the hot path.
+func WithConditionHandler(name string, handler ConditionHandler) Option {
```
eunolabs, line 461: [P2] Update the public custom-condition documentation with this constructor-time API. `docs/capability-manifest-guide.md` still tells users to call `RegisterCondition`, and `site/public/features/index.html` contains an `engine.RegisterCondition("denyBlockedRecipient", ...)` snippet that no longer compiles. The replacement also registers one handler under `capability.ConditionTypeCustom`, with that handler responsible for dispatching on `CustomCondition.Name`; update both the prose and example so users do not migrate to the wrong key.

pkg/capability/enforce.go
```diff
@@ -147,4 +150,27 @@ type CallCounter interface {
 	// most maxEntries most-recent in-window timestamps so storage stays bounded
 	// under sustained traffic. maxEntries must be >= 1 (issue #233).
 	IncrementAndGet(ctx context.Context, key string, windowSec, maxEntries int) (int64, error)
+
+	// Peek reports a key's in-window count WITHOUT recording a call. The
```
eunolabs, line 146: [P2] Rewrite the `CallCounter` godoc above this declaration. Lines 121–147 still describe this as a minimal interface, say the engine discovers optional methods by type assertion, and recommend assertions against the now-removed `enforcement.RateLimiter` / `enforcement.CallHistory` interfaces. That is the exact opposite of the mandatory contract introduced here and will send custom backend authors toward APIs that no longer exist.

pkg/circuitbreaker/breaker.go
```diff
@@ -80,25 +82,27 @@ func (c Config) Validate() error {
 	return nil
 }
 
-// Breaker implements the circuit breaker pattern.
+// Breaker implements the circuit breaker pattern. Its sole public operation is
```
eunolabs: [P3] This public-surface description contradicts the resulting API: `State`, the state constants, `Breaker.State()`, and `Breaker.Stats()` remain exported, so `Do` is not the sole public operation and `State` is not internal as stated above. `do.go` also still says a wedged breaker rejects traffic “until Reset” even though `Reset` is no longer callable outside this package. Please make the godoc describe the actual construction/use plus observability surface.

pkg/circuitbreaker/breaker.go
eunolabs: pre-release. nothing's breaking.

eunolabs (commented):
Reviewed all three commits independently. Each is behavior-preserving; no findings — same fail-explicit-over-fail-silent move as #871, applied cleanly. (Would approve, but GitHub blocks self-approval for this account.)

**1. `05ad8eb` — freeze the registry, drop the mutex (the concurrency-critical one).** Dropping a lock is only safe if the map is provably built-once-before-any-read and never written again, so I checked that directly:
- The only writes to `e.handlers` are the `make` in `New`, `registerBuiltins`, and the new `WithConditionHandler` option — all during construction, all on the constructing goroutine. `RegisterCondition` is fully removed (no callers anywhere), so nothing can mutate the map post-`New`.
- Moving `registerBuiltins()` *before* the option loop is safe because the built-ins are registered as **method values** (`ConditionHandlerFunc(e.handleMaxCalls)`), which read `e.counter`/`e.clock` dynamically via the receiver at eval time — nothing is captured by value at registration, so seeing default fields during `registerBuiltins` doesn't matter. The build-ins-first/options-second order preserves the old override semantics (a custom handler still beats a built-in of the same type).
- After `New` returns the map is read-only, so the lock-free `evalCondition` read is race-free. The enforcement `-race` suite passes.
- Minor forward-looking note (non-blocking): the lock-free read now depends on the "never written after `New`" invariant, which is documented on the field but no longer compile-enforced. A future lazy-registration would silently reintroduce a data race — worth keeping in mind, though `-race` tests would likely catch it.

**2. `a909263` — fold the optional counter capabilities into `capability.CallCounter`.** The fail-closed posture is preserved: the old `if rl, ok := e.counter.(RateLimiter); !ok` branch was only ever reachable by a non-nil counter lacking the method, which never existed (both built-ins implemented all three), so the new `if e.counter == nil` guard reproduces the only trigger that fired in practice. The check-only `maxCalls` path sits after that nil guard, so `e.counter.PeekRetryAfter` can't nil-deref, and for the built-in backends it already took the `RateLimitPeeker` (tight retry-after) branch — the new direct call matches. `recordSessionCall`'s dropped `CallHistory` assertion is equivalent (non-nil counters always implement `Peek` now). Only free-form `Message` strings changed; `Code`/`ConditionType` are unchanged. The `callcounter.Store` → `capability.CallCounter` retyping on the cmd wiring is *necessary* (the narrow producer interface no longer satisfies the widened consumer contract) and safe — concrete backends satisfy it (pinned in `counter_contract.go`), and `StartCleanup` is reached via concrete type-assertion, not the interface.

**3. `c3b6d6c` — trim the circuitbreaker surface.** Pure API-surface reduction. The only external callers (`pdp_jwt.go`, `jwks.go`, `jwkscache.go`) use just `Breaker`/`New`/`DefaultConfig`/`Do`/`ErrOpen` — none touch the now-unexported `Allow`/`AllowProbe`/`Record*`/`Reset`/`Probe`/`AllowFunc`, so unexporting them and deleting `AllowFunc` breaks nothing. `Do` drives the renamed internals from inside the package; the state-machine logic is unchanged, and the white-box `-race` tests confirm it. Good call removing `AllowFunc` — a non-atomic `Allow`-then-`Record` was a footgun.

**Verification:** `go build ./...` + `go vet ./...` clean; `-race` green on `circuitbreaker`, `enforcement`, `callcounter`, `capability`, and `cmd/eunox-mcp` locally; all 43 CI checks green; build passing also confirms no `CallCounter` implementer or circuitbreaker caller was missed. Clean finish to Phase 4 — LGTM.

---

pkg/capability/enforce.go
eunolabs: Fixed in fa35cda. Rewrote the `CallCounter` godoc: it now documents the mandatory full contract (every method required), explains the Phase 4 fold of `Peek`/`IncrementIfBelow`/`PeekRetryAfter` off the now-removed optional mirror interfaces, and recommends `var _ capability.CallCounter = (*MyCounter)(nil)` (producer-side mirror `callcounter.FullStore`) in place of the gone `enforcement.RateLimiter`/`CallHistory` assertions.

---

pkg/enforcement/engine.go
eunolabs: Fixed in fa35cda. Updated both the guide prose and the site's Go-handler snippet to register at construction via `WithConditionHandler(capability.ConditionTypeCustom, …)` with the handler dispatching on `CustomCondition.Name` — no more `RegisterCondition`, and the example now keys on `ConditionTypeCustom` rather than the condition name. The Policy YAML (`type: custom`, `name: denyBlockedRecipient`) is unchanged and lines up with the name-dispatch.

---

pkg/circuitbreaker/breaker.go
eunolabs: Fixed in fa35cda. The `State` and `Breaker` godoc now describe the real surface — construction (`New`/`DefaultConfig`/`Config`/`WithClock`), use (`Do`), and read-only observability (`Breaker.State()`/`Breaker.Stats()`) — rather than calling `Do` the sole public operation with `State` internal. Also dropped do.go's dangling "until Reset" (now "reject all traffic indefinitely … with no further outcome arriving to close or re-open the window") since `Reset` is no longer callable outside the package.

---

PR #882, opened 2026-06-19
refactor(audit): decouple the audit subsystem from the JWT/PDP layer (Phase 5, step 1)

eunolabs:

The complexity audit's final phase promotes the cohesive subsystems out of the ~16K-LOC `package main` into `internal/` packages (`internal/transport|audit|pdp`) with explicit seams — *"once the seams are clean, so the move is a relocation rather than a redesign."* The audit subsystem is the cleanest candidate to go first.

A coupling analysis of the audit files against the rest of `main` found the seam is almost clean — **just two back-references** into the binary that would otherwise force an impossible `internal/audit → package main` import cycle:

| Back-ref | Defined in | Used by audit for |
|---|---|---|
| `mcpMethodToTargetType` | `pdp.go` (but **only** audit calls it) | deriving a record's target type |
| `jwtClaimsFromContext` | `pdp_jwt.go` | stamping `agent_id` / `task_id` (one call site, two fields) |

Per the plan's own sequencing rule — **simplify interfaces before you relocate** — this PR severs both so the audit subsystem has **zero** references back into `main`, unblocking a clean mechanical relocation next. No package moves yet; this is the seam-cleaning step.

## What changed

1. **`mcpMethodToTargetType` moves `pdp.go` → `audit.go`.** `pdp.go` only *declared* it; audit's `deriveTargetFields` is the sole caller. It's a pure `MCP method → capability.TargetType` mapping, so it belongs with the record-target derivation that consumes it.
2. **The JWT identity stamp becomes an injected extractor.** `record()` previously called `jwtClaimsFromContext(ctx)` directly. It now reads identity through a `func(context.Context) (agentID, taskID string)` supplied at construction via a new `withAuditIdentity(...)` option, wired in `main.go` from a small `auditIdentityFromContext` adapter. The dependency now points inward — audit reads identity from the request context without importing the JWT layer that owns the claims.

A `nil` extractor leaves `agent_id`/`task_id` omitted exactly as before, and the attacker-controlled-claim field bounding (#743) stays in `record()`.

## Why it's safe

Behavior-preserving — the JWT-identity audit tests assert the same stamped `agent_id`/`task_id`, now exercised through the **production wiring**: the shared `newTempAuditSink` test helper and the oversized-claim bounding test wire the same extractor `main` uses, so the tests cover the real path rather than the old in-sink coupling.

## Validation

- `go test -race -count=1 ./cmd/eunox-mcp/` — green (the three JWT-identity tests pass through the injected extractor)
- `make lint` (0 issues), `make check-fmt`, `make check-license` — clean
- No `go.mod`/grammar/audit-shape changes

## Next

The mechanical relocation — moving `audit.go` / `audit_rotate.go` / `audit_verify.go` (+ their unit tests) into `internal/audit`, exporting the sink API, and updating `main` (integration tests like the 3.3K-line `audit_test.go` stay in `main` on the exported API) — follows as its own focused PR now that the seam is clean.

---


eunolabs (commented):
Reviewed the current head (`2106cfc1`). No actionable findings. I traced every `openAuditSink` construction path and the JWT-context audit paths: the sole production constructor injects `auditIdentityFromContext`, JWT identity remains bounded in `record()`, nil/no-claims contexts still omit the fields, and the method-to-target mapping move is behavior-preserving. Verification: `git diff --check`, `go vet ./...`, and `go test -race -count=1 ./cmd/eunox-mcp/` all pass.

eunolabs (commented):
Reviewed as a seam-cleaning step — small surface, but the point is whether the audit subsystem truly ends up with zero back-references into `main` and whether the identity-stamp indirection is behavior-preserving. Both hold. (Would approve, but GitHub blocks self-approval for this account.)

**Seam goal achieved (verified):** grepping `audit.go` / `audit_rotate.go` / `audit_verify.go` for `jwtClaimsFromContext`, `JWTClaims`, `*PDP`, `withJWTClaims`, `EnforceTarget`, `httpSession`, `StdioProxy`, etc. returns nothing — the audit files now reference no JWT/PDP/transport symbol. The extractor's `func(context.Context) (agentID, taskID string)` signature returns plain strings (not `*JWTClaims`), so audit doesn't even reference the claims type; the `auditIdentityFromContext` adapter lives in `pdp_jwt.go` (the JWT layer), so the dependency points inward. The next mechanical relocation into `internal/audit` is genuinely unblocked.

**`mcpMethodToTargetType` move** is verbatim (diffed old `pdp.go` body vs new `audit.go` body — identical), and its only *production* caller is audit's `deriveTargetFields` (the other references are direct unit tests, fine in-package). Belongs with the target derivation that consumes it.

**Injected identity extractor is behavior-preserving:**
- The only *production* `openAuditSink` caller is `main.go:492` (`openConfiguredAuditSink`), which now wires `withAuditIdentity(auditIdentityFromContext)` — so the one record-writing sink in the binary stamps identity exactly as before.
- The extractor reproduces the old inline logic precisely: claims present → `(AgentID, TaskID)` then bounded; absent → `("","")` → `omitempty` omits the fields. The #743 attacker-controlled-claim bounding stays in `record()`, applied to the extractor's output — so even a future extractor's values are bounded.
- `nil` extractor (test sinks not wiring it) leaves the fields empty/omitted, matching the "no JWT" path; none of those tests assert on identity.
- The `identity` field is written only by the `withAuditIdentity` option during construction (before the drainer starts) and read-only afterward, so concurrent `record()` reads are race-free.

**Test coverage actually improves:** all four identity-asserting tests — `TestAuditRecord_StampsJWTAgentAndTask`, `TestSamplingAuditRecord_CarriesSessionJWTIdentity`, `TestSamplingAuditRecord_NoJWT_OmitsIdentity` (via `newTempAuditSink`, now extractor-wired), and `TestRecordOversizedAgentTaskIDBounded` (wires it directly) — now exercise the production wiring rather than the old in-sink coupling.

**Verification:** `go build ./...` + `go vet` clean; identity/target-type tests and full `cmd/eunox-mcp` suite green under `-race` locally; all 43 CI checks green. LGTM — clean seam for Phase 5.

---

PR #885, opened 2026-06-19
refactor(audit): relocate the audit subsystem into internal/audit (Phase 5)

eunolabs:

Continues the audit-extraction work from #882 (the seam-decoupling, now merged). With the audit subsystem holding **zero** back-references into the binary, this PR performs the actual relocation: it moves the tamper-evident audit log out of the ~16K-LOC `package main` into its own `internal/audit` package, giving it an explicit, narrow seam instead of being one of dozens of concerns sharing the binary's namespace.

## What moved

`cmd/eunox-mcp/audit.go` / `audit_rotate.go` / `audit_verify.go` → `internal/audit/{audit,rotate,verify}.go` (`package audit`). Git tracked all moves as renames (92–99% similarity). The package **builds standalone** and depends only on `pkg/capability` + the stdlib.

## The seam (de-stuttered exports)

`Sink`, `Open`, `Option`/`WithIdentity`, `VerifyLog`/`VerifyResult`, `NewVerifier`, `KeyID`, `NewLineScanner`, `LoadOrCreateKeys`, `Resolve{Log,Key}Path`, `DeriveTargetFields`, `CloneAndBoundDetails`, `ExpandHome`, `TruncatedKey`, plus the existing `RecordAllow`/`RecordDeny`/`Close`/`VerifyRecord`/`DroppedRecords`/`WriteFailures` methods.

Two exports beyond the obvious rename map, both deliberate:
- **`Record`** — the gateway-aware recorder `routeSink` calls directly.
- **`NewVerifier(keys [][]byte)`** — lets the `audit-verify` subcommand build a multi-key verifier without exposing the sink's raw key fields.

Everything else — `drain`, `rotate`, `pruneRotated`, the chain verifier, `cloneAndBound`, field bounding, the OCSF record type, every internal const — stays **unexported** and travels with its white-box tests (consistent with Phase 4's "un-export the mechanics" direction).

## Tests — split, none dropped

Audit tests split by what they touch:
- The 13 audit-only test files + the white-box test functions (including audit tests that had been scattered through `low_severity_fixes` / `security` / `operational_hardening`) move to `package audit`, next to the internals they exercise.
- Integration tests (transports, PDPs, the real JWT→audit wiring) **stay** in `package main` on the exported API.
- Of `audit_test.go`'s 80 functions, only 3 were genuinely entangled; the one needing the real JWT wiring (`TestRecordOversizedAgentTaskIDBounded`) stays in `main` with its two internal refs rewritten to the public surface (read the record back from the JSONL line; assert the documented 8 KiB bound) — every behavioral assertion preserved.

The repo test-function count is **unchanged at 1994** — no test dropped or weakened.

## Validation

- `go build ./...` — clean
- `go test -race -count=1 ./...` — **all green**, including the new `internal/audit` (which now carries the bulk of the audit tests; `cmd/eunox-mcp` drops from ~70s to ~15s)
- `make lint` (0 issues), `make check-fmt`, `make check-license`, `make check-notice` (7 modules, no new deps) — clean
- `.golangci.yml` grandfather exclusions follow the renamed/moved functions (`verifyAuditLog`→`VerifyLog`, `openAuditSink`→`Open`, the `hugeParam` chain-verifier path); **no new suppressions**
- `CLAUDE.md` architecture section updated to describe the new package
- No `go.mod` / manifest-grammar / audit-shape changes

---


eunolabs:
1. docs/architecture.md still describes the old audit package location.
2. CLAUDE.md incorrectly presents package-level audit.VerifyLog as a Sink method.

cmd/eunox-mcp/main.go
```diff
@@ -901,13 +902,13 @@ func serveStdioHost(ctx context.Context, cfg *GatewayConfig, sink *auditSink, co
...
-	// audit (observe) mode is explicitly acknowledged, mirroring BuildRoutes
+	// auditMode (observe) mode is explicitly acknowledged, mirroring BuildRoutes
 	// (route.go) so both host transports reject the same misconfiguration rather
 	// than the stdio host silently booting allow-all (issue #552).
-	if len(u.Policy) == 0 && !audit {
+	if len(u.Policy) == 0 && !auditMode {
 		return fmt.Errorf("upstream %q: no policy configured and enforcement is not %q — "+
 			"every call would be allowed unenforced. Add 'policy:' to enforce, "+
-			"or set 'enforcement: audit' to run in observe-only (wiretap) mode",
+			"or set 'enforcement: auditMode' to run in observe-only (wiretap) mode",
```
eunolabs: **Bug — corrupted recovery instruction (introduced by this PR).** The `audit` → `auditMode` local rename (added to avoid shadowing the new `audit` import) bled into this user-facing error *string*. The valid config value is `audit` — the `%q` arg is `capability.EnforcementAudit` (`"audit"`) — so this now tells the operator to set `enforcement: auditMode`, which doesn't exist; copying it verbatim yields a config-validation rejection. The sibling guard in `route.go:155` correctly emits `'enforcement: audit'`, and the comment at line 905 even claims to "mirror BuildRoutes" while no longer doing so.

```suggestion
			"or set 'enforcement: audit' to run in observe-only (wiretap) mode",
```

The same rename also left stale prose nearby: **L896** `--auditMode` (the flag is `--audit`), **L905** `auditMode (observe) mode`, **L928** `auditMode route` — all should read `audit`.

---

internal/audit/audit.go
```diff
@@ -97,22 +104,22 @@ func nextRequestID() string {
 	return "req-" + sessionNonce + "-" + strconv.FormatUint(reqCounter.Add(1), 16)
 }
 
-// auditKeyID returns a stable, non-secret identifier for an HMAC key, derived as
+// KeyID returns a stable, non-secret identifier for an HMAC key, derived as
 // the first 16 hex characters of SHA-256(key). It is stamped on every record so
 // audit-verify can select the key that signed a record after rotation (§ 7.4).
 // Being a one-way digest, it discloses nothing about the key material.
-func auditKeyID(key []byte) string {
+func KeyID(key []byte) string {
```
eunolabs: **Needless export.** `KeyID` (was unexported `auditKeyID`) has **no** caller in `cmd/eunox-mcp` — the one prior use (the audit-verify subcommand building its verifier) is now absorbed by `NewVerifier`, and `KeyID` is referenced only within `internal/audit`. Exporting it widens the package's public surface with an internal HMAC-key-id detail for no functional reason. Suggest unexporting back to `keyID`.

---

internal/audit/audit.go
```diff
@@ -918,7 +951,7 @@ const auditTruncatedKey = "_eunox_truncated"
 // stays absent rather than serializing as `{}`. record() and syntheticDenyMarker
 // both apply it before enqueue so the bounded queue never retains an un-truncated
 // or aliased payload (#461).
-func cloneAndBoundDetails(m map[string]interface{}) map[string]interface{} {
+func CloneAndBoundDetails(m map[string]interface{}) map[string]interface{} {
```
eunolabs: **Export forced only by a misplaced test.** `CloneAndBoundDetails` has no production caller in `cmd`; its sole caller is the white-box unit test `TestBoundAuditDetails_TotalCapReplacesMap` (`cmd/eunox-mcp/audit_test.go:1213`), which references only audit symbols and therefore belongs in `internal/audit` next to the other relocated white-box tests. Moving that one test lets this revert to unexported `cloneAndBoundDetails` — it's an internal record-construction helper, not part of the binary's API.

---

internal/audit/audit.go
```diff
@@ -1072,7 +1095,17 @@ func mcpMethodToTargetType(method string) (capability.TargetType, error) {
...
+// the MCP method. target_type is taken from the method (authoritative), never
+// inferred from the overloaded identifier, so an opaque resource URI (urn:…,
+// mailto:…, memory:foo) or an oddly-named tool (one containing "://" or starting
+// with "prompts/") is recorded under its true namespace rather than a guess.
+// Returns empty strings for records with no MCP method — e.g. a JWT rejection
+// before dispatch — which carry no enforcement target.  For post-dispatch
+// records where the method exists but is not a recognized request method (e.g.
+// "tools/execute"), the raw method is preserved so audit consumers can
+// distinguish unmapped-method denials from pre-dispatch denials.
+func DeriveTargetFields(method, identifier string) (mcpMethod, targetType, target string) {
```
eunolabs: **Needless export.** `DeriveTargetFields` (was `deriveTargetFields`) has no call site in `cmd` — the only reference is a *comment* in `suggest.go:106`. It's used solely inside `internal/audit` for record-target derivation. Suggest unexporting to `deriveTargetFields` (and tweaking the suggest.go comment). Together with `KeyID`/`CloneAndBoundDetails`, these three were capitalized purely as a side effect of the move and widen the seam beyond what the binary calls.

---

cmd/eunox-mcp/control_token.go
```diff
@@ -46,7 +47,7 @@ func writeControlTokenFile(path, token string) (string, error) {
 	if path == "" {
 		path = defaultControlTokenPath
 	}
-	expanded, err := expandHome(path)
+	expanded, err := audit.ExpandHome(path)
```
eunolabs: **Layering smell.** The loopback control-token path now imports `internal/audit` *solely* for `audit.ExpandHome` (`~` expansion). On `main`, `expandHome` was a generic path helper that merely happened to live in `audit.go` and was already shared here without any audit coupling. Exporting a generic utility *from the audit subsystem* and pointing an unrelated security path at it is the wrong altitude — prefer relocating the `~`-expander to a neutral home (a small cmd-local helper, or a shared util) over `audit.ExpandHome`.

Minor, same file: the `internal/audit` import (line 5) sits inside the stdlib group — should be grouped with the `github.com/eunolabs/...` imports like `main.go`/`route.go` do.

---

cmd/eunox-mcp/audit_envelope_bound_test.go
```diff
@@ -177,23 +45,33 @@ func TestRecordOversizedAgentTaskIDBounded(t *testing.T) {
...
+	}
 	if err := json.Unmarshal([]byte(line), &rec); err != nil {
 		t.Fatalf("written record does not parse: %v", err)
 	}
-	if len(rec.AgentID) > auditEnvelopeFieldCap {
-		t.Errorf("AgentID not bounded: len=%d, want <= %d", len(rec.AgentID), auditEnvelopeFieldCap)
+	// The audit envelope field cap is a documented 8 KiB bound
+	// (audit.auditEnvelopeFieldCap, unexported); assert it as a literal here since
+	// this test lives in package main.
+	const envelopeFieldCap = 8 << 10
```
eunolabs: Minor coverage drift from keeping this test in `package main` (which it must, to exercise the real `withJWTClaims` + `auditIdentityFromContext` wiring):

- **L69**: the per-field bound is asserted against a hardcoded `8 << 10` rather than the live `auditEnvelopeFieldCap` constant, so if the cap is ever tightened this test silently won't track it.
- **L48**: `lastNonBlankLine` reads the whole file, replacing the original `mustReadLastAuditLine`, which exercised the production *bounded* 4 MiB-tail `readLastAuditLine` — so the "oversized record vs. real tail window" interaction this test's own comment references is no longer covered on the cmd side.

Both survive at full fidelity in the `internal/audit` white-box copies, so this is acceptable — noting it since this is now the only cmd-side check of agent_id/task_id bounding.

---

eunolabs (commented):
Reviewed the audit relocation at high recall — four parallel finder passes (correctness, removed-behavior, cross-file / Go pitfalls, test-split integrity, conventions / over-export) plus direct verification (`go build ./...`, full `go test -race ./...`, `go vet`).

**Verdict: behavior-preserving and well-executed.** `NewVerifier` is equivalent to — and safer than — the old `&auditSink{...}` literal (it guards the empty-keys panic); the gateway `routeSink.Record` call is argument-identical to the old `record`; cross-package const/type identity holds (`TruncatedKey`, the OCSF JSON tags, `VerifyResult` fields); the JWT identity extractor is wired on **both** the stdio and gateway sink-construction paths; and the repo test-function count is unchanged at 1994 with no dropped or weakened security assertion.

**Findings** — 1 correctness, the rest cleanup / altitude / nits (inline):

- 🐞 **`main.go:911` (worth fixing before merge)** — the `audit`→`auditMode` local rename (added to avoid shadowing the new `audit` import) bled into a user-facing recovery instruction (`'enforcement: auditMode'`, an invalid config value) plus three comments (L896/905/928). `route.go:155` — which it claims to mirror — correctly says `'enforcement: audit'`.
- 🧹 **Three needless exports** (`KeyID`, `DeriveTargetFields`, `CloneAndBoundDetails`) — capitalized purely by the move, with no production caller; unexporting keeps the seam to what the binary actually calls (one requires relocating a single white-box unit test into the package).
- 🏗️ **`control_token.go`** now depends on the audit package *solely* for `~`-expansion (`audit.ExpandHome`) — a generic helper exported from a security subsystem; better relocated to a neutral home.
- **nits**: `internal/audit` import landed in the stdlib group in `control_token.go` **and `doctor.go`** (should group with the other `github.com/eunolabs/...` imports, as `main.go`/`route.go` do); minor assertion drift (literal cap + unbounded tail reader) in the one envelope-bound test that stays in `package main`.

None of these block the relocation's correctness. Happy to push fixes for any of them — the `main.go:911` string in particular, since it's a regression in operator-facing guidance.

---

cmd/eunox-mcp/main.go
eunolabs: Fixed in cfe4efa — restored `'enforcement: audit'` (the suggested edit) and reverted the three `auditMode` comments (L896/905/928) back to `audit`.

---

internal/audit/audit.go
eunolabs: Fixed in cfe4efa — unexported. Named it `hmacKeyID` rather than `keyID` to avoid colliding with the existing `keyID` field/param (`keysToTry`).

---

internal/audit/audit.go
eunolabs: Fixed in cfe4efa — unexported to `cloneAndBoundDetails` and moved `TestBoundAuditDetails_TotalCapReplacesMap` into `internal/audit/details_bound_test.go` (white-box).

---

internal/audit/audit.go
eunolabs: Fixed in cfe4efa — unexported to `deriveTargetFields` and updated the `suggest.go` comment.

---

cmd/eunox-mcp/control_token.go
eunolabs: Fixed in cfe4efa — `control_token.go` now has its own small cmd-local `expandHome` and no longer imports `internal/audit` at all (audit's `ExpandHome` is unexported, used only internally by `Resolve*Path`). The import-grouping nit is moot as a result; the package-main expand-home tests now exercise the cmd-local helper.

---

cmd/eunox-mcp/audit_envelope_bound_test.go
eunolabs: Fixed in cfe4efa — added an `internal/audit` white-box test (`TestRecordBoundsAgentTaskID`) that drives the identity extractor and asserts agent_id/task_id bounding against the real `auditEnvelopeFieldCap` via the bounded `mustReadLastAuditLine`. The cmd test drops the `8<<10` literal and stays a pure JWT-wiring smoke check — so the precise bound is now covered with the real constant + bounded reader.

---

eunolabs (commented):
Re-reviewed after the `cfe4efa` fixes. The prior round already caught the substantive issues (the `auditMode` string-corruption bug + three needless exports + a layering smell), all resolved — so I focused on independently confirming the fixes landed *and* sweeping for anything that round might have missed, given a mechanical move-with-rename is exactly where subtle artifacts hide. It's clean. (Would approve, but GitHub blocks self-approval for this account.)

**Prior-fix verification — all six landed correctly:**
- `'enforcement: audit'` restored (`main.go:911`); the three `auditMode` comment artifacts reverted to `audit`.
- `hmacKeyID`, `cloneAndBoundDetails`, `deriveTargetFields` are all unexported again; `ExpandHome` is gone from the surface (control_token.go now has its own cmd-local `expandHome`).

**Independent checks beyond the prior round:**
- **Sibling `auditMode` artifacts:** swept every `auditMode` occurrence across `cmd/`. All are legitimate — the renamed local (`main.go:893`) and the *pre-existing* `auditModeFor` method (confirmed present on base, 5 refs). None bleed into any string literal or comment. The one bug of this class had no siblings.
- **Needless exports:** cross-referenced every package-level export against actual `cmd` usage. Each one (`Open`, `WithIdentity`, `LoadOrCreateKeys`, `NewLineScanner`, `NewVerifier`, `Resolve{Log,Key}Path`, `VerifyLog`, `TruncatedKey`, `Sink`, plus `Option`/`VerifyResult` as returned types and the `Record`/`Record*` methods on `Sink`) has a real cmd consumer — no over-exports remain beyond the three already fixed. The seam is minimal.
- **Clean-seam dependency claim:** `internal/audit`'s *direct* imports are `pkg/capability` only (the `circuitbreaker` that shows in `go list -deps` is transitive via `capability`). Standalone-buildable as advertised.
- **No tests dropped:** base→head delta on `func Test|Fuzz` is **+1** (exactly the `TestRecordBoundsAgentTaskID` added in `cfe4efa` for the coverage-drift fix), so the split moved tests rather than losing them.

**Verification:** `go build ./...` + `go vet ./...` clean; `go test -race ./...` green — `internal/audit` (47s, now carrying the bulk of audit tests) and `cmd/eunox-mcp` (down to ~14s); all 43 CI checks green.

Minor factual nit (non-blocking): the description says the test count is "unchanged at 1994," but a raw `func Test|Fuzz` grep gives 1986 on head (1985 on base). The headline claim — *no test dropped or weakened* — does hold (verified via the +1 base/head delta); the absolute figure is just a different counting method. LGTM — clean relocation.

---

PR #909, opened 2026-06-19
refactor(mcp): promote the MCP protocol types to internal/mcp (Phase 5 foundation)

eunolabs:
## Context — Phase 5 (package restructure)

With `internal/audit` extracted (#882, #885), the remaining Phase 5 promotions are `internal/pdp` and `internal/transport` — but both share a foundation: the **MCP protocol message types** (`mcpToolEntry`, `mcpToolCallResult`, `mcpResourceEntry`, the list/params/result shapes). These DTOs sit *below* both layers — the policy decision points filter list responses with them, and the transports marshal messages with them — so neither extracts cleanly until the types have a home. The correct layering is:

```
internal/mcp  ←  internal/pdp  ←  internal/transport
```

This PR promotes that foundation. It removes 7 of `pdp`'s 13 back-references into `main` and unblocks the transport extraction too.

## What moved

12 pure JSON structs → `internal/mcp/types.go` (`package mcp`), de-stuttered:

| was (package main) | now |
|---|---|
| `mcpInitResult` | `mcp.InitResult` |
| `mcpToolCallParams` / `mcpToolCallResult` / `mcpContent` | `mcp.ToolCallParams` / `mcp.ToolCallResult` / `mcp.Content` |
| `mcpToolsListResult` / `mcpToolEntry` | `mcp.ToolsListResult` / `mcp.ToolEntry` |
| `mcpResourceReadParams` / `mcpResourcesListResult` / `mcpResourceEntry` | `mcp.ResourceReadParams` / `mcp.ResourcesListResult` / `mcp.ResourceEntry` |
| `mcpPromptsListResult` / `mcpPromptEntry` / `mcpPromptGetParams` | `mcp.PromptsListResult` / `mcp.PromptEntry` / `mcp.PromptGetParams` |

The package has **no methods**, depends only on `encoding/json`, and builds standalone with **zero references back into the binary**.

Everything that *uses* the types stays in `cmd/eunox-mcp` (the handshake builders, `decodeParams`, `parseToolsListResult`, the PDP list filters, the drift comparison) — their references are qualified to `mcp.X`. **No test-split**: the DTOs are pure public structs, so every test stays in `package main` and just references `mcp.X`.

## Validation

- `go build ./...` — clean; `internal/mcp` builds standalone (stdlib only)
- `go test -race -count=1 ./...` — all green
- `make lint` (0 issues), `make check-fmt`, `make check-license`, `make check-notice` — clean
[...]

PR #910, opened 2026-06-19
refactor(pdp): decouple the PDP layer from package main for extraction (Phase 5)

eunolabs:

Continues the package promotions: `internal/audit` (#882/#885) and `internal/mcp` (#909) are merged. Next is `internal/pdp`. Following the established **decouple-then-relocate** cadence (as with audit), this PR is the seam-cleaning step so the relocation can be a clean move rather than a redesign.

Re-analyzed on current main, the PDP layer's references into `package main` are down to **3 real ones** (the `internal/mcp` merge removed the 7 wire-type back-deps; `forward`/`sourceIP`/`validateLocalManifest` turned out to be false positives — they appear only in comments or as the PDP-interface *parameter* name). This PR severs all three.

## What changed

**1. `ManifestPDP` no longer depends on `LocalManifest`.** It held a `*LocalManifest` but only ever read `.Capabilities`. It now stores `[]capability.Constraint` directly; `NewManifestPDP` takes the slice and callers pass `manifest.Capabilities`. The binary's `LocalManifest` (loading, validation, the other fields) stays in `package main`.

**2. The two descriptionHash helpers move to `pkg/capability`.** `computeDescriptionHash` and the pinned-exact-tool predicate become `capability.ComputeDescriptionHash` and the `(*Constraint).IsPinnedExactTool()` method. They're capability-domain (they operate on a `Constraint` and its `DescriptionHash` field) and were shared by the PDP runtime check, the startup drift check, and the init/suggest generators — `pkg/capability` is the one package all three already import, so **nothing gains a new dependency direction**. Added table-driven tests (allow / deny / glob / non-tool / nil / malformed).

After this, **`pdp.go` and `pdp_jwt.go` hold zero references** to `LocalManifest`, `computeDescriptionHash`, or `isPinnedExactToolConstraint` (only a `validateLocalManifest` mention survives, in a prose comment).

## Validation

- `go build ./...`, `go test -race -count=1 ./...` — all green
- `make lint` (0 issues), `check-fmt`, `check-license`, `check-notice` — clean
- Zero-back-dep grep over `pdp.go`/`pdp_jwt.go` confirms the seam is clean
- Test-function count **1995 → 1998** — exactly the three new capability tests; no existing test removed
- No `go.mod` / manifest-grammar / audit-shape changes

## Next

With the seam clean, the mechanical relocation of `pdp.go`/`pdp_jwt.go` (+ their tests) into `internal/pdp` — exporting the PDP contract and qualifying the transport call sites — follows as its own focused PR.

---


eunolabs (commented):
Reviewed as the PDP seam-clean (decouple-then-relocate, same cadence as audit #882). Behavior-preserving; no findings. (Would approve, but GitHub blocks self-approval for this account.)

**1. `ManifestPDP` holds `[]capability.Constraint` instead of `*LocalManifest`.** It only ever read `.Capabilities` (confirmed — no leftover `p.manifest` reference survives in `pdp.go`/`pdp_jwt.go`), and `NewManifestPDP(manifest.Capabilities, …)` passes the same slice header, so `p.caps` aliases the same backing array the manifest does — no copy, identical read behavior in `bestMatchingConstraint`. The binary's `LocalManifest` (loading/validation/other fields) correctly stays in `main`.

**2. `computeDescriptionHash` → `capability.ComputeDescriptionHash`** is a verbatim move (same `sha256.Sum256` + `"sha256:"+hex`).

**3. `isPinnedExactToolConstraint` → `(*Constraint).IsPinnedExactTool()`** is the one genuine *reimplementation* (it can't call the package-main helpers from `pkg/capability`), so I checked it closely — it's equivalent:
- The glob check `!strings.ContainsAny(name, "*?[")` is exactly what the old `isGlobPattern` did (`isGlobPattern` is literally `strings.ContainsAny(pattern, "*?[")`).
- The swap from `stripNamespacePrefix(c.Target)` to `ParseTarget`'s bare `name` is immaterial here: the check only looks for `*?[`, and the namespace prefixes (`tool:`/etc.) never contain those, so checking the full target vs the bare name yields the same boolean. And this line is only reached after `tt == TargetTypeTool`, so the nil/empty-hash/non-tool/malformed guards above it are preserved 1:1.
- The new table-driven test (`*`, `?`, `[ab]` globs, resource/prompt, nil, malformed, empty-hash) pins exactly these cases — good coverage for a function that gates FM-5 description-poisoning detection.

**Seam status:** `pdp.go`/`pdp_jwt.go` now hold zero *code* references to `LocalManifest` or the two helpers — only a single `validateLocalManifest` mention in a prose comment (line 730), which is harmless for the relocation. The two helpers landing in `pkg/capability` add no new dependency direction (all three consumers — PDP runtime, drift check, init/suggest — already import it).

**Verification:** `go build ./...` + `go vet ./...` clean; `pkg/capability` and `cmd/eunox-mcp` green under `-race`; all 43 CI checks green; test-function delta is **+3** (exactly the new `ComputeDescriptionHash`/`IsPinnedExactTool` tests — nothing removed). LGTM — clean seam for the `internal/pdp` relocation.

---

pkg/capability/description_hash.go
```diff
@@ -0,0 +1,35 @@
...
+func (c *Constraint) IsPinnedExactTool() bool {
+	if c == nil || c.DescriptionHash == "" {
+		return false
+	}
+	tt, name, err := ParseTarget(c.Target)
+	if err != nil || tt != TargetTypeTool {
+		return false
+	}
+	// A glob target (path.Match metacharacters) is not an exact tool name.
+	return !strings.ContainsAny(name, "*?[")
```
eunolabs, line 34: Nit (optional, non-blocking): this `strings.ContainsAny(name, "*?[")` duplicates the metacharacter set in `isGlobPattern` (cmd/eunox-mcp/drift.go), which is the identical `strings.ContainsAny(pattern, "*?[")`. Since `drift.go` and `manifest.go` already import `pkg/capability`, you could expose a small `capability`-level helper (e.g. `func IsGlobName(name string) bool`) and have all three sites call it, keeping glob detection a single source of truth — in the same spirit as the `splitEnginePrefix` recognized-prefix-set comment in `pkg/enforcement`. Not a blocker; the inline one-liner is perfectly fine as-is.

---

eunolabs (commented):
## Review — refactor(pdp): decouple PDP from `package main` (Phase 5)

Reviewed the full diff plus the surrounding helpers it leans on (`ParseTarget`, `splitEnginePrefix`/`StripEnginePrefix`, `isGlobPattern`). This is a clean, well-scoped seam-cleaning refactor: no behavior change, and the three back-deps into `package main` are severed exactly as described.

### Correctness — the one change with teeth
The only edit with security-relevant behavior is moving `isPinnedExactToolConstraint` → `(*Constraint).IsPinnedExactTool()`, since it gates FM-5 description-hash enforcement. The glob check changed shape:

- old: `!isGlobPattern(stripNamespacePrefix(c.Target))`
- new: `!strings.ContainsAny(name, "*?[")`, where `name` is `ParseTarget`'s bare component

I traced both. `isGlobPattern` is literally `strings.ContainsAny(pattern, "*?[")`, so the metacharacter test is identical. The bare names also agree: both `ParseTarget` and `splitEnginePrefix` split on the **first** `:` and recognize `tool`, and this branch is only reached for `tool:` targets (gated by `tt != TargetTypeTool`). The one spot the two parsers diverge — a prefix-less/unrecognized target, where `splitEnginePrefix` falls back to `("tool", s)` but `ParseTarget` errors — was never observable in the old code either, because the `ParseTarget` type-gate ran first and bailed. The new `malformed target (no namespace) → false` test case locks that in. **Behaviorally equivalent.**

Bonus: the new form is a small improvement — it reuses the bare name `ParseTarget` already returns instead of re-deriving it through a *second*, cross-package parser (`enforcement.splitEnginePrefix`), removing a latent way for the two to drift.

### Aliasing semantics
`ManifestPDP` went from holding `*LocalManifest` (re-dereferenced per read) to capturing the `caps` slice header at construction. That's only observable if something **reassigns** `manifest.Capabilities = …` after the PDP is built. The only such reassignment is `merged.Capabilities = append(...)` during the manifest merge (`manifest.go:135`), which runs *before* `NewManifestPDP` in `loadUpstreamPDP`. No post-construction reassignment exists, so the change is invisible. (Side benefit: `caps == nil` is now safe where a nil `*LocalManifest` would have panicked.)

### Tests & conventions
- New `description_hash_test.go` is table-driven with allow / deny / glob (`*` `?` `[` each) / non-tool / nil / malformed coverage — meets the CONTRIBUTING bar, and pins real SHA-256 vectors (empty string + `hello`) rather than just asserting format.
- License headers present on both new files; placement in `pkg/capability` is right (all three callers already import it, so no new dependency direction); pure refactor, so no docs/grammar/audit-shape changes are owed.
- All 43 CI checks green (build × all platforms, test, lint, gofmt, NOTICE, govulncheck, CodeQL, Trivy, demo + OPA-comparison suites).

### Minor (non-blocking)
One DRY nit left inline on `IsPinnedExactTool`: the `"*?["` metacharacter set is now spelled out in both `isGlobPattern` and this method. A shared `capability`-level helper could keep glob detection a single source of truth. Purely optional.

LGTM — correct and safe to merge as the seam-cleaning step before the mechanical relocation into `internal/pdp`.

---

PR #916, opened 2026-06-19
refactor(config): promote the config/manifest layer to internal/config (Phase 5)

eunolabs:
## Context — Phase 5, prep for the transport extraction

`internal/transport` is the last and largest Phase 5 package. A coupling analysis of the transport/server set (the `stdio*`/`http*`/`forward`/`dispatch`/`jsonrpc`/`route` files) showed its back-references into `package main` reduce to two clusters: **(A1)** the config/manifest *type vocabulary* and **(A2)** a circular drift-checker knot. Because a future `internal/transport` package cannot import `package main`, severing A1 requires giving those types a non-main home first.

This PR does exactly that — and nothing else. It promotes the binary's **config + manifest loading layer** out of `package main` into a new self-contained `internal/config` package. Same shape as the `internal/mcp` extraction (#909): a foundational, behavior-preserving move with no logic changes. The transport decouple (A2) and relocation follow in later PRs.

## What moves

- `manifest.go` / `gateway_config.go` / `schema_version.go` → `internal/config/` (`package config`), tracked as **renames** (94–99% similarity). The package **builds standalone on `pkg/*` + `gopkg.in/yaml.v3` + stdlib, with zero references back into the binary** (`go list -deps ./internal/config/` shows only `pkg/{capability,enforcement,callcounter,circuitbreaker}` + `gopkg.in/yaml.v3` — no `cmd`, no other `internal/`).
- Two straggler helpers `gateway_config.go` depends on come with it: `resolveBool` → `config.ResolveBool` (was `route.go`) and the `maxDurationMs` const → `config.MaxDurationMs` (was `http.go`). Old decls deleted; `cmd` referrers call the exported names.

## Exported seam (compiler-driven — only what `cmd` consumes)

`LocalManifest` / `LoadManifest` / `MergeManifests`; `GatewayConfig` / `UpstreamConfig` / `RouteDefaults` / `LoadGatewayConfig`; the consumed methods `(*GatewayConfig).Validate` / `HostTransport` / `AuditModeFor` / `NoPolicyStartupRejection`; and `HostTransportHTTP` / `HostTransportStdio` / `ResolveBool` / `MaxDurationMs`. The validators (`validateLocalManifest`, `validateDescriptionHashFormat`, `expandEnvInStrings`, …) **stay unexported**. No de-stuttering needed — no name begins with `config`.

## Tests

[...]

PR #917, opened 2026-06-19
refactor(transport): break the drift-check cycle for transport extraction

eunolabs:
## Context — Phase 5, transport extraction (decouple step)

The coupling analysis for extracting the transport/server runtime into `internal/transport` found its back-references into `package main` reduce to two clusters: **(A1)** the config/manifest type vocabulary — severed by the `internal/config` move (#916) — and **(A2)** a circular drift-checker knot. This PR severs A2. No files are relocated yet; this is the decouple that makes the eventual `git mv` a clean move.

### The cycle

The transport (`stdio.go` / `http_session.go` / `http_remote.go`) called `runHTTPDriftCheck` / `runStdioDriftCheck` in `drift.go` at session start, and those functions reached **back** into transport internals (`httpSession.callUpstream`, `StdioProxy.upWriter`/`upReader`/`upHTTP`, `rpcMsg`, `withoutUpstreamTimeout`, `msToDuration`, `msgKey`, …) — a bidirectional dependency.

## The sever — dependency-injected raw-result seam

The probe is transport work (it uses the transport's own upstream connection); the compare is policy (shared with `validate --live`). Split along that line and inject the compare:

- **New `drift_probe.go` (transport):** the `driftCheckFunc` hook type, plus the two `fetchUpstreamToolsRaw` methods (on `*httpSession` and `*StdioProxy`) that send the `tools/list` probe over the transport's connection and return the **raw** JSON result. They reference only transport internals + stdlib.
- **`drift.go` (policy, stays in `package main`, shared with `validate --live`)** keeps `CheckManifestDrift` / `evaluateDrift` / `parseToolsListResult` and gains `makeDriftCheck` + `driftProbeUnavailable`, which **fold the previously-duplicated** `runHTTPDriftCheck`/`runStdioDriftCheck` fail-closed logic into one place. The old `run*`/`fetch*` functions are deleted.
- **Main-staying wiring injects the hook:** `serveHTTPGateway` sets each route's `driftCheck` right after `BuildRoutes`; `serveStdioHost` passes it via `StdioProxyOptions.DriftCheck`. The transport just probes and calls `route.driftCheck` / `p.driftCheck`. `makeDriftCheck` returns `nil` for an unpoliced route, so the `driftCheck != nil` gate is exactly equivalent to the old `manifest != nil` gate.

## Behavior-preserving

Identical error strings; identical fail-closed semantics:
- FM-5 `descriptionHash` pins abort unconditionally;
- `--strict-drift` gates FM-1/2/4;
- a probe **or parse** failure without pins is a logged best-effort skip (parse moved into the hook, but routes through the same `driftProbeUnavailable` path the old fetch error did);
- the HTTP session probe stays `withoutUpstreamTimeout`; the stdio HTTP-bridge probe keeps the per-call `--upstream-timeout`.

Tests: only call sites were rewritten (`StdioProxyOptions`/`HTTPProxyOptions` `Manifest`/`StrictDrift` → `DriftCheck: makeDriftCheck(...)`; the `run*`/`fetch*` test calls → `fetchUpstreamToolsRaw` + `makeDriftCheck`). No test added, dropped, or weakened.

## Validation

- `go build ./...`, `go vet ./...`, `go test -race -count=1 ./cmd/eunox-mcp/ ./internal/...` — green
- **Decouple proof:** the transport-set files now contain **zero** references to `drift.go` symbols (`runHTTPDriftCheck|runStdioDriftCheck|fetch*Tools|evaluateDrift|CheckManifestDrift|parseToolsListResult|manifestHasPinnedDescriptions|\bUpstreamTool\b` → empty)
- Test-function count unchanged at **1905**

## Next

With A1 and A2 severed, the transport set has no remaining `package main` back-references. The final Phase 5 step is the **`git mv` relocation** of the transport files into `internal/transport` (exporting the forward seam + the `proxyVersion` setter, splitting tests white-box vs main).

> Note: this commit is unsigned — SSH commit signing isn't provisioned in this environment (empty key, no agent). The committer identity is correct (`Claude <noreply@anthropic.com>`).

---


eunolabs (commented):
Reviewed as the drift-check decouple (A2) — a genuine refactor, not a relocation, so I focused on whether splitting probe (raw fetch) from compare (parse + evaluate) preserves the fail-closed semantics. It does; clean work, no findings. (Would approve, but GitHub blocks self-approval for this account.)

**Decouple proof confirmed:** grepping the transport set (`stdio.go`/`http*.go`/`forward.go`/`dispatch.go`/`jsonrpc.go`/`route.go`/`drift_probe.go`) for every `drift.go` policy symbol — `runHTTPDriftCheck`, `CheckManifestDrift`, `evaluateDrift`, `parseToolsListResult`, `manifestHasPinnedDescriptions`, `UpstreamTool`, … — comes back **empty**. The transport now references only the injected `driftCheckFunc` hook + its own `fetchUpstreamToolsRaw` probe. The cycle is severed.

**Behavior-preserving, verified branch by branch:**
- **Gate equivalence:** `makeDriftCheck` returns `nil` iff `manifest == nil`, so `route.driftCheck != nil` is exactly the old `route.manifest != nil` gate.
- **All three failure modes converge identically:** an upstream transport error and an RPC error return from the probe as `probeErr`; a **parse** failure (the parse moved out of the probe into the hook) is caught right after — and all three route through `driftProbeUnavailable`, which is byte-identical to the old inline block (fatal when `manifestHasPinnedDescriptions`, else the same `WARN drift check skipped` and `nil`). Nice that this also folds the previously-**duplicated** fail-closed logic from `runHTTPDriftCheck`/`runStdioDriftCheck` into one place.
- **Timeout semantics preserved:** the HTTP-session probe stays `withoutUpstreamTimeout` (session-start budget, not `--upstream-timeout`); the stdio HTTP-bridge probe keeps the per-call `--upstream-timeout`; the subprocess probe keeps the unbounded notification-discard loop. The two `fetchUpstreamToolsRaw` methods are the old `fetch*` bodies with only the final `parseToolsListResult` removed (now returning raw).
- `evaluateDrift` is still handed the transport's `upstreamServerVersion` at each call site.

**Verification:** `go build`/`go vet` clean; `cmd/eunox-mcp` green under `-race`; **all 43 CI checks green**; test-function delta **0** (the `fetch+parse` split in `TestFetchStdioTools_ManyNotifications…` keeps coverage). DCO `Signed-off-by` present (the "unsigned" note is SSH/GPG).

**Minor, forward-looking (non-blocking):** the drift gate moved from the intrinsic `route.manifest` field to a separately-wired `route.driftCheck` field (set in `serveHTTPGateway`'s post-`BuildRoutes` loop / `NewHTTPProxy` opts). The route now carries both, and the gate's correctness depends on that wiring step running — fine today, but worth keeping the `makeDriftCheck` wiring adjacent to route construction through the upcoming `git mv` so a future `BuildRoutes` caller can't leave `driftCheck` nil on a policed route (a silent fail-open). LGTM — both A1 and A2 severed; the transport `git mv` is the finale.

---

PR #920, opened 2026-06-19
refactor(transport): finish relocation prep — opaque routes, config-owned manifest helpers

eunolabs:
## Context — Phase 5, transport-extraction prep (the last prep step)

A forward-seam analysis of the upcoming `internal/transport` `git mv` flagged three remaining cross-boundary references that would otherwise leak into the transport package's exported API or block the move. This PR cleans up all three, so the relocation can be a **pure `git mv`** with a narrow exported surface. Behavior-preserving, no logic changes.

### 1. Manifest policy helpers → `internal/config`

`countAuditOnly` / `manifestDigest` / `manifestHasMaxCalls` operate purely on `*config.LocalManifest` and were used by **both** the transport (`route.go`) and the CLI (`doctor.go` / `main.go`). They move to `internal/config` as methods on `LocalManifest` — `AuditOnlyCount()` / `Digest()` / `HasMaxCalls()` — so the transport package never has to re-export manifest helpers. (`manifestHasSamplingOptIn` and `anyRouteHasMaxCalls` stay in `route.go` — transport-internal / route-typed.)

### 2. `BuildRoutes` wires the drift hook itself (opaque routes)

Previously `serveHTTPGateway` reached into three **unexported** `upstreamRoute` fields after `BuildRoutes`:
```go
for _, rt := range routes { rt.driftCheck = makeDriftCheck(rt.manifest, rt.strictDrift) }
```
`BuildRoutes` now takes `driftCheckFor func(*config.LocalManifest, bool) driftCheckFunc` and sets each route's hook internally; `main` passes `makeDriftCheck`. **`package main` no longer touches any route field**, so `upstreamRoute`'s fields can stay private once the type moves.

### 3. `proxyVersion` setter (one-directional version seam)

`proxyVersion` gains a `SetProxyVersion` setter: `main`'s `init` calls `SetProxyVersion(version)` instead of writing the transport var across the boundary, and the CLI live-upstream probe reports `main`'s own `version` var directly rather than reading `proxyVersion`. Neither the write nor the read crosses into transport state anymore.

## Validation

- `go build ./...`, `go vet ./...`, `go test -race -count=1 ./cmd/eunox-mcp/ ./internal/...` — green
- `internal/config` stays a clean lower layer (`pkg/*` + `gopkg.in/yaml.v3` + stdlib; `go list -deps`)
- `package main` no longer accesses any `upstreamRoute` field (verified by grep)
- Test-function count unchanged at **1905**

## Next

This was the last prep step. The transport set now has **no `package main` back-references and no cross-boundary field/var access**, and the future `internal/transport` API narrows to the construction/wiring seam (~22 symbols). The final Phase 5 step is the `git mv` of the transport files into `internal/transport` (export that seam, split tests white-box vs. main, relocate `enforcement_gaps_test.go` per its CLAUDE.md note).

> Note: this commit is unsigned — SSH commit signing isn't provisioned in this environment (empty key, no agent). The committer identity is correct (`Claude <noreply@anthropic.com>`).

---


eunolabs (commented):
Reviewed the last transport-relocation prep step. All three changes are behavior-preserving; clean work — and #2 directly closes the forward-looking note I left on #917. (Would approve, but GitHub blocks self-approval for this account.)

**1. Manifest helpers → `internal/config` methods.** `HasMaxCalls`/`AuditOnlyCount`/`Digest` are verbatim moves of `manifestHasMaxCalls`/`countAuditOnly`/`manifestDigest` — same nil-handling (`HasMaxCalls` keeps its `m == nil` guard; the other two stay guard-free and every caller is still nil-guarded), same SHA-256-over-canonical-JSON, same fail-closed `#487` semantics. `IsAuditOnly`/`MaxCallsCondition` are `pkg/capability` symbols `internal/config` already imports, so no new dependency direction. Callers in `doctor.go`/`main.go`/`route.go` updated; old decls removed from `cmd`.

**2. `BuildRoutes` owns the drift-hook wiring.** This is exactly the encapsulation I flagged on #917 — the post-`BuildRoutes` loop reaching into three unexported route fields is gone. The injected `driftCheckFor(manifest, r.strictDrift)` runs inside the route loop with the *identical* inputs the old loop used (`manifest` == `r.manifest`, and `r.strictDrift` is set on the line above), so it's behavior-equivalent while making the gate intrinsic to route construction rather than a separable step a future caller could miss. Verified `package main`'s CLI files now access **zero** `upstreamRoute` fields — the type's fields can go private on the `git mv`.

**3. `proxyVersion` setter.** `SetProxyVersion(version)` == the old direct `proxyVersion = version` write; the CLI probe reading `version` directly == the old `proxyVersion` read (both are set to `version` in `init`, never diverge), so the reported `clientInfo` version is unchanged. The var is now write-only-via-setter from `main` and read only inside the transport's `initialize` path — a clean one-directional seam.

**Verification:** `go build`/`go vet` clean; `internal/config` + `cmd/eunox-mcp` green under `-race`; **all 43 CI checks green**; test-function delta **0**. `internal/config`'s direct imports are unchanged (the large `go list -deps` closure is all transitive via `pkg/*`). DCO `Signed-off-by` present.

**One trivial, optional nit:** `(*LocalManifest).Digest()`'s error message still reads `"manifestDigest: marshal failed: %w"` — preserved verbatim (so behavior is unchanged), but the prefix now names a function that no longer exists. Worth a one-word tweak to `"Digest:"` next time this file is touched; not worth a round-trip on its own. LGTM — that's all the prep; the transport `git mv` is the finale.

---

PR #923, opened 2026-06-19
refactor(transport): relocate the transport runtime to internal/transport (Phase 5 finale)

eunolabs:
## Phase 5 finale — relocate the transport runtime to `internal/transport`

This completes the package restructure. The four prior steps severed every transport→`package main` back-reference (config types → `internal/config`; the JSON-RPC envelope → `internal/mcp`; the drift cycle → an injected `DriftCheckFunc` hook; manifest helpers → `LocalManifest` methods; opaque routes via a hook factory; `SetProxyVersion`), so this is a **clean `git mv`**: `internal/transport` builds standalone on `internal/{audit,config,mcp,pdp}` + `pkg/*` with **zero references back into `cmd/eunox-mcp`**.

*(Also carries a one-line fix: `(*LocalManifest).Digest()`'s marshal-error prefix `manifestDigest:` → `Digest:`, naming the function that actually returns it.)*

### What moves
16 sources `git mv`'d to `internal/transport` (`package transport`, history preserved as renames): the proxies (`stdio*`/`http*`), the shared dispatch/forward/enforcement core, the remote-upstream bridge, the JSON-RPC handshake/denial/correlation plumbing, route wiring, the drift probe, and the control-token / OAuth-metadata / health endpoints. The binary now just constructs the proxies and calls `Start`/`Serve`.

### Exported surface — 25 symbols (the construction/wiring seam)
Constructors/options, `BuildRoutes`/`UpstreamRoute`/`LoadUpstreamPDP`/`WrapRoutesWithJWT`/`AnyRouteHasMaxCalls`, the `Resolve*` helpers, the control-token + OAuth funcs, `SetProxyVersion`, `DriftCheckFunc`, and the CLI-probe plumbing the `validate --live`/`suggest` paths reach (`DoMCPHTTP`, `BuildUpstreamClient`, `ProxyName`, `MCPProtocolVersion`, `SessionHeader`, `IsInfraDenialCode`). **Over-exports un-exported**: `HTTPProxyOptions`, `NewHTTPProxy` (test-only) are now private; denial/handshake/correlation/`httpUpstream` stay internal. Every export has a `cmd/*.go` non-test consumer (verified via `go doc`).

### Tests
Split white-box vs main: most move wholesale (with `makeDriftCheck` replaced by an inline `DriftCheckFunc` stub where only a hook is needed); `gateway`/`security`/`integration`/`operational`/`transport_unit`/`pdp_jwt`/`pdp`/`drift`/`main` split their transport halves out; JWT + PDP + re-exec-sentinel helpers duplicated across the split (the documented cost). `enforcement_gaps_test.go` (the mandated home for new MCP-method coverage) moves to `internal/transport`; **`CLAUDE.md` + `CONTRIBUTING.md` path references updated**. `.golangci.yml` path excludes + `docs/architecture.md` updated.

### Validation
- `go build ./...`, `go vet ./...`, `go test -race -count=1 ./...` — **16 packages green** (incl. `internal/transport`)
- **`internal/transport` standalone** (`go list -deps`: no `cmd/eunox-mcp`)
- **Test name-set unchanged** — 1902 unique names in HEAD and working, zero dropped/renamed (raw count 1905→1907 = the two duplicated re-exec sentinel helpers); the binary still builds.

### Known follow-ups (deferred to keep this a focused relocation)
1. **CLI-probe plumbing** — `DoMCPHTTP`/`BuildUpstreamClient`/the protocol consts are exported from transport only because `live_upstream.go`/`suggest.go` reach them. They could move to `internal/mcp` / `pkg/capability` (the analysis recommends it) so those CLI files depend on neither — but that's an architectural call (whether the `mcp` leaf should gain an HTTP client) better made in review. Happy to do it as a follow-up.
2. **Drift-test mirror** — the transport-side HTTP-drift integration tests need both transport internals *and* the drift comparison (which stays in `main`'s `drift.go`), so `internal/transport/drift_helpers_test.go` duplicates the comparison (~334 lines) to preserve their exact assertions. The authoritative logic + its unit tests stay in `main`. Relocating the drift-comparison policy to a shared package would let main and these tests share one implementation and delete the mirror — a clean follow-up.

> Note: commits are unsigned — SSH commit signing isn't provisioned in this environment (empty key, no agent). The committer identity is correct (`Claude <noreply@anthropic.com>`).

---


eunolabs (commented):
Reviewed the Phase 5 finale — the `git mv` of the transport runtime into `internal/transport`. Verified mechanically like the prior relocations (it's a large move, so declaration-identity/leaf-deps/over-export/test-set, not a line read). Clean; no findings. (Would approve, but GitHub blocks self-approval for this account.)

**Standalone + inward-only:** `internal/transport` builds on `internal/{audit,config,mcp,pdp}` + `pkg/{capability,enforcement,killswitch}` with **zero** references back into `cmd/eunox-mcp` (`go list -deps` confirms). The five prior decouple steps clearly paid off — this really is a clean move.

**Export surface — exactly 25, no over-export:** cross-referenced every package-level export against cmd usage. The only two not named explicitly by cmd are `HTTPProxy`/`StdioProxy` — the concrete return types of `NewStdioProxy`/`NewHTTPProxyGateway`, consumed via `:=` + method calls; un-exporting them would make an exported constructor return an unexported type (a `revive` smell), so keeping them public is correct, not a leak. The actual over-exports (`HTTPProxyOptions`, the test-only `NewHTTPProxy`) were correctly un-exported, and denial/handshake/correlation/`httpUpstream` stayed internal.

**No tests lost:** unique `Test`/`Fuzz` name-set is **1974 in both base and head with zero dropped**; raw count is `+2`, exactly the two duplicated re-exec sentinel helpers the PR documents. `enforcement_gaps_test.go` moved with its `CLAUDE.md`/`CONTRIBUTING.md` pointers updated.

**Verification:** `go build`/`go vet` clean; **full `go test -race ./...` green across all 16 packages** (internal/transport 8.8s, cmd down to 5.9s); all 43 CI checks green. And commit 1 (`Digest:` marshal-error prefix) tidies up the exact nit I left on #920 — appreciated.

**On deferred follow-up #1 (you asked for a steer in review):** I'd advise *against* moving `DoMCPHTTP`/`BuildUpstreamClient` into `internal/mcp`. The leaf's value is precisely that it's a pure stdlib, dependency-free protocol-types package — giving it a `net/http` client would change its character and pull `net/http`/TLS into everything that imports the wire types. If the goal is to stop `live_upstream.go`/`suggest.go` depending on `internal/transport` just for the probe client, a small dedicated package (e.g. `internal/mcpclient`) that both transport and the CLI import is the cleaner cut — it keeps `mcp` a leaf and removes the CLI→transport edge. Follow-up #2 (the ~334-line drift-comparison test mirror) is a reasonable cost to defer; sharing the drift-comparison policy into a package both `main` and those tests import would delete it cleanly. Both are honestly disclosed and out of scope here.

LGTM — that completes the package restructure. Nicely sequenced: every step in this arc was independently reviewable and behavior-preserving.

---

PR #933, opened 2026-06-20
Fix review-session bugs (e06470, 24cd12, eb1ff4, e865e5, d471e5)

### LINKED ISSUE #926: [rev-20260620T0006Z-d471e5] pruneRotated silently ignores rotatedSiblings directory-read errors
## Location

`internal/audit/rotate.go`, lines 229–231

## Reference

https://github.com/eunolabs/eunox/blob/eda822826b74bfdc84831d94a04f412d5477de67/internal/audit/rotate.go#L229-L231

## Severity

Low

## Description

`pruneRotated` calls `rotatedSiblings` to list the rotated log siblings and then early-returns on any error from that call:

```go
matches, err := rotatedSiblings(s.logPath)
if err != nil || len(matches) <= s.retain {
    return  // silently swallowed when err != nil
}
```

When `rotatedSiblings` fails (e.g. `os.ReadDir` returns a permission error or the audit directory has been deleted), the function returns without logging anything to stderr. By contrast, individual file removal failures a few lines later **do** emit a diagnostics line:

```go
if err := os.Remove(old); err != nil {
    fmt.Fprintf(os.Stderr, "[eunox-mcp] audit retention: could not delete %q: %v\n", old, err)
}
```

The inconsistency means that a persistent directory-listing failure (e.g. the audit directory's execute bit removed by an admin) produces zero observable signal. Old rotated files accumulate unbounded while the operator has no indication that retention pruning has stopped working entirely.

## Trigger / repro

1. Start `eunox-mcp` with `--audit-log /some/dir/audit.jsonl` and `--audit-retain 3`.
2. Trigger several rotations so there are ≥4 rotated siblings.
3. Remove the execute bit on `/some/dir` (`chmod a-x /some/dir`).
4. Trigger another rotation. `pruneRotated` is called; `os.ReadDir` fails with EACCES. No message appears on stderr; rotated files keep accumulating silently.

## Suggested fix

Log the error before returning, consistent with the remove-failure path:

```go
matches, err := rotatedSiblings(s.logPath)
if err != nil {
    fmt.Fprintf(os.Stderr, "[eunox-mcp] audit retention: could not list rotated siblings of %q: %v\n", s.logPath, err)
    return
}
if len(matches) <= s.retain {
    return
}
```

This makes directory-listing failures as visible as individual file-removal failures.

## Session

`rev-20260620T0006Z-d471e5` — reviewed files: `internal/audit/rotate.go`, `pkg/capability/jwkscache.go`, `internal/transport/http_security.go` @ `eda822826b74bfdc84831d94a04f412d5477de67`

### LINKED ISSUE #927: [rev-20260620T0010Z-e865e5] stdio shutdown deadlocks when --upstream-timeout=0 and an in-flight upstream call never returns
**Location:** `internal/transport/stdio.go:223-230` (Start teardown ordering) and `internal/transport/stdio.go:438` (`serveHost` `wg.Wait()`), interacting with `callUpstream`/`awaitUpstreamResponse` (`internal/transport/jsonrpc.go:94-121`) and `boundUpstreamCall` (`internal/transport/http_handlers.go:37-42`).

**Reference:** https://github.com/eunolabs/eunox/blob/eda822826b74bfdc84831d94a04f412d5477de67/internal/transport/stdio.go#L223-L230

**Severity:** Medium (permanent hang / process + goroutine leak on shutdown; gated behind the non-default `--upstream-timeout=0` setting plus an upstream that does not respond after the host disconnects).

**Description:**

The stdio transport's host-disconnect shutdown path has a circular wait that can deadlock the proxy when the per-call upstream timeout is disabled:

1. The host closes stdin → `serveHost`'s read loop breaks (`stdio.go:398-400`) and the function blocks at `wg.Wait()` (`stdio.go:438`) waiting for any in-flight per-request goroutine.
2. An in-flight `handleHostRequest` goroutine is parked in `callUpstream` → `awaitUpstreamResponse`, whose `select` (`jsonrpc.go:113-120`) unblocks only on (a) the response channel, (b) `ctx.Done()`, or (c) `p.upstreamDone`.
3. With `--upstream-timeout=0`, `withUpstreamTimeout` → `boundUpstreamCall(ctx, 0)` returns the **parent** ctx with a **no-op** cancel (`http_handlers.go:38-39`). The parent ctx is not canceled by a host stdin EOF (that is the normal, signal-free shutdown trigger), so case (b) never fires.
4. `p.upstreamDone` (case c) is closed by `readUpstream` only after the upstream subprocess EOFs its stdout, which for a well-behaved upstream happens after `closeUpstreamInput()` closes the proxy→upstream stdin pipe (`stdio.go:334-340`). But `closeUpstreamInput()` is called at `Start` step 7 (`stdio.go:229`), which runs only **after** `serveHost` returns (`stdio.go:224`).

So: `serveHost` waits on the goroutine → the goroutine waits on `upstreamDone` → `upstreamDone` waits on `closeUpstreamInput` → `closeUpstreamInput` waits on `serveHost`. The process hangs after the host disconnects (it remains killable only via SIGINT/SIGTERM, whose handler is still installed and kills the subprocess).

Under the default config the per-call 30 s timeout (`defaultUpstreamTimeoutMs`) makes `ctx.Done()` eventually fire, breaking the cycle — so the bug only causes a (bounded) delayed shutdown by default and a **permanent** hang when the operator explicitly sets `--upstream-timeout=0` (documented as "pass `0` to disable", README line 313).

The HTTP transport is **not** affected: it handles each POST in its own handler and does not `wg.Wait()` on per-request goroutines at shutdown, so there is no analogous occurrence of this class (shutdown ordering that waits on in-flight work before signaling the dependency that unblocks it).

**Trigger / repro:**
1. Run the stdio proxy with `--upstream-timeout=0` against a subprocess upstream.
2. Host sends an allowed request (e.g. `tools/call`) that the upstream accepts but never answers (or is very slow).
3. Host closes stdin while that request is in flight.
4. The proxy never exits; the in-flight goroutine and the upstream subprocess leak until the process is signaled/killed.

**Suggested fix:**

Break the circular wait so a host EOF can unblock pending upstream calls regardless of `--upstream-timeout`. Closest-to-minimal: close the upstream input (signaling EOF → `readUpstream` returns → `upstreamDone` closes → pending callers unblock) **before** `wg.Wait()`. For example, in `serveHost` after the read loop breaks:

```go
p.closeUpstreamInput() // upstream EOFs -> readUpstream returns -> upstreamDone closes -> pending callers unblock
wg.Wait()
```

and make `Start`'s subsequent `closeUpstreamInput()` / `<-p.upstreamDone` idempotent (the HTTP bridge already guards with `closeOnce`; the subprocess `p.upIn.Close()` double-close returns an already-ignored error). For an upstream that ignores stdin EOF, also consider adding a dedicated `serveDone` teardown channel to the `awaitUpstreamResponse` select so host EOF directly cancels pending stdio calls rather than relying on the upstream exiting.

**Session:** rev-20260620T0010Z-e865e5

### LINKED ISSUE #928: [rev-20260620T0029Z-eb1ff4] readUpstream silently swallows unexpected upstream errors, making abnormal session termination indistinguishable from normal exit
## Location

`internal/transport/http_session.go` — `readUpstream`, lines 424–443 (commit `eda822826b74bfdc84831d94a04f412d5477de67`).

## Reference

https://github.com/eunolabs/eunox/blob/eda822826b74bfdc84831d94a04f412d5477de67/internal/transport/http_session.go#L424-L443

```go
func (s *httpSession) readUpstream(ctx context.Context) {
    defer close(s.done)
    for {
        msg, err := s.upReader.Read()
        if err != nil {
            return  // ← error silently discarded for all err values
        }
        ...
    }
}
```

`MsgReader.Read()` (`internal/mcp/jsonrpc.go:220–231`) returns three distinct non-nil errors:

- `io.EOF` — normal stream end (subprocess stdout closed after exit)
- `bufio.ErrTooLong` — upstream sent a message larger than the 4 MiB scanner limit
- `fmt.Errorf("parsing JSON-RPC message: %w", err)` — upstream sent malformed JSON

All three exit `readUpstream` identically. After `close(s.done)`, the cleanup goroutine logs only:

```
[eunox-mcp] HTTP session X ended.
```

with no error cause attached.

## Severity

**Low** — observability only. The session terminates correctly in all cases (`awaitUpstreamResponse` selects on `s.done` and returns `errUpstreamExited`, so in-flight host requests receive a structured error response and no request is silently lost). The gap is diagnostic: an operator cannot distinguish a normal subprocess exit from a `bufio.ErrTooLong` session kill or a malformed-JSON loop that drove the session into an unrecoverable scanner state.

## Trigger / repro

1. Run eunox-mcp in HTTP/gateway mode against any upstream.
2. Have the upstream send a single newline-delimited JSON-RPC message larger than 4 MiB (e.g. a tool response with a large binary payload embedded in the JSON body).
3. The `httpSession.readUpstream` goroutine exits on `bufio.ErrTooLong`. The scanner is permanently errored; subsequent `Read()` calls return the same error. The session terminates.
4. The only log entry is `[eunox-mcp] HTTP session X ended.` — the 4 MiB limit breach is invisible.

A malformed JSON message produces an equally silent termination via the parse-error path.

## Suggested fix

Log unexpected errors before returning so abnormal termination is distinguishable from normal subprocess exit:

```go
func (s *httpSession) readUpstream(ctx context.Context) {
    defer close(s.done)
    for {
        msg, err := s.upReader.Read()
        if err != nil {
            if !errors.Is(err, io.EOF) {
                fmt.Fprintf(os.Stderr,
                    "[eunox-mcp] HTTP session %s: upstream read error: %v\n",
                    s.id, err)
            }
            return
        }
        ...
    }
}
```

The analogous `StdioProxy.readUpstream` in `internal/transport/stdio.go:586–590` has the same silent-swallow pattern; in stdio mode the proxy exits when this goroutine returns, which is more observable (the process dies), but a stderr log there would also help diagnose large-message or malformed-JSON failures in the stdio path.

## Session

`rev-20260620T0029Z-eb1ff4`

### LINKED ISSUE #929: [rev-20260620T0045Z-24cd12] JWKSClient.VerifyToken samples now() per candidate key, allowing valid token near expiry to be rejected mid-rotation
## Location

`pkg/capability/jwks.go`, lines 371–376 (inside the per-key closure passed to `VerifyWithKeyRotation`)

## Reference

https://github.com/eunolabs/eunox/blob/eda822826b74bfdc84831d94a04f412d5477de67/pkg/capability/jwks.go#L371-L376

## Severity

**Medium** — correctness defect; causes spurious `Terminal` expiry rejection of a valid token during JWKS key rotation when the token is within the clock-leeway window of its `exp` claim.

## Description

Inside `JWKSClient.VerifyToken`, the per-key closure passed to `VerifyWithKeyRotation` constructs `jwt.Expected` with a fresh call to `c.now()` on every invocation:

```go
expected := jwt.Expected{
    Time:        c.now(),   // called once per candidate key
}
if err := claims.ValidateWithLeeway(expected, c.leeway); err != nil {
    return nil, Terminal(fmt.Errorf("validate claims: %w", err))
}
```

`VerifyWithKeyRotation` calls the closure once per candidate key (`tryKeys` inner loop). If the JWKS contains multiple keys matching the token's `kid` (or `kid` is absent and all cached keys are candidates — both normal during rotation), each candidate key trial uses a different wall-clock instant for expiry validation.

This is inconsistent with `JWTPDP.ValidateToken`, which explicitly samples `now` once before entering the key-selection loop to prevent a time-flip at the expiry boundary.

## Trigger / repro

1. JWKS endpoint is in the middle of key rotation: two keys coexist in the key set (old key under which the token was signed + new key).
2. Token has `exp` ≈ `now + DefaultTokenLeeway` (i.e., within the 10-second leeway window).
3. `VerifyWithKeyRotation` tries the new key first → signature fails → retries with old key.
4. The retry happens a few seconds later; `c.now()` now returns a time past `exp − leeway`.
5. Signature verifies against the old key, but `claims.ValidateWithLeeway` rejects the token with an expiry error → `Terminal(...)` → the token is rejected as expired.

The token was valid when the request arrived but is rejected due to time advancing between key trials inside `VerifyWithKeyRotation`.

## Suggested fix

Sample `now` once before the closure and capture it by value, mirroring `JWTPDP.ValidateToken`:

```go
func (c *JWKSClient) VerifyToken(ctx context.Context, tokenStr string) (*TokenPayload, error) {
    // ... parse + kid extraction ...

    now := c.now()   // sample once; prevents time-flip across multi-key trial (#NN)
    return VerifyWithKeyRotation[TokenPayload](ctx, c.cache, kid, func(key *jose.JSONWebKey) (*TokenPayload, error) {
        // ...
        expected := jwt.Expected{
            Time:        now,   // consistent across all key candidates
        }
        // ...
    })
}
```

## Session

`rev-20260620T0045Z-24cd12`

### LINKED ISSUE #930: [rev-20260620T0045Z-24cd12] handleKill checks method before loopback guard, leaking /control/kill endpoint existence to off-host callers
## Location

`internal/transport/http_routing.go`, lines 282–288 (`handleKill`)

## Reference

https://github.com/eunolabs/eunox/blob/eda822826b74bfdc84831d94a04f412d5477de67/internal/transport/http_routing.go#L282-L288

## Severity

**Low** — information-leakage / defence-in-depth defect; no unauthorized action is possible, but the check ordering is inconsistent with every other loopback-only handler and leaks endpoint enumeration information.

## Description

`handleKill` runs the method check **before** the loopback guard:

```go
func (p *HTTPProxy) handleKill(w http.ResponseWriter, r *http.Request) {
    if r.Method != http.MethodPost {
        http.Error(w, "Method Not Allowed", http.StatusMethodNotAllowed)
        return
    }
    if !p.loopbackOnly(w, r) {   // ← loopback check is SECOND
        return
    }
    // ...
}
```

`handleHealth` and `handleMetrics` — the two other loopback-only handlers in the same file — do the opposite: loopback guard first, then method check:

```go
func (p *HTTPProxy) handleHealth(w http.ResponseWriter, r *http.Request) {
    if !p.loopbackOnly(w, r) {   // ← loopback check is FIRST
        return
    }
    if r.Method != http.MethodGet {
        http.Error(w, "Method Not Allowed", http.StatusMethodNotAllowed)
        return
    }
    // ...
}
```

Because of the reversed ordering in `handleKill`, an off-host `GET /control/kill` receives `405 Method Not Allowed` rather than the loopback rejection response, confirming to the caller that the `/control/kill` endpoint exists and responds only to `POST`. In a defence-in-depth posture (e.g. DNS-rebinding mitigations, or a misconfigured server not bound exclusively to loopback), the loopback guard is the primary security boundary and should run first.

## Trigger / repro

```
curl -v http://<non-loopback-address>:<port>/control/kill
# Method: GET
# Response: 405 Method Not Allowed   ← endpoint enumerated
```

With the correct ordering (loopback first), the same request would be rejected by `loopbackOnly` before the method check runs.

## Suggested fix

Swap the two checks in `handleKill` to match `handleHealth`/`handleMetrics`:

```go
func (p *HTTPProxy) handleKill(w http.ResponseWriter, r *http.Request) {
    if !p.loopbackOnly(w, r) {
        return
    }
    if r.Method != http.MethodPost {
        http.Error(w, "Method Not Allowed", http.StatusMethodNotAllowed)
        return
    }
    // remainder unchanged ...
}
```

## Session

`rev-20260620T0045Z-24cd12`

### LINKED ISSUE #931: [rev-20260620T0053Z-e06470] MergeManifests panics on empty input (index out of range on ms[0])
## Location

`internal/config/manifest.go`, lines 129–149 (`MergeManifests`)

## Reference

https://github.com/eunolabs/eunox/blob/eda822826b74bfdc84831d94a04f412d5477de67/internal/config/manifest.go#L129-L149

## Severity

**Low** — panic reachable only via direct library use or a future CLI code path that passes an empty slice; the current CLI requires at least one `--policy` file. Still a latent defect: an exported function should not panic on a plausible caller mistake.

## Description

`MergeManifests` has a fast path for `len(ms) == 1` but no guard for `len(ms) == 0`:

```go
func MergeManifests(ms []*LocalManifest) *LocalManifest {
    if len(ms) == 1 {
        return ms[0]
    }
    merged := &LocalManifest{
        SchemaVersion: ms[0].SchemaVersion,   // ← panics when len(ms) == 0
        Name:          ms[0].Name,
        ...
    }
```

When called with an empty slice the first field access `ms[0].SchemaVersion` panics with:

```
panic: runtime error: index out of range [0] with length 0
```

The current CLI only calls `MergeManifests` when multiple `--policy` flags are supplied (guaranteeing at least one element), so this is not currently reachable in production. However, the function is exported and its signature carries no such contract. A library caller, a future CLI refactor, or a unit-test helper that builds an empty slice before filtering can all trigger the panic.

## Trigger / repro

```go
// panics
_ = config.MergeManifests([]*config.LocalManifest{})
```

## Suggested fix

Add an early return for the empty-input case, consistent with the existing `len(ms) == 1` fast path:

```go
func MergeManifests(ms []*LocalManifest) *LocalManifest {
    if len(ms) == 0 {
        return &LocalManifest{}
    }
    if len(ms) == 1 {
        return ms[0]
    }
    // ... rest unchanged
```

Returning an empty `*LocalManifest` is consistent with the "zero capabilities → deny everything" invariant (an empty Capabilities slice means no tool is allowed). Returning `nil` would be equally defensive but requires callers to nil-check the result.

## Session

`rev-20260620T0053Z-e06470`

eunolabs:
## Bugs addressed

Every open issue whose title carries a review session id, fixed in one PR. One commit per bug.

- Closes #931 — `MergeManifests` panicked (`index out of range`) on an empty slice because the `len(ms)==1` fast path had no `len(ms)==0` guard. Added an early return yielding an empty manifest (zero capabilities = deny everything), consistent with the fail-closed invariant.
- Closes #930 — `handleKill` ran the method check before the loopback guard, so an off-host `GET /control/kill` got `405` (confirming the endpoint exists) instead of the loopback rejection. Swapped the order to match `handleHealth`/`handleMetrics` (loopback first). Regression test added.
- Closes #929 — `JWKSClient.VerifyToken` sampled `c.now()` inside the per-key closure, so wall-clock could advance between key trials during JWKS rotation and reject a token within its expiry leeway window. Sample `now` once before the key loop, mirroring `JWTPDP.ValidateToken`.
- Closes #928 — `httpSession.readUpstream` (and the stdio `StdioProxy.readUpstream`) discarded every upstream read error identically, making a `bufio.ErrTooLong`/malformed-JSON termination indistinguishable from a clean EOF. Now log non-`io.EOF` errors to stderr before returning.
- Closes #927 — stdio shutdown deadlocked when `--upstream-timeout=0` and an in-flight upstream call never returned. Precisely: this fixes the **proxy-internal circular wait** — `serveHost`'s `wg.Wait()` waited on a handler parked in `awaitUpstreamResponse`, whose only remaining wakeup (`upstreamDone`) depended on `closeUpstreamInput`, which `Start` called only after `serveHost` returned. Now close the upstream input inside `serveHost` before `wg.Wait()` (idempotent with `Start`'s later call), and guard `closeUpstreamInput` against a nil subprocess pipe. **Residual edge (out of scope, inherent to opting out of timeouts):** with `--upstream-timeout=0`, a subprocess that holds stdout open without responding *and* does not exit on stdin close still parks `wg.Wait()` — there is no ctx cancellation to fall back on; that is the operator's explicit choice. Documented in a code comment.
- Closes #926 — `pruneRotated` early-returned on a `rotatedSiblings` directory-read error with no log, so a persistent listing failure (e.g. audit dir execute bit removed) silently stopped retention pruning. Now log the error, consistent with the per-file removal-failure path.

## Testing

- `go build ./...` — clean.
- `make check-fmt`, `make check-license`, `make lint` (go vet + golangci-lint) — all clean.
- Added regression tests: `TestMergeManifests_EmptyInput` (#931) and `TestKillEndpoint_OffHostNonPOST_RejectedByLoopbackFirst` (#930).


internal/transport/stdio.go
```diff
@@ -462,6 +468,18 @@ func (p *StdioProxy) serveHost(ctx context.Context) {
...
+	// awaitUpstreamResponse, which unblocks on response / ctx.Done() / upstreamDone.
+	// With --upstream-timeout=0 the context is never canceled by a host stdin EOF,
+	// so without this the only remaining wakeup is upstreamDone — but that closes
+	// only after closeUpstreamInput, which Start otherwise calls only after
+	// serveHost returns, i.e. after wg.Wait(): a circular wait that hangs the
+	// proxy. Closing here lets the upstream EOF, readUpstream return, upstreamDone
+	// close, and pending callers unblock. closeUpstreamInput is idempotent (HTTP
+	// bridge guards with closeOnce; the subprocess double-Close error is ignored),
+	// so Start's later call is a no-op (#927).
+	p.closeUpstreamInput() //nolint:contextcheck // teardown path: the upstream session-termination DELETE (#773) intentionally uses a detached, bounded background context — close/reaper/signal/shutdown carry no request context.
```
eunolabs, line 488: **Non-blocking observation (the fix is correct).** This correctly breaks the *proxy-internal* circular wait: closing the upstream input here makes `upstreamDone` reachable, and `awaitNonced` then unblocks via either the real response or `errUpstreamExited`. Worth noting the residual edge it can't cover: with `--upstream-timeout=0`, if the subprocess holds its stdout open without responding *and* doesn't exit on stdin close, `readUpstream` never returns, `upstreamDone` never closes, and `wg.Wait()` still parks. That's inherent to opting out of timeouts (ctx is never canceled, so there's no other wakeup) and is the operator's choice, not something this PR needs to solve — flagging only so the "deadlock fixed" claim is read as "the self-inflicted circular wait is fixed," which it is.

---

eunolabs (commented):
## Review — review-session bug batch (6 fixes)

Reviewed the full diff and verified at head `ed43448`: `go build ./...` clean, and the touched packages' targeted tests pass (`internal/config`, `internal/transport`, `pkg/capability`). **All six fixes are correct, minimal, and preserve the fail-closed posture.** No blocking issues.

- **#931 `MergeManifests` empty slice** — the `len(ms)==0` early return yielding `&LocalManifest{}` (zero capabilities = deny-all) is the right fail-closed default and is ordered correctly before the `len(ms)==1` fast path. Test covers it. ✔
- **#930 `handleKill` ordering** — running `loopbackOnly` before the method check matches `handleHealth`/`handleMetrics` and stops an off-host caller from confirming the endpoint via a 405. The new test asserts 403-before-405. ✔
- **#929 JWKS clock sampling** — hoisting `now := c.now()` above the per-key closure removes the wall-clock-advances-between-key-trials window during rotation, consistent with `JWTPDP.ValidateToken`. Leeway is still applied via `ValidateWithLeeway`. ✔
- **#928 upstream read-error logging** — non-`io.EOF` errors (`bufio.ErrTooLong`, parse failures) now logged on both the HTTP-session and stdio paths; `io.EOF` stays quiet. Imports already present in both files (build confirms). ✔
- **#927 stdio shutdown deadlock** — closing the upstream input inside `serveHost` before `wg.Wait()` makes `upstreamDone` reachable for a handler parked in `awaitNonced` under `--upstream-timeout=0`; the call is idempotent with `Start`'s later one (sequential, no race) and `closeUpstreamInput` is now nil-guarded. Correct; left one non-blocking note on the residual unresponsive-subprocess edge inline. ✔
- **#926 `pruneRotated` listing error** — logging the `rotatedSiblings` error before returning makes a persistent listing failure as visible as the per-file removal failures, instead of silently halting retention pruning. ✔

LGTM.

---

internal/transport/stdio.go
eunolabs: Agreed — thanks for the precise framing. The claim is exactly "the self-inflicted circular wait is fixed," not "all timeout=0 hangs are eliminated." I've tightened the wording in both the code and the PR body to scope it that way and to document the residual edge: under `--upstream-timeout=0`, a subprocess that holds stdout open without responding *and* doesn't exit on stdin close still parks `wg.Wait()`, because there's no ctx cancellation to fall back on — inherent to opting out of timeouts, and the operator's explicit choice. See a108ac5.

---

PR #934, opened 2026-06-20
fix: static `make build` binary and guard Docker VERSION sentinel

### LINKED ISSUE #838: make build produces a dynamically-linked binary, not the advertised static binary
**Severity (v0.1.0 release audit): NICE-TO-HAVE**

## Summary
The project advertises "one static binary," and the release path is static (`CGO_ENABLED=0` in `.goreleaser.yml` and `deploy/docker/Dockerfile.mcp`). But the local `make build` target does not set `CGO_ENABLED=0`, so on a host with a C toolchain it produces a dynamically-linked binary (links libc via net / os/user).

## Where
- `Makefile` `build:` target — `go build` invoked without `CGO_ENABLED=0`.

Observed on this branch: `file bin/eunox-mcp` → "dynamically linked ... interpreter /lib64/ld-linux-x86-64.so.2"; `ldd` shows `libc.so.6`.

## Impact
Cosmetic / dev-experience parity only — shipped artifacts are static. But a developer running `make build` gets an artifact that doesn't match what ships, which can mask CGO-only behavior differences.

## Suggested fix
Add `CGO_ENABLED=0` to the `build` target (and optionally `-trimpath`) so `make build` matches the release closure.

### LINKED ISSUE #843: Dockerfile ARG VERSION=0.0.0 sentinel can ship a 0.0.0-reporting image on ad-hoc builds
**Severity (v0.1.0 release audit): NICE-TO-HAVE**

## Summary
`deploy/docker/Dockerfile.mcp` (and `.windows`) default `ARG VERSION=0.0.0`. A plain `docker build` without `--build-arg VERSION=...` produces a binary that reports `0.0.0`. Both supported build paths (`make docker-build-mcp`, release.yml) always pass the arg, so production images are correct.

## Where
- `deploy/docker/Dockerfile.mcp:15`
- `deploy/docker/Dockerfile.mcp.windows:23`

## Impact
Only affects ad-hoc local `docker build` invocations; production images are fine.

## Suggested fix
Optional: keep the sentinel (it is intentional) but document that images must be built via `make docker-build-mcp` / release.yml, or fail the build when VERSION is unset.

eunolabs:
## Summary

Two build-hygiene fixes so local builds match the release closure and ad-hoc Docker builds can't ship a misversioned image.

### #838 — `make build` produced a dynamically-linked binary
The `build:` target invoked `go build` without `CGO_ENABLED=0`, so on a host with a C toolchain it linked libc dynamically — diverging from the advertised "one static binary on purpose" and from the release path (`.goreleaser.yml` and `deploy/docker/Dockerfile.mcp` both set `CGO_ENABLED=0`).

**Fix:** prefix the build command with `CGO_ENABLED=0` and add `-trimpath`, matching the release closure. No other target depends on extra env, so nothing else is affected.

Verified locally:
```
$ make build && file bin/eunox-mcp
bin/eunox-mcp: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, ...
```

### #843 — `ARG VERSION=0.0.0` sentinel could ship a 0.0.0-reporting image
Both Dockerfiles defaulted `ARG VERSION=0.0.0`. A plain `docker build` without `--build-arg VERSION=...` produced a binary reporting `0.0.0`.

**Chosen option: fail the build (the issue's preferred option).** I dropped the `0.0.0` default and added a `RUN` guard that errors when `VERSION` is empty or `0.0.0`. This is safe because both supported build paths always pass the arg:
- `make docker-build-mcp` / `docker-build-mcp-multi` → `--build-arg VERSION=$(VERSION)` (Makefile lines 134, 147)
- `.github/workflows/release.yml` → `build-args: VERSION=...` (lines 207, 308, 386)

The guard never fires on a legitimate build; it only catches the ad-hoc-`docker build` footgun. Applied consistently to `Dockerfile.mcp` and `Dockerfile.mcp.windows` — the Windows builder stage is a Debian-based Linux cross-compile image, so the `/bin/sh` guard works there too.

## Verification
- `make check-fmt` — clean
- `make build` — statically linked (see above)
- `go build ./...` — clean
- Docker was not run in this environment; the release/Make Docker paths were verified statically to pass `--build-arg VERSION=...`.

Fixes #838
Fixes #843

---


deploy/docker/Dockerfile.mcp
```diff
@@ -12,10 +12,19 @@
...
 
-ARG VERSION=0.0.0
+ARG VERSION
 ARG TARGETOS=linux
 ARG TARGETARCH=amd64
 
+# Fail closed on an unset version. A plain `docker build` without
+# `--build-arg VERSION=...` would otherwise stamp the binary "0.0.0" and ship an
+# image that misreports its own version. Both supported build paths
+# (`make docker-build-mcp` and .github/workflows/release.yml) always pass it.
```
eunolabs: This claim is incomplete, and it's the source of the red CI. The demo stacks also build this Dockerfile via `docker compose up --build` and **do not** pass `--build-arg VERSION`:

- `demo/docker-compose.yml` (→ `demo / manifest-only`, `demo / JWT mode`)
- `demo/docker-compose.gateway.yml` (→ `demo / gateway`)
- `demo/opa-comparison/docker-compose.yml` (→ all three `OPA comparison /*` jobs)

So the new guard fires during those builds:
```
#18 0.168 VERSION must be passed via --build-arg (use make docker-build-mcp or release.yml)
target eunox-gateway: failed to solve ... exit code: 1
```

The guard itself is fine — the gap is that the demo compose build paths weren't covered. I pushed a fix that adds `args: { VERSION: ${VERSION:-0.1.0-demo} }` to each demo `build:` section (see summary comment).

---

eunolabs (commented):
## Review + CI triage

The two build-hygiene changes themselves are correct and well-reasoned:

- **`make build` (`CGO_ENABLED=0` + `-trimpath`)** — right fix, and it now matches `.goreleaser.yml` / `Dockerfile.mcp`. Verified locally the binary is statically linked.
- **VERSION sentinel guard** — good defensive idea; failing closed on an unset/`0.0.0` version is the right call.

But the PR as-is turns **7 CI jobs red**, and there's an unrelated 8th failure that's pre-existing. I dug into all of them and pushed fixes to **`claude/ci-issues-pr-934-ix565b`** (branched from this PR's head). Details:

### 1. VERSION guard breaks every Docker-based demo job (regression introduced by this PR)
`demo / manifest-only`, `demo / gateway`, `demo / JWT mode`, and all three `OPA comparison /*` jobs fail because the demo stacks build `deploy/docker/Dockerfile.mcp` via `docker compose up --build` without passing `--build-arg VERSION` (see inline comment). The PR description's "both supported build paths always pass it" missed the demo compose paths.

**Fix:** added a `VERSION` build-arg to the three demo compose files, defaulting to a clear non-release marker and overridable from the environment:
```yaml
build:
  context: ..
  dockerfile: deploy/docker/Dockerfile.mcp
  args:
    VERSION: ${VERSION:-0.1.0-demo}
```
(`docker-compose.yml`, `docker-compose.gateway.yml`, `opa-comparison/docker-compose.yml`; the JWT stack is an overlay on `docker-compose.yml` and inherits it.) Confirmed `docker compose config` resolves `VERSION: 0.1.0-demo`, and `0.1.0-demo` passes the guard while `""` / `0.0.0` still fail it.

### 2. `demo / stdio` — pre-existing race, **not** caused by this PR
This one reproduces on `main` (`f04905b`) with none of this PR's changes, so it's independent of the build work — but it's red on the same checks list, so worth flagging. It was introduced by #933's `serveHost` change, which closes the upstream input **before** `wg.Wait()`. A `tools/call` pipelined right behind the host's stdin EOF (exactly what the CI script does) is dispatched in a goroutine that then loses the race to write to the upstream:
```
FAIL  read_file /reports/q3.pdf → ALLOW  (want=allow got=deny)
```
**Fix:** drain in-flight handlers (`wg.Wait()`) before returning to `Start` (which then closes the input), restoring the long-standing pre-#933 ordering. Parked handlers stay bounded by `--upstream-timeout` (30s default); the only residual is the already-documented `--upstream-timeout=0`-against-a-hung-upstream opt-out. Added a regression test (`TestStdioProxy_RequestPipelinedBeforeEOF_NotDropped`) that reproduces the exact `file already closed` drop without the fix.

Since #2 is a separate concern from this PR's scope, feel free to split that commit into its own PR if you'd rather keep this one purely build-hygiene — happy to do that.

### Verification on the fix branch
- `go test -race ./cmd/... ./internal/...` and `go test ./pkg/...` — pass
- `demo/scripts/ci-test-stdio.sh` — 9/9 pass (was 6/9)
- `gofmt` — clean
- Docker builds couldn't be exercised here (no daemon for `docker build` in this sandbox), but the guard condition and compose arg resolution were verified directly.

Two small notes (non-blocking):
- Worth a one-line mention in the PR body / a comment near the guard that demo compose stacks must pass `VERSION` too, so the next person adding a stack doesn't rediscover this.
- The guard message points at `make docker-build-mcp` / `release.yml`; consider also naming the demo path now that it's a supported `--build-arg` consumer.


---

deploy/docker/Dockerfile.mcp
eunolabs: Addressed in 6095871: the demo compose stacks now pass `VERSION` (via `build.args`), and I updated this guard's comment + failure message to name them as a supported build path alongside `make docker-build-mcp` and `release.yml`, so the next person adding a demo stack won't rediscover this. Left `Dockerfile.mcp.windows` as-is since no demo stack builds it.

All three CI fixes are now on this branch (`claude/triage-build`): the demo VERSION build-arg, the stdio drain race, and this doc fix.

---

Issue #891, opened 2026-06-19
[rev-20260619T0830Z-f2a9d1] dispatchToolsCall missing empty-name guard returns AUTHORIZATION_FAILED instead of INVALID_PARAMS and writes blank identifier to audit log


eunolabs:
## Summary

`dispatchResourcesRead` and `dispatchPromptsGet` both validate that the incoming identifier is non-empty and return `-32602 INVALID_PARAMS` if it is not. `dispatchToolsCall` has no equivalent guard: a `tools/call` request with an empty or missing `name` field is silently passed to the PDP, which denies it with `AUTHORIZATION_FAILED` (`-32001`). This returns the wrong error code to the host and writes an audit record with a blank identifier.

## The bug

**`dispatchResourcesRead` (line 118):**
```go
if params.URI == "" {
    return errorResponse(msg.ID, -32602, "resources/read: uri must not be empty")
}
```

**`dispatchPromptsGet` (line 152):**
```go
if params.Name == "" {
    return errorResponse(msg.ID, -32602, "prompts/get: name must not be empty")
}
```

**`dispatchToolsCall` — no equivalent guard.** A `tools/call` with `params.Name == ""` is passed directly to:
```go
dec := d.pdp.Decide(d.decideCtx(ctx), d.sessionID, EnforceTarget{Type: capability.TargetTypeTool, Name: ""}, params.Arguments, d.sourceIP)
```

The PDP finds no manifest entry for `""`, denies with `AUTHORIZATION_FAILED`, and the audit record is written as:
```go
fp.rec.RecordDeny(ctx, fp.sessionID, ""/*auditID*/, "tools/call", capability.ErrCodeAuthorizationFailed, ...)
```

## Impact

1. **Wrong error code**: An invalid-params condition returns `-32001 AUTHORIZATION_FAILED` instead of `-32602 INVALID_PARAMS`. MCP clients (and operators inspecting logs) cannot distinguish a real access-policy denial from a malformed request.

2. **Audit log corruption**: The `identifier` field in the deny record is `""`. Audit analysis tools that group or query by identifier would either silently drop these records or aggregate them incorrectly with other zero-value identifiers.

3. **Denial audit record is charged against the session's policy counters**: If the enforcement engine has any per-session call counters wired at the `""` identifier, an empty-name call consumes one count.

## Relevant code

[...]

Issue #836, opened 2026-06-18
Remote-HTTP upstreams silently drop server-initiated requests (roots/list, elicitation/create hang)


eunolabs:
**Severity (v0.1.0 release audit): SHOULD-FIX (decide before shipping remote-HTTP as supported)**

## Summary
For **subprocess** upstreams, server-initiated requests (`sampling/createMessage`, `roots/list`, `elicitation/create`) are handled fail-closed — `broadcastServerRequest` replies a JSON-RPC error so the upstream unblocks immediately. The **remote-HTTP** upstream bridge has no equivalent: it is request/response only and never surfaces upstream-initiated requests. A remote upstream that issues one over its own SSE channel blocks until the session is torn down.

## Where
- `cmd/eunox-mcp/stdio_http_upstream.go`, `cmd/eunox-mcp/http_remote.go` — no reader path for server-initiated requests.
- `cmd/eunox-mcp/route.go:244` — partial mitigation: a `system:sampling/createMessage` opt-in is refused for an HTTP upstream, but `roots/list` / `elicitation/create` are not covered.
- Subprocess fail-closed path for contrast: `broadcastServerRequest` in `http.go`.

## Impact
A remote upstream that legitimately issues `roots/list` or `elicitation/create` hangs silently rather than receiving a prompt fail-closed error. Currently documented only as an "MVP limitation" in a source comment.

## Suggested fix
Either (a) have the remote bridge drain and reply a JSON-RPC error to any server-initiated request it cannot service (mirroring the subprocess path), or (b) document the limitation in user-facing docs and decide whether remote-HTTP upstreams are "supported" for v0.1.0.

_Found during a v0.1.0 release-readiness audit._

PR #935, opened 2026-06-20
fix: guard empty-name tools/call per transport and document remote-upstream server-initiated limitation

### LINKED ISSUE #837: tools/call lacks the empty-name guard that every other enforced method has
**Severity (v0.1.0 release audit): NICE-TO-HAVE**

## Summary
`resources/read`, `resources/subscribe`, and `prompts/get` all reject an empty identifier with JSON-RPC `-32602` before calling the PDP. `tools/call` does not — an empty `params.Name` flows into `Decide`.

## Where
- `cmd/eunox-mcp/http.go:979` — `handleHTTPToolsCall` (no empty-name guard).
- `cmd/eunox-mcp/stdio.go:515` — `handleToolsCall` (no empty-name guard).
- Precedent to mirror: `http.go:1177-1178` / `stdio.go:748-749` (prompts/get).

## Impact
Not a privilege escalation: with a specific grant, an empty name fails authorization correctly. The only reachable case is a manifest carrying a `tool:*` wildcard, where `matchesResource("*","")` is true, so a nameless `tools/call` matches the wildcard and is forwarded — still within policy, but the proxy forwards a structurally invalid request it should reject with `INVALID_PARAMS`, and the asymmetry with the other three methods is a latent inconsistency. There are empty-identifier tests for the other methods but none for `tools/call`.

## Suggested fix
Add `if params.Name == "" { ... -32602 ... }` after the decode in both handlers, plus a table entry in `cmd/eunox-mcp/enforcement_gaps_test.go`.

eunolabs:
## Summary

Triages two transport-correctness issues.

### Fixes #837 — `tools/call` empty-name guard

The empty-name guard for `tools/call` **already exists** in the shared `dispatchToolsCall` (`internal/transport/dispatch.go:97-99`), added under #891 when the per-transport handlers (`handleHTTPToolsCall` / `handleToolsCall`) were unified into the shared `dispatchRequest`. Both the HTTP and stdio transports route `tools/call` through that single path, so both already reject an empty tool name with `-32602` (INVALID_PARAMS) before the PDP is consulted — matching `resources/read`, `resources/subscribe`, and `prompts/get`.

What was missing was the per-transport regression coverage the issue asks for. Added:
- `TestGap4_HTTPProxy_ToolsCall_EmptyName_Error` — HTTP transport end-to-end: asserts `-32602` and that the upstream is never called (`CountByMethod("tools/call") == 0`), mirroring `TestGap4_HTTPProxy_PromptGet_EmptyName_Error`.
- `TestStdioHandleToolsCall_EmptyName_InvalidParams` — stdio transport via `handleHostRequest`: asserts `-32602`, mirroring `TestStdioHandlePromptsGet_ManifestPDP_EmptyName`. The proxy is built by `closedUpstream` (upstreamDone already closed), so any upstream call would surface `-32603`; asserting `-32602` proves the guard short-circuited first.

### Addresses #836 — remote-HTTP upstreams drop server-initiated requests

**Chose option (b) (document + clarify comments), not option (a) (drain inbound channel).**

Reason: option (a) is not cleanly feasible — there is no inbound channel to drain. Studying the code:
- The stdio host's remote bridge (`stdio_http_upstream.go`) does one POST per host request via `httpUpstream.post` and queues only that POST's own response body onto `incoming`; `Read` returns those queued responses. There is no persistent SSE GET stream from the upstream.
- The gateway's per-session remote path (`http_remote.go` / `doRemoteHTTP` / `DoMCPHTTP`) is identical in shape: one POST, read that POST's body. The SSE handling (`sseResponseForID`) only correlates the *response to the POST by id* and **skips** any interleaved non-matching event — so even an interleaved server-initiated request would be dropped, never handled.
- A repo-wide check confirms no `http.MethodGet` SSE stream is ever opened against a remote upstream.

So a remote upstream has no channel on which to deliver `sampling/createMessage` / `roots/list` / `elicitation/create`. Unlike the subprocess path (`readUpstream` → `handleUpstreamRequest` replies a fail-closed JSON-RPC error so the upstream unblocks), there is nothing to drain. Servicing these would require opening + maintaining a persistent upstream SSE GET stream and wiring a whole new reader path — a large refactor, explicitly out of scope per the issue.

Per option (b):
1. Added a "Limitation: server-initiated requests need a stdio upstream" section to `docs/architecture.md`.
2. Clarified the known-limitation comments in `stdio_http_upstream.go` and `http_remote.go`, cross-referencing the existing startup rejection of a `system:sampling/createMessage` opt-in on an HTTP upstream (`route.go`), which already fails closed.

No behavior change for #836 — comments + docs only.

## Verification

- `make fmt` / `make check-fmt` — clean
- `go build ./...` — clean
- `go test -race -count=1 ./internal/transport/...` — `ok` (8.969s)
- New tests pass individually (`TestGap4_HTTPProxy_ToolsCall_EmptyName_Error`, `TestStdioHandleToolsCall_EmptyName_InvalidParams`, plus existing `TestDispatchToolsCall_EmptyName_InvalidParams`).

---


internal/transport/enforcement_gaps_test.go
```diff
@@ -1832,3 +1832,63 @@ func TestDispatchToolsCall_EmptyName_InvalidParams(t *testing.T) {
...
+	result := decodeRPC(t, resp)
+
+	if result.Error == nil {
+		t.Fatal("expected JSON-RPC error for empty tool name")
+	}
+	if result.Error.Code != -32602 {
+		t.Errorf("error code = %d, want -32602 (INVALID_PARAMS) (#837)", result.Error.Code)
+	}
+	if fu.CountByMethod("tools/call") != 0 {
+		t.Error("upstream must not be called for an empty tool name")
```
eunolabs, line 1885: Nice — this is a strictly stronger assertion than the sibling it mirrors. `TestGap4_HTTPProxy_PromptGet_EmptyName_Error` only checks `result.Error != nil`, whereas this both pins the exact code (`-32602`) and asserts the upstream was never reached. That upstream-not-called check is what actually proves the guard short-circuits *before* the PDP: the manifest grants `tool:*`, so absent the empty-name guard the request would be forwarded and `CountByMethod` would be 1.

Optional follow-up (out of scope here): consider backfilling the same two assertions into the `prompts/get` and `resources/read` empty-identifier siblings so the empty-name contract is covered uniformly rather than just for `tools/call`.

---

internal/transport/enforcement_gaps_test.go
```diff
@@ -1832,3 +1832,63 @@ func TestDispatchToolsCall_EmptyName_InvalidParams(t *testing.T) {
...
+}
+
+// TestStdioHandleToolsCall_EmptyName_InvalidParams is the #837 regression for the
+// stdio transport: an empty tools/call name routed through handleHostRequest must
+// be rejected with INVALID_PARAMS (-32602) before the PDP/upstream are reached,
+// mirroring TestStdioHandlePromptsGet_ManifestPDP_EmptyName. The proxy is built by
+// closedUpstream (upstreamDone already closed), so reaching the upstream would
+// instead surface an upstream-error (-32603, errUpstreamExited): asserting -32602
+// here proves the empty-name guard short-circuited before any upstream call.
+func TestStdioHandleToolsCall_EmptyName_InvalidParams(t *testing.T) {
```
eunolabs, line 1896: Good use of `closedUpstream` here: because `upstreamDone` is already closed, any forward would surface an upstream error (`-32603`), so asserting `-32602` genuinely distinguishes "guard short-circuited" from "reached the upstream." The `t.Fatal` on `len(hw.messages) != 1 || Error == nil` also correctly guards the subsequent `.Error.Code` deref (the `prompts/get` sibling uses `t.Error` but never dereferences, so it doesn't need it — this one does, and gets it right).

---

docs/architecture.md
```diff
@@ -158,6 +158,26 @@ can actually invoke.
...
+upstream. A remote HTTP upstream therefore has no channel on which to deliver a
+server-initiated request, and any such request it attempts is never seen,
+enforced, audited, or forwarded to the host — the upstream would block awaiting a
+response it will never receive. This applies to both the stdio host's remote
+bridge (`internal/transport/stdio_http_upstream.go`) and the gateway's
+per-session remote path (`internal/transport/http_remote.go`).
+
+A manifest `system:sampling/createMessage` opt-in on an HTTP upstream is rejected
+at startup rather than loaded as a silently inert grant
+(`internal/transport/route.go`). If you need server-initiated requests (sampling,
```
eunolabs, line 182: The limitation is described accurately and I verified the mechanics: `DoMCPHTTP` is POST-only (no `http.MethodGet`/SSE stream is opened against a remote upstream), and `sseResponseForID` correlates strictly by id and skips any interleaved event — so a server-initiated request riding the POST response stream would indeed be silently dropped while the upstream blocks.

One suggestion for operator clarity: it may be worth a sentence noting that this is a **liveness/compatibility** limitation, not a policy bypass — no unenforced server-initiated action reaches the host; the request simply deadlocks upstream-side, and the `route.go` startup rejection prevents the misleading `system:sampling/createMessage` opt-in from looking enforced. Readers scanning a security-sensitive proxy's docs will want that distinction made explicit.

---

eunolabs (commented):
## Review — looks solid, ship it

A tight, well-scoped triage PR. I verified the substantive claims against the code rather than taking them at face value:

**#837 (empty-name `tools/call` guard) — confirmed already-present, coverage added correctly**
- The guard does live in the shared `dispatchToolsCall` (`dispatch.go:97-99`) and runs before the PDP, so both transports already reject `-32602`. The PR is purely backfilling the per-transport regression tests the issue asked for.
- Both new tests are correct and actually *stronger* than the siblings they mirror (exact `-32602` + upstream-never-called for HTTP; `-32602` vs the `-32603` a closed upstream would yield for stdio). The `t.Fatal`/`t.Error` nil-deref guarding is right.

**#836 (remote-upstream server-initiated requests) — analysis verified, docs accurate**
- Confirmed `DoMCPHTTP` is POST-only with no GET/SSE stream against a remote upstream, and `sseResponseForID` correlates strictly by id and skips interleaved events — so a server-initiated request really is dropped while the upstream blocks. Option (b) (document + comment, no behavior change) is the right call; option (a) genuinely has nothing to drain.
- The `route.go` cross-reference is accurate: a `system:sampling/createMessage` opt-in on an HTTP upstream is rejected at startup, so this isn't a silent fail-open — consistent with the repo's fail-closed invariant.

**Non-blocking suggestions** (left inline):
1. Consider backfilling the stronger assertions onto the `prompts/get` / `resources/read` empty-name siblings for uniform coverage.
2. The architecture-doc section could note explicitly that this is a liveness/compatibility limitation, not a policy bypass.

Nothing here blocks merge. Test-and-docs only, no runtime behavior change, fail-closed preserved.

---

Issue #842, opened 2026-06-18
Idle SSE-subscribed session pins its upstream subprocess until the client disconnects


eunolabs:
**Severity (v0.1.0 release audit): NICE-TO-HAVE**

## Summary
The idle-session reaper exempts any session with an open GET/SSE stream (`hasSubscribers()`), by design — a well-behaved host holds SSE open for the whole session. A client that sends `initialize` + `GET` and then goes silent holds its upstream subprocess indefinitely.

## Where
- `cmd/eunox-mcp/http.go:1675-1679` — `hasSubscribers()` exemption from the idle reaper.

## Impact
Blast radius is bounded by `maxSessions` (default 512) and the behavior is documented in a code comment, so it's acceptable for v0.1.0. But on a network-exposed bind it is a cheap way to pin up to `maxSessions` upstream subprocesses with no further traffic.

## Suggested fix
Consider an absolute session-lifetime ceiling, or count "no host *requests* for N×idle" toward reaping even when an SSE stream is open.

_Found during a v0.1.0 release-readiness audit._

eunolabs:
Fixed via a hard idle ceiling in the session reaper. Each session now tracks `lastRequest` (atomic), and a session with no host *request* for `hardIdleMultiplier` (4) × the idle window is reaped even with an open SSE stream — exactly the "count no host requests for N×idle toward reaping" approach suggested here (`internal/transport/http_session.go`). A well-behaved host that polls or calls tools keeps resetting `lastRequest` and is never hard-reaped; only a genuinely silent SSE-only session hits the ceiling, so it can no longer pin its upstream subprocess indefinitely. Closing as completed.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #936, opened 2026-06-20
fix: join Redis kill-switch goroutines and bound idle SSE sessions

### LINKED ISSUE #835: Redis kill-switch goroutines are never joined on shutdown (ksRedis.Stop() not called)
**Severity (v0.1.0 release audit): SHOULD-FIX**

## Summary
`cmdProxy` calls `ksRedis.Start(ctx)` but never calls the matching `ksRedis.Stop()`. The Redis kill-switch's pub/sub listener and reconcile loop were given a `Stop()` specifically so a caller can `wg.Wait()` them before freeing the Redis client; that join never happens in production.

## Where
- `cmd/eunox-mcp/main.go:392-393` — `if ksRedis != nil { ksRedis.Start(ctx) }`, with no corresponding `defer ksRedis.Stop()`.

## Impact
Bounded to process lifetime (ctx cancellation does stop the goroutines eventually), so this is not an unbounded leak. But on shutdown the goroutines can still be mid-`refreshState` (an open Redis round-trip) while the process tears down, and the documented "safe to free the client after Stop" contract is never exercised. Only affects Redis-backed (multi-instance) deployments.

## Suggested fix
Add `if ksRedis != nil { defer ksRedis.Stop() }` immediately after `Start`, ordered to run before the Redis client is closed.

eunolabs:
## Summary

Two operational-lifecycle fixes.

### #835 — Redis kill-switch goroutines never joined on shutdown
`cmdProxy` called `ksRedis.Start(ctx)` but never the matching `Stop()`, so the Redis kill-switch's pub/sub listener and reconcile loop outlived the proxy and could touch the shared Redis client and logger after the process believed teardown was done.

Fix: register `defer ksRedis.Stop()` immediately after `Start`. `Stop()` cancels the lifecycle context and blocks on `wg.Wait()`, so both background goroutines have exited before the process returns. `Stop()` is idempotent (guarded by `lifeMu`/`stopped`, with a `cancel != nil` check), and the existing `pkg/killswitch` start/stop tests already cover the join — so no new test is added there (no clean unit seam exists for the `cmdProxy` main flow itself).

### #842 — Idle SSE-subscribed session pins its upstream subprocess
The idle-session reaper spared any session holding an open SSE stream (`hasSubscribers()`), so a client that sent `initialize` + `GET` and then went silent pinned its upstream subprocess indefinitely (bounded only by `maxSessions`, default 512).

Fix: track the last host **request** (`lastRequest`, advanced only on `/mcp` POSTs via a new `touchRequest()`) distinct from SSE liveness (`lastActive`, still advanced when a GET/SSE stream is opened or held). The reaper keeps the normal SSE-open exemption, but adds a conservative **hard idle ceiling**: a subscribed session is reaped once it has received no host request for `hardIdleMultiplier` × the idle window, even with an open SSE stream.

**Threshold chosen: `hardIdleMultiplier = 4`.** A host that calls tools even infrequently keeps resetting `lastRequest` and is never hard-reaped; only a genuinely silent SSE-only session hits the ceiling. Picked a constant rather than a new config knob to keep the safety net's scope small and predictable.

## Tests
- `internal/transport/operational_hardening_test.go`: `TestIdleReaper_SparesSubscribedSessionWithRecentRequest` (SSE-open + recent request → spared) and `TestIdleReaper_ReapsSubscribedSessionPastHardCeiling` (SSE-open + past the hard ceiling → reaped).
- Existing reaper/start-stop tests still pass.

## Docs
- README operational-controls table updated to describe the hard ceiling.

## Verification
`make fmt`, `make check-fmt`, `make check-license`, `go build ./...`, `go vet`, and `go test -race -count=1 ./internal/transport/... ./cmd/eunox-mcp/... ./pkg/killswitch/...` all pass.

Fixes #835, #842

---


eunolabs:
## Review + CI triage

I reviewed the two lifecycle fixes and traced the red `demo / stdio` check.

### The two changes in this PR look correct

**#835 — `defer ksRedis.Stop()`**
- `Stop()` cancels the kill-switch's *own* lifecycle context (the `r.cancel` it stored from `context.WithCancel(ctx)` inside `Start`), not the outer `ctx`. So the defer reliably joins the pub/sub listener + reconcile loop even on a normal stdin-EOF shutdown where the outer `ctx` is never cancelled — not only on a signal.
- Defer LIFO ordering is right: `ksRedis.Stop()` (registered after `Start`) runs before the earlier-registered `sink.Close()`, so the Redis goroutines are joined before the audit tape is flushed/closed.

**#842 — hard idle ceiling for SSE-only sessions**
- The `touch()` / `touchRequest()` split is consistent: GET/SSE advances `lastActive` only, POST advances both, both are seeded in `registerSession`, and the reaper reads them atomically without the proxy lock.
- `reapOnce`'s `switch` reaps every qualifying session — the two arms only differ in the log reason — so nothing slips through. `hardIdleMultiplier = 4` as a constant rather than a config knob is a reasonable call for a safety net.
- Tests cover both the spared (recent request) and reaped (past ceiling) subscribed-session cases.

Nit (non-blocking): the `lastActive` doc comment says it's updated "on opening/holding a GET (SSE) stream," but `touch()` only fires on stream *open*, not continuously while held. It doesn't matter functionally — a held stream keeps `hasSubscribers()` true, which is what actually spares it from the normal reaper — but the wording slightly oversells it.

### The CI failure is pre-existing and unrelated to this PR

`demo / stdio` fails with three allowed calls returning `upstream error` (-32603). Root cause: in `serveHost` (stdio transport), the upstream's stdin was closed *before* in-flight request handlers had written their forwarded request:

```
```

A host that pipelines `initialize` + `tools/call` and then closes stdin (which the demo, and any batch client, does) races the handler's outbound write against `p.closeUpstreamInput()`, so an allowed call comes back as a spurious error.

I bisected it: `b48e667` passes, `f04905b` (#933) fails. The premature `closeUpstreamInput()` was added in #933 (referencing #927) to break a `--upstream-timeout=0` + hung-upstream shutdown wait, but it broke the common fast-close path. It is **not** introduced by this PR — #936 just inherits a red base.

**Fix:** I pushed a one-file change to branch `claude/ci-issues-pr-936-lxpnwv` that restores the pre-#927 ordering — `wg.Wait()` with the upstream input still open, then `Start` closes it after `serveHost` returns. The `--upstream-timeout=0` edge it was guarding is the operator's explicit opt-out of per-call timeouts (the default 30s timeout still cancels a stuck handler via `ctx.Done`), and a healthy upstream always answers or exits, so it doesn't arise in normal shutdown.

Verified with the fix:
- `demo / stdio`: 9/9 pass (was 6/9)
- `demo / e2e`: 19 legs + 9 audit assertions pass
- `go test -race ./internal/transport/...`: green; `go vet` + `gofmt` clean

That fix lives on a separate branch per my branch constraints — it should be folded into this PR's base (or cherry-picked onto `claude/triage-lifecycle`) to turn `demo / stdio` green here.

---

PR #937, opened 2026-06-20
fix: align gateway JSON Schema to the config loader and strengthen parity test

### LINKED ISSUE #839: Gateway JSON Schema diverges from the Go config loader; schema-parity test only diffs field names
**Severity (v0.1.0 release audit): NICE-TO-HAVE (all divergences fail safe)**

## Summary
`schemas/eunox-gateway-config.schema.json` is stricter than the Go loader in several places, so a config the loader accepts can be flagged invalid by a schema-aware editor/CI (never the reverse — all fail safe). The schema-parity test doesn't catch this because it only compares a flat set of field names.

## Where / specifics
- Schema sets `port` `minimum: 1` (schema:32-36) but `gateway_config.go:348-352` treats `port: 0` as "unset → default 3000".
- Schema enums for `transport` (schema:18) and `enforcement` (schema:134,201) exclude `""`, but `validate()` accepts `transport: ""` (`gateway_config.go:332`) and `enforcement: ""` (`validEnforcementValue`, ~line 472).
- Schema sets `minimum: 0` on `audit.rotateSizeBytes` (schema:89) and `defaults.upstreamTimeoutMs` (schema:141); the loader has no lower-bound check (negatives are coerced to defaults at runtime).
- `TestGatewaySchema_MatchesStructs` (`gateway_test.go:1002`) collects yaml tags vs schema property keys into two **flat name sets** and diffs names only — type / enum / parent-object / `required` / `additionalProperties:false` drift is invisible to it.

## Impact
No security impact (the Go loader is the enforcing surface and fails safe). It's a developer-confusion / tooling-accuracy gap, and the weak parity test is why the divergences went unnoticed.

## Suggested fix
Either align the schema to accept `0` / `""`, or have `validate()` reject explicit `port: 0` / empty enums / negatives so the two surfaces match exactly. Strengthen the parity test to assert per-object (parent-qualified) property sets and `additionalProperties:false` on every object node.

eunolabs:
Fixes #839.

## Problem

The Go config loader (`internal/config/gateway_config.go`) is the enforcing surface; the published JSON Schema (`schemas/eunox-gateway-config.schema.json`) is what editors and CI validate against. Four schema constraints were **stricter** than the loader, so a config the loader accepts could be flagged invalid by a schema-aware editor/CI (never the reverse — every divergence failed safe). The parity test only diffed flat field-name sets, so this drift was invisible.

## Schema changes (aligned to the loader — never loosen the loader)

I read `validate()` / `validEnforcementValue` / the port/timeout/rotateSize handling first to confirm exactly what the loader accepts before editing.

1. **`listen.port`: `minimum: 1` → `minimum: 0`.** The loader treats `port: 0` as "unset → default 3000" and only range-checks `[1, 65535]` when the value is non-zero (`validate()`, the `cfg.Listen.Port != 0` guard).
2. **`transport` and `enforcement` enums: add `""`.** `validate()` accepts `transport: ""` (`case "", HostTransportHTTP, HostTransportStdio`) and `validEnforcementValue("")` returns true. Empty string means "default/inherit". Added to all three enums (top-level `transport`, `defaults.enforcement`, per-upstream `enforcement`).
3. **`audit.rotateSizeBytes`: drop `minimum: 0` entirely.** The loader has *no* lower-bound check on this field; a non-positive value is coerced to the default at runtime. Removing the bound (rather than setting a negative minimum) is the faithful representation of "no lower bound enforced".
4. **`defaults.upstreamTimeoutMs`: drop `minimum: 0`, keep `maximum`.** The loader treats `<= 0` as "unset → flag/default" (`resolveUpstreamTimeout`) and only enforces the upper overflow bound. So only `maximum` is asserted (still pinned to `config.MaxDurationMs` by the existing `TestGatewaySchema_DurationMaxMatchesConstant`).

Each changed field's `description` now documents the loader behavior inline.

## Strengthened parity test

The old `TestGatewaySchema_MatchesStructs` collected yaml tags and schema property keys into two **flat global name sets** and diffed names only — blind to wrong-parent placement, enum/type drift, and missing `additionalProperties: false`.

- **`TestGatewaySchema_MatchesStructs` rewritten** to assert **per-object, parent-qualified** property sets — `root`, `listen`, `audit`, `defaults` (via `$ref` resolution), and `upstreams[*]` (via `items` `$ref`). A field on the wrong parent object now fails (verified: moving `keyPath` from `audit` to `listen` in the schema makes the test fail with a clear per-object message). `$ref` is resolved to `$defs`; transport `then`-branch `"<field>": false` entries are treated as forbidden-on-branch, not declared properties.
- **New `TestGatewaySchema_EveryObjectForbidsAdditionalProperties`** walks every node and requires `additionalProperties: false` on every genuine object **definition** (`type: object`). It deliberately skips `if`/`then`/`else`/`anyOf`/`oneOf` constraint fragments, which carry a `properties` block but are partial constraints layered onto a base object — sealing them would break validation.

## Verification

- `make fmt`, `make check-fmt`, `make check-license`, `go build ./...`, `go vet` — all clean.
- `go test -race -count=1 ./internal/config/... ./cmd/eunox-mcp/...` — green.
- Schema validated as well-formed JSON.
- Negative check: a wrong-parent mutation fails the strengthened test; the unmutated aligned schema passes.

---


eunolabs:
## Review + CI triage

### The schema/test changes look correct

I checked each of the four schema relaxations against the enforcing surface (`internal/config/gateway_config.go`) and they're all faithful — the schema now matches what the loader actually accepts, never looser on the enforcing side:

- **`listen.port` `minimum: 0`** — loader only range-checks `[1, 65535]` when the port is non-zero and HTTP (`gateway_config.go:349-351`); `0` means unset → default. ✓
- **`transport` enum `""`** — `validate()` has `case "", HostTransportHTTP, HostTransportStdio` (`:333`). ✓
- **`enforcement` enums `""`** — `validEnforcementValue("")` returns `true` (`:471-474`). ✓
- **`audit.rotateSizeBytes` no lower bound** — there is no validation; a non-positive value is coerced to the default in `internal/audit/audit.go:302` (`if rotateSizeBytes <= 0`). ✓
- **`defaults.upstreamTimeoutMs` drop `minimum`, keep `maximum`** — loader only asserts `> MaxDurationMs` (`:378`); `<= 0` is treated as unset. ✓

The per-object, parent-qualified rewrite of `TestGatewaySchema_MatchesStructs` and the new `EveryObjectForbidsAdditionalProperties` walk are a real improvement over the old flat name-set diff. No concerns.

### The red check (`demo / stdio`) is pre-existing and unrelated to this PR

This PR only touches `schemas/eunox-gateway-config.schema.json` and `cmd/eunox-mcp/gateway_test.go` — neither is on the stdio request path. The failure **reproduces on the base commit `f04905b`** (and on `main`), so it is not introduced here.

**Root cause** (`internal/transport/stdio.go`, `serveHost`): on host stdin EOF, the proxy closed the upstream input *before* waiting on in-flight request handlers — the #927 deadlock-break. A `tools/call` that arrives just before EOF is still mid-flight in its handler goroutine (PDP decision, then the upstream write over `p.upIn`, the very pipe `closeUpstreamInput` closes). Closing it first yanks the pipe out from under the forward:

```
{"jsonrpc":"2.0","id":2,"error":{"code":-32603,"message":"upstream error"}}
```

The demo harness pipes `initialize` + `initialized` + `tools/call` in one shot then EOFs immediately, so every **allow** case races the teardown and fails deterministically (deny cases pass because they never touch the upstream). The integration suite missed it because its harness drains each response *before* sending EOF, so nothing is ever in flight at shutdown.

**Fix** (prepared on branch `claude/ci-issues-pr-937-7jr9wp`, on top of this PR's head): `serveHost` now waits for in-flight handlers before closing the upstream input, bounded by `shutdownMs` so the #927 deadlock-break is preserved (if a handler is wedged on an upstream that neither answers nor exits under `--upstream-timeout=0`, the grace expires and we close anyway, unblocking it via its `<-done` arm). Added `TestStdioProxy_InFlightRequestSurvivesHostEOF`, which mirrors the demo (close stdin immediately after a `tools/call` without draining) and asserts a result rather than `-32603` — verified it fails on the old ordering and passes with the fix. `demo / stdio` and `demo / e2e` go 9/9 green locally, and `internal/transport` / `internal/config` / `cmd/eunox-mcp` pass under `-race`.

Since the fix is in a different file than this PR and addresses a separate base-branch bug, I kept it on its own branch rather than pushing here. Happy to push it onto `claude/triage-schema` so this PR's CI goes fully green, or land it as its own PR — let me know which you'd prefer.

---

eunolabs:
Pushed the stdio shutdown-race fix onto this branch as `20ed397` (on top of `0f35de4`), so the `demo / stdio` check should go green on the next run. The two commits are independent — the schema/test alignment and the transport fix touch disjoint files.

---

Issue #840, opened 2026-06-18
Threat-model audit-field table (section 6.1) omits 5 emitted top-level fields


eunolabs:
**Severity (v0.1.0 release audit): NICE-TO-HAVE**

## Summary
The audit-record field table in `docs/threat-model-mcp.md` §6.1 lists 18 fields, but the record struct emits 23. Missing from the table: `seq`, `prev_hmac`, `upstream`, `policy_version`, `policy_sha256`. CONTRIBUTING.md requires the threat model to be updated when audit fields are added.

## Where
- `docs/threat-model-mcp.md:522-542` — the §6.1 field table.
- `cmd/eunox-mcp/audit.go:130-155` — the emitted record struct (23 fields).

## Impact
Documentation completeness only. All five fields *are* described elsewhere in the repo (`seq`/`prev_hmac` in the HMAC-chain section §3.4; `upstream`/`policy_version`/`policy_sha256` in the gateway prose and `docs/architecture.md`), so this is a table gap, not an undocumented field.

## Suggested fix
Add the five rows to the §6.1 table (or add a pointer to §3.4 for the chain fields and the gateway section for the route/policy fields).

_Found during a v0.1.0 release-readiness audit._

eunolabs:
The §6.1 audit-field table in `docs/threat-model-mcp.md` now lists all five previously-missing rows: `seq`, `prev_hmac`, `upstream`, `policy_version`, and `policy_sha256`. The table is back in sync with the emitted record struct. Closing as completed.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #833, opened 2026-06-18
Audit pipeline is fail-open on the request path: permitted calls can be forwarded with no durable audit record


eunolabs:
**Severity (v0.1.0 release audit): SHOULD-FIX / decision**

## Summary
A permitted (`allow`) request is forwarded to the upstream and recorded only *after* the upstream returns; the record is enqueued non-blockingly and can be dropped (full queue) or lost (write error) without blocking or reversing the call. The result is that a privileged call can reach the host with **no durable audit entry**, which is in tension with the product's "complete, tamper-evident audit trail" guarantee.

## Where
- `cmd/eunox-mcp/forward.go:65` — `callUpstream(ctx, msg)` runs first.
- `cmd/eunox-mcp/forward.go:85` — the `allow` record is `Record(...)`'d *after* the upstream call.
- `cmd/eunox-mcp/audit.go` — `Record(...)` is non-blocking; a full queue drops the record and a write error in the drainer loses it.
- `--require-audit` (main.go) only gates sink-open at **startup**, not per-record loss at runtime.

## Impact
Under disk pressure, a full audit queue, or a write error, an `allow` decision executes and is forwarded with no on-disk record. Drops are tamper-evident via an in-band `AUDIT_RECORDS_DROPPED` marker **only if the writer later recovers**; a writer broken straight through shutdown surfaces the loss only out-of-band (the `eunox_audit_dropped_records_total` / `eunox_audit_write_failures_total` counters and a non-zero exit).

This is a deliberate, documented tradeoff today (CLAUDE.md: "warn rather than abort so a corrupted local log does not block enforcement"). Filing so it becomes an explicit, signed-off decision rather than a surprise post-release.

## Suggested direction
1. Document the fail-open posture prominently in `docs/threat-model-mcp.md` (the exact conditions under which an `allow` can be unaudited).
2. Consider a `--require-audit=strict` mode that fails the request **closed** when the drop/write-failure counters increment — at minimum on the synchronous deny path, ideally gating the allow path too.

_Found during a v0.1.0 release-readiness audit._

eunolabs:
[...]

PR #938, opened 2026-06-20
sec: reject all-zero audit HMAC key and document audit-trail posture

### LINKED ISSUE #834: Audit HMAC key is not checked for weak/all-zero material
**Severity (v0.1.0 release audit): SHOULD-FIX**

## Summary
`parseAuditKeys` validates only that an operator-supplied HMAC key decodes to exactly 32 bytes; it accepts an all-zero or otherwise low-entropy key. A predictable key lets anyone forge audit records that pass `audit-verify`, defeating the tamper-evidence guarantee.

## Where
- `cmd/eunox-mcp/audit.go:2369` — `parseAuditKeys(...)` performs length-only validation.
- Auto-generated keys correctly use `crypto/rand` (`audit.go:2208`), so this only affects externally supplied keys via `--audit-key-path` / `EUNOX_AUDIT_KEY_PATH`.
- Several tests construct sinks with `key := make([]byte, 32)` (all-zero), which the current code accepts.

## Impact
An operator who supplies a weak key (e.g. 64 hex zeros) gets a log that *looks* signed but is trivially forgeable — silently failing the "tamper-evident" property.

## Suggested fix
Reject an all-zero key in `parseAuditKeys` (cheap, fail-closed), and document a minimum-entropy expectation for externally supplied keys. Update the affected tests to use non-zero key material.

eunolabs:
## Summary

A small bucket of audit-hardening fixes across the audit subsystem and the threat model.

### Issue #834 (security) — reject weak/all-zero audit HMAC key
`parseAuditKeys` (internal/audit/audit.go) now rejects an all-zero 32-byte HMAC key in addition to the existing exact-32-byte length check, failing closed at startup. An all-zero key is the canonical no-entropy placeholder a misconfigured deployment supplies (a zeroed file, an unset secret rendered as zeros) and makes HMAC-SHA256 signatures trivially forgeable, defeating the tamper-evidence control. Only externally-supplied keys (`--audit-key-path` / `EUNOX_AUDIT_KEY_PATH`) reach this path; auto-generated keys use `crypto/rand` and are unaffected. Adds focused `parseAuditKeys` subtests (all-zero active key, all-zero retired key, non-zero accepted) and a direct `isAllZeroKey` test, and moves the direct-construction test keys off all-zero `make([]byte, 32)` placeholders onto a shared non-zero test key helper. Documents the minimum-entropy expectation for operator-supplied keys in the threat model (§3.4) and `docs/architecture.md`.

### Issue #840 (docs) — complete the §6.1 audit-field table
Adds rows for the five fields the table omitted but the `auditRecord` struct emits: `seq`, `prev_hmac` (the HMAC chain, §3.4) and `upstream`, `policy_version`, `policy_sha256` (gateway routing), cross-checked against the emitting struct in internal/audit/audit.go. Formatting matches the existing rows.

### Issue #833 (security, docs-only) — document the allow-path fail-open posture
Adds threat-model §5.4.1 documenting that a permitted (`allow`) request is forwarded and recorded only after the upstream returns, via a non-blocking enqueue onto the bounded audit queue, so under a full queue / write error / process loss an `allow` can be forwarded with no durable on-disk record — never blocking or reversing the call. Notes the existing mitigations (in-band `AUDIT_RECORDS_DROPPED` marker, the `eunox_audit_dropped_records_total` / `eunox_audit_write_failures_total` counters, non-zero exit, and that `--require-audit` gates only sink-open at startup) and references `internal/transport/forward.go` and `internal/audit/audit.go`. A stricter `--require-audit=strict` mode is noted as possible future work only — not implemented in this PR.

## Verification
- `make fmt` / `make check-fmt` — clean
- `go build ./...` — clean
- `go test -race -count=1 ./internal/audit/...` — ok (47.7s)
- `go test -race -count=1 ./cmd/... ./internal/transport/...` — ok (audit-key load paths)

Fixes #834, #840; documents #833.

---


internal/audit/audit.go
```diff
@@ -1745,6 +1745,19 @@ func parseAuditKeys(data []byte) ([][]byte, error) {
...
+		// entropy: it is the canonical "uninitialized"/placeholder value a
+		// misconfigured deployment is most likely to supply (a zeroed file, a
+		// `head -c32 /dev/zero | xxd`, an unset secret rendered as zeros), and it
+		// makes the HMAC-SHA256 signature trivially forgeable by anyone who knows
+		// the (public) construction. Externally-supplied keys reach parseAuditKeys
+		// via --audit-key-path / EUNOX_AUDIT_KEY_PATH; auto-generated keys come from
+		// crypto/rand and never produce all-zero material in practice. Fail closed so
+		// a weak key is rejected at startup instead of silently signing a forgeable
+		// audit log.
+		if isAllZeroKey(key[:n]) {
```
eunolabs, line 1758: Well-scoped fix, and good that the check runs per-line so a weak *retired* key is rejected too.

One optional thought (no change required): this catches the all-zero placeholder but not the other equally-canonical no-entropy value a misconfig produces — an **all-same-byte** key (a secret template rendered as repeated `0x01`, a file filled with `0xFF`). Generalizing to "reject a key with a single distinct byte" would be a one-liner, but it's also reasonable to leave out to avoid the slippery slope toward a general entropy estimator. If the narrow scope is deliberate, no change needed — the threat-model wording ("zeros are the canonical placeholder") frames it honestly rather than over-claiming that all weak keys are rejected. Flagging only so the choice is explicit.

---

internal/audit/audit.go
```diff
@@ -1753,6 +1766,19 @@ func parseAuditKeys(data []byte) ([][]byte, error) {
 	return keys, nil
 }
 
+// isAllZeroKey reports whether key consists entirely of zero bytes. An all-zero
+// HMAC key is the canonical placeholder a misconfigured deployment supplies and
+// carries no entropy, so parseAuditKeys rejects it (fail closed). Kept as a tiny
+// helper so the weak-key check has a single, testable definition.
+func isAllZeroKey(key []byte) bool {
+	for _, b := range key {
```
eunolabs, line 1774: Nit (non-blocking, fine as-is): the early `return false` means this isn't constant-time — it leaks via timing how many leading zero bytes the key has. That's acceptable here: it runs once at startup on a local operator-supplied file, there's no attacker-timed oracle, and a key reaching this branch is being rejected *because* it has no secrecy. Noting it only so the non-constant-time choice is on record for a crypto-adjacent path; `subtle.ConstantTimeCompare(key, make([]byte, 32))` would sidestep the question entirely if you'd prefer to keep this corner timing-clean.

---

eunolabs (commented):
## Review — audit-hardening bucket (#834 / #840 / #833)

Solid, well-targeted PR. The code change is small and fails closed in the right place, the docs are factually precise, and the tests cover the new behavior. I verified the claims rather than taking them at face value — details below. Two **optional, non-blocking** notes left inline; nothing here needs to change to merge.

### What I checked

**#834 — reject all-zero HMAC key**
- The check sits inside the per-line loop in `parseAuditKeys`, after the length check, so it rejects an all-zero key on **any** line (active *or* retired) — the `all-zero retired key fails closed` test confirms this.
- It covers every entry point: both the Sink-open path (`audit.go:316`) and `audit-verify` (`cmd/eunox-mcp/main.go:1725`) load keys through `LoadOrCreateKeys → parseAuditKeys`. There's no inline-key flag/env that bypasses it — only `--audit-key-path` / `EUNOX_AUDIT_KEY_PATH` (path-based) reach this code, so the PR's "only externally-supplied keys reach this path" is accurate. Auto-generated keys come from `crypto/rand` and don't pass through the check on generation.
- Tests are well-formed: all-zero active key, all-zero retired key, non-zero accepted, plus a direct `isAllZeroKey` predicate test. The shared `nonZeroTestKey()` helper lives in `audit_more_test.go` and is used from `partialwrite_test.go` — both are `package audit`, so it compiles.

**#840 — §6.1 audit-field table**
- Cross-checked all five new rows against the `auditRecord` struct (`audit.go:45–64`): `seq`, `upstream`, `policy_version`, `policy_sha256`, `prev_hmac` all exist, the `omitted when…` notes match the `omitempty` tags (`seq`/`prev_hmac` have none and are documented as always-present; the other three are `,omitempty` and documented as omitted), and the table order matches struct order. Accurate.

**#833 — §5.4.1 allow-path fail-open (docs-only)**
- The central behavioral claim holds: `enforcedForwardCore` (`internal/transport/forward.go`) calls `callUpstream` (line 75), applies redaction, then `RecordAllow` (line 109) — i.e. records strictly after the upstream has answered.
- Secondary references are correct too: the queue is `auditChannelSize = 4096`, `Sink.Record` is the non-blocking enqueue, and `eunox_audit_dropped_records_total` / `eunox_audit_write_failures_total` / the `/healthz` `auditWriteFailed` field all exist as described.

### Notes (optional)
1. Scope of the weak-key check — all-zero only, not other canonical placeholders (e.g. all-same-byte). Inline on `audit.go:1758`. Fine to leave as-is; the threat-model wording scopes it honestly.
2. `isAllZeroKey` isn't constant-time. Inline on `audit.go:1774`. Acceptable here (startup-only, local, no attacker-timed oracle) — noted for the record.

### Heads-up (not a concern)
Because `audit-verify` shares `parseAuditKeys`, a pre-existing log that was (mis)signed with an all-zero key becomes unverifiable after this lands. That's consistent with the pre-1.0 "no back-compat shims" stance and is arguably the correct outcome — such a tape was never tamper-evident to begin with — so I'd only mention it in the release notes, not gate on it.

---

Issue #902, opened 2026-06-19
Engineering code review: correctness, simplification, and polish (June 2026)


eunolabs:
## Health summary

eunox-mcp is a well-engineered, security-first proxy. The fail-closed invariant is consistently applied across every code path — missing manifest, unmapped method, malformed JWT, unset condition arg all produce structured denials. The HMAC audit chain design is sound (double-marshal, chain-resume on restart, integrity markers). The generic `filterListResult[E,L]` scaffolding eliminates the classic copy-paste divergence trap across all `*/list` paths. Non-obvious defenses — `descriptionHash` pinning, length-prefixed counter keys, exhaustive JWT capability allowlists, DNS-rebinding guards — reflect principled security thinking and are correctly implemented. The highest-leverage improvements are a real hot-path allocation concern in the JSON-RPC parser, a latent lock-ordering issue in the session reaper, and several simplification opportunities that would reduce the cognitive surface area of the PDP and session code.

## What is already done well

- Fail-closed enforced everywhere without exception; structured error codes distinguish `CAPABILITY_DENIED` from `AUTHORIZATION_FAILED`
- HMAC audit chain is robust: double-marshal body-then-sign pattern, chain-resume on restart with HMAC verification of tail, integrity markers on open/close
- `filterListResult[E,L]` generic scaffolding is fail-closed on parse/marshal error and eliminates per-method copy-paste
- `compositeCounterKey` length-prefixed encoding prevents cross-namespace counter-key injection (#273)
- `rpcMsg.UnmarshalJSON` correctly distinguishes notification (absent `id`) from null-id request via two-pass probe — a subtle correctness requirement
- DNS-rebinding guard is conservative: absent Origin passes freely; present-but-unparseable is denied
- `descriptionHash` pinning prevents prompt injection via tool description rewrite
- `redactFields` obligations are fail-closed on unparseable containers; `SetEscapeHTML(false)` avoids mangling JSON values (#481)
- `awaitUpstreamResponse` three-way select (response / context cancel / upstream-exited) is the correct structure for upstream round-trips; dup-ID guard is properly implemented
- `cloneAndBoundDetails` / `boundAuditObligations` deep-clone before enqueue prevents TOCTOU on mutable caller data (#461)
- `decideTarget` stray-schema guard runs before the audit-mode defer — correctly by design, and the comment explains why

---

### (a) Correctness & Risk

---

#### CR-1: `rpcMsg.UnmarshalJSON` allocates a full map on every incoming message

[...]

Issue #899, opened 2026-06-19
[rev-20260619T0905Z-45d49e] broadcast() silently drops upstream notifications for slow SSE subscribers with no observability


eunolabs:
# `broadcast()` silently drops upstream notifications for slow SSE subscribers with no observability

**Location:** `cmd/eunox-mcp/http_session.go`
**Reference:** https://github.com/eunolabs/eunox/blob/8fbcb5ba04237648f20dd98b0cef33c23f572599/cmd/eunox-mcp/http_session.go
**Severity:** Medium
**Session:** rev-20260619T0905Z-45d49e

## Description

`broadcast` delivers upstream SSE notifications to all active subscribers using a non-blocking channel send:

```go
// http_session.go — broadcast()
for _, ch := range s.subs {
    select {
    case ch <- msg:
    default:
        // silent drop
    }
}
```

Each subscriber channel is buffered at size 16. When a subscriber is slow (e.g. a client with a stalled TCP connection, a slow consumer, or a momentary burst of upstream events), the buffer fills and subsequent messages are dropped without any log line, metric, counter increment, or returned error indicating loss occurred.

The messages routed through `broadcast` include critical MCP state-change notifications:

- `notifications/tools/list_changed`
- `notifications/resources/list_changed`
- `notifications/prompts/list_changed`

A dropped `notifications/tools/list_changed` leaves the client operating on a stale tool list with no indication that its view is out of date. The client will not know to re-issue `tools/list` because it never received the trigger. This is a silent data-loss path that can cause clients to call deprecated or removed tools, receive unexpected errors, and generally behave incorrectly without any operator visibility.

`broadcastServerRequest` (server-initiated requests requiring a response) does have a fail-closed path: if delivery returns `false` it writes an error back upstream. However plain `broadcast` for notifications has no such fallback — the return value `delivered bool` is ignored at every call site that uses it for notifications.

## Trigger / repro

1. Start the HTTP proxy with multiple SSE subscribers.
2. Have one subscriber consumer pause reads (or throttle the SSE connection artificially).
3. Generate >16 upstream notifications in quick succession (e.g. a rapidly changing tool list).
4. Observe that the paused subscriber misses notifications with no log output and no error returned to the upstream MCP server.

## Suggested fix

At minimum, increment an atomic counter (similar to the existing `s.dropped` counter on `auditSink`) and expose it via the `/stats` subcommand so operators can detect subscriber lag:

```go
default:
[...]

PR #939, opened 2026-06-20
fix: triage safe code-review findings (#902 CR-1..CR-4, PO-1..PO-3)

eunolabs:
Addresses #902: CR-1..CR-4, PO-1..PO-3.

Implements the SAFE subset of the #902 code-review meta-issue. The larger refactors (SR-1..SR-4) are deliberately left out of this PR.

Note: code was relocated since the issue was written; symbol locations were confirmed by grep before editing (e.g. `jsonrpc.go` is split between `internal/mcp/jsonrpc.go` and `internal/transport/jsonrpc.go`; `http_session.go`/`http.go`/`http_security.go` moved to `internal/transport/`; `pdp.go` to `internal/pdp/`).

## Implemented

- **CR-1 — single-pass JSON-RPC decode (no probe map).** `internal/mcp/jsonrpc.go`: `RPCMsg.UnmarshalJSON` now does one struct decode into an alias whose `id` is a value `json.RawMessage`. A zero-length raw means the `"id"` key was absent (notification); a present key — including the literal `null` — yields non-empty bytes (request/response with a present-null id preserved). The probe map and the second full unmarshal are gone. Exact semantics preserved; the `#788` regression test and the `FuzzMsgReader` fuzz target still pass, and a string-id and an integer-id-response case were added.
- **CR-2 — reapOnce lock-ordering.** `internal/transport/http_session.go`: `hasSubscribers` takes `s.notifMu`, so calling it under `p.mu` established the order `p.mu -> s.notifMu` (a latent deadlock against `broadcast()`, which holds `s.notifMu`). Now the session refs are snapshotted under `p.mu`, the lock is released, then the idle/subscriber checks run on the snapshot. Reaping behavior is unchanged (`lastActive` is atomic; `hasSubscribers` re-reads live state).
- **CR-3 — serverReqTracker capacity drop observability.** `internal/transport/jsonrpc.go`: the 1024-entry bounded set evicts an arbitrary in-flight ID when full. `trackServerReqID` now reports whether it evicted, and `serverReqTracker.track` logs a single warn line on the first eviction (then counts silently), so a re-hung server-initiated request (#648) is no longer invisible. No sink counter (out of scope); a log line only. Existing `TestTrackServerReqID` extended for the new return value + eviction flag.
- **CR-4 — invariant test for stray-schema + audit-only.** `internal/pdp/pdp.go`: the explanatory comment on the stray-`argumentSchema` guard (which runs before the `IsAuditOnly()` defer) is elevated to an `// INVARIANT:` marker. New test `TestDecideTarget_StraySchema_AuditOnly_StillDenies` asserts a stray-schema match with `enforcement: audit` still produces a hard `ENFORCEMENT_ERROR` deny with `AuditOnly == false`. No behavior change.
- **PO-1 — document serveCtx fallback.** `internal/transport/http.go`: added `// serveCtx returns the serve-lifetime context. Only nil in tests that bypass Serve.`
- **PO-2 — session id in checkOrigin denial record.** `internal/transport/http_security.go`: `checkOrigin` now reads `Mcp-Session-Id` from the request headers and passes it as the session id on the pre-session `ORIGIN_REJECTED` audit record. New test `TestOriginRejection_AuditRecordCarriesSessionID` verifies the `session_id` field is populated.
- **PO-3 — clarify msToDuration overflow comment.** `internal/transport/http.go`: saturation comment reworded to name the threat — `// Prevent overflow: values above ~292 years saturate to MaxInt64.`

## Deliberately left out

- **SR-1..SR-4** — the larger refactors. Out of scope for this safe-subset PR.

## Verification

- `make fmt`, `make check-fmt`, `make check-license` — clean.
- `go build ./...` — clean; `go vet` on the touched packages — clean.
- `go test -race -count=1 ./internal/mcp/... ./internal/transport/... ./internal/pdp/...` — all pass.
- `FuzzMsgReader` — 8s run, PASS (confirms CR-1 preserves decode semantics).

No security behavior was weakened: fail-closed paths, structured errors, and the append-only audit shape are all preserved. No new top-level audit field was added (PO-2 only changes the value of an existing `session_id` field).

---


internal/transport/jsonrpc.go
```diff
@@ -57,13 +61,34 @@ func trackServerReqID(ids map[string]struct{}, key string) map[string]struct{} {
...
+	var evicted bool
+	t.ids, evicted = trackServerReqID(t.ids, key)
+	warn := evicted && !t.warnedEviction
+	if warn {
+		t.warnedEviction = true
+	}
 	t.mu.Unlock()
+	if warn {
+		fmt.Fprintf(os.Stderr,
+			"[eunox-mcp] WARNING: server-initiated request tracker reached its %d-entry cap; evicting an in-flight request ID to make room (the evicted request may hang until session teardown). Further evictions counted but not individually logged.\n",
```
eunolabs: The message says `Further evictions counted but not individually logged`, and the PR description says it "then counts silently" — but nothing actually counts them. `serverReqTracker` only gained a `warnedEviction bool`; there's no eviction counter, and the `track` doc comment just above even states the drop is "otherwise silent."

The #899 dropped-notification path this explicitly says it mirrors *does* keep a real tally (`s.droppedNotifs.Add(1)`). For true parity, add an `atomic.Uint64` eviction count here; otherwise drop "counted" from the message (e.g. "…evicting an in-flight request ID to make room; subsequent evictions are suppressed.") so the log line doesn't promise a number that's never recorded.

Minor — not blocking.

---

internal/transport/http_session.go
```diff
@@ -329,14 +329,26 @@ func (p *HTTPProxy) reapIdleSessions(ctx context.Context) {
 // return), keeping the function deterministic for tests.
 func (p *HTTPProxy) reapOnce(idle time.Duration) {
 	cutoff := time.Now().Add(-idle).UnixNano()
-	var stale []*httpSession
+	// Snapshot the session set under p.mu, then release it BEFORE the idle and
+	// subscriber checks. hasSubscribers takes s.notifMu, so calling it while holding
+	// p.mu would establish the lock order p.mu -> s.notifMu; broadcast() already
+	// holds s.notifMu, so any path that reached for p.mu under it would deadlock
```
eunolabs: Correct fix and good hygiene — but worth being precise about the framing. None of the `s.notifMu` critical sections (`hasSubscribers`, `broadcast`, `addSub`, `removeSub`) ever reach for `p.mu`, so there is no `notifMu -> p.mu` path anywhere today and therefore no *currently reachable* deadlock. `broadcast()` holds `notifMu` but never acquires `p.mu`, so the old code couldn't actually deadlock against it.

The comment's conditional phrasing ("any path that reached for p.mu under it would deadlock") is technically accurate, but the PR description's "a latent deadlock against `broadcast()`" overstates it. Suggest framing this as preventing a *future* lock-order inversion rather than fixing a present deadlock, so a maintainer doesn't go hunting for a reverse path that isn't there.

The snapshot itself is behavior-preserving — `lastActive` is atomic and `hasSubscribers` re-reads live state, so a session that goes active or gains a subscriber after the snapshot is still spared. 👍

---

internal/pdp/pdp.go
```diff
@@ -405,15 +405,17 @@ func (p *ManifestPDP) decideTarget(ctx context.Context, sessionID string, target
...
+	// stray-argumentSchema match is a hard DENY even when the matched constraint is
+	// audit-only (enforcement: audit). It is an ENFORCEMENT_ERROR — an engine bug,
+	// not a policy verdict — so "fail closed on ambiguity" wins over "audit mode
+	// never blocks", exactly as collectObligations deliberately does NOT mark its
+	// analogous unwired-directive deny AuditOnly. Were this return inside the
+	// defer's scope (as the action-check and INVALID_PARAMS returns deliberately
+	// are, since those ARE downgradable policy verdicts), the blanket defer would
+	// stamp AuditOnly=true and the transport would log-and-forward the request
+	// guarded by an unenforced schema — the exact opposite of fail closed (issue
+	// #671). TestDecideTarget_StraySchema_AuditOnly_StillDenies pins this.
```
eunolabs, line 418: Nice regression test — it genuinely pins the ordering. I confirmed the audit-mode `defer func() { resp.AuditOnly = true }()` is registered *after* this guard (and after the descriptionHash pin), so relocating it above this `return` flips `AuditOnly` to true and fails `TestDecideTarget_StraySchema_AuditOnly_StillDenies`.

One optional follow-up: the descriptionHash pin a few lines down (`if schema == validateSchema && matched.IsPinnedExactTool()`, #884) relies on the *identical* "must run before the audit-mode defer" invariant, but isn't pinned by any test. A single refactor that relocates the defer would silently re-open both fail-opens, yet only the stray-schema one now has a guard test. A sibling case for the descriptionHash-in-audit-mode path would be cheap insurance — out of scope for this safe subset, just flagging.

---

eunolabs (commented):
Reviewed the full diff against the current `internal/{mcp,transport,pdp}` sources. This is a clean, well-scoped safe-subset PR — the comments are unusually thorough and every functional change is backed by a test (the rest are comment-only). CI is green across all 35 checks.

Verified each item:

- **CR-1 (single-pass JSON-RPC decode)** — ✅ correct. The value-`json.RawMessage` alias distinguishes an absent id (zero-length → nil ID → notification) from a present-null (`null` bytes → non-nil ID → request/response), because `*json.RawMessage`'s `UnmarshalJSON` fires for a `null` literal on a *value* field but is skipped on a *settable pointer* field — which is exactly why the old code needed the probe map. No second unmarshal and no input-buffer aliasing (`RawMessage.UnmarshalJSON` copies). The new string-id and integer-id-response cases are exercised by the table harness via `IsRequest`/`IsResponse`/`IsNotification`.
- **CR-2 (reapOnce lock-ordering)** — ✅ correct and behavior-preserving. One framing nit inline: there's no *currently reachable* deadlock, so this is preventive rather than a fix for an existing one.
- **CR-3 (eviction observability)** — ✅ correct; the warn-once latch is computed under `mu` and logged after unlock, so no double-log race. One wording nit inline: the message claims subsequent evictions are "counted," but nothing counts them.
- **CR-4 (stray-schema invariant test)** — ✅ the test pins the real invariant (`defer` registered after the guard). One optional follow-up inline: the sibling descriptionHash pin shares the identical invariant but isn't similarly pinned.
- **PO-1 / PO-3** — comment-only clarifications, fine.
- **PO-2 (session id in ORIGIN_REJECTED)** — ✅ a real correlation improvement. The recorded id is client-claimed/unauthenticated, which the comment correctly calls out, and JSON-escaping in the JSONL sink neutralizes any log-injection concern. No new top-level audit field, so no threat-model update is required — agree with the PR's reasoning.

Nothing blocking — the three inline notes are all minor (one small wording fix on the CR-3 log line is the only concrete code change I'd suggest). Nice work.

---

internal/transport/jsonrpc.go
eunolabs: Addressed in 69e6ad5: replaced the `warnedEviction bool` with an `evictions uint64` tally (guarded by `mu`) that increments on every eviction, so the "counted" claim is now true — the first eviction logs, the rest are tallied. Reconciled the doc comments and added `TestServerReqTracker_TalliesEvictions` to pin it. Went with a plain `uint64` under the existing mutex rather than `atomic.Uint64`, since `track`/`take` always hold `mu` and there's no lock-free reader.

---

internal/transport/http_session.go
eunolabs: Addressed in 69e6ad5: reworded the comment to state plainly that no `notifMu -> p.mu` path exists today (broadcast/hasSubscribers/addSub/removeSub are all `notifMu` leaves), so there is no current deadlock, and that keeping the acquisitions disjoint is *preventive* against a future inversion — no longer implying an existing deadlock against `broadcast()`.

---

internal/pdp/pdp.go
eunolabs: Correction — this is already covered, so no new test was added. The descriptionHash-pin-in-audit-mode invariant is pinned by the pre-existing `TestDecideTarget_DescriptionHashDenyNotDowngradedByAuditMode` (`internal/pdp/pdp_test.go`, #884), which asserts a hash mismatch on an `enforcement: audit` constraint stays a hard `AUTHORIZATION_FAILED` deny with `AuditOnly == false` — the exact sibling case I claimed was missing. My original note that it was unpinned was wrong.

---

Issue #887, opened 2026-06-19
[rev-20260619T0710Z-6e6d16] applyInitializeResult treats a malformed/empty upstream initialize result as a successful handshake (fail-open vs #708)


eunolabs:
**Location:** `cmd/eunox-mcp/jsonrpc.go:314-328` (`applyInitializeResult`)
**Reference:** https://github.com/eunolabs/eunox/blob/8fbcb5ba04237648f20dd98b0cef33c23f572599/cmd/eunox-mcp/jsonrpc.go#L314-L328
**Severity:** low-medium (fail-open in the upstream handshake; contradicts the #708 fail-closed intent)

**Description**

`applyInitializeResult` validates an upstream's `initialize` response for all three handshakes (stdio, local-HTTP, remote-HTTP). It fails closed on an explicit JSON-RPC **error** response (#708), but silently accepts two other malformed shapes as success:

```go
func applyInitializeResult(resp rpcMsg) (caps map[string]interface{}, serverVersion string, err error) {
    if resp.Error != nil {
        return nil, "", fmt.Errorf("upstream initialize rejected: ...")   // #708: fail closed
    }
    if resp.Result != nil {
        var result mcpInitResult
        if json.Unmarshal(resp.Result, &result) == nil {   // unmarshal error is DISCARDED
            caps = result.Capabilities
            ...
        }
    }
    return caps, serverVersion, nil   // err == nil even when Result was malformed or absent
}
```

Two fail-open paths:

1. **Malformed `result`.** If `resp.Result` is present but does not unmarshal into `mcpInitResult` (e.g. the upstream returns `"result": "garbage"` or `"result": 42`), the `== nil` guard simply skips the assignment and the function returns `(nil, "", nil)` — a *success* with empty capabilities.
2. **No `result` and no `error`.** A response carrying neither (`{"jsonrpc":"2.0","id":N}`) is a malformed JSON-RPC reply, yet `isResponse()` accepts it, it matches the initialize id, and `applyInitializeResult` returns `(nil, "", nil)`.

In both cases the caller proceeds as though initialize succeeded: it sends `notifications/initialized`, registers the session, and hands the host a session backed by an upstream whose initialize was never actually verified. That is precisely the outcome #708 set out to prevent ("proceeding to notifications/initialized would signal a successful handshake to an upstream that explicitly failed and hand the client a session backed by an un-initialized upstream"). It also conflicts with the project-wide *fail closed on any ambiguity* invariant.

[...]

PR #953, opened 2026-06-20
fix: nil-safe AuditOnlyCount matching HasMaxCalls contract

eunolabs:
## Summary

- `AuditOnlyCount()` on `*LocalManifest` dereferenced `m.Capabilities` without checking for a nil receiver, unlike its sibling `HasMaxCalls()` which has an explicit nil guard.
- Every call site worked around this by wrapping in `if manifest != nil`, but the correct fix is for the method to be nil-safe so callers can't forget the guard.

## Changes

- `internal/config/manifest_policy.go`: add `if m == nil { return 0 }` guard at the top of `AuditOnlyCount`, matching the pattern in `HasMaxCalls`
- `internal/config/manifest_policy_test.go`: extend the nil-receiver test to cover `AuditOnlyCount` alongside `HasMaxCalls`

## Test plan

- [ ] `make test` passes (all packages, race detector)
- [ ] `make lint` clean
- [ ] Nil-receiver test explicitly asserts `AuditOnlyCount() == 0` for a nil `*LocalManifest`

---
_Generated by [Claude Code](https://claude.ai/code/session_013LzCuFaMq8QX5rYKgG6JRu)_


CLAUDE.md
```diff
@@ -66,3 +66,4 @@ Three layers: `cmd/eunox-mcp/` is the binary (CLI, transport wiring, PDP wiring)
 - Tests: new condition types need table-driven tests under `pkg/capability/` (allow + deny + malformed-input cases); new MCP method coverage needs a test in `internal/transport/enforcement_gaps_test.go`; new audit-record fields need a sign-and-verify round-trip test.
 - Behavior changes need a docs update in the same PR. Manifest grammar changes also update `docs/capability-manifest-guide.md` and the spec repo (`eunolabs/mcp-capability-manifest`). Schema changes update `schemas/` plus a roundtrip test.
 - Security vulnerabilities go through `SECURITY.md`, not public issues.
+- Never reference issue numbers, bug IDs, or internal milestones (stages, phases, epics, sprints) in code comments, documentation, or site copy. Those references rot and belong in the PR description or commit message only.
```
eunolabs, line 69: This new convention is unrelated to the nil-safety fix this PR is titled for, and it's the rule that PR #954 actually enforces (that PR strips the `#NNN` refs out of comments). Bundling a repo-wide guideline into a one-line bugfix couples two independent changes and makes the history harder to read — consider moving it to #954 (or a standalone docs commit) so the rule lands alongside the cleanup that satisfies it.

Separately, as written the rule is immediately violated by a large body of existing comments (`#318`, `#506`, `#668`, `#708`, `#810`, `#887`, …), and #954 only cleans up a handful of them. Either the cleanup needs to be comprehensive, or the rule should be phrased as forward-looking ("new comments must not …") so it isn't documenting a convention the tree broadly breaks on day one.

---

eunolabs (commented):
### Review — nil-safe `AuditOnlyCount`

The core fix is correct and exactly the right shape: `AuditOnlyCount()` now carries the same `if m == nil { return 0 }` guard as its sibling `HasMaxCalls()`, so the two `*LocalManifest` policy-property methods present a consistent nil-safe contract, and the extended nil-receiver test locks it in.

I checked the three call sites and they're all already safe, so this is pure defense-in-depth with no behavior change:
- `internal/transport/route.go:178` — inside a `case manifest != nil:` switch arm
- `cmd/eunox-mcp/main.go:947` — inside `if manifest != nil`
- `cmd/eunox-mcp/doctor.go:371` — inside an `else` branch where `merged` is already dereferenced (`merged.ServerVersion`, `merged.Digest()` above it)

So nothing here regresses; the method just stops being the odd one out. 👍

One scope note on the `CLAUDE.md` hunk (inline) — the added convention is unrelated to this fix and is the rule #954 enforces; worth landing it with that PR instead. Not blocking.

---

PR #954, opened 2026-06-20
fix: preserve envelope fields in */list filtering and forward upstream instructions

eunolabs:
## Summary

Two proxy-transparency fixes: pagination fields (`nextCursor`, `_meta`) and unmodeled entry fields (`annotations`, `outputSchema`, `size`) were silently dropped when the proxy filtered `*/list` responses; and the upstream's `instructions` field from the `initialize` handshake was never captured or forwarded to the host.

### `*/list` field preservation (`internal/pdp/pdp.go`, `internal/pdp/jwt.go`)

The old `filterListResult[E, L any]` generic round-tripped through typed structs (`ToolsListResult`, `ResourcesListResult`, `PromptsListResult`) that only modeled the list array. Any sibling field (`nextCursor`, `_meta`) and any entry field not in the struct (`annotations`, `outputSchema`, `title`) was silently dropped on re-marshal.

Replaced with `filterListResult(resultBytes, fieldName, keep)` that works through `map[string]json.RawMessage`: all envelope fields survive untouched; only the target array is decoded entry-by-entry (using a minimal anonymous struct for name/URI matching) and each kept entry's original raw JSON bytes are preserved in the output.

Updated all six callers in `pdp.go` and `jwt.go`; removed the now-unused `internal/mcp` import from both files.

New regression tests in `internal/transport/enforcement_gaps_filter_test.go`:
- `TestFilterListResult_NextCursorPreserved`
- `TestFilterListResult_MetaPreservedInPrompts`
- `TestFilterListResult_ExtraEntryFieldsPreserved`

### Upstream `instructions` pass-through (`internal/mcp/types.go`, `internal/transport/`)

`mcp.InitResult` had no `Instructions` field, so the upstream's `instructions` string (model guidance) was silently dropped during the initialize handshake.

- Added `Instructions string \`json:"instructions,omitempty"\`` to `mcp.InitResult`
- Updated `applyInitializeResult` to return a 4th value `instructions string`
- Added `upstreamInstructions` to `StdioProxy` and `httpSession`
- All three handshake paths (stdio, HTTP subprocess, HTTP remote) capture and store the value
- `handleInitialize` (stdio) and `buildInitResponse` (HTTP) include it in the response sent to the host

New test `TestApplyInitializeResult_PreservesInstructions`; three existing tests updated for the new 4-return-value signature.

## Test plan

- [ ] `make test` passes (all packages, race detector)
- [ ] `make lint` clean
- [ ] `TestFilterListResult_NextCursorPreserved` — `nextCursor` survives tools/list filter
- [ ] `TestFilterListResult_MetaPreservedInPrompts` — `_meta` survives prompts/list filter
- [ ] `TestFilterListResult_ExtraEntryFieldsPreserved` — `annotations`/`outputSchema` survive on kept entries
- [ ] `TestApplyInitializeResult_PreservesInstructions` — instructions round-trip through initialize

---
_Generated by [Claude Code](https://claude.ai/code/session_013LzCuFaMq8QX5rYKgG6JRu)_


internal/mcp/types.go
```diff
@@ -15,6 +15,7 @@ type InitResult struct {
 	ProtocolVersion string                 `json:"protocolVersion"`
 	Capabilities    map[string]interface{} `json:"capabilities"`
 	ServerInfo      map[string]interface{} `json:"serverInfo"`
+	Instructions    string                 `json:"instructions,omitempty"`
```
eunolabs, line 18: Capturing and relaying `instructions` is reasonable, but two follow-ups worth recording:

1. **No integrity/drift coverage.** Tool descriptions get pinned via `descriptionHash` (the FM drift checks), but `instructions` is upstream-controlled model guidance — a plausible prompt-injection surface — now forwarded to the host verbatim with no manifest constraint and no way for the operator to detect it changing between sessions. That's a defensible first cut (it mirrors how an unpinned description flows through), but it's a new trust conduit and deserves at least a mention in `docs/threat-model-mcp.md`.

2. **Docs.** Per CLAUDE.md ("behavior changes need a docs update in the same PR"), this user-visible change — the proxy now surfaces upstream `instructions` to the host — should land with a one-line note in the threat model and/or capability guide. The diff is code+tests only.

Not blocking the mechanics, which look correct.

---

eunolabs (commented):
### Review — `*/list` field preservation + `instructions` pass-through

Both fixes are real and the approach is sound.

**`filterListResult` rewrite.** Moving from the typed-struct round-trip to `map[string]json.RawMessage` is the right way to stop silently dropping envelope siblings (`nextCursor`, `_meta`) and unmodeled entry fields (`annotations`, `outputSchema`, `title`). The fail-closed posture is preserved end-to-end: empty body short-circuits, every error path returns `{"<field>":[]}` (never the unfiltered bytes), and a fully-filtered list still marshals to `[]` rather than `null` (the #546 invariant). I also confirmed the cleanup is clean: the `internal/mcp` import is now genuinely unused in both `pdp.go` and `jwt.go` (the only remaining `mcp.` tokens are in comments), and the typed list/entry structs stay in use as test fixtures, so nothing is orphaned. CI is green (Test + Lint).

A few things worth addressing or recording:

1. **Subtle — and good — semantic change in the per-entry path; please confirm it's intentional.** The old generic decoded the whole array into `[]ToolEntry`, so a single malformed entry failed the entire `Unmarshal` and fail-closed the *whole* list to `[]`. The new code decodes each entry as `json.RawMessage`, so only the per-entry `keep` unmarshal can fail — one bad entry is now dropped individually while valid siblings survive. This is still fail-closed per entry (parse failure → `keep` returns false → entry dropped, no policy bypass), and it's arguably better for availability, but it's a behavior change from "one bad apple drops the list" to "drop only the bad apple." Worth a line in the function comment if intended.

2. **`instructions` has no integrity/drift coverage and no threat-model note** — see inline on `internal/mcp/types.go`. The mechanics are correct; the gap is that a new upstream-controlled, host-visible field is being relayed without a `docs/threat-model-mcp.md` update (CLAUDE.md requires docs for behavior changes).

3. **Nit — output field ordering.** `json.Marshal(map[string]json.RawMessage)` emits keys in sorted order, so the upstream result object's original *top-level* field ordering isn't preserved (each kept entry's internal bytes are, since they're carried raw). Semantically irrelevant for JSON, but since the PR is framed around fidelity it's worth being aware of.

4. **Nit — PR description path.** The description says the regression tests live in `internal/transport/enforcement_gaps_filter_test.go`, but the diff adds them to `internal/pdp/enforcement_gaps_filter_test.go` — which is the correct home (that's where `filterListResult` lives). Just fix the description.

Tests look good and the instructions round-trip / envelope-preservation cases cover the regressions well.

---

PR #964, opened 2026-06-20
fix: resolve open automated-review bugs (sessions rev-20260620T0310Z–T0610Z)

### LINKED ISSUE #940: [rev-20260620T0310Z-b6f730] */list filtering drops nextCursor and other unmodeled fields, breaking MCP pagination
**Location:** `internal/pdp/pdp.go:847-884` (`filterListResult`), with the under-modeled structs in `internal/mcp/types.go:38-79`

**Reference:** https://github.com/eunolabs/eunox/blob/f04905b54943fb70d1ef6b0723665dd8e2da9e09/internal/pdp/pdp.go#L847-L884
and https://github.com/eunolabs/eunox/blob/f04905b54943fb70d1ef6b0723665dd8e2da9e09/internal/mcp/types.go#L38-L79

**Severity:** High

**Description:**
Every `*/list` response is filtered by `filterListResult[E,L]`, which unmarshals the upstream result into a typed envelope (`mcp.ToolsListResult`, `mcp.ResourcesListResult`, `mcp.PromptsListResult`), filters the slice, and **re-marshals the typed envelope**. Those envelope structs only declare the list slice:

```go
type ToolsListResult struct { Tools []ToolEntry `json:"tools"` }
type ResourcesListResult struct { Resources []ResourceEntry `json:"resources"` }
type PromptsListResult struct { Prompts []PromptEntry `json:"prompts"` }
```

The MCP spec defines a paginated result shape: each of `tools/list`, `resources/list`, `prompts/list` may return a top-level **`nextCursor`** (and an optional `_meta`). Because those fields are not modeled, the re-marshal **silently drops `nextCursor`**. The host therefore never learns there is another page and never sends the follow-up `cursor` request — so every tool/resource/prompt beyond the first upstream page becomes invisible and effectively unusable, even though the policy would allow it. This is a fail-*open*-to-data-loss correctness bug, not just cosmetic.

The same struct round-trip also drops per-entry fields that are not modeled:
- `ToolEntry` lacks `annotations` (incl. `readOnlyHint`/`destructiveHint`), `title`, `outputSchema`, `_meta`.
- `ResourceEntry` lacks `annotations`, `size`, `_meta`.
- `PromptEntry` lacks `title`, `_meta`.

Dropping tool `annotations` is also a safety-relevant loss (the host loses destructive/read-only hints).

Notably this is the **exact bug class that was already fixed on the redaction path**: `ApplyRedactObligs` (`internal/pdp/pdp.go:1075-1087`) decodes into a generic `map[string]interface{}` specifically because "the previous `mcp.ToolCallResult` round-trip silently dropped structuredContent, non-text content payloads, annotations, and _meta." The list-filter path never got the same treatment. (Issue #902 even praises `filterListResult` without noticing this.)

**Trigger / repro:**
Point the proxy at any upstream whose `tools/list` (or `resources/list`/`prompts/list`) returns more entries than its page size, so the first response carries a `nextCursor`. Observe that the host-facing response has no `nextCursor`; the host stops after page one and never sees the remaining tools. A minimal repro: have the mock upstream return `{"tools":[...],"nextCursor":"abc"}` and assert the proxied response retains `nextCursor` (it won't).

**Suggested fix:**
Filter the list **in place on a generic map** rather than round-tripping through a partial struct — mirror the approach already used in `ApplyRedactObligs`. Decode into `map[string]interface{}` (with `UseNumber`), filter only the `tools`/`resources`/`prompts` array (decoding just enough of each entry to evaluate the keep predicate), then re-marshal the whole map so `nextCursor`, `_meta`, and entry-level fields survive. Keep the fail-closed-to-empty behavior on parse/marshal error. Alternatively, add `NextCursor string` + `Meta json.RawMessage` to each envelope struct and the missing fields (`annotations`, `title`, `outputSchema`, `size`) to the entry structs — but the generic-map approach is more robust against future spec additions and matches the already-fixed redaction path. Add a roundtrip test asserting `nextCursor`/`_meta`/`annotations` are preserved through filtering.

**Session:** rev-20260620T0310Z-b6f730

### LINKED ISSUE #941: [rev-20260620T0310Z-b6f730] initialize response reconstruction drops the upstream `instructions` field
**Location:** `internal/transport/http_session.go:116-136` (`buildInitResponse`) and `internal/transport/stdio.go:528-544` (`handleInitialize`); root cause in `internal/mcp/types.go:13-18` (`InitResult`) and `internal/transport/jsonrpc.go:136-152` (`applyInitializeResult`)

**Reference:** https://github.com/eunolabs/eunox/blob/f04905b54943fb70d1ef6b0723665dd8e2da9e09/internal/transport/http_session.go#L116-L136
and https://github.com/eunolabs/eunox/blob/f04905b54943fb70d1ef6b0723665dd8e2da9e09/internal/mcp/types.go#L13-L18

**Severity:** Medium

**Description:**
Both transports answer the host's `initialize` themselves by reconstructing the response from `mcp.InitResult`:

```go
type InitResult struct {
    ProtocolVersion string                 `json:"protocolVersion"`
    Capabilities    map[string]interface{} `json:"capabilities"`
    ServerInfo      map[string]interface{} `json:"serverInfo"`
}
```

`serverInfo` is *deliberately* replaced with the proxy's own name/version, and `capabilities` are forwarded from the upstream (captured by `applyInitializeResult`). But the MCP `initialize` result also has an optional top-level **`instructions`** field — server-authored guidance the host injects into the model's context to describe how to use the server. `InitResult` does not model `instructions`, and `applyInitializeResult` only extracts `capabilities` + `serverInfo.version`, so the upstream's `instructions` are **silently discarded** and never reach the host's model.

Unlike `serverInfo` (intentionally rewritten), there is no reason to drop `instructions`; it's lost purely because the struct doesn't model it. This degrades the behavior of any upstream that ships usage instructions (the model loses guidance it would otherwise get), and it's the same "round-trip through a partial struct drops fields" class already fixed on the redaction path (`ApplyRedactObligs`, `pdp.go:1075`).

**Trigger / repro:**
Configure an upstream MCP server whose `initialize` result includes `"instructions": "..."`. Connect a host through the proxy and inspect the proxied `initialize` response: the `instructions` field is absent.

**Suggested fix:**
Capture `instructions` from the upstream initialize result (extend `applyInitializeResult` to return it, or capture the raw result) and include it in the reconstructed `InitResult`. Add `Instructions string` with `json:"instructions,omitempty"` to `mcp.InitResult` and thread it through `buildInitResponse` / `handleInitialize` (store it on the session/proxy alongside `upstreamCaps`). Add a test asserting upstream `instructions` survive to the host.

**Session:** rev-20260620T0310Z-b6f730

### LINKED ISSUE #942: [rev-20260620T0410Z-744dc5] (*LocalManifest).AuditOnlyCount panics on a nil receiver, breaking the nil-safety contract its sibling HasMaxCalls upholds
**Location:** `internal/config/manifest_policy.go:40-48` (`AuditOnlyCount`); contrast with `HasMaxCalls` at `internal/config/manifest_policy.go:19-36`

**Reference:** https://github.com/eunolabs/eunox/blob/f04905b54943fb70d1ef6b0723665dd8e2da9e09/internal/config/manifest_policy.go#L19-L48

**Severity:** Low (latent — not reached by current call sites, but a real robustness/consistency defect)

**Description:**
The three `(*LocalManifest)` policy-property methods in `manifest_policy.go` do not agree on nil-receiver handling:

- `HasMaxCalls()` explicitly guards nil (`if m == nil { return false }`) and the package even has a regression test asserting it (`manifest_policy_test.go`: `nilManifest.HasMaxCalls()` must return false).
- `Digest()` is incidentally nil-safe — `json.Marshal((*LocalManifest)(nil))` yields `null` without panicking (though it then stamps a digest of the literal `null`, arguably its own minor wart).
- `AuditOnlyCount()` is **not** nil-safe: it goes straight to `for i := range m.Capabilities`, which dereferences the nil pointer and panics.

```go
func (m *LocalManifest) AuditOnlyCount() int {
	n := 0
	for i := range m.Capabilities { // panics if m == nil
		if m.Capabilities[i].IsAuditOnly() {
			n++
		}
	}
	return n
}
```

The existence of an explicit nil guard plus a dedicated nil test on the sibling method establishes an implicit contract that these `*LocalManifest` predicates tolerate a nil receiver (a nil manifest = "no policy"). `AuditOnlyCount` silently breaks that contract.

This is reinforced by the call-site asymmetry in `cmd/eunox-mcp/main.go`: `manifest.HasMaxCalls()` is called **unconditionally** at `main.go:957` (relying on the method's own nil guard), whereas every `AuditOnlyCount()` call site (`main.go:938`, `internal/transport/route.go:178`) is wrapped in `if manifest != nil`. So today no panic occurs — but the safety depends on every caller remembering to guard, exactly the footgun `HasMaxCalls`'s internal guard was added to remove.

**Trigger / repro:**
Call `AuditOnlyCount()` on a nil `*LocalManifest` (the value `LoadUpstreamPDP` returns for a policyless upstream):

```go
var m *config.LocalManifest // nil, e.g. the no-policy return from LoadUpstreamPDP
_ = m.HasMaxCalls()    // ok → false
_ = m.AuditOnlyCount() // panics: runtime error: invalid memory address or nil pointer dereference
```

A future caller that mirrors the unconditional `manifest.HasMaxCalls()` pattern (dropping the `if manifest != nil` wrapper, reasonably assuming parity with the sibling) turns this latent defect into a live crash.

**Suggested fix:**
Add the same nil guard to `AuditOnlyCount` (and, for full parity, decide whether `Digest` on a nil manifest should error rather than return the digest of `null`):

```go
func (m *LocalManifest) AuditOnlyCount() int {
	if m == nil {
		return 0
	}
	n := 0
	for i := range m.Capabilities {
		if m.Capabilities[i].IsAuditOnly() {
			n++
		}
	}
	return n
}
```

Extend `manifest_policy_test.go` with a `nilManifest.AuditOnlyCount()` case mirroring the existing `nilManifest.HasMaxCalls()` assertion so the three methods' nil-receiver behavior stays in lock-step.

**Session:** rev-20260620T0410Z-744dc5

### LINKED ISSUE #943: [rev-20260620T0510Z-fe8d36] validateRedactFields misses empty path segments (e.g. "users.", ".ssn", "a..b", "") — silent no-op redaction, same fail-open class as #326
**Location:** `internal/config/manifest.go:592-599` (`validateRedactFields`)
**Reference:** https://github.com/eunolabs/eunox/blob/344f9b0da0a080ee6f62942a8b4bd61a02e5d873/internal/config/manifest.go#L592-L599
**Runtime redactor:** `internal/pdp/pdp.go:1360-1368` (`redactDotPathRec`) — https://github.com/eunolabs/eunox/blob/344f9b0da0a080ee6f62942a8b4bd61a02e5d873/internal/pdp/pdp.go#L1360-L1368
**Severity:** medium

**Description:**
`validateRedactFields` only rejects redactFields paths that contain array-index notation (`[` / `]`), the fail-open class fixed in #326. It does **not** reject paths with empty `.`-delimited segments or an empty field name. The runtime redactor splits a path on `.` and looks up each segment as a literal object key, so an empty segment never resolves and the redaction silently no-ops — exactly the bug #326 describes ("the path silently no-ops and the field is forwarded unredacted, a declared security control with no runtime effect"). Worse, the audit path still records the `redactFields` obligation as discharged (`enforcedForwardCore` records `oblNames` regardless of whether anything matched, and `ApplyRedactObligs` returns the unchanged result with no error), so the tape falsely reports a secret was redacted.

Concretely, with `redactDotPathRec`:
- `"users."` → `SplitN(".",2)=["users",""]` → descends into `users`, then `delete(map, "")` → no-op.
- `".ssn"` → `["","ssn"]` → looks up key `""` (absent) → no-op.
- `"a..b"` → middle empty segment → no-op.
- `""` (empty field) → `delete(obj, "")` → no-op.

All four pass `validateRedactFields` today (none contain `[`/`]`), so a typo'd path ships a redaction directive that silently redacts nothing while the operator and audit log believe a secret is being scrubbed. This is the same fail-open security-control bypass as #326, just a different malformed-path shape.

**Trigger / repro:**
Author a tool constraint with `directives: [{type: redactFields, fields: ["user.", ".ssn"]}]` (or an empty `""` entry). Manifest loads clean. A `tools/call` whose response contains `user`/`ssn` is forwarded **unredacted**, and the allow record lists the `redactFields` obligation as applied.

**Suggested fix:**
In `validateRedactFields`, after the existing `[`/`]` check, reject any field whose `.`-split produces an empty segment (and reject a wholly empty/whitespace field name). For example, split on `.` (after stripping an optional leading `$`/`$.` to match `RedactDotPath`) and error if any segment is empty, mirroring the array-index rejection message and pointing the author at a well-formed dot path. Add a load-time test alongside `TestManifestDirective_RejectsRedactFieldsArrayIndexPath`.

**Session:** rev-20260620T0510Z-fe8d36

### LINKED ISSUE #944: [rev-20260620T0510Z-fe8d36] validateAllowedExtensions accepts an all-blank extensions list (e.g. [""]) that silently denies every call — inconsistent with sibling validators
**Location:** `internal/config/manifest.go:544-552` (`validateAllowedExtensions`)
**Reference:** https://github.com/eunolabs/eunox/blob/344f9b0da0a080ee6f62942a8b4bd61a02e5d873/internal/config/manifest.go#L544-L552
**Runtime handler:** `pkg/enforcement/handlers.go:751-766` (`handleAllowedExtensions` normalization) — https://github.com/eunolabs/eunox/blob/344f9b0da0a080ee6f62942a8b4bd61a02e5d873/pkg/enforcement/handlers.go#L751-L766
**Severity:** low

**Description:**
`validateAllowedExtensions` rejects an empty `extensions` list (`len == 0`) but does **not** reject a list whose entries are all blank/whitespace, e.g. `extensions: [""]` or `extensions: ["  "]`. At runtime, `handleAllowedExtensions` skips blank entries when building the allowlist (`n := strings.ToLower(strings.TrimSpace(ext)); if n == "" { continue }`), so an all-blank list collapses to an **empty allowlist that denies every call**.

This is exactly the "silently-wrong policy → learn at validate-time, not on the first denied live call" class this file is built around, and it is handled for the sibling conditions but missed here:
- `validateRecipientDomain` rejects blank/whitespace `domains` entries (manifest.go:719-723).
- `validateAllowedOperations` rejects a `*` entry (manifest.go:531-535).
- `validateAllowedTables` rejects empty column allowlists (manifest.go:509-511).

`allowedExtensions` is the odd one out: a `extension:`/`extensions:` typo or a stray blank entry ships a policy that blocks 100% of calls with a `CONDITION_FAILED` that reads like a real enforcement event. (The same gap exists for `validateAllowedTables`' `tables` list, which checks `len == 0` but not blank entries — a blank-only `tables` list also denies every call once `parseTableArgument` rejects blank table names.)

The direction is fail-closed (deny-all), so this is a usability/consistency defect rather than a security hole — but the whole point of these load-time guards is to catch it before runtime.

**Trigger / repro:**
Load a manifest with a tool constraint carrying `conditions: [{type: allowedExtensions, argument: path, extensions: [""]}]`. It validates clean. Every matching `tools/call` is then denied at runtime.

**Suggested fix:**
In `validateAllowedExtensions`, after the empty-list check, reject the condition when every entry is blank after `strings.TrimSpace` (or reject any individual blank entry), with a message mirroring `validateRecipientDomain`'s blank-entry guard. Apply the same per-entry blank check to the `tables` list in `validateAllowedTables`.

**Session:** rev-20260620T0510Z-fe8d36

### LINKED ISSUE #948: [rev-20260620T0517Z-e84f90] Numeric manifest literals above 2^53 authorize the wrong value
**Location:** `pkg/capability/conditions.go:205-207`
**Reference:** https://github.com/eunolabs/eunox/blob/d5f8055ef400dad3c630df14e9ce5c7f9a6a6634/pkg/capability/conditions.go#L205-L207
**Severity:** high
**Description:** Numeric policy literals decoded into interface fields use plain `json.Unmarshal`, so integers above 2^53 are rounded to `float64`. For example, an `allowedValues` entry authored as `9007199254740993` becomes `9007199254740992`. Incoming request arguments are correctly preserved as `json.Number`, so the policy denies the value the operator wrote and authorizes the adjacent value instead. `ArgumentSchema.Enum` has the same partial-struct decode problem through `Constraint.UnmarshalJSON` at `pkg/capability/constraint.go:327-342`.
**Trigger / repro:** Load a YAML manifest with `allowedValues.values: [9007199254740993]`, then evaluate a request containing JSON number `9007199254740992`. The lower, distinct integer is allowed. A focused regression probe reproduced this through `LoadManifest` and `MatchAllowedValue`.
**Suggested fix:** Decode polymorphic condition bodies and constraint/argument-schema bodies with a `json.Decoder` using `UseNumber`, rather than plain `json.Unmarshal`, so numeric literals in `allowedValues` and `enum` remain `json.Number`. The existing exact `numericEqual` handling can then compare the preserved values. Add manifest round-trip tests for adjacent integers above 2^53.
**Session:** rev-20260620T0517Z-e84f90

### LINKED ISSUE #949: [rev-20260620T0517Z-e84f90] Out-of-range JSON numbers bypass argumentSchema bounds
**Location:** `pkg/enforcement/schema.go:139-173`
**Reference:** https://github.com/eunolabs/eunox/blob/d5f8055ef400dad3c630df14e9ce5c7f9a6a6634/pkg/enforcement/schema.go#L139-L173
**Severity:** high
**Description:** The schema validator converts `json.Number` to `float64` before applying `minimum` and `maximum`, but silently leaves the original value untouched when conversion fails. A valid JSON number outside the float64 range, such as `1e400`, makes `toFloat64` return false. If the schema uses a numeric bound without an explicit `type` keyword, the remaining `json.Number` matches no switch case and validation returns nil. JSON Schema numeric bounds apply independently of `type`, so this is a fail-open policy bypass.
**Trigger / repro:** Validate `{"amount": 1e400}` against an argument schema whose `amount` property has `maximum: 100` and no explicit type. The request is allowed. Request decoding uses `UseNumber`, and the original params bytes can then be forwarded upstream unchanged.
**Suggested fix:** Detect numeric inputs explicitly. If a numeric value cannot be represented by the current comparison implementation, fail closed instead of falling through; preferably compare JSON numbers and bounds with an arbitrary-precision representation. Add cases for `1e400`, `-1e400`, and typeless minimum/maximum schemas.
**Session:** rev-20260620T0517Z-e84f90

### LINKED ISSUE #950: [rev-20260620T0517Z-e84f90] Stdio mock ignores its required tool arguments
**Location:** `demo/mock-mcp-server-stdio/main.go:122-148`
**Reference:** https://github.com/eunolabs/eunox/blob/d5f8055ef400dad3c630df14e9ce5c7f9a6a6634/demo/mock-mcp-server-stdio/main.go#L122-L148
**Severity:** low
**Description:** The mock advertises `path`, `content`, and `query` as required string arguments in its tool input schemas, but `dispatchTool` ignores failed type assertions and substitutes empty strings. All three tools therefore return successful `isError:false` results when required arguments are missing or have the wrong type. This makes the mock violate its own advertised contract and lets demo/integration paths report success for malformed calls.
**Trigger / repro:** Send `tools/call` for `read_file`, `write_file`, or `query_db` with `arguments:{}`. The server returns fake contents, a zero-byte write, or an executed empty query instead of rejecting invalid params. A focused probe reproduced all three successes.
**Suggested fix:** Validate each tool argument before dispatch. Require the advertised fields and string types, returning JSON-RPC `-32602` or a tool result with `isError:true` for invalid input. Add table-driven tests for missing, null, and wrong-type arguments.
**Session:** rev-20260620T0517Z-e84f90

### LINKED ISSUE #951: [rev-20260620T0517Z-e84f90] Stdio mock responds to initialize notifications
**Location:** `demo/mock-mcp-server-stdio/main.go:154-166`
**Reference:** https://github.com/eunolabs/eunox/blob/d5f8055ef400dad3c630df14e9ce5c7f9a6a6634/demo/mock-mcp-server-stdio/main.go#L154-L166
**Severity:** low
**Description:** The `initialize` method branch runs before the notification guard. An `initialize` message without an `id` is therefore treated as a request and produces a JSON-RPC response with no id. JSON-RPC servers must never respond to notifications, regardless of the method name. The neighboring notification test covers `notifications/initialized` but misses this ordering case.
**Trigger / repro:** Write `{"jsonrpc":"2.0","method":"initialize","params":{}}` to the mock server stdin. It emits an initialize result instead of emitting nothing. A focused probe reproduced the id-less response.
**Suggested fix:** Check `msg.isNotification()` before dispatching any request method, or require `msg.isRequest()` on every branch that emits a response. Add an initialize-notification regression test.
**Session:** rev-20260620T0517Z-e84f90

### LINKED ISSUE #952: [rev-20260620T0517Z-e84f90] mcp-call.sh exits successfully on HTTP errors
**Location:** `demo/scripts/mcp-call.sh:55-66`
**Reference:** https://github.com/eunolabs/eunox/blob/d5f8055ef400dad3c630df14e9ce5c7f9a6a6634/demo/scripts/mcp-call.sh#L55-L66
**Severity:** low
**Description:** The initialize request checks its HTTP status, but the subsequent tool call uses `curl -s` without `--fail` or status capture. Curl exits zero for HTTP 4xx/5xx responses, so the helper prints the server error body and exits successfully. Automation or users invoking the helper cannot distinguish a completed call from a transport/server failure by exit status. The gateway sibling already captures `%{http_code}` and rejects non-2xx responses.
**Trigger / repro:** Point `MCP_HOST` at a server that initializes successfully and returns HTTP 500 for the tool call. `mcp-call.sh` prints the 500 response body and exits 0. A stubbed curl repro produced a JSON-RPC internal-error body with exit status 0.
**Suggested fix:** Mirror `mcp-call-gateway.sh`: use `-sS -w '\n%{http_code}'`, split the status trailer from the body, and exit nonzero outside 2xx. Alternatively use `--fail-with-body` where supported, while retaining a portable status-capture fallback.
**Session:** rev-20260620T0517Z-e84f90

### LINKED ISSUE #955: [rev-20260620T0527Z-49eb63] redactFields rounds large integers in tool results
**Location:** `internal/pdp/pdp.go:1043-1099`
**Reference:** https://github.com/eunolabs/eunox/blob/d5f8055ef400dad3c630df14e9ce5c7f9a6a6634/internal/pdp/pdp.go#L1043-L1099
**Severity:** medium
**Description:** `ApplyRedactObligs` decodes the entire tool result into `map[string]interface{}` with `json.Unmarshal`. Every JSON number therefore becomes `float64`; integers above 2^53 are rounded before the result is re-serialized. A redaction directive targeting an unrelated field can silently change IDs, counters, monetary values, or other structured output. A focused probe changed `9007199254740993` to `9007199254740992` in both text JSON and `structuredContent`.
**Trigger / repro:** Return `{"content":[{"type":"text","text":"{\"secret\":\"x\",\"id\":9007199254740993}"}],"structuredContent":{"secret":"x","id":9007199254740993}}` under any active `redactFields` obligation. The forwarded response contains `9007199254740992`.
**Suggested fix:** Decode the result envelope and JSON text bodies with `json.Decoder.UseNumber`, then preserve `json.Number` values through the redaction and re-marshal path. Add regression coverage for integers above 2^53 in text content and `structuredContent`.
**Session:** rev-20260620T0527Z-49eb63

### LINKED ISSUE #956: [rev-20260620T0527Z-49eb63] Malformed text content bypasses redactFields
**Location:** `internal/pdp/pdp.go:1055-1073`
**Reference:** https://github.com/eunolabs/eunox/blob/d5f8055ef400dad3c630df14e9ce5c7f9a6a6634/internal/pdp/pdp.go#L1055-L1073
**Severity:** high
**Description:** The redaction path skips a `type:"text"` content item when its `text` field is not a string. That is an invalid MCP content shape, but forwarding it unchanged violates the documented fail-closed guarantee: an untrusted upstream can put the named secret inside an object-valued `text` field and the proxy returns it without applying the redaction or raising an error. Non-object content items and items with a missing or unexpected `type` are similarly skipped rather than treated as structurally unverifiable.
**Trigger / repro:** Under a `redactFields` obligation for `secret`, return `{"content":[{"type":"text","text":{"secret":"leak"}}],"isError":false}`. `ApplyRedactObligs` returns success and forwards the `secret` value unchanged.
**Suggested fix:** Validate every `content` item sufficiently to prove the redaction surface: require an object with a recognized type and require `text` to be a string for text items. Fail the response closed on malformed or ambiguous items instead of continuing. Add malformed-item regression tests.
**Session:** rev-20260620T0527Z-49eb63

### LINKED ISSUE #957: [rev-20260620T0527Z-49eb63] Redis startup can miss kill events before subscribing
**Location:** `pkg/killswitch/redis.go:197-233`
**Reference:** https://github.com/eunolabs/eunox/blob/d5f8055ef400dad3c630df14e9ce5c7f9a6a6634/pkg/killswitch/redis.go#L197-L233
**Severity:** high
**Description:** `Redis.Start` completes its initial Redis snapshot before it subscribes to the kill-switch pub/sub channel. A kill written and published after the final snapshot read but before the subscription becomes active is absent from the committed cache and its event has no subscriber. Because the successful snapshot clears `lastRefreshErr`, fail-closed mode still reports healthy and permits the killed subject until the periodic reconciliation runs (30 seconds by default).
**Trigger / repro:** Block the final session-key SCAN after it has returned an empty result, SET `killswitch:agent:alice` and PUBLISH `agent:kill:alice` from another client, then release Start. With a long reconcile interval, `ShouldBlock(ctx, "alice", "")` returns false even though the durable Redis kill exists. A focused miniredis probe reproduces this deterministically.
**Suggested fix:** Establish and confirm the pub/sub subscription before taking the initial snapshot, then reconcile after subscription so any event in the handoff window either updates `cacheGen` or is included in the snapshot. Alternatively use an explicit subscribe-ready handshake followed by a second snapshot before marking the backend healthy. Add a deterministic startup-window test.
**Session:** rev-20260620T0527Z-49eb63

### LINKED ISSUE #958: [rev-20260620T0601Z-28e3f6] Doctor leaks URL passwords when parsing fails
**Location:** `cmd/eunox-mcp/doctor.go:299-308`
**Reference:** https://github.com/eunolabs/eunox/blob/344f9b0da0a080ee6f62942a8b4bd61a02e5d873/cmd/eunox-mcp/doctor.go#L299-L308
**Severity:** medium
**Description:** `redactURL` returns the original string when `url.Parse` fails. `doctor` deliberately reads and prints the raw config even when that config is invalid, so a malformed `upstreamUrl` containing userinfo is emitted verbatim in a support bundle labeled "Config (redacted)". This can disclose an embedded username/password when the operator shares the bundle for help diagnosing the malformed config.
**Trigger / repro:** Put `upstreamUrl: "https://alice:super-secret@example.com/%"` in a config and run `eunox-mcp doctor --config <file> --audit-tail 0`. The invalid `%` escape makes `url.Parse` fail, and the generated bundle still contains `super-secret`.
**Suggested fix:** On parse failure, never return a URL-looking value verbatim when its authority may contain userinfo. Either conservatively redact everything before the last `@` in the authority, or replace the whole malformed URL with a redacted placeholder. Add a regression test using invalid percent escapes plus userinfo.
**Session:** rev-20260620T0601Z-28e3f6

### LINKED ISSUE #959: [rev-20260620T0601Z-28e3f6] Init emits invalid YAML for upstream-defined names
**Location:** `cmd/eunox-mcp/init_manifest.go:162-225`
**Reference:** https://github.com/eunolabs/eunox/blob/344f9b0da0a080ee6f62942a8b4bd61a02e5d873/cmd/eunox-mcp/init_manifest.go#L162-L225
**Severity:** medium
**Description:** The `init` scaffold interpolates upstream-controlled tool names, schema property names, required names, and type strings into YAML without quoting them. Prefixing every physical line with a comment prevents immediate active-YAML injection, but the documented workflow is to uncomment selected entries. At that point YAML-significant names can make the manifest unparsable; `#` in a tool name can also truncate the target and make the entry refer to a different tool.
**Trigger / repro:** Introspect a tool named `read: report # prod` whose input schema has a required property named `header: name`. After uncommenting the generated entry, it contains `- target: tool:read: report # prod`, `header: name: { type: string }`, and `required: [header: name]`. `eunox-mcp validate` fails with `mapping values are not allowed in this context`.
**Suggested fix:** Render the scaffold through a YAML encoder or apply YAML-safe quoting to every upstream-derived scalar: the full target, property keys, required entries, and type values. Add a round-trip test that uncomments entries containing colon-space, `#`, quotes, newlines, and YAML keywords.
**Session:** rev-20260620T0601Z-28e3f6

### LINKED ISSUE #960: [rev-20260620T0601Z-28e3f6] Remote HTTP accepts mismatched JSON-RPC response IDs
**Location:** `internal/transport/http_remote.go:280-333`
**Reference:** https://github.com/eunolabs/eunox/blob/344f9b0da0a080ee6f62942a8b4bd61a02e5d873/internal/transport/http_remote.go#L280-L333
**Severity:** medium
**Description:** `DoMCPHTTP` correlates SSE responses with the outbound JSON-RPC ID, but the normal `application/json` path only decodes and returns the body. The remote HTTP gateway path then treats that result as the response to the current call, and `enforcedForwardCore` overwrites its ID with the host's ID. A mismatched or missing upstream response ID is therefore masked and the wrong result is bound to the caller's request.
**Trigger / repro:** Send a request with JSON-RPC `id: 1` through a remote-HTTP route and have the upstream return HTTP 200 with `{"jsonrpc":"2.0","id":999,"result":{"ok":true}}`. `DoMCPHTTP` returns success, and the proxy forwards the result as `id: 1` instead of rejecting the protocol violation. A focused custom-`RoundTripper` probe reproduces this without network I/O.
**Suggested fix:** For a JSON response to a request, require `out.IsResponse()` and exact `mcp.MsgKey(out.ID) == mcp.MsgKey(msg.ID)` before returning it. Share the same correlation check with the SSE and stdio-HTTP paths so all remote transports reject mismatches consistently.
**Session:** rev-20260620T0601Z-28e3f6

### LINKED ISSUE #961: [rev-20260620T0601Z-28e3f6] Null initialize results are accepted as valid handshakes
**Location:** `internal/transport/jsonrpc.go:164-179`
**Reference:** https://github.com/eunolabs/eunox/blob/344f9b0da0a080ee6f62942a8b4bd61a02e5d873/internal/transport/jsonrpc.go#L164-L179
**Severity:** medium
**Description:** `applyInitializeResult` treats any JSON value that unmarshals into `mcp.InitResult` without an error as a successful MCP handshake. In Go, unmarshalling JSON `null` into a struct succeeds and leaves every field at its zero value, so an upstream response with `result: null` is accepted. The proxy proceeds to send `notifications/initialized` and creates a client session with default/empty capabilities even though the upstream never supplied a valid InitializeResult. The one-shot `init`/`validate --live` path has the same permissive behavior in `cmd/eunox-mcp/live_upstream.go:58-79`.
**Trigger / repro:** Call `applyInitializeResult(mcp.RPCMsg{Result: json.RawMessage("null")})`; it returns `(nil, "", nil)`. A remote upstream returning `{"jsonrpc":"2.0","id":1,"result":null}` is therefore treated as initialized.
**Suggested fix:** Require the result to be a JSON object and validate the required MCP InitializeResult fields, at minimum `protocolVersion`, `capabilities`, and `serverInfo`, before accepting the handshake. Reuse that validator in the live-introspection path so proxy and CLI behavior cannot diverge.
**Session:** rev-20260620T0601Z-28e3f6

### LINKED ISSUE #962: [rev-20260620T0610Z-33d3e4] Server-initiated MCP requests (sampling/createMessage) are fanned out to every SSE subscriber on a session
**Location:** `internal/transport/http_session.go:646-701` (`broadcast` / `broadcastServerRequest`) and `internal/transport/http_routing.go:263-265` (`handleMCPGet` → `addSub`)

**Reference:**
- https://github.com/eunolabs/eunox/blob/d42c51ba480b252f34cb193f056a86541eb7978e/internal/transport/http_session.go#L646-L701
- https://github.com/eunolabs/eunox/blob/d42c51ba480b252f34cb193f056a86541eb7978e/internal/transport/http_routing.go#L263-L265

**Severity:** Low

**Description:**

A server-initiated JSON-RPC **request** (most importantly `sampling/createMessage`, but also `roots/list`, `elicitation/create`, …) is delivered through `httpSession.broadcast`, the same primitive used for fire-and-forget **notifications**. `broadcast` iterates *every* subscriber in `s.notifSubs` and sends the message to each:

```go
func (s *httpSession) broadcast(msg mcp.RPCMsg) bool {
	s.notifMu.Lock()
	defer s.notifMu.Unlock()
	delivered := false
	for _, ch := range s.notifSubs {   // sends to ALL subscribers
		select {
		case ch <- msg:
			delivered = true
		default: /* slow-subscriber drop */
		}
	}
	return delivered
}
```

`handleMCPGet` registers a new subscriber for **every** `GET /mcp` SSE stream via `addSub(ch)`, with no per-session cap and no dedup:

```go
ch := make(chan mcp.RPCMsg, 16)
sess.addSub(ch)
defer sess.removeSub(ch)
```

```go
func (s *httpSession) addSub(ch chan mcp.RPCMsg) {
	s.notifMu.Lock()
	s.notifSubs = append(s.notifSubs, ch)  // unbounded, no dedup
	s.notifMu.Unlock()
}
```

Fan-out is correct for notifications, but a server-initiated **request** expects to be delivered to a single client and answered by a single response. When a session has more than one open SSE stream, `broadcastServerRequest` delivers the request to all of them. The response correlation in `handleMCPPost` only routes the **first** matching response back to the upstream (`serverReqs.take` removes the id), so subsequent responses for the same id are silently discarded:

```go
if msg.IsResponse() && sess.serverReqs.take(mcp.MsgKey(msg.ID)) && sess.upWriter != nil {
	_ = sess.upWriter.Write(msg)
}
```

Consequences:
- **Duplicate / unintended disclosure:** a `sampling/createMessage` payload carries the conversation/context the upstream wants the host LLM to complete. It is copied to every open SSE stream on the session rather than to one. If two consumers share a session id (or a single consumer holds multiple streams), the sensitive sampling payload reaches all of them.
- **Wasted work / nondeterminism:** each stream may compute and POST a response with the same JSON-RPC id; only the first is forwarded upstream and the rest are dropped, so which client's completion the upstream receives is effectively a race.

The MCP Streamable HTTP model treats a server→client request as a single-stream, single-response exchange; broadcasting it to all subscribers does not match that. The surrounding code is aware `broadcast` is shared (`broadcast` doc comment), but the multi-subscriber **request** case is not addressed.

**Trigger / repro:**
1. Start the HTTP transport with a local (subprocess) upstream that issues a `sampling/createMessage` server-initiated request.
2. As one client, `initialize` a session, then open **two** concurrent `GET /mcp` SSE streams with the same `Mcp-Session-Id`.
3. Drive the upstream to emit `sampling/createMessage`.
4. Observe the request is written to **both** SSE streams. Have both respond with the same id — only the first reaches the upstream; the second is dropped.

**Suggested fix:**
- Deliver server-initiated **requests** to a single subscriber instead of broadcasting. e.g. add a `deliverToOne(msg) bool` that sends to the first subscriber whose buffer accepts it (returning `false` to keep the existing fail-closed path when none can), and call it from `broadcastServerRequest` while leaving `broadcast` (true fan-out) for notifications.
- Optionally enforce a single (or bounded) SSE stream per session in `handleMCPGet`/`addSub` so the fan-out surface and the per-session goroutine/connection footprint are bounded.

**Session:** rev-20260620T0610Z-33d3e4

eunolabs:
This PR resolves every open issue carrying a review-session id in its title (the
`[rev-20260620T...]` batch, #940–#962). Each bug is one commit referencing its
issue; the five issues already resolved by earlier work on this branch (relative
to `main`) are listed too so merging closes them.

### Fixed in this PR (one commit each)
- Closes #943 — `validateRedactFields` accepted paths with empty `.`-segments (`"users."`, `".ssn"`, `"a..b"`, `""`); the runtime redactor no-ops on them while the audit log reports the obligation discharged → reject empty segments at load (after stripping the optional `$`/`$.` prefix).
- Closes #944 — `validateAllowedExtensions`/`validateAllowedTables` accepted an all-blank list (`[""]`) that collapses to a deny-all allowlist at runtime → reject any blank entry, matching the sibling `recipientDomain` guard.
- Closes #948 — polymorphic condition and constraint/argumentSchema bodies decoded with plain `json.Unmarshal`, rounding numeric literals above 2^53 (an `allowedValues`/`enum` of `9007199254740993` authorized `…992`) → decode with `UseNumber` so `numericEqual`/`asInt64` compare the preserved values exactly.
- Closes #950 — the stdio mock substituted empty strings for missing/wrong-typed required args and returned success → validate advertised required string args before dispatch, reject with `-32602`.
- Closes #951 — the stdio mock answered an id-less `initialize` (a notification) with a response → move the `isNotification` guard ahead of the method branches.
- Closes #952 — `mcp-call.sh` used `curl -s` without status capture, exiting 0 on HTTP 4xx/5xx → capture `%{http_code}`, split it from the body, exit nonzero outside 2xx (mirrors `mcp-call-gateway.sh`).
- Closes #956 — `ApplyRedactObligs` skipped malformed content items (non-object, non-string `text`, missing/unrecognized `type`), forwarding a secret hidden in an object-valued `text` field → require an object with a recognized type and a string `text` body, else fail the response closed.
- Closes #957 — the Redis kill switch took its initial snapshot before subscribing, losing a kill published in the handoff window → subscribe and confirm the subscription before snapshotting (the existing `cacheGen` retry covers the listener/snapshot race).
- Closes #958 — `doctor`'s `redactURL` returned the raw string when `url.Parse` failed, leaking embedded userinfo in a support bundle → conservatively strip everything before the last `@` in the authority on parse failure.
- Closes #959 — the `init` scaffold interpolated upstream-controlled tool/property/required/type names into YAML unquoted, breaking or retargeting entries once uncommented → render every upstream-derived scalar through the YAML encoder.
- Closes #960 — the remote-HTTP gateway path forwarded a JSON upstream response without checking its JSON-RPC id, so `enforcedForwardCore` rebound a mismatched result to the caller → correlate in `callRemoteUpstream` (refuse a mismatched result, preserve a mismatched error's diagnostic), matching the stdio HTTP bridge.
- Closes #961 — `applyInitializeResult` accepted a `result: null` handshake (null unmarshals to all-zero fields) → require `protocolVersion`/`capabilities`/`serverInfo`, and reuse the check in the CLI live-upstream probe via the new exported `ApplyInitializeResult`.
- Closes #962 — server-initiated requests (`sampling/createMessage`, …) were broadcast to every SSE subscriber on a session, disclosing the payload and racing the response → add `deliverToOne` and route server requests through it; `broadcast` stays for true notification fan-out.

### Already resolved on this branch (relative to `main`)
These were fixed by earlier commits on this branch (#953/#954 and siblings) and merging closes them:
- Closes #940 — `*/list` filtering now preserves `nextCursor`/`_meta`/entry fields (generic-map filter).
- Closes #941 — the upstream `initialize` `instructions` field is now threaded through both transports.
- Closes #942 — `(*LocalManifest).AuditOnlyCount` is now nil-safe, matching `HasMaxCalls`.
- Closes #949 — the argument-schema validator now fails closed on an un-representable `json.Number` (e.g. `1e400`) under a typeless `minimum`/`maximum`.
- Closes #955 — `ApplyRedactObligs` now decodes with `UseNumber`, so unrelated integers above 2^53 are not rounded during redaction.

## Testing
- Added focused regression tests for every fix in this PR: load-time validator cases (#943/#944), `MatchAllowedValue`/condition round-trip for large integers (#948), malformed-content fail-closed cases (#956), a deterministic Redis startup-window test that fails without the reorder (#957), `redactURL` malformed-input cases (#958), a hostile-name init round-trip (#959), gateway id-mismatch reject/preserve (#960), null/empty handshake rejection (#961), single-subscriber delivery (#962), and stdio-mock notification/arg-validation cases (#950/#951).
- `go build ./...`, `go test ./...`, and `go vet ./...` all pass; `go test -race` passes on the changed `pkg/killswitch`, `internal/transport`, and `pkg/capability` packages.
- `gofmt` clean; Apache-2.0 headers present on the new files.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QCYEWBD5pHJcyyGxfG7jy8)_


internal/pdp/pdp.go
```diff
@@ -1156,6 +1175,27 @@ func ApplyRedactObligs(resultBytes []byte, obligs []capability.Obligation) ([]by
...
+// structuredContent); the others carry no top-level JSON object body a dot path
+// could name, so they are preserved verbatim. A content item whose type is none
+// of these is structurally unverifiable and fails the response closed rather than
+// being forwarded unredacted.
+var recognizedContentTypes = map[string]struct{}{
+	"text":          {},
+	"image":         {},
+	"audio":         {},
+	"resource":      {},
+	"resource_link": {},
```
eunolabs, line 1197: The fail-closed tightening here is good and correctly scoped (only runs when a `redactFields` obligation is present). One residual case worth confirming: `resource` (and `resource_link`) items are preserved verbatim via the `continue` above, but an embedded resource carries its own object (`resource.text` / `resource.blob`, plus uri/mimeType). If a `redactFields` path could ever name a field that surfaces inside an embedded resource body, that's still a fail-open for those types — unlike `text`, which is now inspected. Since the rest of this change deliberately fails closed on anything "structurally unverifiable," it's worth a sentence in the comment stating *why* `resource`/`resource_link` are considered safe to pass through (i.e. dot paths are only matched against the top-level `text` body / `structuredContent`, never the embedded resource), so a future reader doesn't read the preservation as an oversight.

---

cmd/eunox-mcp/doctor.go
```diff
@@ -293,21 +293,58 @@ func redactString(v interface{}) string {
...
+	}
+	authStart := schemeEnd + len(sep)
+	rest := s[authStart:]
+	authority := rest
+	tail := ""
+	if end := strings.IndexAny(rest, "/?#"); end >= 0 {
+		authority = rest[:end]
+		tail = rest[end:]
+	}
+	at := strings.LastIndex(authority, "@")
```
eunolabs, line 343: Minor edge case (low severity): the authority is cut at the first `/?#`, then the `@` is sought only within that prefix. For a value like `https://user:pa/ss@host/x` the `/` truncates the authority to `user:pa`, no `@` is found, and the string is returned unchanged — leaking the userinfo. In practice this only matters when `url.Parse` *fails* (this fallback's precondition) yet the string still contains an unescaped `/` inside userinfo, which is unusual, so I wouldn't block on it. If you want to close it fully, search for `@` before splitting on `/?#` (an `@` in the authority always precedes the path on a real URL). Worth at most a one-line note acknowledging the conservative-but-imperfect handling.

---

eunolabs (commented):
## Review — automated-review bug batch (#940–#962)

Reviewed all 24 files. This is a clean, high-quality batch: each fix is small, narrowly scoped, fails closed, and lands with a focused regression test. No correctness blockers found. Summary below, plus two non-blocking inline notes.

### CI
The **Lint** job was the only red check (everything else green). Two findings, both introduced by this PR:
- `unparam`: `(*httpSession).broadcast` returned a `bool` that became dead once the #962 fix routed server-initiated requests through the new `deliverToOne`; its doc comment also still claimed `broadcastServerRequest` relied on that report.
- `staticcheck QF1008`: `wc.Client.Close()` in the #957 startup-window test — `windowClient` has no overriding `Close`, so the embedded selector is redundant (note `w.Client.Scan` is correctly *not* flagged because `windowClient` does override `Scan`).

I pushed a fix for both to the PR branch (`chore: fix lint findings`, commit `507d4b3`): dropped the dead return + refreshed the comment, and switched to `wc.Close()`. `go build`, `go vet`, `go test ./pkg/killswitch/... ./internal/transport/...`, and `golangci-lint run` on both packages are clean locally. Lint should now go green.

### Spot-checks that hold up
- **#960 (remote-HTTP id correlation)** — `callRemoteUpstream` now refuses a mismatched *result* and stamps the right id onto a mismatched *error*, matching the stdio bridge. The split-by-shape rationale (don't bind possibly-wrong data; preserve a genuine diagnostic) is sound and the early `err` return keeps the happy path intact.
- **#961 (null handshake)** — centralizing the check in `validateInitializeResultFields` and exporting `ApplyInitializeResult` is the right call. Confirmed `applyInitializeResult` already handles `resp.Error` (returns the upstream code/message) and the empty-result case — so removing the duplicate `initResp.Error` check in `live_upstream.go` actually *improves* error reporting rather than losing it, and `json`/`mcp` remain used (no orphaned imports).
- **#957 (Redis startup window)** — subscribing and confirming via `pubsub.Receive` *before* the snapshot, with a clean fall-back to periodic reconcile if the subscription can't be confirmed, closes the handoff-window race. The deterministic `windowClient.Scan` injection test fails without the reorder — nice.
- **#948/#955 (numeric precision)** — decoding policy literals with `UseNumber` in both `unmarshalCondition` and `Constraint.UnmarshalJSON`, matched against `UseNumber`-decoded args, correctly stops integers > 2^53 collapsing onto a neighbour.
- **#943/#944 (load-time validators)** — empty-segment / blank-entry rejection mirrors the existing `recipientDomain` guard and fails at validate-time instead of silently denying at runtime. The `$`/`$.` prefix strip before segment-checking is a nice touch.
- **#959 (init scaffold)** — routing every upstream-derived scalar through `yamlScalar` (YAML encoder, with a `strconv.Quote` fallback for multi-line) closes the YAML-injection/retarget hole.

### Inline notes (both non-blocking)
1. `internal/pdp/pdp.go` — worth a sentence documenting *why* `resource`/`resource_link` are safe to pass through verbatim, since the rest of the change fails closed on anything unverifiable.
2. `cmd/eunox-mcp/doctor.go` — `redactURLFallback` has a narrow leak path (slash inside userinfo of an otherwise-unparseable URL); very low severity, optional to harden.

Nice work — once Lint is green this looks ready.

---

internal/pdp/pdp.go
eunolabs: Addressed in `0bd5347`. Expanded the `recognizedContentTypes` doc comment to state explicitly that `resource`/`resource_link` are passed through by design — a `redactFields` dot path is matched only against the top-level `text` body and `structuredContent`, never an embedded resource's `text`/`blob` — and noted that if redaction scope is ever widened those types would have to leave the recognized-and-skipped set.

---

cmd/eunox-mcp/doctor.go
eunolabs: Addressed in `0bd5347` — opted to close it rather than just note it. When the authority carries no `@` but one appears past its boundary (the `/`-before-`@` malformed case), `redactURLFallback` now replaces the whole value with the `<redacted unparseable URL>` placeholder instead of returning it verbatim, so the embedded secret can't leak. Added a regression case (`https://alice:super-secret/x@example.com/%`) to `TestRedactURL_MalformedNeverLeaksPassword`.

---

PR #1005, opened 2026-06-20
fix: address code-review findings (kill-switch wiretap gap, nil-check traps, dispatch projection)

eunolabs:
Resolves the maintenance and operator-facing findings from the June 2026
review:

- A1: a global kill on a gateway with policyless wiretap routes (enforcement:
  audit, no policy) does not stop those routes in non-JWT mode. Surface the gap
  loudly without touching enforcement logic: the startup NOTICE now says so, the
  /control/kill `all` response carries `wiretap_routes_unaffected: true` when it
  applies, and `eunox-mcp kill` prints a stderr warning so an operator is not
  misled by an "ok" during an incident.
- A2: filterToolsListResult computes the pinned-exact-tool predicate once with
  an explicit c == nil check, removing the hidden nil-receiver contract and the
  duplicate IsPinnedExactTool call.
- A3: forwardServerRequest's observe gate now mirrors enforcedForwardCore
  exactly: (fp.audit || dec.AuditOnly).
- B1: dispatchParams embeds forwardParams, deleting the field-by-field forward()
  projection.
- C1: AnyRouteHasMaxCalls guards rt.manifest != nil explicitly.
- C2: dispatchList's RecordAllow notes why AuditOnly never applies to */list.

Adds handleKill tests for the wiretap callout and documents the gap in the
client-integration revocation section.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>


internal/transport/route.go
```diff
@@ -304,10 +307,26 @@ func manifestHasSamplingOptIn(m *config.LocalManifest) bool {
...
 }
 
+// anyWiretapRoute reports whether any route runs in policyless wiretap mode
+// (enforcement: audit with no manifest). The global kill switch does not stop
+// these routes unless JWT auth wraps them (the JWTPDP wrapper holds the kill
+// manager; AlwaysAllowPDP does not), so callers surface the partial coverage on
+// kill activation.
+func anyWiretapRoute(routes map[string]*UpstreamRoute) bool {
+	for _, rt := range routes {
+		if rt.manifest == nil {
```
eunolabs: `anyWiretapRoute` keys the kill-coverage signal off `rt.manifest == nil`, but `manifest` is only a reliable wiretap indicator in gateway mode. It is set in exactly one place — `BuildRoutes` (`route.go:161`). The single-upstream constructor `newHTTPProxy` (`http.go:248`) synthesizes its route with `pdp: opts.PDP` but never sets `manifest`, so `manifest == nil` there even when `opts.PDP` is a real `ManifestPDP` (a fully policed, kill-switch-covered route).

What actually determines coverage is whether the route's PDP honors the kill switch: `AlwaysAllowPDP.CheckKill` is a no-op, `ManifestPDP` enforces it. So the precise predicate is the PDP type, not the manifest field:

```go
if _, wiretap := rt.pdp.(pdp.AlwaysAllowPDP); wiretap {
    return true
}
```

This isn't a live production bug — the binary serves every HTTP config (including single-upstream) through `BuildRoutes`/`NewHTTPProxyGateway`, where `manifest` is set correctly. But the signal is coupled to a gateway-only invariant, and `newHTTPProxy` (used pervasively in tests, and a real construction path) breaks it: a policed proxy built that way would emit `wiretap_routes_unaffected: true` and trip the `eunox-mcp kill` stderr warning during an incident — the exact "don't mislead the operator" failure this PR exists to prevent, inverted. Keying off the PDP type closes that gap and makes the predicate correct under both construction paths.

---

eunolabs (commented):
Reviewed the six findings. A2 (single `pinned` predicate with explicit `c == nil`), A3 (observe gate now `fp.audit || dec.AuditOnly`, mirroring `enforcedForwardCore`), B1 (`dispatchParams` embedding `forwardParams` — the field-by-field `forward()` projection is gone and the dispatch handlers pass `d.forwardParams` straight through), C1 (`AnyRouteHasMaxCalls` nil guard), and C2 (the `*/list` AuditOnly rationale) all look correct and are net simplifications. The kill-switch gap analysis itself is accurate: `AlwaysAllowPDP.CheckKill` is a no-op while `ManifestPDP` honors the kill switch, and JWT mode rewraps every route's PDP with the kill manager, so `p.jwtPDP == nil` is the right gate.

One substantive concern on A1, left inline on `anyWiretapRoute`:

- The wiretap predicate is `rt.manifest == nil`, but `manifest` is a gateway-only invariant — it's set exclusively in `BuildRoutes` (`route.go:161`). The single-upstream constructor `newHTTPProxy` (`http.go:248`) builds its route with a real PDP but leaves `manifest` nil, so a policed single-upstream route reads as a wiretap there. What actually determines kill coverage is the PDP type (`AlwaysAllowPDP` ignores kill, `ManifestPDP` enforces it), so keying off `rt.pdp.(pdp.AlwaysAllowPDP)` is the correct and construction-path-independent predicate. Not a live production bug (the binary serves all HTTP configs through `BuildRoutes`), but it inverts this PR's own "don't mislead the operator" goal on the `newHTTPProxy` path.

Related, on the new tests (`http_test.go`): both fixtures set `manifest` directly on a hand-built `UpstreamRoute` and leave `pdp` unset (nil). That exercises the `manifest`-based branch but not the shape either real constructor produces — production routes always carry a non-nil `pdp`, and `newHTTPProxy` carries a non-nil `pdp` with a nil `manifest`. If `anyWiretapRoute` is switched to a PDP-type check, set `pdp: pdp.AlwaysAllowPDP{}` / `pdp: <a ManifestPDP>` in the fixtures so the tests track what the code keys off. As written, the green tests would not catch the `newHTTPProxy` false-positive.

Everything else LGTM.

---

PR #1010, opened 2026-06-20
Fix automated-review bugs (sessions rev-20260620T1505Z–T1910Z)

### LINKED ISSUE #996: [rev-20260620T1505Z-f540ba] initRemoteUpstream leaks upstream session when applyInitializeResult fails
**Location:** `internal/transport/http_remote.go`, function `initRemoteUpstream`, lines 246–253

**Reference:** commit `6824022c7b40a0282efbf9ffe0419c6215ea7933`

**Severity:** Medium

**Description:**
`initRemoteUpstream` captures `s.upstreamSessID` from the `initialize` response headers _after_ calling `applyInitializeResult`. When `applyInitializeResult` returns an error (malformed result, JSON-RPC error response, nil `result` field), the function returns early without ever setting `s.upstreamSessID`. The upstream server has already allocated a session by this point and sent its session ID in the `Mcp-Session-Id` response header. Because `close()` guards the DELETE call with `if s.upstreamSessID != ""`, the cleanup is skipped entirely and the upstream session is permanently leaked.

```go
// Current (buggy) ordering:
respMsg, respHdr, err := s.doRemoteHTTP(ctx, initReq, "")
if err != nil {
    return fmt.Errorf("sending initialize: %w", err)
}

caps, sv, instructions, err := applyInitializeResult(respMsg)
if err != nil {
    return err  // s.upstreamSessID never set; close() skips DELETE
}
s.upstreamSessID = respHdr.Get(SessionHeader)  // too late
```

The `close()` guard that relies on this field:
```go
if s.upstreamSessID != "" && s.route != nil {
    DeleteMCPHTTPSession(...)
}
```

**Trigger / repro:**
Any condition that causes `applyInitializeResult` to return an error after the upstream HTTP call succeeds. Concrete cases:

1. Upstream returns HTTP 200 but the body is a JSON-RPC error object — `applyInitializeResult` detects the error code and returns `err`.
2. Upstream returns HTTP 200 with a valid JSON-RPC response but `result` is `null` — nil-result path returns `err`.
3. Upstream returns HTTP 200 but the `result` field is not a valid `InitializeResult` JSON — unmarshal error returns `err`.
4. Any transient server misbehaviour during the `initialize` handshake that the upstream has already booked a session for.

In a gateway deployment with many short-lived or misbehaving upstream connections this causes unbounded session accumulation on the upstream server.

**Suggested fix:**
Move the session ID capture to immediately after `doRemoteHTTP` succeeds, before the result is validated, so `close()` can always send the DELETE even when result parsing fails:

```go
respMsg, respHdr, err := s.doRemoteHTTP(ctx, initReq, "")
if err != nil {
    return fmt.Errorf("sending initialize: %w", err)
}
// Capture session ID before validating the result so close() can DELETE it on any error path.
s.upstreamSessID = respHdr.Get(SessionHeader)

caps, sv, instructions, err := applyInitializeResult(respMsg)
if err != nil {
    return err  // close() will now correctly DELETE the upstream session
}
s.upstreamCaps, s.upstreamServerVersion, s.upstreamInstructions = caps, sv, instructions
```

**Session:** rev-20260620T1505Z-f540ba

### LINKED ISSUE #1000: [rev-20260620T1545Z-d3f21e] expandEnvRefs silently expands $identifier inside "pa$$word"-style values when the identifier is a set env var
## Location

`internal/config/gateway_config.go` — `expandEnvRefs` (line 66–82), comment at line 70; `LoadGatewayConfig` docstring lines 243–245; `docs/client-integration.md` line 157.

## Reference

Commit `6824022c7b40a0282efbf9ffe0419c6215ea7933`

## Severity

**Low** — requires an operator to have a config value whose `$<identifier>` substring accidentally matches a set environment variable; the outcome is a corrupted (not bypassed) auth token, producing a detectable authentication failure rather than a silent bypass.

## Description

`expandEnvRefs` applies the regex

```
\$\{[A-Za-z_][A-Za-z0-9_]*\}|\$[A-Za-z_][A-Za-z0-9_]*
```

to every decoded config string and replaces each `$VAR` / `${VAR}` match with the environment value (or leaves the match unchanged when the variable is unset).

The code comment at line 70 and the `LoadGatewayConfig` docstring both claim that a password like `pa$$word` is not silently corrupted:

```go
// so secrets containing '$'
// (e.g. "pa$$word") are not silently corrupted.
```

```go
// literal '$' that is not a well-formed reference (e.g. "$$", a password
// containing '$') are left untouched, so values are never silently corrupted.
```

The user-facing documentation repeats this claim (`docs/client-integration.md` line 157: "a password like `pa$$w0rd`, say").

**The claim is incorrect.** In the string `pa$$word`:

- Position 2: `$` followed by `$` (not `[A-Za-z_]`) → no regex match; this `$` is left alone. ✓
- Position 3: `$` followed by `w` (an identifier start) → `$word` **matches** the regex. ✗

If the environment variable `word` is set, `$word` is replaced with its value, yielding `pa$<env:word>` instead of `pa$$word`. The same applies to `pa$$w0rd` (the docs example): `$w0rd` matches because `w` satisfies `[A-Za-z_]` and `0rd` satisfies `[A-Za-z0-9_]*`.

The existing test (`cmd/eunox-mcp/gateway_test.go:102`) validates `pa$$-` (second `$` followed by `-`, a non-identifier character) and correctly asserts it is left unchanged. It does **not** test `pa$$word`, so this case has no regression coverage.

## Trigger / Repro

```go
// In a test or direct call:
os.Setenv("word", "foo")
got := expandEnvRefs("pa$$word")
// got == "pa$foo", not "pa$$word"
```

Config scenario:
```yaml
listen:
  authToken: pa$$word  # operator intended the literal string
```
With `word=foo` in the environment, the proxy starts with `authToken = "pa$foo"`, silently breaking authentication.

## Suggested Fix

Two options:

**Option A — Correct the documentation only.** The function's behavior (expand all `$VAR`) is reasonable; the misleading claim should be removed. Update the comment to: "A bare `$` followed by a non-identifier character (e.g. `$-`, `$!`) is not expanded; a `$` followed by a letter or `_` is always treated as a reference." Add a test case for `pa$$word` with `word` set, verifying it gets expanded (documenting the actual behavior as the contract).

**Option B — Implement `$$` as an escape sequence.** Change `expandEnvRefs` to recognise `$$` as a literal `$`, expanding `$$word` to `$word` (not to `<env:word>`). This is the shell convention. The regex can be updated to match `$$` first as a no-expand escape, or the function can pre-process `$$` → sentinel → expand → restore. This makes the documented behavior actually true, and the existing test (`pa$$-` → `pa$$-`) continues to pass.

Option A has lower risk (no behavior change). Option B fixes the semantic gap the comment implies.

## Session

`rev-20260620T1545Z-d3f21e`

### LINKED ISSUE #1002: [rev-20260620T1710Z-1d0da7] DirectiveWrapper.MarshalJSON nil guard is ineffective for a typed-nil *RedactFieldsDirective (nil-pointer panic)
**Location:** `pkg/capability/directives.go:22-61` (and the parallel path `pkg/enforcement/engine.go:907-924`)
**Reference:** https://github.com/eunolabs/eunox/blob/9c1c48dacbd98a6bdf7532559eece1c8d26a0d46/pkg/capability/directives.go#L22-L61
**Severity:** low

**Description:**
`DirectiveWrapper.MarshalJSON` guards against a nil directive:

```go
func (w DirectiveWrapper) MarshalJSON() ([]byte, error) {
	if w.Directive == nil {
		return []byte("null"), nil
	}
	return marshalDirective(w.Directive)
}
```

That guard is the classic Go nil-interface gotcha: it only catches a *nil interface*, not an interface that holds a **typed-nil pointer**. A `DirectiveWrapper{Directive: (*RedactFieldsDirective)(nil)}` has `w.Directive != nil` (the interface carries a concrete type), so the guard is skipped and `marshalDirective` runs:

```go
case *RedactFieldsDirective:
	type alias RedactFieldsDirective
	return json.Marshal(struct {
		directiveEnvelope
		*alias
	}{directiveEnvelope{Type: typed.DirectiveType()}, (*alias)(typed)})
```

`typed.DirectiveType()` is a **value-receiver** method (`func (RedactFieldsDirective) DirectiveType()`), so calling it on the nil `*RedactFieldsDirective` dereferences nil → panic. The explicit `== nil` guard in `MarshalJSON` signals the author intended to handle the nil case, but it is incomplete.

The same class of access exists in `(*Engine).collectObligations`, where `case *capability.RedactFieldsDirective:` also matches a nil pointer and then reads `d.Fields`, panicking on a nil `*RedactFieldsDirective`.

**Trigger / repro:**
```go
w := capability.DirectiveWrapper{Directive: (*capability.RedactFieldsDirective)(nil)}
_, _ = json.Marshal(w) // panics: nil pointer dereference
```
Reachable in practice only via a programmatically-constructed directive slice that contains a nil pointer (e.g. a library caller setting `req.Directives`, or a `Constraint.Directives` built in-process). The normal manifest-load / `DirectiveWrapper.UnmarshalJSON` path always produces a non-nil pointer, so the shipped proxy does not hit this today — hence low severity. It is a latent robustness defect: a panic in JSON marshaling rather than a clean error/`null`.

**Suggested fix:**
Make the guard catch a typed-nil pointer too, e.g. in `MarshalJSON`:

```go
if w.Directive == nil {
	return []byte("null"), nil
}
if rv := reflect.ValueOf(w.Directive); rv.Kind() == reflect.Ptr && rv.IsNil() {
	return []byte("null"), nil
}
```

or, more locally, guard the nil pointer inside `marshalDirective`'s `*RedactFieldsDirective` case (and mirror it in `collectObligations`) so a nil pointer returns an error / `null` instead of dereferencing. A small unit test marshaling a typed-nil wrapper pins the behavior.

**Session:** rev-20260620T1710Z-1d0da7

### LINKED ISSUE #1003: [rev-20260620T1810Z-c81f82] redactFields cannot redact a top-level field whose name begins with "$" (e.g. $ref/$schema) — silent fail-open + wrong key deleted
**Location:** `internal/pdp/pdp.go:1399-1428` (`RedactDotPath`); validation mirror `internal/config/manifest.go:629-652` (`validateRedactFields`)
**Reference:** https://github.com/eunolabs/eunox/blob/9c1c48dacbd98a6bdf7532559eece1c8d26a0d46/internal/pdp/pdp.go#L1399-L1428
**Severity:** medium

**Description:**
`RedactDotPath` strips the optional JSONPath root marker once, at entry, with two sequential `TrimPrefix` calls:

```go
func RedactDotPath(obj map[string]interface{}, dotPath string) {
	dotPath = strings.TrimPrefix(dotPath, "$.")
	dotPath = strings.TrimPrefix(dotPath, "$")
	redactDotPathRec(obj, dotPath)
}
```

The second `TrimPrefix(dotPath, "$")` fires on **any** leading `$`, including one that is part of a real field name rather than the root marker. The recursive worker `redactDotPathRec` was deliberately changed to **not** re-strip `$` (its doc comment explains this is so a *nested* field named `$schema`/`$ref`/`$id` is matched literally) — but the entry-point double-strip leaves the **top-level** case still broken and inconsistent with that fix.

Concretely, for a path naming a top-level key that begins with `$`:
- `"$ref"` → strip `"$."` (no match) → strip `"$"` → `"ref"` → deletes top-level key `ref`.
- `"$.$ref"` → strip `"$."` → `"$ref"` → strip `"$"` → `"ref"` → deletes top-level key `ref`.

There is **no** spelling that targets a literal top-level `$ref`. The consequences are both:
1. **Fail-open:** the field the operator declared for redaction (`$ref`) is forwarded to the host **unredacted**, while the engine/audit record reports the redaction obligation as discharged — a silent bypass of a configured security control.
2. **Collateral deletion:** an unrelated sibling key (`ref`) is silently deleted from the response.

Top-level `$`-prefixed keys are common in JSON-Schema-like tool outputs (`$schema`, `$ref`, `$id`, `$defs`), so this is reachable with ordinary tool results.

`validateRedactFields` uses the same double-strip (`strings.TrimPrefix(strings.TrimPrefix(field, "$."), "$")`), so the bad path passes load-time validation — the misconfiguration surfaces only as a silent runtime leak, exactly the failure mode `validateRedactFields` was written to prevent.

**Trigger / repro:**
Manifest directive on a `tool:` target: `redactFields` with `fields: ["$ref"]` (or `["$.$ref"]`). Upstream tool returns a text content item whose JSON body (or `structuredContent`) is `{"$ref": "secret", "ref": "keep"}`. After enforcement the host receives `{"$ref":"secret"}` (secret leaked) with `ref` removed, and the audit log shows the redaction applied.

**Suggested fix:**
Strip exactly one marker form and treat a `$`-prefixed segment as a literal field name otherwise — mirroring the nested behavior:

```go
func RedactDotPath(obj map[string]interface{}, dotPath string) {
	switch {
	case strings.HasPrefix(dotPath, "$."):
		dotPath = dotPath[len("$."):]
	case dotPath == "$":
		dotPath = ""
	}
	redactDotPathRec(obj, dotPath)
}
```

Apply the identical change to `validateRedactFields` so load-time validation and runtime agree, and add a redaction test for a top-level `$ref`/`$schema` key (allow path: redacted; plus a sibling `ref` left intact).

**Session:** rev-20260620T1810Z-c81f82

### LINKED ISSUE #1004: [rev-20260620T1810Z-c81f82] JWKSClient.VerifyToken samples the claim-validation clock before a potentially slow JWKS network fetch, widening the expired-token acceptance window on a cache miss
**Location:** `pkg/capability/jwks.go:314-325` (clock sample) and `pkg/capability/jwks.go:377-383` (the `now`-based expiry check)
**Reference:** https://github.com/eunolabs/eunox/blob/9c1c48dacbd98a6bdf7532559eece1c8d26a0d46/pkg/capability/jwks.go#L314-L325
**Severity:** low

**Description:**
`VerifyToken` samples the validation clock once, *before* it calls `VerifyWithKeyRotation`:

```go
now := c.now()                                   // line 320

return VerifyWithKeyRotation[TokenPayload](ctx, c.cache, kid, func(key *jose.JSONWebKey) (*TokenPayload, error) {
	...
	expected := jwt.Expected{ Time: now, AnyAudience: []string{c.audience} }
	if err := claims.ValidateWithLeeway(expected, c.leeway); err != nil { ... }
	...
})
```

The comment justifies sampling once "before entering the key-selection loop" so expiry is consistent across candidate keys during rotation. But `VerifyWithKeyRotation` performs the JWKS retrieval (`cache.GetKeys` / `ForceRefreshForKID` / `ForceRefreshForVerify`) **after** `now` is captured and **before** the per-key verifier runs. On a cache miss (TTL elapsed, or a forced refresh during key rotation) that retrieval is a network round-trip bounded only by the HTTP client timeout (default 10s) or `maxJWKSFetch` (30s). During that fetch, `now` goes stale.

`claims.ValidateWithLeeway` uses `expected.Time = now` as the reference instant for `exp`/`nbf`/`iat`. A stale (earlier) `now` makes the `exp` check **more lenient**: a token whose `exp` falls between `now` and `real_now` is accepted as still-valid. The extra acceptance window equals the fetch duration (up to ~10-30s), on top of the configured leeway — a fail-open for a freshly-expired token, gated by a cache miss coinciding with a slow JWKS endpoint.

This is also inconsistent with the sibling `TokenCache`, which was deliberately reworked to sample its clock **late** (under the lock, immediately before the expiry check) precisely to avoid serving a token past its TTL while waiting on contention (see `token_cache.go` `Get`/`Put`/`purgeExpired` comments). The token-verification path samples early, before a much longer (network) wait.

**Trigger / repro:**
1. Present a capability token that expires in `T` seconds, with `T` < (JWKS fetch latency + leeway).
2. Ensure the JWKS cache is cold or due for a forced refresh (e.g. first use after TTL, or a kid-miss rotation path) and the JWKS endpoint responds slowly (a few seconds).
3. The fetch elapses past the token's real `exp`, but `ValidateWithLeeway` compares against the pre-fetch `now`, so the expired token verifies. A test can drive this deterministically via `cfg.Now` (step-advancing clock) plus a slow/blocking JWKS HTTP handler.

**Suggested fix:**
Sample the validation clock after the key set is in hand but still once for all candidates. Cleanest is to let `VerifyWithKeyRotation` invoke the verifier with a freshly-sampled time, or split the verifier so claim/leeway validation re-reads `c.now()` immediately before `ValidateWithLeeway` (one sample, captured after `GetKeys` returns, reused across the retry). Either keeps the cross-candidate consistency the current comment wants while closing the fetch-duration leniency window. Mirror the late-sampling discipline already applied in `token_cache.go`.

**Session:** rev-20260620T1810Z-c81f82

### LINKED ISSUE #1007: [rev-20260620T1910Z-8127a6] HTTP live-introspection (fetchLiveTools) ignores tools/list nextCursor — init/validate --live/doctor see only the first page of tools
**Location:** `cmd/eunox-mcp/live_upstream.go:81-96` (`fetchLiveTools`)
**Reference:** https://github.com/eunolabs/eunox/blob/0dc5eab5934e4bb1280d3d3e0890f439fe148792/cmd/eunox-mcp/live_upstream.go#L81-L96
**Severity:** medium

**Description:**
`fetchLiveTools` (the HTTP one-shot live-introspection client used by `init`, `validate --live`, and `doctor --live`) sends a single `tools/list` request and passes the result straight to `drift.ParseToolsListResult`, never following the MCP `nextCursor` pagination field:

```go
listResp, _, err := transport.DoMCPHTTP(ctx, client, endpoint, mcp.RPCMsg{
    JSONRPC: "2.0", ID: mcp.RawJSON(`2`), Method: "tools/list",
}, sessID, authHeader)
...
tools, err := drift.ParseToolsListResult(listResp.Result)
```

Every other `tools/list` consumer in the codebase paginates to exhaustion via `drift.FetchAllToolPages`:
- The stdio CLI peer in the same file, `runStdioHandshake` (`live_upstream.go:218-241`).
- The transport runtime drift probes for **both** HTTP and stdio upstreams (`internal/transport/drift_probe.go:24` and `:61`).

`FetchAllToolPages`' own doc comment states it is *"Shared by the HTTP, stdio, and CLI live-introspection paths so all three paginate identically"* (`internal/drift/drift.go:531`). The CLI HTTP path is the lone exception that violates that contract.

**Trigger / repro:**
Run `eunox-mcp init --upstream-url <url>`, `eunox-mcp validate --live --upstream-url <url>`, or `eunox-mcp doctor --live --config <cfg>` against an HTTP MCP upstream whose `tools/list` returns a `nextCursor` (a paginated tool catalog). Only the first page of tools is observed; all subsequent pages are silently dropped.

**Impact:**
- `init` scaffolds a deny-all starter manifest that is missing every tool on page 2+, so an operator who "uncomments the tools they need" never even sees those tools exist.
- `validate --live` / `doctor --live` drift reporting is wrong: tools on later pages are invisible, so newly-added later-page tools are never surfaced as NOT COVERED, and a manifest entry that does match a later-page tool can be misreported as STALE (FM-2) — potentially prompting an operator to delete a live, in-use entry.
- The stdio path against the same logical server reports correctly, so the two transports disagree, undermining trust in the drift output.

**Suggested fix:**
Replace the single-shot call with the shared paginator, mirroring `runStdioHandshake`:

```go
merged, err := drift.FetchAllToolPages(func(cursor string) (json.RawMessage, error) {
    req := mcp.RPCMsg{JSONRPC: "2.0", ID: mcp.RawJSON(`2`), Method: "tools/list"}
    if cursor != "" {
        req.Params, _ = json.Marshal(map[string]string{"cursor": cursor})
    }
    listResp, _, err := transport.DoMCPHTTP(ctx, client, endpoint, req, sessID, authHeader)
    if err != nil {
        return nil, fmt.Errorf("tools/list: %w", err)
    }
    if listResp.Error != nil {
        return nil, fmt.Errorf("tools/list: server error %d: %s", listResp.Error.Code, listResp.Error.Message)
    }
    return listResp.Result, nil
})
if err != nil {
    return LiveUpstreamInfo{}, err
}
tools, err := drift.ParseToolsListResult(merged)
```

**Session:** rev-20260620T1910Z-8127a6

### LINKED ISSUE #1008: [rev-20260620T1910Z-8127a6] init --pin-descriptions skips tools with an empty description, leaving a tool-poisoning detection gap
**Location:** `cmd/eunox-mcp/init_manifest.go:204-206` (`toolEntryYAMLLines`)
**Reference:** https://github.com/eunolabs/eunox/blob/0dc5eab5934e4bb1280d3d3e0890f439fe148792/cmd/eunox-mcp/init_manifest.go#L204-L206
**Severity:** low

**Description:**
`init --pin-descriptions` is the mechanism that emits a `descriptionHash` for each tool so the proxy can detect upstream "tool poisoning" (a tool's description silently changing) at startup (FM-5). The generator gates the pin on a non-empty description:

```go
if pinDescriptions && tool.Description != "" {
    lines = append(lines, fmt.Sprintf("  descriptionHash: %q", capability.ComputeDescriptionHash(tool.Description)))
}
```

A tool whose live description is the empty string therefore gets **no** `descriptionHash` even though the operator explicitly asked for descriptions to be pinned. The runtime FM-5 path itself handles a hash of `""` correctly (`ComputeDescriptionHash("")` is a well-defined value and the drift check compares it like any other), so the only thing preventing coverage is this init-time guard.

The consequence is a detection gap: if an upstream later changes an empty description into a poisoned, prompt-injecting one, nothing detects it — there is no pin to compare against, and a description change is not surfaced by any other drift failure mode. The operator who ran `--pin-descriptions` reasonably believes every tool is pinned.

**Trigger / repro:**
1. Point `eunox-mcp init --pin-descriptions` at an upstream that exposes a tool with an empty `description` field.
2. Inspect the generated manifest: that tool has `actions`/`argumentSchema` but no `descriptionHash` line, while sibling tools with non-empty descriptions do.
3. The upstream later sets that tool's description to a malicious value; `eunox-mcp proxy` startup raises no FM-5 drift abort for it.

**Suggested fix:**
Drop the `&& tool.Description != ""` guard so every tool is pinned when `--pin-descriptions` is set (the hash of `""` is a valid, verifiable pin):

```go
if pinDescriptions {
    lines = append(lines, fmt.Sprintf("  descriptionHash: %q", capability.ComputeDescriptionHash(tool.Description)))
}
```

Alternatively, if empty-description tools are intentionally excluded, document that they must be pinned manually and emit a stderr notice naming them so the gap is not silent.

**Session:** rev-20260620T1910Z-8127a6

### LINKED ISSUE #1009: [rev-20260620T1910Z-8127a6] Duplicate exact tool targets aren't rejected at load; first-in-order wins on a specificity tie, letting a non-pinned constraint shadow a descriptionHash-pinned one (FM-5 skipped)
**Location:** `internal/drift/drift.go:400-419` (`BestManifestConstraint`); manifest loader `internal/config/manifest.go` (no duplicate-exact-target rejection)
**Reference:** https://github.com/eunolabs/eunox/blob/0dc5eab5934e4bb1280d3d3e0890f439fe148792/internal/drift/drift.go#L400-L419
**Severity:** low

**Description:**
`BestManifestConstraint` selects the constraint covering a tool by specificity score, breaking ties by **first in manifest order**:

```go
if s := resSpecificity(c.Target, toolName); s > bestScore {  // strict >, so the FIRST max-scoring entry wins
    bestScore = s
    best = i
}
```

Two constraints with the identical exact target (e.g. two `tool:read_file` entries) both score the maximum exact-match specificity (1000), so the tie is resolved purely by which appears first in the manifest. The manifest loader rejects various other duplicate/ambiguous constructs (e.g. duplicate condition windows/tables) but does **not** reject two constraints with the same exact `tool:` target.

This interacts badly with descriptionHash pinning (FM-5, the tool-poisoning defense). If an operator authors a non-pinned `tool:read_file` *before* a pinned `tool:read_file`, `BestManifestConstraint` returns the non-pinned one. The drift check's FM-5 verification only runs when the selected constraint `IsPinnedExactTool()` (`drift.go` FM-5 loop), so the pin is silently skipped — a poisoned description for that tool is never detected, even though the operator clearly intended to pin it. The covering constraint chosen for the COVERED report also disagrees with what an operator reading the manifest would expect.

**Trigger / repro:**
Author a manifest with two entries for the same tool, the non-pinned one first:
```yaml
capabilities:
  - target: tool:read_file
    actions: [call]
  - target: tool:read_file
    actions: [call]
    descriptionHash: "sha256:...."   # intended poisoning pin
```
Load passes validation (no duplicate-target error). At startup the first (non-pinned) entry wins the specificity tie, so `read_file`'s description is never hash-verified; a later upstream description change goes undetected.

**Suggested fix (either):**
1. Reject duplicate exact `tool:` targets at manifest load (fail closed on the ambiguity, matching the loader's treatment of other duplicate constructs), **or**
2. On a specificity tie, deterministically prefer the pinned constraint (and/or surface a load-time warning), so a pin can never be shadowed by a less-restrictive sibling.

Add a drift test covering "non-pinned exact entry precedes pinned exact entry for the same tool → FM-5 still verifies the pin."

**Session:** rev-20260620T1910Z-8127a6

eunolabs:
## Bugs addressed

Eight session-tagged automated-review issues across five review sessions, one commit per bug.

- Closes #996 — `initRemoteUpstream` set `s.upstreamSessID` only after `applyInitializeResult` succeeded, so a malformed/error/nil-result initialize response made `close()` skip the DELETE and permanently leak the already-allocated upstream session. Capture the session ID from the response header before validating the result.
- Closes #1000 — The `expandEnvRefs` comment, `LoadGatewayConfig` docstring, and `docs/client-integration.md` wrongly claimed `pa$$word`-style values survive expansion. A `$` followed by an identifier always expands (`$word` inside `pa$$word`), so corrected all three to the actual contract (no `$$` escape) and added a regression test.
- Closes #1002 — `DirectiveWrapper.MarshalJSON`'s nil-interface guard missed a typed-nil `*RedactFieldsDirective`, so `marshalDirective` panicked calling the value-receiver `DirectiveType()` on nil. Guard the nil pointer in `marshalDirective` (emit `null`) and in `Engine.collectObligations` (skip — nothing to redact).
- Closes #1003 — `RedactDotPath` stripped any leading `$` unconditionally, so a top-level `$`-prefixed field (`$ref`, `$schema`) was never redacted (silent fail-open) while an unrelated sibling (`ref`) was deleted. Strip only the real root-marker spellings (`$.` and a lone `$`); mirrored in `validateRedactFields` so load-time and runtime agree.
- Closes #1004 — `JWKSClient.VerifyToken` (capability-token path) sampled the validation clock before the JWKS fetch, so a cold-cache/rotation round-trip widened the `exp` acceptance window by the fetch duration (fail-open for a freshly-expired token). Sample lazily on the first per-key verify call, once. The IdP-JWT path keeps its deliberate up-front sampling (existing regression test); a note documents the intentional divergence on both sides.
- Closes #1007 — `fetchLiveTools` (HTTP live introspection for `init`/`validate --live`/`doctor --live`) never followed `tools/list` `nextCursor`, so only the first page of a paginated tool catalog was observed (incomplete `init` manifests, wrong drift reporting). Routed through the shared `drift.FetchAllToolPages`, matching the stdio path and both transport probes.
- Closes #1008 — `init --pin-descriptions` skipped the `descriptionHash` for any tool with an empty live description, leaving a tool-poisoning detection gap. The hash of `""` is a valid pin; drop the guard so every tool is pinned.
- Closes #1009 — `BestManifestConstraint` broke an exact-target specificity tie by manifest order, so a non-pinned `tool:` entry placed before a pinned one with the same target won — and FM-5 only verifies the selected constraint, silently skipping the pin. Deterministically prefer the pinned constraint on a tie.

## Testing

- `make lint` (go vet + golangci-lint) — 0 issues; `gofmt` clean on all changed files.
- New regression tests added per fix: typed-nil directive marshal, top-level `$ref` redaction, capability-token clock-sampled-after-fetch, FM-5 pin-not-shadowed-on-tie, `$word`-after-`$` env expansion, and empty-description pinning (existing empty-description test updated to assert the new behavior).

---
_Generated by [Claude Code](https://claude.ai/code/session_01VQ4juwZvGTBbfmfSGJF259)_


internal/pdp/pdp.go
```diff
@@ -1409,15 +1409,22 @@ func redactJSONValue(val interface{}, paths []string) bool {
...
+// leaked the declared $-prefixed field (it was never redacted) and wrongly
+// deleted an unrelated sibling (e.g. "$ref" became "ref").
 func RedactDotPath(obj map[string]interface{}, dotPath string) {
-	dotPath = strings.TrimPrefix(dotPath, "$.")
-	dotPath = strings.TrimPrefix(dotPath, "$")
+	switch {
+	case strings.HasPrefix(dotPath, "$."):
+		dotPath = dotPath[len("$."):]
+	case dotPath == "$":
+		dotPath = ""
```
eunolabs, line 1426: The fix is correct for the `$ref`/`$schema` case, and mirroring it in `validateRedactFields` (manifest.go) is the right call — load-time and runtime now agree.

One behavior change worth calling out: a path written as `$secret` (a single `$` + name, **no** dot) used to be stripped to `secret` by the old unconditional `TrimPrefix`; it now falls through to `default` and is treated as a literal field named `$secret`. If a response has a top-level `secret`, that path no longer redacts it — a **silent fail-open** for any existing manifest relying on the old loose stripping (validation passes, the audit still logs "redaction discharged").

This is defensible (canonical forms are `$.x` / bare `$`), but it's a manifest-grammar behavior change. `docs/capability-manifest-guide.md` (around line 1571) currently says *"the leading `$.` or `$` is optional"*, which reads as if `$secret` ≡ `secret`. Per the repo conventions (manifest-grammar changes update the guide + a migration note), this PR should tighten that doc line so authors know only `$.`-prefix and a lone `$` are stripped now.

---

pkg/capability/jwks.go
```diff
@@ -311,18 +311,30 @@ func (c *JWKSClient) VerifyToken(ctx context.Context, tokenStr string) (*TokenPa
...
+	// Sample the validation clock once, lazily, on the first per-key verify call —
+	// which runs only AFTER VerifyWithKeyRotation has fetched the key set
+	// (cache.GetKeys / a forced refresh during key rotation). Sampling before that
+	// call, as this path used to, captured `now` ahead of a potentially slow JWKS
+	// network round-trip (bounded only by the HTTP-client / maxJWKSFetch timeout);
+	// the stale earlier `now` widened the exp acceptance window by the fetch
+	// duration, letting a freshly-expired token verify on a cache miss. Sampling
+	// once (the closure is invoked serially, never concurrently) keeps expiry
+	// consistent across every candidate key and the rotation retry, while moving
+	// the sample past the fetch — mirroring the late-sampling discipline in
```
eunolabs: Verified the mechanics: `VerifyWithKeyRotation` calls `cache.GetKeys` (and, on a kid miss, `ForceRefreshForKID`) **before** invoking the verifier, and `tryKeys` calls the closure serially in a `for` loop — so lazy sampling is past the initial fetch and the non-atomic `now`/`nowSampled` are race-free. Good fix for the cold-cache case.

One precision note on "moving the sample past the fetch": there can be a **second** fetch after the sample. When the cached/refreshed keys all fail signature verification, the function does one more `ForceRefresh...` round-trip and re-runs `tryKeys`; by then `nowSampled` is already true, so `now` is the instant captured *before* that second network round-trip. So the original window-widening still exists in the (narrower) rotation-retry-after-signature-failure path. Re-sampling there would reintroduce the cross-key inconsistency you're deliberately avoiding, so sampling-once is the right trade-off — but the comment slightly over-claims. Consider softening "past the fetch" to "past the initial key fetch" so the residual case isn't mistaken for covered.

---

internal/pdp/jwt.go
```diff
@@ -440,6 +440,12 @@ func (p *JWTPDP) ValidateToken(ctx context.Context, authHeader string) (context.
 	// token valid when received could be rejected after the refresh retry (or vice
 	// versa). Capturing one instant matches the single-sample pattern the engine,
 	// TokenCache, and AlwaysAllowPDP already use.
+	//
+	// NOTE: this IdP-JWT path deliberately samples UP FRONT (a token valid when
+	// received stays valid even if a forced refresh round-trip crosses its exp).
```
eunolabs: The cross-reference note between the two paths is genuinely helpful. But I'd push on the *substance* of the divergence, not just the documentation of it.

Both paths run through the same `VerifyWithKeyRotation` (same `GetKeys`/force-refresh round-trips), so the IdP-JWT path has the **exact same** fail-open that #1004 just fixed on the capability-token side: a JWT valid at receipt but whose `exp` passes during a cold-cache JWKS fetch will still verify, because `now` is sampled up front here. For a proxy whose stated posture is "fail closed," accepting a freshly-expired IdP token because the key fetch was slow is the same class of issue you're treating as a bug next door.

The stated rationale ("a token valid when received stays valid even if a forced refresh crosses its exp") is a real design stance, but it's asserted rather than justified against the fail-closed default. Could you say a bit more about *why* the IdP path should be lenient where the capability path is strict? If there isn't a strong reason, sampling late on both (with the same "sample once" discipline) would be the more consistent and safer outcome. Not blocking — but it deserves a deliberate answer, since the note currently just says "don't unify."

---

eunolabs (commented):
## Review — automated-review bug-fix batch

Reviewed all 8 fixes against the source (not just the diff). This is a clean, well-scoped batch: one commit per bug, each with a focused regression test and load-bearing comments explaining the *why*. CI is fully green (41/41; `unstable` is branch protection, not a failure). Nothing here is a blocker — comments are a docs gap and two design questions.

### Verified correct
- **#996 (session leak)** — Confirmed `httpSession.close()` gates the upstream DELETE on `s.upstreamSessID != "" && s.route != nil`. Capturing the session ID from the response header *before* `applyInitializeResult` means a malformed/error initialize result now still gets cleaned up. Empty-header case stays a no-op. Correct.
- **#1002 (typed-nil directive)** — Both guards are right (marshal → `null`, `collectObligations` → skip). See minor note below on `null` vs. error.
- **#1003 (redactFields `$`-prefix)** — Runtime and `validateRedactFields` now strip exactly the same two spellings (`$.` and lone `$`); they agree. One behavior-change caveat inline.
- **#1004 (JWKS clock)** — Traced `VerifyWithKeyRotation`: keys are fetched before the verifier runs, and the closure is invoked serially, so the lazy `now`/`nowSampled` are correct and race-free. One precision note inline + a divergence question on the IdP path.
- **#1007 (tools/list pagination)** — Routes through `drift.FetchAllToolPages` (cursor dedup + page/tool caps), `encoding/json` already imported, ignoring the `json.Marshal(map[string]string)` error is fine. Matches the stdio path. Correct.
- **#1008 (pin empty descriptions)** — Dropping the `tool.Description != ""` guard closes a real poisoning gap; hash of `""` is a valid pin. Correct.
- **#1009 (FM-5 tie-break)** — Tie-break logic is sound across all orderings (pinned-first stays, non-pinned-then-pinned flips, two-pinned keeps first). `IsPinnedExactTool()` correctly gates on exact tool target + non-empty hash. Correct.

### Points worth addressing (in priority order)
1. **Docs gap (#1003)** — The redactFields path-parsing change is a manifest-grammar behavior change with a possible silent fail-open for the undocumented `$name` form; `docs/capability-manifest-guide.md` (~line 1571) still implies `$secret` ≡ `secret`. Repo convention says grammar changes update the guide. See inline.
2. **Auth-path divergence (#1004)** — The IdP-JWT path retains the same cold-cache fail-open #1004 fixes for capability tokens. The "deliberately diverge" note documents the asymmetry but doesn't justify it against the fail-closed default. See inline.
3. **Comment precision (#1004)** — "past the fetch" doesn't cover the rotation-retry refresh that can happen after sampling. See inline.

### Minor / non-blocking
- **#1002 marshal → `null`**: emitting `null` for a typed-nil directive is a reasonable, fail-safe choice and matches the nil-interface round-trip, but since the codepath is "programmatic-construction only," an `error` return would arguably be more in keeping with the fail-closed ethos. Either is fine — flagging for awareness.
- **#1000** is comment/doc-only (the regex was already correct); the clarified contract and the `pa$$word` → `pa$<value>` regression test are a good catch and accurately documented. Note this does highlight a real foot-gun (a `$name`-containing secret gets partially expanded if `name` is set in the env) — the doc now says so, which is the right resolution.

Nice work — the comments-as-rationale style here makes the security intent of each fix easy to verify.

---

Issue #969, opened 2026-06-20
Code review: maintenance traps, wiretap kill-switch gap, and minor simplifications


eunolabs:
## Health summary

The codebase is in excellent shape. The fail-closed principle is consistently applied across every ambiguous path; security-sensitive surfaces (kill-switch ordering, JWT interception, Origin/DNS-rebinding validation, audit HMAC chaining) are handled with precision. The shared `dispatchRequest`/`enforcedForwardCore` core prevents the two transports from drifting, and the two-phase `maxCalls` evaluation, injection-safe `compositeCounterKey`, and atomic `IncrementIfBelow` are all exactly correct. The only substantive finding below is an operator-facing gap — kill-switch activation has no effect on wiretap routes in non-JWT mode — which the code comments acknowledge as intentional but do not surface loudly at runtime. The remaining findings are low-priority: a maintenance trap in `filterToolsListResult`, a minor asymmetry in `forwardServerRequest`, a pointless struct projection, and two nil-guard legibility issues.

**Highest-leverage changes (in order):**
1. Add a startup/kill-endpoint warning that wiretap routes are not affected by kill-switch (no code change to enforcement logic needed).
2. Resolve `filterToolsListResult` nil-check inversion with a local `pinned` variable.
3. Collapse `forwardParams` / `dispatchParams` projection (delete ~10 lines).

## What's done well

- Fail-closed default: unmapped methods, missing manifests, malformed JWTs, unset condition args all deny with a structured error code.
- Two-phase `maxCalls`: pure-predicate pass first, consuming-condition commit pass second — quota is never burned on calls denied by later conditions.
- `compositeCounterKey` uses length-prefixed parts; no delimiter collision between caller-controlled sessionID and tool names.
- `enforcedForwardCore` is the single shared deny/observe/forward/record core for both transports — no behavioral drift.
- `JWTPDP.CheckKill` correctly wires the kill-switch even for JWT-wrapped wiretap routes; `WrapRoutesWithJWT` threads the manager through.
- `checkOrigin` correctly distinguishes absent header (pass) from present-but-empty / opaque "null" (must satisfy allowlist); rejects userinfo-prefixed Origins.
- Audit HMAC chain writes `AUDIT_INTEGRITY_FAILURE` markers on parse failure, HMAC mismatch, and I/O error — integrity failures are themselves on the tape.
[...]

PR #1011, opened 2026-06-20
feat: add --require-audit=strict fail-closed audit gate (#833)

eunolabs:
## Summary

Requested: *"Fix issues #833 to #969 in a single PR."*

On investigation, **five of the six open issues in that range were already
resolved** by intervening changes (they stayed open only because this repo's
convention forbids issue numbers in commits, so nothing auto-closed them). The
**one genuinely-remaining actionable item** was #833's suggested
`--require-audit=strict` runtime mode — this PR implements it.

### Status of each issue in #833–#969

| Issue | State | Evidence |
|------|-------|----------|
| **#833** Audit fail-open on the allow path | **Fixed here** (doc half already done in §5.4.1; the suggested `--require-audit=strict` mode was missing) | this PR |
| #836 Remote-HTTP drops server-initiated requests | Already resolved — the maintainers' chosen option (b): documented + the `system:sampling` opt-in is rejected at startup for HTTP upstreams | `internal/transport/http_remote.go`, prior PR #935 |
| #840 Threat-model §6.1 omits 5 audit fields | Already resolved — `seq`, `prev_hmac`, `upstream`, `policy_version`, `policy_sha256` are all in the table | `docs/threat-model-mcp.md` §6.1 |
| #842 Idle SSE session pins upstream | Already resolved — hard idle ceiling (`hardIdleMultiplier` + `lastRequest`) | `internal/transport/http_session.go`, prior PR #936 |
| #841 Migrate off `gopkg.in/yaml.v3` | Explicitly deferred by the issue ("No action for v0.1.0") | n/a |
| #969 Code-review findings A1–C2 | Already resolved — all six findings landed | prior PR #1005 |

## What this PR implements (#833)

`--require-audit` becomes a **three-state** flag:

- `off` (default) — a sink-open failure warns and the proxy continues unaudited.
- `on` / bare `--require-audit` / `=true` — a sink-open failure is **fatal at startup** (prior behavior, unchanged).
- `strict` — `on`, **plus a runtime fail-closed gate**.

In `strict` mode, before forwarding any authorized request — an enforced method
(`tools/call`, `resources/read`, `resources/subscribe`, `prompts/get`) or a
`*/list` enumeration — the shared forward core consults `Sink.AuditDegraded()`.
Once a record has been **dropped under back-pressure** or a **log write has
failed**, the call is denied fail-closed with a new **`AUDIT_UNAVAILABLE`** code
(JSON-RPC `-32603`) and **the upstream is never contacted** — so a privileged
call can no longer be applied with no durable record of the decision (the
allow-path fail-open documented in threat-model §5.4.1).

The gate is **coarse and sticky by design** (it keys on the cumulative loss
counters, so one drop denies all later forwards for the process lifetime),
deliberately trading data-plane availability for audit completeness — the inverse
of the default posture. Use it only where an unauditable privileged call is worse
than a denied one.

## Changes

- **audit**: `Sink.AuditDegraded() (bool, string)` — reports dropped/write-failure state with a value-free reason; nil-safe.
- **capability**: new `ErrCodeAuditUnavailable`, mapped to `-32603`.
- **transport**: `auditRecorder` grows `AuditDegraded()`; `enforcedForwardCore` and `dispatchList` gate the forward via the shared `forwardParams.strictAuditDenial`; `requireAuditStrict` threaded through both proxies + `routeSink`.
- **cmd**: `auditRequirement` `flag.Value` (`IsBoolFlag` preserves bare `--require-audit`; a typo fails closed at parse time).
- **docs**: threat-model §5.4.1 rewritten from "future direction (not implemented)" to the implemented gate; `conformance.md`, `README.md`, `CHANGELOG.md` updated.

## Tests

- Strict gate: degraded → `AUDIT_UNAVAILABLE` + upstream skipped; healthy → forwards; non-strict + degraded → still forwards (default posture preserved); `*/list` path.
- `AuditDegraded` unit tests (dropped, write-failure, nil receiver).
- Three-state flag parsing (`off`/`on`/`true`/`bare`/`strict`/typo).

## Verification

`go build ./...`, `go vet ./...`, `go test -race -count=1 ./...`, `make lint`, `make check-fmt check-license check-notice` — all green. Binary smoke-tested: `--help` renders the flag, invalid values fail closed at parse, `--require-audit=strict` parses.

https://claude.ai/code/session_01LgD1taNoVi6k9un1LEcpK6

---
_Generated by [Claude Code](https://claude.ai/code/session_01LgD1taNoVi6k9un1LEcpK6)_


internal/transport/forward.go
```diff
@@ -72,6 +113,15 @@ func enforcedForwardCore(ctx context.Context, fp forwardParams, msg mcp.RPCMsg,
 		)
 	}
 
+	// --require-audit=strict: do not forward an authorized call (a clean allow, or
+	// an audit-mode observe downgraded from a would-be deny above) once the audit
+	// trail has degraded. Fail closed instead, so audit completeness is a hard
+	// precondition for the upstream side effect rather than a best-effort
+	// afterthought recorded only after the call already happened.
+	if denied, blocked := fp.strictAuditDenial(ctx, msg, auditID, method, denialTarget); blocked {
```
eunolabs, line 157: **The gate is retrospective — the boundary call is still forwarded unaudited.**

`strictAuditDenial` reads the cumulative loss counters via `AuditDegraded()`, which only reflect *prior* calls. But this call's own record isn't enqueued until `RecordAllow` runs *after* `callUpstream` (line 125, then ~line 169). So for the first call whose `RecordAllow` is dropped under back-pressure:

1. gate checks `AuditDegraded()` → `false` (no prior drop)
2. `callUpstream` forwards — the privileged side effect happens
3. `RecordAllow` is enqueued → dropped → counter goes 0→1

That call was applied to the upstream *and* its record was lost; only the *next* call sees `degraded=true`. Under concurrency, every call already in flight when the queue saturates is in the same window.

This is inherent to the async/non-blocking audit design, and I don't think making the allow path block on a durable write is worth it. But it means the absolute wording in the threat model / CHANGELOG / the `AuditDegraded` doc comment — "a privileged call can **never** / **no longer** be applied with no durable record" — overstates the guarantee. The accurate statement is "**once the first lost record is observed, every subsequent forward is denied**"; the boundary call(s) can still be forwarded unaudited. Suggest aligning the docs to the actual guarantee, or calling out the boundary window explicitly.

---

internal/transport/forward.go
```diff
@@ -34,6 +41,40 @@ type forwardParams struct {
...
+// enforced-forward core and the */list dispatch path. When strict mode is active
+// and the audit trail has degraded, it records a fail-closed AUDIT_UNAVAILABLE
+// deny and returns the JSON-RPC denial to deliver to the host (ok=true) so the
+// caller returns it instead of forwarding; otherwise it returns ok=false and the
+// caller proceeds normally.
+//
+// The deny record enqueued here may itself be dropped — that is acceptable: the
+// host-visible call is denied either way, so no unaudited privileged call reaches
+// the upstream. auditID names the call for the record; denialTarget names it for
+// the host error. The gate runs only on the forward path: a hard deny already
```
eunolabs, line 102: **`sampling/createMessage` is an enforced method but its allow path is not gated.**

The gate is wired into `enforcedForwardCore` and `dispatchList`, but `sampling/createMessage` — one of the five enforced, deny-by-default methods — is handled by `forwardServerRequest` (further down this file, unchanged by this PR). Its allow leg goes straight to `fp.forward(msg)` → `recordForwardOutcome` → `RecordAllow`, with no `strictAuditDenial` check. So under `--require-audit=strict` with a degraded trail, an *allowed* sampling call is still forwarded to the host — exactly the unaudited-privileged-forward this feature is meant to close.

Scope is narrow (sampling requires the `system:sampling` manifest opt-in, which is rejected at startup for HTTP upstreams, so this only bites stdio subprocess upstreams), but it's still a real gap, and the docs' blanket "before forwarding **any** authorized request … a privileged call can never be applied with no durable record" doesn't hold for it. Either add the gate to the sampling allow leg in `forwardServerRequest`, or document explicitly why sampling is exempt.

---

internal/transport/forward.go
```diff
@@ -34,6 +41,40 @@ type forwardParams struct {
...
+func (fp forwardParams) strictAuditDenial(ctx context.Context, msg mcp.RPCMsg, auditID, method, denialTarget string) (mcp.RPCMsg, bool) { //nolint:gocritic // hugeParam: rpcMsg by value, matching the forward core
+	if !fp.requireAuditStrict || fp.rec == nil {
+		return mcp.RPCMsg{}, false
+	}
+	degraded, reason := fp.rec.AuditDegraded()
+	if !degraded {
+		return mcp.RPCMsg{}, false
+	}
+	fp.rec.RecordDeny(ctx, fp.sessionID, auditID, method, capability.ErrCodeAuditUnavailable, "", map[string]interface{}{"reason": reason}, false)
+	fmt.Fprintf(os.Stderr,
```
eunolabs: **Per-request stderr flood once degraded (minor).**

Because the gate is sticky — it keys on cumulative counters that only ever increase — once the trail degrades this `Fprintf` fires on *every* subsequent enforced call and `*/list` for the rest of the process lifetime (plus a dropped deny record per request). The threat model describes the existing drop / write-failure signals as "one-shot stderr warnings"; this per-call line diverges from that pattern and could bury logs under load.

Consider making the strict-denial stderr warning one-shot — log on the healthy→degraded transition only — and leaning on the per-call audit deny records (`AUDIT_UNAVAILABLE`) as the durable per-request signal.

---

internal/audit/audit_test.go
```diff
@@ -3008,6 +3008,74 @@ func TestAuditSink_WriteFailuresCounted(t *testing.T) {
 	}
 }
 
+// TestAuditDegraded_HealthyThenDroppedThenWriteFailure pins the AuditDegraded
```
eunolabs: Nit: the doc comment opens with `TestAuditDegraded_HealthyThenDroppedThenWriteFailure`, but the function is `TestAuditDegraded_Dropped` (the write-failure axis is a separate test, `TestAuditDegraded_WriteFailure`). Looks like a rename left the comment behind — Go convention (and `go doc`) wants the comment to start with the actual function name.

---

eunolabs (commented):
## Review: `--require-audit=strict` fail-closed audit gate

Thanks for the unusually thorough write-up — the issue-by-issue triage and the threat-model rewrite make this easy to follow. The implementation is clean and the wiring is tasteful:

- The three-state `auditRequirement` `flag.Value` keeps the bare-flag ergonomics (`IsBoolFlag`), accepts `=off`/`=strict`, and **fails closed on a typo** at parse time rather than silently disabling the requirement.
- The gate lives in one shared `strictAuditDenial` consumed by both `enforcedForwardCore` and `dispatchList`, so the two transports and the `*/list` path can't drift.
- `AUDIT_UNAVAILABLE` correctly maps to the `-32603` internal-error class, distinct on the wire from the `-32001/-32002/-32003` policy codes.
- Nil-receiver safety is handled at both `*Sink` and `routeSink`, and the placement after the deny→observe downgrade means audit/wiretap forwarding is also blocked when degraded.
- Good test coverage: degraded-denies-and-skips-upstream / healthy-forwards / non-strict-still-forwards (default posture preserved) / `*/list` path / three-state flag parsing / `AuditDegraded` axes incl. nil receiver.

A few things worth addressing before merge — the two substantive ones are about the *guarantee the feature actually provides* vs. what the docs claim (which matters a lot for a compliance/forensics control):

**1. The gate is retrospective — the boundary call is still forwarded unaudited.** (inline, `forward.go`) `AuditDegraded()` reads cumulative counters reflecting only *prior* calls, and a call's own `RecordAllow` is enqueued *after* `callUpstream`. So the first call whose record is dropped is forwarded *and* loses its record; only the next call is denied. The absolute "a privileged call can **never** be applied with no durable record" wording (threat model / CHANGELOG / `AuditDegraded` doc) overstates this — the real guarantee is "**after the first detected loss, every subsequent forward is denied**." Suggest aligning the wording (closing it fully would mean a synchronous/blocking allow path, which I don't think is worth it).

**2. `sampling/createMessage` is not gated.** (inline, `forward.go`) It's an enforced, deny-by-default method, but its allow path runs through `forwardServerRequest`, which this PR doesn't touch — so an allowed sampling call still forwards under a degraded trail. Narrow in practice (stdio-only; needs the `system:sampling` opt-in), but it contradicts the "any authorized request" framing. Gate it, or document the exemption.

**3. Per-request stderr flood once degraded (minor).** (inline) The sticky gate prints a SECURITY line per request for the rest of the process lifetime; consider a one-shot/transition-only warning.

**4. Test doc-comment name mismatch (nit).** (inline, `audit_test.go`)

One more minor UX note: the `IsBoolFlag` quirk means `--require-audit strict` (space-separated) silently parses as `on` plus a stray positional. You've documented this in both the help text and a code comment — the right call — but it remains an easy footgun for anyone used to space-separated flag values; worth keeping in mind if support questions show up.

Net: no correctness bug on the happy path — the gate does what it says for `tools/call` / `resources/*` / `prompts/get` / `*/list`. Items 1 and 2 are about closing (or accurately scoping) the guarantee for a feature where the precise wording is the product.

---

eunolabs:
Thanks for the review — all four points addressed in bd46821.

**1. Retrospective guarantee (overstated wording).** Agreed, and I took the "align the docs" route rather than a blocking allow path. The threat model (§5.4.1), conformance, CHANGELOG, and the `AuditDegraded` / gate doc comments now state the accurate guarantee — *"after the first observed loss, every subsequent forward is denied"* — and explicitly call out the boundary call (and the in-flight-at-saturation window under concurrency), pointing operators who need zero-loss at strict mode **+** the real-time external sink.

**2. `sampling/createMessage` not gated.** Closed it rather than documenting an exemption — it's the fifth enforced method, so leaving it out contradicted the framing. The gate condition is now a shared `auditGateTripped` helper consumed by `enforcedForwardCore`, `dispatchList`, and the sampling allow leg of `forwardServerRequest`; an allowed sampling call under a degraded trail returns `AUDIT_UNAVAILABLE` to the upstream initiator instead of forwarding to the host. Added two `forwardServerRequest` tests (degraded-denies-to-upstream, healthy-forwards). The unenforced server-initiated passthroughs (`roots/list`, `elicitation/create`) remain ungated — no policy decision — and that scope is now stated in §5.4.1.

**3. Per-request stderr flood.** The strict-denial warning is now **one-shot** (healthy→degraded transition only), via a proxy-held `atomic.Bool` threaded through `forwardParams`/`serverRequestParams`; the durable per-request signal is the `AUDIT_UNAVAILABLE` audit record, as you suggested.

**4. Test doc-comment name.** Fixed.

On the `--require-audit strict` (space-separated) footgun: kept the `IsBoolFlag` ergonomics + the documented `=` requirement, since dropping `IsBoolFlag` would break the bare `--require-audit` form. Local `go test -race ./...`, `make lint`, and `make check-fmt check-license check-notice` are green.

---

internal/transport/forward.go
```diff
@@ -190,6 +282,19 @@ func forwardServerRequest(ctx context.Context, msg mcp.RPCMsg, fp serverRequestP
...
 	dec := pdp.DecideSampling(ctx, fp.pdp, fp.sessionID, fp.sourceIP)
 	if dec.Decision == capability.DecisionAllow {
+		// --require-audit=strict also gates this enforced method: once the audit
+		// trail has degraded, an allowed sampling request is failed closed —
+		// AUDIT_UNAVAILABLE back to the upstream initiator — rather than forwarded to
+		// the host unaudited, mirroring enforcedForwardCore for the other enforced
+		// methods. (Scope: sampling needs the system:sampling manifest opt-in, which
+		// is rejected at startup for HTTP upstreams, so this only bites stdio
+		// subprocess upstreams.)
+		if tripped, reason := auditGateTripped(fp.rec, fp.requireAuditStrict); tripped {
```
eunolabs: Follow-up on the sampling fix (the clean-allow leg is gated correctly now — good). One remaining asymmetry: this gate only covers the `DecisionAllow` branch. The **observe branch just below** — a would-be-denied sampling request downgraded to observe-and-forward under route-level audit mode (`observe := !pdp.IsKillSwitchDenial(...) && (fp.audit || dec.AuditOnly)`) — still calls `fp.forward(msg)` with no `auditGateTripped` check.

`enforcedForwardCore` gates *after* its observe downgrade, so for the four host-initiated methods an audit-mode observe forward **is** blocked when degraded; sampling's observe leg isn't, so the two diverge. The threat model now states strict mode "also overrides route-level audit/wiretap forwarding when the trail is degraded" and scopes the gate to "exactly the enforced decision paths plus `*/list`" — the observe-sampling leg is an enforced decision path that doesn't honor that.

Reachability is narrow (manifest-policed **stdio** route in audit mode + a would-be-denied `sampling/createMessage` + a degraded trail), so this is low priority — but for full consistency either run the same `auditGateTripped` check before the observe-mode `fp.forward(msg)`, or carve out the exception in the doc. Your call.

---

eunolabs (commented):
## Re-review of `bd46821` — all four prior points addressed ✅

Verified each fix against the head tree; CI is green (41/41 checks passing).

**1. Sampling gate (substantive) — fixed.** `forwardServerRequest`'s `DecisionAllow` leg now runs `auditGateTripped` and, when degraded, replies `AUDIT_UNAVAILABLE` to the upstream initiator, records a deny, and skips the host forward (`forward.go:292`). The condition is factored into a shared `auditGateTripped` helper consumed by `enforcedForwardCore`, `dispatchList`, and the sampling leg — nice DRY, and `fp.rec` is provably non-nil inside the `tripped` block (the helper returns false on a nil sink), so no nil-panic. Two new tests cover degraded-denies-to-upstream and healthy-forwards.

**2. Retrospective guarantee (substantive) — fixed.** The absolute "can never be applied with no durable record" wording is gone from CHANGELOG, conformance, threat model, and the `AuditDegraded` / `forwardParams` / `enforcedForwardCore` doc comments — all now state the accurate guarantee ("after the first observed loss, every subsequent forward is denied; the boundary call is still forwarded") and point at the real-time external sink for a zero-loss guarantee. The new "What the gate does and does not guarantee" paragraph in §5.4.1 is exactly right.

**3. stderr flood (minor) — fixed.** `warnStrictAuditOnce` uses a per-proxy `atomic.Bool` + `CompareAndSwap`, firing once on the healthy→degraded transition; the per-call durable signal is the `AUDIT_UNAVAILABLE` record. The atomic lives on `*HTTPProxy` / `*StdioProxy` (pointer-only, so copylocks-safe) and is threaded as `*atomic.Bool` through both `forwardParams` and `serverRequestParams`. Correct and race-free.

**4. Doc-comment nit — fixed.**

**One small follow-up (low priority, inline):** the sampling gate covers the clean-allow leg but not the audit-mode *observe* leg just below it, which diverges from `enforcedForwardCore` (gates after its observe downgrade) and from the doc's "overrides audit/wiretap forwarding" claim. Reachability is narrow (manifest-policed stdio route in audit mode + would-be-denied sampling + degraded trail) — gate it or note the exception. Details inline.

Nicely done overall — clean factoring, accurate docs, and good test coverage on the new path. Modulo the optional follow-up above, this looks ready.

---

internal/transport/forward.go
eunolabs: Gated it (rather than carving out a doc exception) — in 9b2d01d. The sampling gate is now a shared `serverRequestParams.strictSamplingDenial` helper called on **both** the clean-allow leg and the audit-mode observe leg, placed after the kill-switch hard-deny so a kill still surfaces `AUTHORIZATION_FAILED` rather than `AUDIT_UNAVAILABLE`. This matches `enforcedForwardCore` (which gates after its own observe downgrade) and the §5.4.1 claim that strict mode overrides audit/wiretap forwarding when degraded. Added a test for the observe-leg path (manifest-policed route, audit mode, would-be-denied sampling, degraded trail → fails closed to the upstream initiator, single `AUDIT_UNAVAILABLE` record, no host forward) and updated the §5.4.1 wording to name both legs. `go test -race ./internal/transport/…`, `make lint`, and the fmt/license/notice checks are green.

---

PR #1016, opened 2026-06-20
sec: fail-closed enforcement hardening (nil-PDP, list-filter, wiretap kill switch)

eunolabs:
## Fail-closed enforcement hardening (audit group 2/5)

Three fail-closed / enforcement-semantics fixes from the readiness audit.

### S1 — Transport constructors defaulted an omitted PDP to allow-all
`NewStdioProxy` and `newHTTPProxy` substituted `AlwaysAllowPDP{}` when the caller omitted a PDP — inverting the deny-by-default direction. A library caller that forgot to wire a PDP would forward every request verbatim.
- Add exported **`DenyAllPDP`** (full `PolicyDecisionPoint`): every `Decide*` denies with `AUTHORIZATION_FAILED`, every `Filter*List` drops every entry.
- Both constructors default to it. Production unaffected (`main.go` always sets `PDP`; gateway enforces per-route). `newTestRemoteProxy` keeps the plumbing-test passthrough via explicit `AlwaysAllowPDP{}`.

### N1 — `filterListResult` returned original bytes on an empty result
The `len == 0` early return forwarded the (empty) upstream bytes instead of failing closed. Now returns `{"field":[]}` uniformly.

### S2 — Kill switch did not stop policyless wiretap routes
A `enforcement: audit` route with no policy kept forwarding after `/control/kill` unless JWT auth wrapped it — the emergency stop didn't stop a wiretap route (surfaced only as a JSON field operators could miss).
- `AlwaysAllowPDP` gains an optional `killswitch.Checker` (`NewAlwaysAllowPDP`); its `Decide*`/`CheckKill` now consult it. A global or session-targeted kill hard-blocks even a policyless route, matching audit-mode-with-policy (which already hard-blocked kills). A zero-value `AlwaysAllowPDP{}` stays pure-passthrough.
- Removed the now-obsolete partial-coverage scaffolding: `WarnWiretapKillGap`/`anyWiretapRoute`, the startup NOTICE, the **`/control/kill` `wiretap_routes_unaffected` response field**, and the kill-CLI stderr warning.
- **Migration (CHANGELOG):** `/control/kill` no longer returns `wiretap_routes_unaffected`. Threat-model + client-integration docs updated.

### Verification
`go test -race ./internal/pdp/... ./internal/transport/... ./cmd/...` all pass; `make check-fmt` / `check-license` clean; `golangci-lint` 0 issues. New tests: `deny_all_test.go`, `always_allow_kill_test.go`, repurposed kill response-shape tests, and an end-to-end `BuildRoutes` wiretap-kill test.


internal/pdp/pdp.go
```diff
@@ -265,6 +265,70 @@ func (p AlwaysAllowPDP) FilterPromptsList(_ context.Context, result json.RawMess
...
+// response and filters every */list result down to an empty list. It is the
+// fail-closed default the transport constructors substitute when no PDP is wired,
+// so a caller that forgets to set a PDP denies every request instead of forwarding
+// it verbatim — the deny-by-default direction a policy-enforcement proxy must take.
+//
+// The shipped binary always wires a concrete PDP (ManifestPDP, JWTPDP, or an
+// explicitly opted-in AlwaysAllowPDP, and gateway mode enforces per-route), so this
+// is a defense-in-depth backstop for the exported library seam, never a runtime
+// enforcement path. Unlike AlwaysAllowPDP it implements every contract method as a
+// deny, so no method silently falls through to "forward verbatim".
```
eunolabs, line 282: Optional: add a co-located compile-time assertion next to this type, e.g.

```go
var _ PolicyDecisionPoint = DenyAllPDP{}
```

Right now the only thing guaranteeing `DenyAllPDP` satisfies the *full* `PolicyDecisionPoint` contract is the `opts.PDP = pdp.DenyAllPDP{}` assignment over in `internal/transport`. Since this type's entire reason for existing is to be a complete fail-closed implementation, an assertion in this package documents that intent and keeps a future transport refactor from silently dropping the only check. (None of the existing PDPs have one, so this is a nudge, not a blocker.)

---

eunolabs (commented):
## Review — fail-closed PDP default + empty-list hardening

Overall this is tight, well-scoped defense-in-depth and I'd ship it. CI is green (Test, Lint, Enforcement compliance, gofmt, License Headers, NOTICE, govulncheck all passing; the only non-success is `Analyze (go)` CodeQL still in progress — unrelated to this change).

### What it does
- Adds `DenyAllPDP`, a complete fail-closed `PolicyDecisionPoint`: every `Decide*` returns a structured `AUTHORIZATION_FAILED` deny, every `Filter*List` drops all entries, `CheckKill` is a no-op.
- `NewStdioProxy` / `newHTTPProxy` now default an omitted PDP to `DenyAllPDP` instead of `AlwaysAllowPDP` — the deny-by-default direction a policy-enforcement proxy must take.
- `filterListResult` now fails closed to `{"<field>":[]}` on an empty result instead of returning the (empty) upstream bytes, so every non-happy path through the helper is uniform.

### Correctness — verified against the tree
- **Contract is complete.** `DenyAllPDP` implements `Decide`, `DecideResourceRead`, `DecidePromptGet`, `DecideSampling`, `CheckKill`, and the three `Filter*List` methods — the full `PolicyDecisionPoint` surface. The `opts.PDP = pdp.DenyAllPDP{}` assignments are the de-facto compile-time interface check (see inline note suggesting a co-located assertion as well).
- **The empty-result change is safe for `ManifestPDP`/`JWTPDP`.** The only transport call site (`internal/transport/dispatch.go:199`) guards with `upResp.Result != nil` before filtering, so an upstream *error* response (no `result`) never reaches this helper and won't be rewritten into a success `{"tools":[]}`. A non-nil-but-zero-length `result` isn't reachable through normal JSON unmarshaling either, so in practice this branch is exercised only by `DenyAllPDP` and the new test — where returning an empty list is strictly safer than forwarding empty bytes. Good change.
- **Ordering is correct.** `failClosed` is now declared before the `len(resultBytes) == 0` early return that calls it.
- **Gateway mode is untouched.** `NewHTTPProxyGateway` constructs `HTTPProxy` directly and never routes through `newHTTPProxy`, so the new default only affects single-upstream construction — consistent with the inline comment ("gateway mode … enforces per-route").
- **Production binary unaffected.** `main.go` always wires a concrete PDP, and these are `internal/` packages (not importable outside the module), so this is purely a backstop on the in-module seam, exactly as the PR describes.

### Conventions
- License headers present, no emojis, `sec:` prefix appropriate, every deny carries a structured `Code` + message (`AUTHORIZATION_FAILED` / "no policy configured: denying by default"). Matches the fail-closed invariant.
- Tests cover deny-on-every-method, `CheckKill`, and per-`*/list` filtering including the empty-input case. `newTestRemoteProxy` / `TestNewStdioProxy_Defaults` are updated coherently.

### Minor / optional (none blocking)
1. Co-located `var _ PolicyDecisionPoint = DenyAllPDP{}` assertion — see inline comment.
2. `TestDenyAllPDP_DeniesEveryMethod` asserts the denial `Code` but not the `Message`; if you want the "no policy configured" wording pinned, it's a one-line `assert`. Optional.

Nice, surgical hardening of the deny-by-default seam.

---

internal/pdp/pdp.go
eunolabs: Addressed in 6788fdd — added `var _ PolicyDecisionPoint = DenyAllPDP{}` next to the type. Builds and `go test ./internal/pdp/...` pass, gofmt clean.

---

PR #1020, opened 2026-06-20
fix: bound stdio→HTTP bridge goroutines and cap SSE subscribers

eunolabs:
## Transport resource-exhaustion bounds (audit group 3/5)

Two unbounded-resource paths on the HTTP/bridge transports.

### S3 — Unbounded fire-and-forget POST goroutines on the stdio→remote-HTTP bridge
`httpUpstream.Write` and `postWithCtx` (`stdio_http_upstream.go`) spawned `go h.post(...)` per host request/notification with no concurrency limit. A pipelining host could spawn an unbounded number of in-flight POSTs, each holding a request body and an upstream connection (`incoming` is only buffered at 64).
- Add a `maxInflightPosts` (64) semaphore via `spawnPost`: acquiring blocks for a slot (backpressure) once the cap is reached, and drops on `close`. The releasing `post()` runs on an independent goroutine drained by `readUpstream`, so a blocked spawn never deadlocks against its own release.

### N6 — SSE subscriber count per session was uncapped
`handleMCPGet` appended a subscriber to the session unconditionally, so a client could open an unbounded number of GET streams on one session, each with a 16-slot buffer (memory bounded only indirectly by `maxSessions`).
- Cap at `maxSubsPerSession` (8). `addSub` now returns `false` at the cap, and the handler returns `429 Too Many Requests` **before** writing the SSE 200 headers (so an error status is still possible).

### Verification
`go test -race ./internal/transport/...` passes. New tests: `TestHTTPUpstream_BoundsInflightPosts` (peak in-flight ≤ cap under 2× load) and `TestHTTPSession_AddSubCapsSubscribers`. `make check-fmt` / `check-license` clean; `golangci-lint` 0 issues.

---


internal/transport/stdio_http_upstream.go
```diff
@@ -85,15 +100,37 @@ func newHTTPUpstream(parent context.Context, baseURL, authHeader string, tlsSkip
...
 }
 
+// spawnPost runs h.post under the in-flight semaphore, bounding the number of
+// concurrent POST goroutines to maxInflightPosts. When the cap is reached the
+// caller blocks until a slot frees (backpressure) or the bridge closes — at which
+// point the message is dropped, matching the fire-and-forget contract of Write and
+// postWithCtx. The slot is released when post() returns. The releasing post()
+// runs on an independent goroutine and completes as the proxy's readUpstream loop
+// drains `incoming`, so a blocked spawn never deadlocks against its own release.
+func (h *httpUpstream) spawnPost(ctx context.Context, msg mcp.RPCMsg) { //nolint:gocritic // hugeParam: rpcMsg matches the sink signature
```
eunolabs, line 115: The semaphore + escape-hatch design is correct and the no-deadlock reasoning holds — I traced it through: `spawnPost` holds no locks while blocked on the acquire, the releasing `post()` runs on an independent goroutine that `readUpstream` keeps draining via `incoming`, and the `<-ctx.Done()`/`<-h.done` cases guarantee shutdown and per-call cancellation still unblock. The new silent-drop path (returning before `post()` runs when the slot can't be acquired) is also safe for the request path: `callUpstream` → `awaitNonced` already selects on `ctx.Done()` and `upstreamDone`, so a dropped POST never leaves the caller hung. Nice touch matching `maxInflightPosts` to the `incoming` buffer (64) so the two bounds don't fight.

Two things worth tightening, neither blocking:

1. **`Write`/`postWithCtx` are no longer non-blocking, but their doc comments still say so.** `Write` (just below) still reads "fire-and-forget ... where no per-call timeout applies"; it can now block on the semaphore under backpressure. A one-liner on `Write` and `postWithCtx` noting they may block until a slot frees (or the bridge closes) would save the next reader from assuming the old contract.

2. **The host read loop can now block here — safe today only by a property documented in another file.** `serveHost` forwards host→upstream notifications via `p.upWriter.Write` directly on the read-loop goroutine (`stdio.go:453`), which on this bridge now routes through the blocking `spawnPost`. The big warning at `stdio.go:458-464` says the read loop must not block because it's also the only path that routes host replies to server-initiated requests back upstream. That's not a deadlock *here* purely because this bridge never delivers server-initiated requests (the KNOWN LIMITATION at the top of this file → `incoming` only ever carries responses → `serverReqs` stays empty → the `stdio.go:479` reply path never fires). That safety is load-bearing and non-obvious. Worth a sentence in the commit/PR (or a brief code note) tying the two together, so a future change that adds upstream-SSE support to this bridge doesn't silently reintroduce the deadlock that comment warns about.

---

internal/transport/http_routing.go
```diff
@@ -264,6 +264,17 @@ func (p *HTTPProxy) handleMCPGet(w http.ResponseWriter, r *http.Request, route *
 		return
 	}
 
+	// Register the SSE subscriber BEFORE writing the 200 + SSE headers, so a
+	// session already at maxSubsPerSession is rejected with an error status
+	// (rather than a 200 stream that delivers nothing). This bounds how many
+	// concurrent GET streams one session can hold open.
+	ch := make(chan mcp.RPCMsg, 16)
+	if !sess.addSub(ch) {
```
eunolabs, line 272: Ordering is correct: `addSub` now runs before any `WriteHeader(200)`, so the cap path can legitimately return `429`. The discarded `ch` (buffer 16) on the reject path is just GC'd — no leak — and `defer removeSub(ch)` frees the slot on disconnect (`r.Context().Done()`), so a flapping client recovers its slot. Good.

Minor: the comment says it bounds "how many concurrent GET streams one session can hold open," which is exactly right, but note the cap is also what `deliverToOne` iterates over for server-initiated requests — capping `notifSubs` at 8 also bounds that scan. No change needed, just confirming the cap is the right knob.

---

eunolabs (commented):
## Review: bound stdio→HTTP bridge goroutines and cap SSE subscribers

Both bounds are correct, fail-safe, and well-tested. I traced the concurrency carefully — no deadlock, no caller hangs, no leaks. Net assessment: **looks good to merge**, with a couple of non-blocking doc/coupling nits left inline.

### S3 — `maxInflightPosts` semaphore on the bridge
- The acquire is decoupled from the release (release runs in the spawned goroutine's `defer`), so a blocked acquire waits on *other* goroutines' releases, never its own — the no-self-deadlock claim checks out.
- `<-ctx.Done()` / `<-h.done` escapes mean shutdown and per-call timeouts still unblock a stuck acquire, and the resulting silent drop is safe: `awaitNonced` (`callUpstream`) already selects on `ctx.Done()` + `upstreamDone`, so a dropped request POST returns an error to the host rather than hanging.
- The full-`incoming` backpressure chain (host stops reading stdout → `readUpstream` stalls → `post()` blocks on `incoming` → semaphore fills → `Write` blocks) is correct end-to-end flow control, not a deadlock — and `close()` unblocks every link. Matching the cap (64) to the `incoming` buffer (64) is the right call so the first wave never blocks on `incoming`.
- One coupling worth documenting: `Write` can now block on the read-loop goroutine, which is only deadlock-free because this bridge never delivers server-initiated requests. Details inline on `spawnPost`.

### N6 — `maxSubsPerSession` cap
- `addSub` returning `false` at the cap with the handler emitting `429` **before** the `200`/SSE headers is the correct shape (an error status is still possible). Slot is freed on disconnect via `defer removeSub`. `addSub` cap + `removeSub` free-slot are both covered by the new test.
- `addSub` now takes the lock with `defer Unlock` and checks `len(notifSubs) >= maxSubsPerSession` under it — race-free, consistent with `deliverToOne`/`removeSub`.

### Tests
`TestHTTPUpstream_BoundsInflightPosts` (peak in-flight ≤ cap under 2× load, with a 50ms grace window for an over-cap goroutine to slip through) and `TestHTTPSession_AddSubCapsSubscribers` (cap enforced + slot reclaimed after `removeSub`) both target the right invariants.

### One question
Per the repo convention that behavior changes ship with a docs update — does this group warrant a short note in `docs/threat-model-mcp.md` on the new transport resource-exhaustion bounds (the `429` and the bridge backpressure)? No new audit field, so the threat-model field invariant isn't triggered, but a mitigation note seems in scope for a security-hardening change. Your call.

---

PR #1021, opened 2026-06-20
docs: fix audit-trust citations, redactFields framing, and document missing flags

eunolabs:
## Docs accuracy (audit group 4/5)

Self-contained documentation-accuracy fixes (one `.go` change is comment-only).

### S6 — Audit code cited threat-model sections that don't exist in the eunox TM
`internal/audit/audit.go` used the **spec repo's** section numbering:
- `agent_id`/`task_id` cited `§5.8` (no such section in the eunox TM) → now **§2.1** ("What the proxy trusts after verification", which explicitly lists `mcp.agent_id`/`mcp.task_id`).
- `key_id` rotation cited `§7.4` (which is "Reporting security issues") → now **§3.4** ("Audit Log Tampering"). The threat-model audit-shape table mirrored the bad `§5.8`; fixed.

### S7 — `redactFields` framed as a future "v0.5" change
The manifest guide said `redactFields` "reclassifies … in v0.5" and is rejected "once directive enforcement ships." It **already is** a directive and `redactFields`-in-`conditions` is **already rejected at load**. Rewrote both passages to present tense.

### N2 — Catalog-disclosure callout
Added an explicit note that route-level audit mode returns the **full, unfiltered** `*/list` even with a restrictive manifest (observe-everything), so audit mode is not a way to hide the catalog (kill switch still blocks enumeration).

### S8 — Document missing flags
README now documents `--max-call-counter-keys`, `--killswitch-reconcile-interval`, `--shutdown-timeout`, `--jwt-leeway`, the OAuth/JWKS flags, and the `stats`/`suggest`/`audit-verify` `--since`/`--request-id`/`--session-id` filters. Added the `suggest` subcommand to `CLAUDE.md`.

### N9 — Circular citation
Dropped the `request_id` row's `(§6)` self-citation.

Build + `check-fmt` + `check-license` clean.

---


README.md
```diff
@@ -465,6 +468,11 @@ session shapes MCP servers actually expose.
 | `eunox-mcp audit-verify`                                              | Verify HMAC-SHA256 signatures in the audit log               |
 | `eunox-mcp doctor [--config <eunox.yaml>] [--live]`                    | Print a redacted support bundle (binary identity, config, manifests, audit tail) for bug reports. Nothing is uploaded. |
 
+The audit-log readers — `stats`, `suggest`, and `audit-verify` — accept filters to
+scope the records they read: `--since <RFC3339|duration>`, `--request-id <id>`, and
+`--session-id <id>`. (`audit-verify` always runs the full HMAC chain check; the
+filters only narrow which records are reported.)
```
eunolabs: This new paragraph doesn't match the code (`cmd/eunox-mcp/main.go`), and since the PR's whole purpose is doc accuracy it's worth getting exactly right:

- **`stats` accepts none of these filters.** Its flag set (`main.go:1889`) registers only `--config` and `--audit-log`.
- **`suggest` accepts none of these filters.** Its flag set (`main.go:1569`) registers only `--audit-log`, `--name`, `--output`, `--max-values`.
- **`--session-id` is not a reader filter at all.** The only `session-id` flag is on `proxy` (`main.go:321`, "Session ID to use"). No reader registers it.
- **`--since` takes an RFC3339 timestamp only, not a duration.** The `audit-verify` flag help (`main.go:1799`) says "records after this RFC3339 timestamp", so `<RFC3339|duration>` overstates it.

In practice only **`audit-verify`** accepts any of these, and only `--request-id` + `--since` (`main.go:1798-1799`) — not `--session-id`.

Two ways to fix:
1. Narrow the docs to what exists, e.g.: "`audit-verify` accepts `--request-id <id>` and `--since <RFC3339>` to scope which records it reports; the full HMAC chain is always verified regardless."
2. Or, if the intent was for `stats`/`suggest` and the `--session-id` filter to exist, that's a code change, not a docs change — out of scope for this PR.

---

eunolabs (commented):
## Review — docs accuracy

I verified each claim against the code/headings at the merge base. Most of the PR is solid; one new README paragraph is inaccurate (inline comment).

### Verified correct
- **S6** — `§5.8` genuinely doesn't exist as a section in this repo's threat model (the only matches were the citations being fixed), and `§2.1 "What the proxy trusts after verification"` does list `mcp.agent_id`/`mcp.task_id`. `key_id`'s old `§7.4` pointed at "Reporting security issues"; `§3.4 "Audit Log Tampering"` is the right target. Table row + `internal/audit/audit.go` comments are consistent. ✅
- **S7** — `redactFields` is already a directive and `redactFields`-in-`conditions` is already rejected at load (`pkg/capability/conditions.go:195`), so dropping the "v0.5 / once enforcement ships" framing for present tense is accurate. ✅
- **N2** — The catalog-disclosure callout matches the code: `internal/transport/dispatch.go:199` bypasses the manifest filter when the route is in audit mode ("only the manifest filter is bypassed"), the kill-switch check above it stays unconditional, and per-entry `AuditOnly` "never applies to list methods" — so an otherwise-enforcing route still filters `*/list`. ✅
- **N9** — The audit-record table lives under `§6 Data Classification and Handling`, so the `request_id` row's `(§6)` was a self-citation; dropping it is correct. ✅
- **S8 flag defaults** — all check out: `--max-call-counter-keys` `defaultMaxCallCounterKeys = 1_000_000` (`main.go:176`), `--shutdown-timeout` 5000ms = 5s, `--killswitch-reconcile-interval` effective 30s, `--jwt-leeway` `DefaultJWTLeeway` (10s), `--oauth-authorization-server` defaults to `--jwt-issuer`, `--oauth-resource` unset. The new JWT/OAuth table matches the flag help text. `suggest` exists and is correctly added to `CLAUDE.md`. ✅

### Needs a fix
- The new **"audit-log readers — `stats`, `suggest`, and `audit-verify` — accept filters …"** paragraph in `README.md` is wrong on several counts (see inline). `stats` and `suggest` accept none of these filters, `--session-id` isn't a reader filter anywhere, and `--since` is RFC3339-only (not a duration). Only `audit-verify` takes `--request-id`/`--since`. Worth correcting since the goal here is accuracy.

Otherwise this is a clean, well-targeted accuracy pass. Nothing else blocking.

---

README.md
eunolabs: Fixed in 95ae977 (option 1 — narrow the docs to what exists). The paragraph now reads:

> `eunox-mcp audit-verify` accepts `--request-id <id>` and `--since <RFC3339>` to scope which records it reports. The full HMAC chain is always verified regardless — these filters only narrow which records are counted and printed, not what is checked.

Dropped the `stats`/`suggest` claims, the nonexistent `--session-id` reader filter, and the `|duration` on `--since`.

---

PR #1022, opened 2026-06-20
sec: kill switch hard-blocks policyless wiretap routes

eunolabs:
## Kill switch stops wiretap routes (audit group 2b/5)

> This is the **S2** change from the fail-closed batch. PR #1016 was merged before this commit landed, orphaning it on the branch — this PR re-targets it cleanly onto `main`. (S1/N1 from that batch are already in `main`.)

### S2 — Kill switch did not stop policyless wiretap routes
A route running in `enforcement: audit` with no policy (wiretap: forward-and-log) kept forwarding after `/control/kill` unless JWT auth wrapped it — an operator's emergency stop did **not** stop a wiretap route (surfaced only as a JSON field operators could miss). Per maintainer decision, the stronger fix:

- `AlwaysAllowPDP` gains an optional `killswitch.Checker` (`NewAlwaysAllowPDP`); its `Decide*` and `CheckKill` now consult it. A global or session-targeted kill **hard-blocks even a policyless route**, matching audit-mode-with-policy (which already hard-blocked kill-switch denials). A zero-value `AlwaysAllowPDP{}` (no checker) stays pure-passthrough for tests.
- `route.go` wires the shared kill switch into the wiretap PDP (`LoadUpstreamPDP`).
- Removed the now-obsolete partial-coverage scaffolding: `WarnWiretapKillGap` / `anyWiretapRoute`, the startup NOTICE, the **`/control/kill` `wiretap_routes_unaffected` response field**, and the kill-CLI stderr warning.

### Migration (in CHANGELOG)
The `/control/kill` JSON response no longer carries `wiretap_routes_unaffected`. Tooling that parsed it should drop it. Threat-model + client-integration docs updated.

### Verification
`go test -race ./internal/pdp/... ./internal/transport/... ./cmd/...` pass; `make check-fmt` clean; `golangci-lint` 0 issues. New tests: `always_allow_kill_test.go` (allow→deny after kill, bare PDP never blocks), an end-to-end `BuildRoutes` wiretap-kill test, and repurposed kill response-shape tests.

---


internal/pdp/pdp.go
```diff
@@ -230,19 +255,21 @@ func (p AlwaysAllowPDP) DecidePromptGet(_ context.Context, _, _, _ string) capab
 // --audit — the silent break the pre-fold deny-by-default (missing-method)
 // fallback caused in a mode whose whole purpose is "observe everything, block
 // nothing except kill-switch". Like the other AlwaysAllowPDP methods this carries
-// no kill-switch check — a pure wiretap route has no policy or kill manager wired
-// to it; the kill switch is a manifest/route concern (ManifestPDP.DecideSampling)
-// that this passthrough mode does not participate in.
-func (p AlwaysAllowPDP) DecideSampling(_ context.Context, _, _ string) capability.EnforceResponse {
+// the kill switch (when wired via NewAlwaysAllowPDP) still hard-blocks sampling so
```
eunolabs: Nit (comment grammar): the rewritten sentence reads "...Like the other AlwaysAllowPDP methods this carries / the kill switch (when wired via NewAlwaysAllowPDP) still hard-blocks sampling so...". The verb `carries` is now dangling — the prior line's "this carries" no longer connects to anything. Given how precise the comments are elsewhere in this file, suggest tightening to something like:

> Like the other AlwaysAllowPDP methods, this consults the kill switch: when one is wired (NewAlwaysAllowPDP) it hard-blocks sampling so a killed session cannot keep driving server-initiated sampling; a bare AlwaysAllowPDP{} (nil ks) still allows.

Behavior is correct — kill-switch denials carry `ErrCodeKillSwitch`, so `forward.go`'s `!IsKillSwitchDenial(...) && fp.audit` gate hard-blocks even the sampling observe leg. Only the prose needs a fix.

---

internal/transport/route.go
```diff
@@ -229,7 +228,10 @@ func LoadUpstreamPDP(u *config.UpstreamConfig, counter capability.CallCounter, k
 		if u.ExpectVersion != "" {
 			return nil, nil, "", "", fmt.Errorf("upstream %q: expectVersion %q set but no policy is configured", u.Name, u.ExpectVersion)
 		}
-		return pdp.AlwaysAllowPDP{}, nil, "", "", nil
+		// Wire the shared kill switch into the wiretap PDP so an operator's
+		// /control/kill halts even this policyless route — it forwards everything
+		// else, but an emergency stop must stop it too.
+		return pdp.NewAlwaysAllowPDP(ks), nil, "", "", nil
```
eunolabs, line 234: Worth calling out (not a defect): because `LoadUpstreamPDP` is the shared PDP-construction seam, this change also reaches the **stdio single-upstream `--audit` path** — `serveStdioHost` builds its wiretap PDP through this same function (`cmd/eunox-mcp/main.go:1017`), not just the gateway routes built by `BuildRoutes`. So with a Redis-backed kill switch, an `eunox-mcp kill <session>` would now hard-block a stdio wiretap session too.

That's arguably the more correct behavior and a nice side effect of wiring it here. But the CHANGELOG / threat-model / client-integration docs all frame this strictly as gateway `/control/kill` + routes. Could you confirm the stdio broadening is intended, and consider a one-line mention in the docs so the documented surface matches the actual one? (Conversely, if `ks` is always nil on the stdio path without `--redis-addr`, a note that it's a no-op there would resolve the ambiguity.)

---

eunolabs (commented):
## Review: `sec: kill switch hard-blocks policyless wiretap routes`

Read through the full diff and traced the change end-to-end against the codebase. **The fix is correct, minimal, and well-targeted.** A couple of small things below; nothing blocking.

### What it does
Wires the shared kill switch into the wiretap decision point (`AlwaysAllowPDP` gains an optional `killswitch.Checker` via `NewAlwaysAllowPDP`), so a global or per-session `/control/kill` now hard-blocks even a policyless `enforcement: audit` route — closing the gap where an operator's emergency stop silently kept forwarding. It then removes the now-obsolete partial-coverage scaffolding (`WarnWiretapKillGap`, `anyWiretapRoute`, the startup NOTICE, the `/control/kill` `wiretap_routes_unaffected` field, and the kill-CLI stderr warning).

### Why it's correct
- **End-to-end leg holds.** A wiretap route runs with `r.audit = true`, but `AlwaysAllowPDP.CheckKill` now returns a denial carrying `ErrCodeKillSwitch`. `forward.go`'s observe gate is `observe := !pdp.IsKillSwitchDenial(denial) && (fp.audit || dec.AuditOnly)`, so a kill-switch denial is *not* downgraded and the route hard-blocks (upstream never contacted). The fix elegantly reuses the gate already in place for the `ManifestPDP` audit-with-policy path rather than adding a new one.
- **Both transports covered by one change.** Gateway routes (`BuildRoutes`) and the stdio single-upstream host (`serveStdioHost`) both construct their PDP through `LoadUpstreamPDP`, so the single edit at the no-policy branch covers both. (See the inline note on the stdio side effect.)
- **Pure-passthrough preserved.** `NewAlwaysAllowPDP(nil)` / the zero-value `AlwaysAllowPDP{}` keep `ks == nil`, and `killCheck` returns `nil` for a nil checker — so the many test call sites using `AlwaysAllowPDP{}` are unaffected.
- **Types line up.** `killswitch.Manager` embeds `Checker`, so passing the route's `Manager` into `NewAlwaysAllowPDP(killswitch.Checker)` is fine.
- **Breaking change handled per repo conventions.** Dropping `wiretap_routes_unaffected` is a clean pre-1.0 grammar change with the migration documented in CHANGELOG + threat-model + client-integration docs (no back-compat shim, as required).
- **CI is green** (Test, Lint, gofmt, all platform builds succeed; `mergeable_state: unstable` is only because CodeQL `Analyze (go)` was still running).

### Tests
Good coverage at the PDP boundary: `always_allow_kill_test.go` (allow→deny after kill across all four `Decide*` methods + `CheckKill`; bare PDP never blocks) and the `BuildRoutes`/`LoadUpstreamPDP` e2e test. The repurposed kill-response-shape tests correctly assert the field is now absent.

**Suggestion (non-blocking):** consider one transport-level test that drives a killed *wiretap* route through `enforcedForwardCore` and asserts the host receives the `KILL_SWITCH` error and the upstream is never contacted. The current tests stop at the PDP decision; the user-visible guarantee in the PR title depends on the `forward.go` observe gate, which is only exercised for the ManifestPDP audit path today.

### Minor
- One dangling-verb comment in `DecideSampling` (inline).
- A docs-scope observation about the stdio path (inline).

Nice, surgical change — it deletes more than it adds and removes a footgun (an "ok" that didn't mean "stopped").

---

PR #1023, opened 2026-06-20
fix: re-validate merged manifests, anchor schema pattern, add config tests

eunolabs:
## Config & robustness polish (audit group 5/5)

The remaining NICE-TO-HAVE items. Two audit items in this group are intentionally **no-ops** (documented below) because acting on them would degrade a deliberately-engineered design.

### N4 — Re-validate after `MergeManifests`
`MergeManifests` now re-validates the merged result and returns `(*LocalManifest, error)`. Each input is already validated by `LoadManifest` and merge only unions valid capabilities, so this never fails in the shipped binary — it's defense-in-depth against a future merge change producing an invalid manifest. The four callers (`validate`, `validate --live`, `doctor`, transport `LoadUpstreamPDP`) handle the new error. CHANGELOG migration note added.

### N3 — Anchor the `expectVersion` schema pattern
`schemas/eunox-gateway-config.schema.json` pattern `^[0-9]+\.[0-9]+\.[0-9]+` → adds trailing `$`, so `1.2.3-rc` is rejected at author time (it already failed closed at runtime — a malformed pin matches no valid manifest version).

### N5 — Co-locate `gateway_config_test.go`
Adds `internal/config/gateway_config_test.go` testing `LoadGatewayConfig` (strict decode + unknown-key rejection + multi-doc rejection) and `expandEnvRefs` (set→expanded, unset→literal) directly in-package, matching the package's standalone-testability goal.

### N7 — Audit `Close()` drop accounting: intentionally NOT changed
The audit agent suggested folding the `dropped` counter into `Close()`'s error (it currently folds only `writeFailures`). **I did not make this change**: the audit subsystem **deliberately** distinguishes `dropped` ("queue can't keep up" on a healthy file — recorded via an in-chain tamper-evident `AUDIT_RECORDS_DROPPED` marker + counter + stderr + `eunox_audit_dropped_records_total` + the `--require-audit=strict` gate) from `writeFailures` ("the file is broken" — fails `Close()`). Tests pin this (`sink.dropped.Store(n)` then `Close()` must return nil). Folding `dropped` into the error would reverse that design and break those tests. The shutdown-window residual-drop gap is inherent to an async audit path and is already documented + observable.

### S5 — Version cut: a release-time action, not faked here
The CHANGELOG `[Unreleased]` section is the correct state for unreleased work; `VERSION=0.1.0` is the next version. Cutting `[Unreleased]` → `[0.1.0] - <date>` and `git tag v0.1.0` is the maintainer's release step (dating an unreleased section now would be a false record). Flagged in the final summary as the one remaining release action.

### Verification
`go test -race ./internal/config/... ./cmd/... ./internal/transport/...` pass; schema tests pass; `make check-fmt` / `check-license` clean; `golangci-lint` 0 issues.

---


schemas/eunox-gateway-config.schema.json
```diff
@@ -191,7 +191,7 @@
         "expectVersion": {
           "type": "string",
           "description": "Pin the manifest's `version`. When set and the loaded version differs, the gateway fails closed (refuses to start). Requires `policy` to be set, and is only supported with a single policy file (the pin is unambiguous only then).",
-          "pattern": "^[0-9]+\\.[0-9]+\\.[0-9]+"
+          "pattern": "^[0-9]+\\.[0-9]+\\.[0-9]+$"
```
eunolabs: Anchoring with `$` is the right call. Two small, non-blocking notes:

1. **Pattern still diverges from the canonical semver regex.** `semverRe` in `internal/config/manifest.go` is `^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)$` (rejects leading zeros), whereas this accepts `01.2.3`. Not a bug — `expectVersion` must equal the manifest `version` (which forbids leading zeros), so a leading-zero pin fails closed at runtime — but if the intent is "schema mirrors what the runtime accepts," reusing the stricter pattern here would close the last gap. Optional.

2. **No value-level test for this pattern.** Per CONTRIBUTING ("Schema changes update `schemas/` plus a roundtrip test"), I'd expect a case asserting `1.2.3` passes and `1.2.3-rc` is rejected. The existing schema tests (`TestGatewaySchema_*`) are structural only (well-formed / struct-parity / `additionalProperties`), and the Go loader doesn't run JSON-Schema value validation — so a real test would need a json-schema dependency, which cuts against the one-static-binary rule. Probably fine to skip, but worth a sentence in the PR explaining why no value test accompanies an author-time validation change.

---

eunolabs (commented):
## Review — config & robustness polish (group 5/5)

Clean, low-risk PR. I verified each claim against the tree rather than taking the description at face value; everything holds up.

### Overview
- `config.MergeManifests` → `(*LocalManifest, error)`, re-validating the merged result (N4).
- Anchors the `expectVersion` schema pattern with `$` (N3).
- Adds in-package `internal/config/gateway_config_test.go` (N5).
- Documents N7/S5 as deliberate no-ops.

### Correctness — verified ✅
- **"Re-validation never fails today" is accurate.** `validateLocalManifest` has *no* cross-capability uniqueness/conflict check — it validates each capability independently plus top-level `name`/`version`. `MergeManifests` only unions already-valid capabilities and copies `ms[0]`'s already-valid metadata, so the merged result is valid by construction. The defense-in-depth framing is correct: this guards a *future* merge change, not a current gap.
- **All callers updated.** Confirmed exactly four production call sites (`main.go` validate + `validateConfigRoutes`, `doctor.go`, `route.go` `LoadUpstreamPDP`) and the three `manifest_test.go` cases — that's every `MergeManifests` reference in the tree, so the signature change compiles cleanly. Error handling is idiomatic at each site (the two `validate` paths set `worst >= 2` / exit non-zero and `continue`; `LoadUpstreamPDP` wraps with the upstream name).
- **Test fixture bump `"1"` → `"1.0.0"` is necessary and telling** — it's direct evidence the new re-validation is live (the old non-semver `Version` would now be rejected by `semverRe`).

### Tests
- `TestExpandEnvRefs` is a genuine assertion of the fail-closed contract (unset ref → literal preserved, both `$VAR` and `${VAR}` forms), matching the `os.LookupEnv` impl — not a tautology. The unknown-key and multi-doc rejection tests both assert the error *names* the offending key / mentions multi-doc, which is the right level of specificity.
- One gap (see inline): the N3 schema-pattern change has no value-level roundtrip test, which CONTRIBUTING nominally asks for on schema changes. Likely acceptable given the loader doesn't do JSON-Schema validation, but worth a note in the PR.

### Design no-ops (N7 / S5)
Both judgment calls are sound. N7: folding `dropped` into `Close()`'s error would collapse the deliberate `dropped` (queue pressure, observable via marker/counter/metric/`--require-audit=strict`) vs `writeFailures` (broken file → fail `Close()`) distinction — correctly left alone. S5: dating `[Unreleased]` now would be a false record; cutting the version is a maintainer release step. Agreed on both.

### Risk: low
No hot-path impact (merge runs at startup/validate/doctor only). The one nit worth your call is the schema-pattern consistency/test point inline. Nice, well-justified cleanup.

---

schemas/eunox-gateway-config.schema.json
eunolabs: Both addressed in 3461cba:

1. **Pattern now mirrors the runtime semver.** Changed to `^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)$` — same form as `semverRe` in `internal/config/manifest.go`, so leading zeros (`01.2.3`) are now rejected at author time too.
2. **Added the value-level test** (`TestGatewaySchema_ExpectVersionPattern` in `cmd/eunox-mcp/main_test.go`): it compiles the published pattern and asserts `0.1.0`/`1.2.3`/`10.20.30` match while `1.2.3-rc`, `1.2.3+build`, `1.2`, `1.2.3.4`, `v1.2.3`, leading zeros, and surrounding whitespace are rejected. It exercises the pattern directly (stdlib `regexp`) rather than pulling in a json-schema validator, so no new dependency.

`go test -race ./cmd/eunox-mcp/... ./internal/config/...`, `make check-fmt`, and `make check-license` all pass.

---

Issue #1019, opened 2026-06-20
[rev-20260620T2320Z-24375b] callcounter window-boundary semantics differ between Redis and in-memory `Peek`


eunolabs:
**Location:** `pkg/callcounter/redis.go:74-84`, `pkg/callcounter/memory.go:86-103`

**Reference:**
- https://github.com/eunolabs/eunox/blob/194877e40276f30c8c58b2e721987a83c2ac4a99/pkg/callcounter/redis.go#L74-L84
- https://github.com/eunolabs/eunox/blob/194877e40276f30c8c58b2e721987a83c2ac4a99/pkg/callcounter/memory.go#L86-L103

**Severity:** low

**Description:**
The two `Store` backends disagree on whether an entry whose timestamp equals the window cutoff is inside the window.

- `InMemory.Peek` (and `InMemory.IncrementAndGet`) keep entries with `ts.After(cutoff)` — strictly greater than the cutoff, so an entry exactly at the cutoff is **excluded**.
- `Redis.Peek` counts with `ZCount(ctx, windowKey, fmt.Sprintf("%f", cutoff), "+inf")`, and Redis `ZCOUNT` min/max are **inclusive** by default, so an entry exactly at the cutoff is **included**.

`Redis.IncrementAndGet` happens to agree with the in-memory side (its `ZRemRangeByScore "-inf" cutoff` removes the boundary entry before `ZCard`), so the inconsistency is specifically in `Peek`, which the `sequenceBlock` condition uses to test "has tool A already run this session?". A boundary entry can therefore be counted as present under Redis but absent under the in-memory backend, yielding different allow/deny decisions for the same call sequence depending on which backend is deployed. Because counters are security-relevant (rate limits, sequence gating) and the project goes to lengths to keep the two backends behaviorally identical, this divergence is a latent correctness/consistency bug. Probability is low (requires a microsecond-exact boundary match), hence low severity.

**Trigger / repro:**
Record a call at time `T`, then `Peek` at exactly `T + windowSec` so the recorded score equals the computed cutoff. Redis returns 1; the in-memory backend returns 0.

**Suggested fix:**
Make the cutoff exclusive on both sides consistently. For Redis, use an exclusive lower bound: `ZCount(ctx, windowKey, "("+fmt.Sprintf("%f", cutoff), "+inf")` (the `(` prefix makes the bound exclusive), matching `ts.After(cutoff)` in the in-memory backend. Add a cross-backend table test pinning the boundary semantics.

[...]

Issue #1018, opened 2026-06-20
[rev-20260620T2320Z-24375b] HTTP upstream client (`doMCPHTTP`) omits the spec-required `Accept` header and cannot parse `text/event-stream` responses


eunolabs:
**Location:** `cmd/eunox-mcp/http_remote.go:186-219`

**Reference:** https://github.com/eunolabs/eunox/blob/194877e40276f30c8c58b2e721987a83c2ac4a99/cmd/eunox-mcp/http_remote.go#L186-L219

**Severity:** high

**Description:**
`doMCPHTTP` is the single HTTP-upstream client shared by the gateway remote path (`doRemoteHTTP`), `validate --live` / `init` (`fetchLiveTools`), and the stdio→HTTP bridge (`httpUpstream.do`). It is not compliant with the MCP Streamable HTTP transport on two counts:

1. **Missing `Accept` header (request side).** The MCP Streamable HTTP spec requires the client to send `Accept: application/json, text/event-stream` on every POST. `doMCPHTTP` only sets `Content-Type`:

   ```go
   req.Header.Set("Content-Type", ctJSON)
   if sessID != "" { req.Header.Set(sessionHeader, sessID) }
   // no Accept header set anywhere
   ```

   A spec-conformant server is entitled to reject such a request with `406 Not Acceptable`, which `doMCPHTTP` then surfaces as a generic `upstream returned 406` error — every call fails.

2. **Cannot read `text/event-stream` responses (response side).** The spec explicitly permits a server to answer a POST'd JSON-RPC request with an SSE stream (`Content-Type: text/event-stream`) rather than `application/json`. `doMCPHTTP` ignores the response `Content-Type` entirely and unconditionally does:

   ```go
   var out rpcMsg
   if err := json.NewDecoder(resp.Body).Decode(&out); err != nil {
       return rpcMsg{}, resp.Header, fmt.Errorf("decoding upstream response: %w", err)
   }
   ```

   When the body is SSE-framed (`event:` / `data:` lines) the JSON decode fails and the call errors out. This is not the same as #836 (which concerns *server-initiated* requests pushed over SSE); this is the ordinary request→response path failing whenever the upstream chooses the SSE response form the spec allows. eunox's own server emits `text/event-stream` (`http.go:48`, `http.go:1051`), so the project already relies on SSE — but its client cannot consume it.

[...]

Issue #1017, opened 2026-06-20
[rev-20260620T2320Z-24375b] `init` / `validate --live` hang forever on a slow or hostile upstream (no timeout)


eunolabs:
**Location:** `cmd/eunox-mcp/main.go:740`, `cmd/eunox-mcp/main.go:926`, `cmd/eunox-mcp/live_upstream.go:108-148`

**Reference:**
- https://github.com/eunolabs/eunox/blob/194877e40276f30c8c58b2e721987a83c2ac4a99/cmd/eunox-mcp/main.go#L738-L743
- https://github.com/eunolabs/eunox/blob/194877e40276f30c8c58b2e721987a83c2ac4a99/cmd/eunox-mcp/main.go#L922-L929
- https://github.com/eunolabs/eunox/blob/194877e40276f30c8c58b2e721987a83c2ac4a99/cmd/eunox-mcp/live_upstream.go#L108-L148

**Severity:** medium

**Description:**
Both the `init` subcommand and `validate --live` (without `--config`) introspect the upstream with **`context.Background()`** and no deadline:

```go
// main.go:740 (validate --live) and main.go:926 (init)
info, err = fetchLiveToolsStdio(context.Background(), spec.Command, spec.Args)
...
info, err = fetchLiveTools(context.Background(), spec.URL, spec.AuthHeader, *tlsSkipVerify)
```

Neither path bounds the handshake:

- **HTTP path (`fetchLiveTools`)** builds its client via `buildUpstreamClient`, which sets **no `http.Client.Timeout`**, and the request context is `context.Background()`. A remote MCP server that accepts the TCP connection and then never responds (or trickles bytes) blocks `init`/`validate` indefinitely.
- **Stdio path (`fetchLiveToolsStdio`)** is worse: the SIGKILL safety-net timer is created **inside the deferred teardown** (`live_upstream.go:132-141`), so it only arms *after* `runStdioHandshake` returns. If the spawned subprocess answers `initialize` but then never replies to `tools/list` and keeps its stdout open, `runStdioHandshake` → `readResponseWithID` blocks forever in `r.Read()`. `readResponseWithID` only checks `ctx.Err()` *before* each blocking `Read` (`live_upstream.go:220-240`), and the context is `Background`, so it never fires. `exec.CommandContext(ctx, …)` is likewise tied to that never-canceled context, so the process is never killed. The kill timer never runs because the function never returns. The CLI hangs permanently.

[...]

PR #1025, opened 2026-06-21
Fix automated-review session bugs (#1012, #1014)

### LINKED ISSUE #1012: [rev-20260620T2059Z-356feb] Repeated UTF-8 BOMs bypass redactFields
**Location:** `internal/pdp/pdp.go:1222-1293`

**Reference:** https://github.com/eunolabs/eunox/blob/92907fef4cffca81a87041f78847c6943a261514/internal/pdp/pdp.go#L1222-L1293

**Severity:** high

**Description:** `stripBOM` and `stripBOMString` remove exactly one leading UTF-8 BOM. `redactJSONText` then decodes the remaining text and, on a decode error, treats it as harmless prose unless the first non-space byte is `{` or `[`. An upstream can prefix a JSON object with two BOMs: the first is removed, the second still makes `encoding/json` fail, and the classifier sees the second BOM instead of the container delimiter. The function returns the original text as prose, including every field that `redactFields` was required to remove. This is an incomplete edge case in the fix for #866 and restores the same fail-open data leak for repeated BOMs.

**Trigger / repro:** Configure a `redactFields` obligation for `ssn`, then have the upstream return a text item containing `BOM + BOM + {"ssn":"123-45-6789","name":"alice"}`. A focused test calling `ApplyRedactObligs` succeeds without an error and returns output still containing `123-45-6789`. The same payload with one BOM is redacted correctly.

**Suggested fix:** Strip all consecutive leading UTF-8 BOMs before parsing and classification, for example with a small loop around `bytes.TrimPrefix` / `strings.TrimPrefix`, and use the fully normalized value consistently. Add regression tests with two or more BOMs on both the response envelope and text content so the result either redacts the field or fails closed.

**Session:** rev-20260620T2059Z-356feb

### LINKED ISSUE #1014: [rev-20260620T2240Z-3d6842] Dead code StdioProxy.forwardToUpstream omits audit allow record; stale test comment claims it is a live path
**Location:** `internal/transport/stdio.go:564-577`

**Reference:** https://github.com/eunolabs/eunox/blob/9136b28b2a459f7ea739c24100aeb2a7e8ff0a5a/internal/transport/stdio.go#L564-L577

**Severity:** low

**Description:**
`(*StdioProxy).forwardToUpstream` is never called from any production path. The stdio request path is `handleHostRequest` → `dispatchRequest`, whose `default` case is `dispatchUnmapped` (a fail-closed deny), and every mapped method is handled by the shared enforced-forward core in `forward.go`. A repo-wide search shows `forwardToUpstream` is referenced only by its own definition and by unit tests:

```
$ grep -rn "forwardToUpstream" --include=*.go | grep -v _test.go
internal/transport/stdio.go:564:// forwardToUpstream sends a request to the upstream and returns its response.
internal/transport/stdio.go:565:func (p *StdioProxy) forwardToUpstream(ctx context.Context, msg mcp.RPCMsg) {
```

Two concrete problems flow from this:

1. **Latent audit-completeness defect.** The function records a deny on the upstream-error path (`p.sink.RecordDeny(...)`) but writes the successful upstream response to the host **without any `RecordAllow`**:

   ```go
   func (p *StdioProxy) forwardToUpstream(ctx context.Context, msg mcp.RPCMsg) {
       resp, err := p.callUpstream(ctx, msg)
       if err != nil {
           ... p.sink.RecordDeny(...) ...
           return
       }
       resp.ID = msg.ID
       _ = p.hostWriter.Write(resp)   // <-- forwards to host, no allow record
   }
   ```

   This violates the project's load-bearing invariant that *every* decision is written to the audit log. It is dormant today because the function is unreachable, but if it is ever re-wired into a forward path it silently re-introduces an audit gap (the surrounding enforced-forward core, by contrast, always calls `RecordAllow`).

2. **Misleading test comment.** `TestStdioHandlePromptsGet_NotManifestPDP` (`internal/transport/stdio_test.go:1224-1235`) carries the comment `// alwaysAllowPDP is not *ManifestPDP → calls forwardToUpstream`, but that request actually flows through `dispatchRequest` → `dispatchPromptsGet` → `enforcedForwardCore`, not `forwardToUpstream`. The comment describes a pre-refactor architecture and can mislead a maintainer into believing `forwardToUpstream` is exercised in production. Go does not flag unused methods, and `go vet` / the linter pass clean, so nothing else surfaces the staleness.

**Trigger / repro:**
- `grep -rn "forwardToUpstream" --include=*.go | grep -v _test.go` returns only the definition — no caller.
- `TestStdioForwardToUpstream_Success` exercises the success path and asserts the response is written, but never asserts an audit allow record exists — confirming the missing `RecordAllow` is unnoticed.

**Suggested fix:**
Remove `forwardToUpstream` (and its two direct tests) now that all stdio methods route through the shared `dispatchRequest`/`enforcedForwardCore` path, and correct the stale comment in `TestStdioHandlePromptsGet_NotManifestPDP` to describe the actual `dispatchRequest` → `enforcedForwardCore` flow. If the method is intended to be kept for a future use, add the `RecordAllow` on its success path so it cannot resurface as an audit gap.

**Session:** rev-20260620T2240Z-3d6842

eunolabs:
## Bugs addressed

This batch covers the open `automated-review` issues whose code is not already fixed on `main`.

- Closes #1012 — **Repeated UTF-8 BOMs bypassed `redactFields` (high).** `stripBOM`/`stripBOMString` removed only one leading BOM, so an upstream prefixing a JSON container with two or more BOMs had the first stripped, the next still made `encoding/json` fail, and the classifier then saw a residual BOM byte (`0xEF`) instead of the `{`/`[` delimiter — forwarding the body as prose **unredacted**, reopening the fail-open data leak. Both helpers now loop until no leading BOM remains, so the body is either parsed-and-redacted or recognized as a container and failed closed. Added regression tests for repeated BOMs on both the text content item and the response envelope.
- Closes #1014 — **Dead `StdioProxy.forwardToUpstream` + stale test comment (low).** The method had no production caller (the stdio path is `handleHostRequest` → `dispatchRequest` → shared `enforcedForwardCore`, with a fail-closed default), and it wrote the upstream response to the host with **no `RecordAllow`** — a latent audit-completeness gap that would resurface if it were ever re-wired. Removed the method and its two direct tests, and corrected the stale comment in `TestStdioHandlePromptsGet_NotManifestPDP` so it describes the actual `dispatchRequest` flow.

### Already resolved on `main` (no code change needed)

While triaging the session I verified the remaining three open `automated-review` issues are **already fixed** in the current tree (they reference the pre-refactor `cmd/eunox-mcp/*` paths):

- #1017 (`init`/`validate --live` no-timeout hang) — both call sites now wrap `context.WithTimeout(context.Background(), liveUpstreamTimeout)` (`cmd/eunox-mcp/main.go:1237,1440`) and the stdio kill-timer is armed before the handshake (`live_upstream.go`).
- #1018 (HTTP upstream `Accept` header / SSE response) — `DoMCPHTTP` now sends `Accept: application/json, text/event-stream` and branches on the response Content-Type to parse SSE (`internal/transport/http_remote.go:319,348-360`).
- #1019 (callcounter `Peek` boundary divergence) — `Redis.Peek` now uses an exclusive lower bound `"("+cutoff`, matching the in-memory `ts.After(cutoff)`, with regression tests (`pkg/callcounter/redis.go:345`).

I'll close those three separately as already-fixed with the verifying references, rather than attach a closing keyword here since this diff doesn't touch them.

## Testing

- `make build`, `make check-fmt`, `make lint` (`go vet` + golangci-lint) — clean.
- New regression tests in `internal/pdp/pdp_test.go` cover 2/3/5 leading BOMs on a text item and a doubly-BOM-prefixed envelope; both redact or fail closed (never leak).

https://claude.ai/code/session_01GGSCcusqD2aGmL14tL4w2D

---
_Generated by [Claude Code](https://claude.ai/code/session_01GGSCcusqD2aGmL14tL4w2D)_


internal/pdp/pdp.go
```diff
@@ -1321,14 +1321,26 @@ func isRecognizedContentType(t string) bool {
...
-// stripBOM removes a single leading UTF-8 BOM from b, if present.
+// stripBOM removes ALL consecutive leading UTF-8 BOMs from b. Stripping only one
+// left a fail-open hole: an upstream prefixing a JSON container with two or more
+// BOMs would have the first removed, the second still make encoding/json fail, and
+// the classifier then see a BOM byte (0xEF) instead of the '{'/'[' delimiter and
+// wave the body through as prose UNREDACTED. Trimming every leading BOM means the
+// body is either parsed-and-redacted or recognized as a container and failed closed.
 func stripBOM(b []byte) []byte {
-	return bytes.TrimPrefix(b, []byte(utf8BOM))
+	for bytes.HasPrefix(b, []byte(utf8BOM)) {
```
eunolabs: **Correct for the reported case, but the same fail-open class is still reachable via a BOM behind leading whitespace.**

`stripBOM`/`stripBOMString` only trim BOMs at byte offset 0, but both `encoding/json` and the `looksLikeUnparseableJSONContainer` classifier tolerate *leading JSON whitespace* — and `strings.TrimSpace` does **not** treat U+FEFF as space (`unicode.IsSpace(rune(0xFEFF)) == false`). So an adversarial upstream can re-open the exact bypass this PR closes by separating the BOM from the start with a space/tab/CR/LF.

I applied this PR's loop to the current tree and called the redaction path directly with `{"ssn":"123-45-6789","keep":"ok"}` bodies:

```
single_BOM             LEAKED=false
double_BOM             LEAKED=false   <- fixed by this PR
tab_then_single_BOM    LEAKED=true    <- "\t" + BOM + {...}
space_then_single_BOM  LEAKED=true    <- " "  + BOM + {...}
BOM_ws_BOM             LEAKED=true    <- BOM + " " + BOM + {...}
```

Each leaking case is a well-formed envelope whose text content item is `<ws> + BOM + {...}`: `stripBOMString` sees a leading space/tab (not a BOM) and no-ops, the internal `dec.Decode` skips the whitespace then fails on `0xEF`, and `looksLikeUnparseableJSONContainer` `TrimSpace`s the whitespace but keeps the BOM, so `t[0] == 0xEF` → classified as prose → field forwarded unredacted. Same fail-open as the bug being fixed, just a different trigger.

Closing it fully probably belongs at the classifier rather than only in the strippers — e.g. trim leading BOM(s) *after* `TrimSpace` inside `looksLikeUnparseableJSONContainer` (and normalize whitespace+BOM together before the parse), so the strippers' notion of "leading" can't diverge from the parser's. Given it's the same severity as #1012, worth handling here or as an immediate fast-follow.

Minor, unrelated nit: `[]byte(utf8BOM)` reallocates on each loop iteration — hoisting `bom := []byte(utf8BOM)` above the loop avoids it. Negligible (loop rarely runs more than once), just noting.

---

eunolabs (commented):
## Review — Fix automated-review session bugs (#1012, #1014)

Well-scoped and well-tested. Both fixes are correct as far as they go, the dead-code removal is verified safe, and the regression tests assert the right invariant (redact-or-fail-closed, never forward the field). I checked the changes against the tree locally. One substantive finding on the BOM fix (left inline) that I think is worth resolving before #1012 is considered fully closed — flagging rather than blocking.

### #1012 — repeated-BOM redaction bypass
- The loop correctly closes the reported case: I confirmed `double_BOM` leaks on `main` and is redacted with this PR applied.
- **The fail-open class is broader than _consecutive_ BOMs, though.** `stripBOM*` only trims at offset 0, while the JSON parser and the `looksLikeUnparseableJSONContainer` classifier both tolerate leading whitespace (and `TrimSpace` doesn't treat U+FEFF as space). A single BOM behind a leading space/tab/CR/LF — e.g. `"\t" + BOM + {...}` — still reaches the prose path and forwards the field unredacted. Empirical results + a suggested classifier-side fix are in the inline comment. Same severity as the bug being fixed.

### #1014 — dead `forwardToUpstream` removal
- Verified safe: on `main` the only references are the definition + the two deleted tests; there's no production caller (live path is `handleHostRequest → dispatchRequest → enforcedForwardCore`, fail-closed default). Removing a method that wrote host responses with no `RecordAllow` is a real audit-completeness improvement, not just tidy-up.
- The corrected comment on `TestStdioHandlePromptsGet_NotManifestPDP` now matches the actual dispatch flow, and removing the two tests doesn't orphan the `encoding/json` import or `respondingProxy` (both still used elsewhere).

### Conventions
- Code/test comments correctly avoid issue numbers (kept to the PR body), and the fail-closed framing, test placement, and no-emoji rules all look right.

Net: good change that strictly improves on `main` — the one thing I'd weigh before merge is whether to close the whitespace+BOM variant here vs. as a fast follow.

---

internal/pdp/pdp.go
eunolabs: Addressed in 0a218c2.

- Replaced the offset-0 strippers (`stripBOM`/`stripBOMString`) with `trimLeadingSpaceAndBOM[Bytes]`, which trim the **whole** leading run of BOMs and JSON whitespace — so a BOM behind a space/tab/CR/LF (or `BOM ws BOM`) no longer survives into the parser.
- Hardened `looksLikeUnparseableJSONContainer` to skip leading BOMs as well as whitespace, so classification can't diverge from the parser even if a caller forgets to normalize (defense-in-depth).
- `redactJSONText` now normalizes a separate parse input and forwards the **original** text byte-for-byte on the prose/scalar paths, so non-redacted content (including a leading BOM/whitespace) is preserved unchanged.

Net behavior: the `tab/space/CRLF + BOM` and `BOM + ws + BOM` cases now redact (or fail closed if the body is malformed) instead of leaking. New regression tests cover those shapes on both a text content item and the response envelope, a malformed container behind whitespace+BOM (must fail closed), and byte-for-byte prose preservation.

The alloc nit is handled too: the bytes variant hoists `bom := []byte(utf8BOM)` above the loop, and the string variant compares against the `utf8BOM` string directly (no per-iteration conversion).

---

eunolabs (commented):
## Re-review — head `0a218c2` (after the BOM-bypass follow-up)

Took another full pass over the updated PR, with a critical self-review of the follow-up commit that addressed my earlier finding.

### Earlier finding (whitespace+BOM bypass) — resolved
The fix replaces the offset-0 strippers with `trimLeadingSpaceAndBOM[Bytes]`, which trim the **whole** leading run of BOMs and JSON whitespace, and additionally hardens `looksLikeUnparseableJSONContainer` to skip leading BOMs (so classification can't diverge from the parser even if a caller forgets to normalize). I re-traced every shape I'd flagged — `"\t"+BOM`, `" "+BOM`, `"\r\n"+BOM`, `BOM+" "+BOM`, `"\t"+BOM+BOM` — and they now redact (or fail closed when the container is malformed) instead of leaking. The new regression tests cover all of these on both a text content item and the response envelope, plus a malformed-behind-whitespace+BOM fail-closed case.

### Self-review of the new code — no issues found
- **Termination/efficiency:** both trim loops strictly shorten on each `continue` (and `strings`/`bytes.TrimLeft` reslice without copying), so they terminate and don't allocate per iteration; the bytes variant hoists `[]byte(utf8BOM)` out of the loop.
- **Byte-fidelity (nice catch by the change):** `redactJSONText` now decodes/classifies a separate `parseInput` and returns the **original** `text` on the prose/scalar paths, so non-redacted content — including a leading BOM/whitespace — is preserved verbatim. That's strictly more faithful than the prior reassign-and-strip, and there's a regression test asserting byte-for-byte preservation.
- **Fail-closed preserved:** a malformed container behind whitespace+BOM still returns an error → caller fails the response closed.
- The classifier's switch logic is unaffected by dropping `TrimSpace`'s trailing-trim (only `t[0]`, the `rest` computation, and `ContainsRune` are used).

### Verification
- Local: `go build ./...`, `go vet ./...`, and the full `go test -race ./...` suite all pass on `0a218c2`; `golangci-lint` reported 0 issues on the package.
- CI: all 33 checks green (Test, Lint, gofmt, Build matrix, CodeQL, Trivy, govulncheck, Enforcement compliance, License/NOTICE); `mergeable_state: clean`.

### #1014 (dead `forwardToUpstream` removal) — still sound
Unchanged by the follow-up; verified earlier that it had no production caller and that removing the two tests orphans neither the `encoding/json` import nor `respondingProxy`.

**Verdict:** no blocking issues — the fail-open class is fully closed and well-tested. LGTM; ready to merge whenever you are.

---

PR #1044, opened 2026-06-21
Fix session rev-20260621 review bugs (#1026–#1041)

### LINKED ISSUE #1026: [rev-20260621T0010Z-88f6da] Unused symbolic denial code ErrCodeInvalidRequest (dead constant, no wire-code mapping)
**Location:** `pkg/capability/errors.go:23`

**Reference:** https://github.com/eunolabs/eunox/blob/64b2814d89fc749d43164a9b4049e7839850c2ae/pkg/capability/errors.go#L23

**Severity:** low

**Description:**
`ErrCodeInvalidRequest = "INVALID_REQUEST"` is declared in the exported denial-code block but is never referenced anywhere in the codebase. A repo-wide search finds only its own declaration:

```
$ grep -rn "ErrCodeInvalidRequest\|INVALID_REQUEST" --include=*.go | grep -v _test.go
pkg/capability/errors.go:23:	ErrCodeInvalidRequest  = "INVALID_REQUEST"
```

Unlike every other constant in that block, it is:

1. **Never emitted** — no PDP, enforcement handler, or transport path produces a denial with this code (compare the heavily-used `ErrCodeMissingContext`, `ErrCodeRateLimited`, etc.).
2. **Absent from the wire-code mapping** — `denialToJSONRPCCode` in `internal/transport/jsonrpc.go:222-256` has a `case` for every other symbolic code but none for `ErrCodeInvalidRequest`. If it were ever returned as a denial code it would silently fall through the `default` arm to `JSONRPCCodeCapabilityDenied` (-32002), i.e. be reported on the wire as a *capability denial* rather than the standard JSON-RPC -32600 "Invalid Request" the name implies.

It is genuinely orphaned: malformed-request/param paths (e.g. `dispatch.go:81/107/125/141`) return raw `-32602` via `mcp.ErrorResponse` and carry no symbolic `denial_code` at all, so there is no caller that "should" have used this constant either. This is dead code in an exported, pre-1.0 public API surface — exactly the kind of thing the project's "no back-compat shims, change cleanly" convention says to remove.

**Trigger / repro:**
- `grep -rn "ErrCodeInvalidRequest" --include=*.go` returns only the declaration line — no producer, no test, no mapping entry.
- `go build` / `go vet` do not flag unused package-level constants, so the staleness is invisible to CI.

**Suggested fix:**
Remove the `ErrCodeInvalidRequest` constant. If a JSON-RPC "Invalid Request" denial path is actually intended in the future, reintroduce it together with a matching `JSONRPCCodeInvalidRequest = -32600` and an explicit `case` in `denialToJSONRPCCode`, plus a producing call site and test — so the symbolic code can never silently map to CAPABILITY_DENIED (-32002).

**Session:** rev-20260621T0010Z-88f6da

### LINKED ISSUE #1027: [rev-20260621T0110Z-34ddf7] init: generated config emits upstreamUrl unquoted while every other string field is %q-quoted
**Location:** `cmd/eunox-mcp/init_manifest.go:164-177`

**Reference:** https://github.com/eunolabs/eunox/blob/64b2814d89fc749d43164a9b4049e7839850c2ae/cmd/eunox-mcp/init_manifest.go#L164-L177

**Severity:** low

**Description:**
`generateInitConfigYAML` scaffolds a runnable gateway config. Every string-valued field it writes is YAML-safely quoted via Go's `%q` (and the file goes to considerable lengths elsewhere — `yamlScalar`, the line-break re-prefixing of untrusted tool names — to avoid YAML injection):

- `command: %q` (line 160)
- `args: %s` via `yamlStringList`, each element `%q` (line 162)
- `upstreamAuthHeader: %q` (line 168)
- `policy: [%q]` (line 180)

The single exception is the upstream URL:

```go
fmt.Fprintf(&sb, "    upstreamUrl: %s\n", u.URL)   // line 166 — unquoted
```

`u.URL` is operator-supplied (the `--upstream-url` flag). Emitted as a bare YAML plain scalar, a value containing YAML-significant syntax — a `": "` sequence, a leading indicator character (`!`, `&`, `*`, `@`, `` ` ``, `[`, `{`, `#`, quote chars), or a `" #"` run — would either fail to parse when the generated config is loaded by `LoadGatewayConfig`, or round-trip to a *different* string than the operator intended. The other fields are quoted precisely to prevent this; the URL field is the lone gap.

In practice an ordinary `http(s)://host:port/path` URL is a valid plain scalar, so this is benign for the common case — hence low severity — but it is a genuine inconsistency and a latent YAML-safety hazard in a security tool's generated config.

**Trigger / repro:**
Run `eunox-mcp init` against an upstream whose URL contains a YAML-significant character (or hand-construct an `initUpstreamSpec` with `URL` set to e.g. `http://h/p #c` or a value beginning with `@`/`[`) and load the emitted config — it parses to a different URL or errors, whereas the same value placed in `upstreamAuthHeader` (which is `%q`-quoted) is preserved verbatim.

**Suggested fix:**
Quote the URL like every sibling field:

```go
fmt.Fprintf(&sb, "    upstreamUrl: %q\n", u.URL)
```

**Session:** rev-20260621T0110Z-34ddf7

### LINKED ISSUE #1028: [rev-20260621T0110Z-34ddf7] Remote HTTP upstream returning 202 Accepted to an enforced REQUEST yields a bogus allow + malformed empty response instead of failing closed
**Location:** `internal/transport/http_remote.go:307-361` (the 202 branch at `337-339`); consumed by `callRemoteUpstream` (`267-299`) and `enforcedForwardCore` (`internal/transport/forward.go:161-198`)

**Reference:** https://github.com/eunolabs/eunox/blob/64b2814d89fc749d43164a9b4049e7839850c2ae/internal/transport/http_remote.go#L337-L361

**Severity:** low

**Description:**
`DoMCPHTTP` treats every `202 Accepted` upstream response as a no-body notification ack:

```go
if resp.StatusCode == http.StatusAccepted {
    return mcp.RPCMsg{}, resp.Header, nil   // empty RPCMsg, nil error
}
```

The doc comment justifies this as "A 202 Accepted response (notifications) returns an empty rpcMsg without error." But `DoMCPHTTP` is the single shared HTTP-upstream call used for **enforced requests too** (`tools/call`, `resources/read`, `prompts/get` → `callRemoteUpstream` → `doRemoteHTTP` → `DoMCPHTTP`). Per the MCP Streamable HTTP spec a server must answer a JSON-RPC **request** with `200` (JSON or SSE body); `202` with no body is reserved for inputs containing only responses/notifications. A non-compliant or hostile upstream (the threat model explicitly contemplates "a hostile or compromised upstream" — see the `maxUpstreamRespBytes` comment) that returns `202` to a `tools/call` drives this path:

1. `DoMCPHTTP` returns an empty `RPCMsg{}` with `err == nil`.
2. `callRemoteUpstream`'s id-correlation guard is skipped: `resp.IsResponse()` is false for an empty message (nil ID), so the mismatch check at line 291 does not fire. It returns the empty message, `nil`.
3. `enforcedForwardCore` sees `fwdErr == nil`, so it **records an `RecordAllow`** for a call the upstream never actually answered, then stamps `upResp.ID = msg.ID` and returns `{"jsonrpc":"2.0","id":<host id>}` — a JSON-RPC message carrying **neither `result` nor `error`**, which is malformed and which the host MCP client may mishandle.

So a spec-violating upstream converts an enforced request into (a) a false "allow" audit record and (b) a malformed host response, rather than the fail-closed transport error the surrounding code produces for every other upstream anomaly. The subprocess transport has no equivalent gap (it waits for a framed response or times out).

**Trigger / repro:**
Point the proxy at a remote HTTP upstream (gateway `transport: http`, upstream `upstreamUrl`). Have the upstream answer a `tools/call` POST with `HTTP/1.1 202 Accepted` and an empty body. Observe: the host receives a response with no `result`/`error`, and the audit log records an `allow` for the call.

**Suggested fix:**
Distinguish "request expecting a response" from a notification in `DoMCPHTTP` (the caller already knows `msg.ID`): when `msg.ID != nil` and the upstream returns `202` (or otherwise yields an empty, non-response body), return a transport error so `enforcedForwardCore` records a deny and returns a structured upstream error to the host — fail closed. Only treat `202`/empty as a valid no-body ack when `msg.ID == nil` (a genuine notification).

**Session:** rev-20260621T0110Z-34ddf7

### LINKED ISSUE #1029: [rev-20260621T0110Z-34ddf7] allowedValues glob load-validation uses whole-string path.Match but runtime splits on "/" for "**", so some patterns pass validation yet match nothing at runtime
**Location:** `internal/config/manifest.go:600-625` (`validateAllowedValues`), vs runtime `pkg/enforcement/glob.go:46-98` (`MatchValueGlob`)

**Reference:** https://github.com/eunolabs/eunox/blob/64b2814d89fc749d43164a9b4049e7839850c2ae/internal/config/manifest.go#L607-L623

**Severity:** low

**Description:**
`validateAllowedValues` exists to surface a broken glob at load time rather than letting it "silently [be] more restrictive than intended, with no error to tell the operator the pattern is broken." It validates each string value by running `path.Match` on the **whole pattern string**:

```go
if _, err := stdpath.Match(s, ""); err != nil {
    return fmt.Errorf("...invalid glob pattern %q...", ...)
}
```

But the runtime matcher `MatchValueGlob` does **not** evaluate a `**`-bearing pattern with whole-string `path.Match`. When the pattern contains `**`, it splits the pattern on `/` and runs `path.Match` **per segment** (`matchGlobSegments`). These two validation models disagree for a pattern where a `path.Match` bracket character-class spans a `/` while the pattern also contains `**`. For example:

```yaml
- type: allowedValues
  argument: path
  values: ["[a/b]/**"]
```

- **Load time:** `path.Match("[a/b]/**", "")` parses `[a/b]` as a single valid character class (matching `a`, `/`, or `b`), so it returns `(false, nil)` — **no error, accepted.**
- **Runtime:** the pattern contains `**`, so `MatchValueGlob` splits it on `/` into segments `["[a", "b]", "**"]`. `path.Match("[a", …)` returns `path.ErrBadPattern`, which `matchGlobSegments` treats as a non-match. The condition therefore **matches nothing and denies every call**, with no diagnostic.

This is exactly the "silently dead policy" failure mode the validator was written to prevent — it just misses it because it validates with different glob semantics than the runtime uses. It fails *closed* (deny-all), so there is no authorization bypass; the impact is an availability/correctness surprise plus a misleading `CONDITION_FAILED` on every call that reads like a real enforcement event. The bracket-spanning-slash case is contrived, but the underlying defect — load validation and runtime matcher using divergent glob models — is real.

**Trigger / repro:**
Add an `allowedValues` condition whose string value is `[a/b]/**` (a closed `[...]` class containing `/`, plus a `**`). `eunox-mcp validate` / proxy startup accepts the manifest, but every live call evaluating that condition is denied unconditionally.

**Suggested fix:**
Validate `allowedValues` string patterns through the same code path the runtime uses, so the two cannot diverge: when the pattern contains `**`, split on `/` and `path.Match`-check each segment (mirroring `matchGlobSegments`), otherwise check the whole string. Factor a shared `ValidateValueGlob(pattern) error` helper out of `glob.go` and call it from both `validateAllowedValues` and `MatchValueGlob`'s call sites, rather than re-deriving the validity test with whole-string `path.Match`.

**Session:** rev-20260621T0110Z-34ddf7

### LINKED ISSUE #1030: [rev-20260621T0105Z-f82896] audit rotate(): fsync of rotated file placed after os.OpenFile, creating avoidable crash-recovery record loss window
**Location:** `internal/audit/rotate.go:183-214`
**Reference:** https://github.com/eunolabs/eunox/blob/64b2814/internal/audit/rotate.go#L183-L214
**Severity:** Medium
**Session:** rev-20260621T0105Z-f82896

---

## Description

In `rotate()`, `os.OpenFile` (line 183) that creates the new log file is called **before** `s.f.Sync()` (line 204) that makes the rotated file durable. The code comment on line 197-203 correctly identifies the problem—"a power loss or hard kill between the os.Rename above and the page-cache flush could leave the rotated file missing its last records"—but the sync is placed in the wrong position relative to `os.OpenFile`, so a crash window remains open.

Current order:
1. `os.Rename(activePath, rotated)` — rename old file
2. **`os.OpenFile(logPath, O_CREATE…)` — new current log created**  ← crash here leaves new file + unsynced rotated
3. `s.f.Sync()` — attempt to flush old fd
4. `s.f.Close()` — close old fd
5. `s.f = f` — switch to new fd

If the process is killed (SIGKILL, OOM, power loss) between steps 2 and 3:
- The new (empty) `logPath` file exists on disk.
- The rotated file's tail records are still only in the OS page cache; they may or may not survive the crash.
- On restart, `Open()` opens the empty `logPath`, falls back to `newestRotatedSibling()`, reads the rotated file's tail, and runs `VerifyRecord`. If the page cache was not flushed, the tail HMAC doesn't match the actual last written record → `tail_hmac_mismatch` marker written; those records are permanently gone.

## Why the existing fallback does not prevent this

The fallback path (lines 184-195) relies on `s.f` remaining open when `os.OpenFile` fails. That constraint is valid, but it does **not** require the sync to happen after `os.OpenFile`. Moving the sync to just before `os.OpenFile` preserves the fallback entirely:

```go
// Sync the old fd while it is still open — fallback still works
// because s.f is not closed until after os.OpenFile succeeds.
    s.writeFailures.Add(1)
    fmt.Fprintf(os.Stderr, "[eunox-mcp] audit rotate error (sync): %v\n", err)
}
f, err := os.OpenFile(s.logPath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o600)
if err != nil {
    // s.f is still open and already synced — safe to fall back
    s.activePath = rotated
    s.written = 0
    return
}
_ = s.f.Close()
s.f = f
```

With this order:
- Crash between Sync and OpenFile → rotated file is durable, `logPath` doesn't exist yet; on restart `Open()` creates a fresh `logPath`, resumes from the rotated sibling with no HMAC mismatch.
- Crash between OpenFile and Close → same chain-resume path, still no record loss because the rotated file was already synced.

## Trigger / repro

1. Configure a small `--rotate-size` (e.g. 1 KiB) so rotations are frequent.
2. Write enough records to trigger a rotation.
3. Kill the process with `kill -9` or pull power immediately after the `os.OpenFile` call in `rotate()` returns (e.g. via a debugger breakpoint at line 204) but before `s.f.Sync()` completes.
4. Restart and run `eunox-mcp audit-verify`; observe a `tail_hmac_mismatch` integrity marker for the last N records before the rotation point.

## Suggested fix

Move `s.f.Sync()` and its error handling to **before** `os.OpenFile` (as shown above). The fallback path is unchanged in behaviour because `s.f` is not closed until after `os.OpenFile` succeeds. Update the code comment to reflect the new ordering.

### LINKED ISSUE #1031: [rev-20260621T0105Z-f82896] http_routing: upstream startup error string leaked verbatim to MCP client in 500 response
**Location:** `internal/transport/http_routing.go:164-168`
**Reference:** https://github.com/eunolabs/eunox/blob/64b2814/internal/transport/http_routing.go#L164-L168
**Severity:** Medium
**Session:** rev-20260621T0105Z-f82896

---

## Description

In `handleMCPPost`, when session creation fails for any reason other than `errSessionLimit`, the raw Go error string is returned verbatim in the HTTP 500 body:

```go
http.Error(w, "failed to start upstream: "+err.Error(), http.StatusInternalServerError)
```

`newSession` and `newRemoteSession` can return errors from multiple internal subsystems. Examples of what an MCP client would see in the response body:

- `failed to start upstream: exec: "/usr/local/bin/secret-tool": stat /usr/local/bin/secret-tool: no such file or directory` — reveals the upstream command path configured in the proxy
- `failed to start upstream: fork/exec /opt/internal/mcp-backend: permission denied` — reveals internal deployment paths
- `failed to start upstream: dial tcp 10.0.1.42:9999: connect: connection refused` — reveals an internal IP and port of a remote upstream
- `failed to start upstream: tls: failed to verify certificate: x509: certificate signed by unknown authority` — reveals TLS infrastructure details

This endpoint is on the MCP-client-facing `/mcp` POST path, reachable by any authenticated (or, if no auth is configured, any) client. Unlike `/control/kill` (loopback-only + control token), this is the hot path every session initialization traverses.

## Trigger / repro

1. Start `eunox-mcp proxy` with a command that does not exist on `$PATH`.
2. Send a JSON-RPC `initialize` request (no `Mcp-Session-Id` header) to `POST /mcp`.
3. Observe the 500 response body; it contains the full `exec.LookPath` error including the configured command name.

## Suggested fix

Replace the raw error with a generic client-facing message and log the detail to stderr (consistent with how other subsystems expose sensitive details only to operators):

```go
fmt.Fprintf(os.Stderr, "[eunox-mcp] failed to start upstream: %v\n", err)
http.Error(w, "upstream unavailable", http.StatusInternalServerError)
```

The same pattern should be applied to the analogous `kill switch activation failed: "+err.Error()` and `kill switch session kill failed: "+err.Error()` messages in `handleKill`, though those are behind the loopback+control-token guard and therefore lower urgency.

### LINKED ISSUE #1032: [rev-20260621T0105Z-f82896] checkOrigin validates only the first Origin header; multiple Origin headers allow DNS-rebinding guard bypass
**Location:** `internal/transport/http_security.go:115-120`
**Reference:** https://github.com/eunolabs/eunox/blob/64b2814/internal/transport/http_security.go#L115-L120
**Severity:** Medium
**Session:** rev-20260621T0105Z-f82896

---

## Description

`checkOrigin` correctly distinguishes an absent `Origin` header (allowed) from a present one (must pass `originAllowed`), using `r.Header.Values("Origin")` to detect presence. However, only the **first** value is then validated:

```go
vals := r.Header.Values("Origin")
if len(vals) == 0 {
    return true           // absent → OK
}
origin := vals[0]        // ← only the first header is ever checked
if p.originAllowed(origin) {
    return true
}
```

The code comment on the function states: "Any present value — including the empty string and the opaque 'null' origin…— must satisfy `originAllowed` or the request is rejected with 403." The implementation does not match this intent: if a client sends two `Origin` headers, only the first is evaluated. The audit record (line 130) also records only `vals[0]`, so the second header leaves no trace.

### Attack scenario

A non-browser MCP client (e.g. a compromised tool subprocess with local network access to the loopback listener) can craft a request with:

```
Origin: http://localhost
Origin: http://evil.example.com
```

The proxy checks `vals[0] = "http://localhost"`, which matches `allowedOriginHosts`, and admits the request. The second `Origin: http://evil.example.com` is silently ignored. While this is not the classic DNS-rebinding attack vector (which requires a browser, and browsers enforce the single-Origin rule), it bypasses the guard for any client that can reach the proxy directly.

The MCP Streamable HTTP spec requires servers to validate the Origin header precisely to prevent a malicious same-host process from impersonating a browser session — which is exactly the class of client that can send arbitrary headers.

## Trigger / repro

1. Start `eunox-mcp proxy` in HTTP mode with no `listen.allowedOrigins` override (default loopback-only allowlist).
2. Send a POST to `/mcp` with two headers: `Origin: http://localhost` and `Origin: http://evil.com`.
3. Observe that the request is accepted (200/202) rather than rejected with 403.

## Suggested fix

Reject the request if **any** Origin header fails `originAllowed`, and also reject if more than one Origin header is present (RFC 6454 forbids multiple Origin headers; their presence is itself anomalous):

```go
vals := r.Header.Values("Origin")
if len(vals) == 0 {
    return true
}
// Reject if multiple Origin headers are present (RFC 6454 §7.1 forbids this;
// a legitimate client never sends more than one).
if len(vals) > 1 {
    // treat the same as a rejected origin
    // ... write audit record and 403
    return false
}
origin := vals[0]
if p.originAllowed(origin) {
    return true
}
// ... existing rejection path
```

Alternatively, iterate all values and require every one to pass `originAllowed` — this is strictly equivalent since a legitimate client sends exactly one, but makes the "all present values must satisfy the check" intent from the comment explicit in the code.

### LINKED ISSUE #1033: [rev-20260621T0128Z-e57291] `serverVersion` pin syntax undocumented; FM-4 log example implies unsupported semver format that is permanently unsatisfiable
## Location

- `docs/capability-manifest-guide.md` line 1737 (FM-4 log example)
- `internal/drift/drift.go` lines 623–650 (`matchServerVersion`)

## Reference

https://github.com/eunolabs/eunox/blob/64b2814d89fc749d43164a9b4049e7839850c2ae/internal/drift/drift.go#L623-L650

https://github.com/eunolabs/eunox/blob/64b2814d89fc749d43164a9b4049e7839850c2ae/docs/capability-manifest-guide.md#L1737

## Severity

**Medium** — can cause a full session-blocking DoS under `strictDrift: true` for any operator who follows the docs example

## Description

`matchServerVersion` only accepts dot-separated version components where any component may be `"*"` as a "wildcard from here on" — for example `"1.2.*"`, `"1.*"`, or `"*"`. It does **not** support comparison operators like `>=`, `>`, `<=`, `~`, or any semver range syntax.

The FM-4 example log line in `capability-manifest-guide.md` shows:

```
[eunox-mcp] WARN drift=fm4 serverVersion=">=2.0.0" actual="1.4.2" — server version does not satisfy manifest pin
```

The `serverVersion=">=2.0.0"` portion is the value from the manifest's `serverVersion` field as echoed in the log. An operator reading this example will reasonably conclude that `>=2.0.0` is valid syntax.

It is not. `matchServerVersion(">=2.0.0", anyRealVersion)` splits on `.` to yield `[">=2", "0", "0"]`. No real server version string begins with the literal component `">=2"`, so this constraint can **never** be satisfied. Under `strictDrift: true` (FM-4 is fatal) the proxy aborts every session establishment with a startup error — a complete traffic blackout for the route — until the manifest is corrected.

The `capability-manifest-guide.md` does not document the valid `serverVersion` syntax anywhere. The only authoritative description is buried in `matchServerVersion`'s godoc:

```
// "1.2.3"  — exact match
// "1.2.*"  — major 1, minor 2, any patch
// "1.*"    — major 1, any minor and patch
// "*"      — any version
```

## Trigger / Repro

1. Write a manifest with `serverVersion: ">=2.0.0"` (following the docs example)
2. Run `eunox-mcp proxy --strict-drift`
3. Every client session immediately fails — the pin is never satisfied

## Suggested Fix

Two changes:

1. **In `capability-manifest-guide.md`**: Add a subsection or inline explanation of the `serverVersion` pin syntax:
   > `serverVersion` accepts a dot-separated version string. Any component may be `*` to wildcard that position and everything to its right: `"1.2.3"` (exact), `"1.2.*"` (any patch), `"1.*"` (any minor/patch), `"*"` (any version). Semver range operators (`>=`, `~`, `^`) are not supported.

2. **In `capability-manifest-guide.md` line 1737**: Change the FM-4 example to use a valid constraint format, e.g.:
   ```
   [eunox-mcp] WARN drift=fm4 serverVersion="1.4.*" actual="1.5.2" — server version does not satisfy manifest pin
   ```

Optionally, add validation in `LoadManifest` / `validateLocalManifest` to reject `serverVersion` values that contain comparison operators or other invalid characters, returning an error at manifest-load time rather than silently misbehaving at runtime.

### LINKED ISSUE #1034: [rev-20260621T0128Z-e57291] `conditionArgumentNames` in drift.go lacks exhaustive handling — new argument-carrying condition types silently bypass FM-3 coverage
## Location

`internal/drift/drift.go` lines 481–515 (`conditionArgumentNames`)

## Reference

https://github.com/eunolabs/eunox/blob/64b2814d89fc749d43164a9b4049e7839850c2ae/internal/drift/drift.go#L481-L515

## Severity

**Low / Latent** — not a current bug (all five argument-carrying condition types are handled), but a structural trap that silently degrades FM-3 coverage when a new condition type is added

## Description

`conditionArgumentNames` extracts argument field values from a condition list for FM-3 drift checking. It is a type-switch that currently covers:

- `capability.AllowedValuesCondition` / `*capability.AllowedValuesCondition`
- `capability.AllowedOperationsCondition` / `*capability.AllowedOperationsCondition`
- `capability.AllowedExtensionsCondition` / `*capability.AllowedExtensionsCondition`
- `capability.AllowedTablesCondition` / `*capability.AllowedTablesCondition`
- `capability.RecipientDomainCondition` / `*capability.RecipientDomainCondition`

All five currently-defined argument-carrying condition types in `pkg/capability/condition.go` are covered. However, the switch has **no `default` branch** — any condition type not listed is silently skipped with no warning or compile-time error.

When a new condition type with an `Argument` field is added to `pkg/capability/condition.go` (and wired into the engine), a developer who forgets to update `conditionArgumentNames` will produce a build that:

1. Accepts the new condition type in manifests
2. Evaluates it at runtime (the engine gets it through separate handler registration)
3. **Silently omits it from FM-3 schema-drift checking**

The manifest author's `argument: "newParam"` pin will no longer be verified against the live `inputSchema`, so if the upstream renames `newParam`, the condition silently stops enforcing — without FM-3 warning. This contradicts the load-bearing fail-closed invariant that underpins FM-3.

The same function's sibling `pinnedArgumentNames` also loops through `conditionArgumentNames`, so the gap propagates to both the drift check and the `validate --live` report.

## Trigger / Repro

1. Add a new condition type with an `Argument` string field to `pkg/capability/condition.go`
2. Register its handler in `pkg/enforcement`
3. Write a manifest using the new condition with `argument: "someParam"`
4. Run `eunox-mcp proxy` against an upstream that has renamed `someParam` to `newParam`
5. FM-3 **does not fire** — the drift is invisible

## Suggested Fix

Add a `default` branch to the type switch in `conditionArgumentNames` that panics (or logs a hard warning) on an unrecognized condition type implementing the `Condition` interface with a non-empty `Argument`-like field. Since this is a build-time-detectable gap, a compile-time approach is preferable:

Define an `argumentNamer` interface:

```go
type argumentNamer interface {
    ArgumentName() string
}
```

Implement it on every condition type that carries an `Argument` field, then replace the type switch with:

```go
for _, cond := range conditions {
    if n, ok := cond.(argumentNamer); ok {
        add(capability.ArgumentRootKey(n.ArgumentName()))
    }
}
```

This makes the coverage automatic: any future condition type that implements `argumentNamer` is picked up without a manual `conditionArgumentNames` update, and any that does NOT implement it is correctly excluded.

### LINKED ISSUE #1035: [rev-20260621T0128Z-e57291] JWT `resource:https://api/*` claims silently deny all requests with query strings — no wildcard syntax exists for the query component
## Location

`internal/pdp/jwt.go` lines 986–996 (`matchClaimBare`)

## Reference

https://github.com/eunolabs/eunox/blob/64b2814d89fc749d43164a9b4049e7839850c2ae/internal/pdp/jwt.go#L986-L996

## Severity

**Medium** — a silent enforcement gap that causes operators to under-grant access in JWT mode, resulting in unexpected `AUTHORIZATION_FAILED` denials for legitimate requests

## Description

`matchClaimBare` applies special split logic for http(s) resource claims: the resource URI is split at its first `?` and the query component is compared **exactly** (not as a glob). The path portion retains glob semantics.

```go
pPath, pQuery, pHasQuery := strings.Cut(bareName, "?")
tPath, tQuery, tHasQuery := strings.Cut(targetName, "?")
if pHasQuery != tHasQuery || pQuery != tQuery {
    return false
}
return matchBare(pPath, tPath)
```

The consequence is that a JWT capability claim like:

```
"resource:https://api/search/*"
```

only matches requests whose target URI has **no query string**. A request to `https://api/search/results?page=2` is denied because `pHasQuery=false` but `tHasQuery=true`.

There is no way to write a JWT claim that grants access to `https://api/*` regardless of query string:

- `"resource:https://api/*"` — denied for any URL with a query string (`pHasQuery != tHasQuery`)
- `"resource:https://api/*?*=*"` — because the claim is an http(s) resource, `splitV2Claim` treats the whole string including `?*=*` as the resource name (not a condition suffix); `matchClaimBare` then compares query `"*=*"` literally against the target's query, which fails for `"page=2"`

This creates a footgun where operators writing broad-access claims for REST APIs — where nearly every real resource request includes query parameters — are surprised to find access entirely denied despite having issued what looks like a permissive claim.

The `matchClaimBare` comment says "the path portion keeps glob semantics (so an intended `https://api/*` still works)" — but "still works" means only for query-string-free URLs, a restriction the comment does not state.

## Trigger / Repro

1. Issue a JWT with `mcp.capabilities: ["resource:https://api/*"]`
2. Call `resources/read` on URI `https://api/data?format=json`
3. Receive `AUTHORIZATION_FAILED` despite the wildcard claim

## Suggested Fix

**Option A (behavior change, broadens scope):** When the CLAIM has no query (`pHasQuery=false`), treat the match as "any query string is allowed" — the claim author wrote a path-only wildcard and implicitly accepts all query strings:

```go
if pHasQuery && (pHasQuery != tHasQuery || pQuery != tQuery) {
    return false
}
return matchBare(pPath, tPath)
```

This makes `"resource:https://api/*"` match both `https://api/data` and `https://api/data?format=json`, while an explicit `"resource:https://api/data?format=json"` still requires the exact query.

**Option B (documentation only, no behavior change):** Document the exact-match query behavior clearly in:
- The JWT PDP package-level comment (the "Exception for http(s) resources" section)
- `matchClaimBare`'s godoc (fix the misleading "still works" wording)
- `docs/capability-manifest-guide.md` (add a note explaining http(s) resource claim query semantics)

Option A is the more operationally useful fix; Option B preserves the current fail-closed posture but at minimum must be clearly documented so operators are not surprised.

### LINKED ISSUE #1036: [rev-20260621T0206Z-1aacde] MatchValueGlob path-confinement guard normalizes backslashes but not percent-encoding, so %2e/%2f traversal smuggles past a "/subtree/**" allowedValues constraint
**Location:** `pkg/enforcement/glob.go:62-77` (the path-confinement guard inside `MatchValueGlob`)

**Reference:** https://github.com/eunolabs/eunox/blob/64b2814d89fc749d43164a9b4049e7839850c2ae/pkg/enforcement/glob.go#L62-L77

**Severity:** medium (security / defense-in-depth; conditional on the upstream percent-decoding the value)

**Description:**

`MatchValueGlob` confines a path-style `allowedValues` pattern (one containing `/`, e.g. `/reports/**`) to its subtree by rejecting any value whose `/`-split segments contain `.` or `..`. Before that scan it deliberately normalizes one separator-smuggling vector — backslashes — to forward slashes, with this rationale:

```go
// An upstream (Windows, or any path library that calls filepath.Clean)
// treats '\' as a separator, so a value like
// "/reports/..\\..\\etc\\passwd" splits on '/' into segments none of
// which equal "..", smuggling a traversal past the guard while still
// matching "/reports/**". Folding '\' to '/' first surfaces those
// segments ...
value = strings.ReplaceAll(value, "\\", "/")
for _, seg := range strings.Split(value, "/") {
    if seg == ".." || seg == "." {
        return false
    }
}
```

The exact same reasoning applies to **percent-encoding**, which the guard does not handle. An upstream that URL/URI-decodes the argument value before using it as a path — which is the normal behavior for any tool that fetches a URL, reads a `file://`/URI resource, or otherwise treats the value as a percent-encoded path — sees `%2e` as `.` and `%2f` as `/`. A value that encodes its traversal segments therefore passes the guard (no segment literally equals `..`) yet matches `/reports/**` and resolves, after the upstream decodes it, to a path outside the confined subtree.

This is the identical smuggling class the backslash normalization was added to close; percent-encoding is simply the other common separator/dot alias. The guard's own stated purpose ("a value carrying a `.` or `..` segment can traverse OUT of that subtree on the upstream's filesystem") is defeated.

The same un-decoded-value gap exists in the sibling `handleAllowedExtensions` (`pkg/enforcement/handlers.go:789`), which likewise normalizes only `\` → `/` before deriving the extension/base name; a `%2e`-encoded path could similarly defeat the "match the file name, not the directory path" property there. Both share the root cause: the matcher compares the literal (still-encoded) text while the upstream acts on the decoded text.

**Trigger / repro:**

Manifest constraint on a path/URI-style argument:

```yaml
- type: allowedValues
  argument: path
  values: ["/reports/**"]
```

Call the tool with `path = "/reports/..%2f..%2fetc%2fpasswd"` (or `"/reports/%2e%2e/%2e%2e/etc/passwd"`).

- `MatchValueGlob("/reports/**", "/reports/..%2f..%2fetc%2fpasswd")`:
  1. not the whole-pattern `*`/`**` wildcard;
  2. pattern contains `/` → normalize backslashes (none), split value on `/` → `["", "reports", "..%2f..%2fetc%2fpasswd"]`; **no segment equals `..`**, so the guard passes;
  3. pattern contains `**`, roots agree → `matchGlobSegments(["","reports","**"], ["","reports","..%2f..%2fetc%2fpasswd"])` → `**` matches the final segment → **true (allowed)**.
- If the upstream percent-decodes `path`, it opens `/reports/../../etc/passwd` → `/etc/passwd` — traversal out of the confined subtree, despite the guard.

Because this is a runtime value (not a manifest pattern), no load-time validation catches it.

**Suggested fix:**

Before the segment scan (and before the subsequent match), normalize percent-encoding the same way backslashes are normalized — at minimum decode the dot/slash aliases so the guard sees the segments the upstream will. Concretely, fold `%2e`/`%2E` → `.` and `%2f`/`%2F` → `/` (and ideally a single `url.PathUnescape` pass, failing closed on a malformed escape), then run the existing `.`/`..` segment rejection on the decoded value. Apply the identical normalization in `handleAllowedExtensions` so the two stay in lock-step (the file already notes that handler "normalizes for the identical smuggling class"). Add allow+deny table tests covering `%2e%2e`, `..%2f`, mixed-case `%2E`/`%2F`, and double-encoding.

Alternatively, if percent-decoding is deemed out of scope, document explicitly (in `glob.go` and the manifest guide's allowedValues section) that the confinement guard operates on the literal value and that operators must not rely on `"/subtree/**"` for upstreams that percent-decode — but the backslash precedent strongly suggests closing it rather than documenting around it.

**Session:** rev-20260621T0206Z-1aacde

### LINKED ISSUE #1037: [rev-20260621T0205Z-7144ee] MEDIUM: PDP-layer deny responses lack RequestID and DecidedAt — incomplete EnforceResponse for library callers
**Location:** `internal/pdp/pdp.go:142-153` (`denyResponse` helper) and all inline deny constructions throughout `internal/pdp/pdp.go` and `internal/pdp/jwt.go`

**Reference:** https://github.com/eunolabs/eunox/blob/main/internal/pdp/pdp.go#L142-L153

**Severity:** medium

**Description:**

`denyResponse()` is the central deny-path helper for the PDP layer. It returns an `EnforceResponse` with no `RequestID` or `DecidedAt`:

```go
func denyResponse(code, condType, message string) capability.EnforceResponse {
            Code:          code,
            ConditionType: condType,
            Message:       message,
        },
    }
}
```

By contrast, `newAllowResponse()` (lines 214–220) explicitly stamps both fields:

```go
func newAllowResponse(clock enforcement.Clock) capability.EnforceResponse {
        DecidedAt: clockNow(clock).UTC().Format(time.RFC3339Nano),
    }
}
```

The comment above `newAllowResponse` (lines 203–213) even warns that a blank `RequestID`/`DecidedAt` breaks "join-by-requestID, per-request latency, and replay detection" for library callers — yet the same logic was never applied to deny responses.

The engine layer in `pkg/enforcement/engine.go` correctly stamps both fields on every deny response it produces (e.g. `ValidateAction` generates `requestID := uuid.New().String()` at line 590 and stamps it on all deny returns). The PDP layer is inconsistent with this established pattern.

Affected deny paths (non-exhaustive):

| Location | Trigger |
|---|---|
| `pdp.go:116-130` (`killCheck`) | kill-switch activation |
| `pdp.go:483-490` | no-match deny |
| `pdp.go:514-521` | stray-schema deny |
| `pdp.go:547-556` | description-hash mismatch |
| `pdp.go:591-598` | action-check failure |
| `pdp.go:611-621` | schema validation failure |
| `pdp.go:644-647` | evaluateAndRecord error |
| `pdp.go:866` | `ManifestPDP.DecideSampling` denial |
| All deny branches in `jwt.go` | JWT condition failures |

**Impact:** Any library caller that embeds the enforcement engine and inspects `EnforceResponse.RequestID` or `EnforceResponse.DecidedAt` for correlation, latency measurement, or replay detection gets zero-value strings for every deny decision that flows through the PDP layer. The `pkg/capability` struct definition and the engine documentation both imply these fields are populated on all responses. The internal transport path (`internal/transport/forward.go`) does not use these fields (it generates its own IDs via `audit.nextRequestID()`), so there is no impact on the internal proxy behavior — but the exported contract is broken for library callers.

**Trigger / repro:**

```go
pdp := pdp.NewManifestPDP(...)
resp, _ := pdp.Decide(ctx, req)
if resp.Decision == capability.DecisionDeny {
    fmt.Println(resp.RequestID) // prints "" — zero value, useless for correlation
    fmt.Println(resp.DecidedAt) // prints "" — zero value, useless for latency
}
```

**Suggested fix:**

Change `denyResponse` to accept an `enforcement.Clock` and stamp both fields, mirroring `newAllowResponse`:

```go
func denyResponse(clock enforcement.Clock, code, condType, message string) capability.EnforceResponse {
        DecidedAt: clockNow(clock).UTC().Format(time.RFC3339Nano),
            Code:          code,
            ConditionType: condType,
            Message:       message,
        },
    }
}
```

Update all call sites of `denyResponse` (and the inline deny constructions) to pass the clock. For `killCheck`, thread the clock in or use `time.Now().UTC()` directly if the clock is not available there. Add a test asserting that every PDP deny response has a non-empty `RequestID` and `DecidedAt`.

**Session:** rev-20260621T0205Z-7144ee

### LINKED ISSUE #1038: [rev-20260621T0205Z-7144ee] LOW: "SAMPLING_DENIED" error code is a magic string — not exported as a constant in pkg/capability/errors.go
**Location:** `internal/pdp/jwt.go:659` and `internal/pdp/pdp.go:866`

**Reference:**
- https://github.com/eunolabs/eunox/blob/main/internal/pdp/jwt.go#L659
- https://github.com/eunolabs/eunox/blob/main/internal/pdp/pdp.go#L866

**Severity:** low

**Description:**

`pkg/capability/errors.go` defines sixteen denial codes as exported constants:

```go
const (
    ErrCodeAuthorizationFailed  = "AUTHORIZATION_FAILED"
    ErrCodeCapabilityDenied     = "CAPABILITY_DENIED"
    ErrCodeInvalidParams        = "INVALID_PARAMS"
    ErrCodeConditionFailed      = "CONDITION_FAILED"
    ErrCodeKillSwitch           = "KILL_SWITCH"
    ErrCodeKillSwitchError      = "KILL_SWITCH_ERROR"
    ErrCodeRevoked              = "REVOKED"
    ErrCodeExpired              = "EXPIRED"
    ErrCodeMissingContext       = "MISSING_CONTEXT"
    ErrCodeInvalidRequest       = "INVALID_REQUEST"
    ErrCodeRateLimited          = "RATE_LIMITED"
    ErrCodeNoJWTClaims          = "NO_JWT_CLAIMS"
    ErrCodeOperationNotPermitted = "OPERATION_NOT_PERMITTED"
    ErrCodeValueNotPermitted    = "VALUE_NOT_PERMITTED"
    ErrCodeEnforcementError     = "ENFORCEMENT_ERROR"
    ErrCodeAuditUnavailable     = "AUDIT_UNAVAILABLE"
)
```

`"SAMPLING_DENIED"` is not in this list. It appears as a bare string literal in two separate call sites:

```go
// internal/pdp/jwt.go:659
return denyResponse("SAMPLING_DENIED", "", "server-initiated sampling requires ...")

// internal/pdp/pdp.go:866
return denyResponse("SAMPLING_DENIED", "", "sampling is not permitted ...")
```

This violates the established pattern where every denial code in the codebase is an exported constant from `pkg/capability/errors.go`. Consequences:

1. **Typo silently accepted.** A typo in one of the two literals produces a subtly different denial code observable on the wire and in audit logs — and goes unnoticed because the compiler cannot detect string-constant mismatches.

2. **`denialToJSONRPCCode` mapping.** The wire-code mapping function in `internal/transport/jsonrpc.go` has an explicit `case` for every symbolic code. `"SAMPLING_DENIED"` is not in that switch, so it silently falls through to the `default` arm (`JSONRPCCodeCapabilityDenied`, -32002) rather than whatever code was intended. Adding a constant would force a decision about the correct mapping at the same time.

3. **Tests reference it by literal.** Any test that asserts the denial code for sampling denials must duplicate the string literal (or happen to import it from where it is used), rather than referencing the constant — a maintenance hazard.

**Trigger / repro:**

```
$ grep -rn "SAMPLING_DENIED" --include="*.go"
internal/pdp/jwt.go:659:    return denyResponse("SAMPLING_DENIED", "", ...)
internal/pdp/pdp.go:866:    return denyResponse("SAMPLING_DENIED", "", ...)
```

No definition found in `pkg/capability/errors.go`.

**Suggested fix:**

Add to `pkg/capability/errors.go`:

```go
ErrCodeSamplingDenied = "SAMPLING_DENIED"
```

Replace both literal usages:

```go
// internal/pdp/jwt.go:659
return denyResponse(capability.ErrCodeSamplingDenied, "", "server-initiated sampling requires ...")

// internal/pdp/pdp.go:866
return denyResponse(capability.ErrCodeSamplingDenied, "", "sampling is not permitted ...")
```

Add an explicit `case capability.ErrCodeSamplingDenied:` in `denialToJSONRPCCode` with the intended JSON-RPC error code (likely `-32002` to match other policy denials, or a documented choice), so the mapping is explicit rather than falling through the `default` arm.

**Session:** rev-20260621T0205Z-7144ee

### LINKED ISSUE #1039: [rev-20260621T0219Z-2f5e1f] TokenCache.Put runs an O(n) scan under the write lock on every new-key insert into a full live cache
**Location:** `pkg/capability/token_cache.go:395-410`
**Reference:** https://github.com/eunolabs/eunox/blob/64b2814d89fc749d43164a9b4049e7839850c2ae/pkg/capability/token_cache.go#L395-L410
**Severity:** medium

**Description:**
`Put` is carefully structured into a phase-1 read-locked O(n) scan and a phase-2 write-locked O(1) reclaim/evict, specifically so the expensive scan stays *off* the exclusive write lock that the validation hot path (`Get`) contends on. The "fallback O(n) sweep" at lines 400-410 is gated only on:

```go
    for el := c.insertOrder.Front(); el != nil; { ... } // O(n) under c.mu.Lock()
}
```

The comment claims this branch "runs only if STILL at capacity, which … means the cache filled between the read-lock scan and now (so phase 1 under-collected). In the steady state phase 1 already collected the expired middle/rear entries and this is skipped."

That reasoning is wrong for the most common steady state of a busy cache: **a full cache of live (non-expired) entries.** In that case nothing is expired, so:
- phase 1 collects `maybeExpired = nil`,
- the contiguous front-reclaim (360-372) removes nothing (front is live),
- the `maybeExpired` delete loop (387-394) removes nothing,
- `c.insertOrder.Len()` is therefore still `>= MaxSize`, so the **fallback O(n) sweep runs**, scanning all `MaxSize` entries under the write lock and deleting none,
- the O(1) eviction loop (413-421) then actually makes room.

So the O(n) sweep — the exact work the two-phase design was built to keep off the write lock — executes on **every** new-key `Put` whenever the cache is full of live entries, defeating the optimization and serializing all concurrent `Get` calls behind an O(MaxSize) scan.

**Trigger / repro:**
A gateway under sustained load with more than `MaxSize` (default 4096) distinct tokens active within `MaxEntryTTL` (default 30s) keeps the cache full of live entries. Each newly-seen token is a new-key `Put` that hits this path. Concretely:
1. `NewTokenCache` with `MaxSize: N`.
2. Insert N distinct tokens, all with TTLs > a few seconds (so none expire).
3. Insert the (N+1)-th distinct token → the write-locked loop at 400-410 scans all N entries and removes nothing before the O(1) eviction runs. Repeats for every subsequent distinct token.

**Suggested fix:**
Only run the fallback sweep when there is actual evidence of uncollected expired entries (e.g. track whether phase 1 was forced to give up / the cache grew between phases), or drop the fallback sweep entirely and rely on the O(1) eviction loop plus background `purgeExpired` — an expired middle/rear entry surviving one extra `Put` only costs a single avoidable live-entry eviction, which is far cheaper than an O(n) write-locked scan on every insert. At minimum, gate it on `len(maybeExpired) > 0 || grewSincePhase1` rather than on capacity alone.

**Session:** rev-20260621T0219Z-2f5e1f

### LINKED ISSUE #1040: [rev-20260621T0205Z-7144ee] LOW: Duplicated phrase in comment in internal/pdp/jwt.go
**Location:** `internal/pdp/jwt.go` approximately lines 1558–1560

**Reference:** https://github.com/eunolabs/eunox/blob/main/internal/pdp/jwt.go

**Severity:** low (cosmetic / documentation quality)

**Description:**

A comment in `evaluateJWTConditions` contains a duplicated phrase:

```
// keeping the two paths from diverging keeping the two paths from diverging
```

The phrase "keeping the two paths from diverging" appears twice in the same comment line, making the sentence nonsensical. This is almost certainly a copy-paste artifact from a refactor.

**Trigger / repro:**

```
$ grep -n "keeping the two paths" internal/pdp/jwt.go
```

Returns a line where the phrase appears twice.

**Suggested fix:**

Deduplicate the comment to read:

```go
// keeping the two paths from diverging
```

or incorporate it into a complete sentence that describes the invariant it is guarding, consistent with the codebase's convention of writing comments only when the WHY is non-obvious.

**Session:** rev-20260621T0205Z-7144ee

### LINKED ISSUE #1041: [rev-20260621T0219Z-2f5e1f] MsgKey does not canonicalize numerically-equal JSON-RPC ids (5 vs 5.0 vs 5e0 map to different keys)
**Location:** `internal/mcp/jsonrpc.go:139-147`
**Reference:** https://github.com/eunolabs/eunox/blob/64b2814d89fc749d43164a9b4049e7839850c2ae/internal/mcp/jsonrpc.go#L139-L147
**Severity:** low

**Description:**
`MsgKey` exists to produce a stable correlation key so that "two spellings of the same value … produce the same key", preventing a server-initiated request from being dropped "when a conforming host re-serializes the same value differently". The string branch achieves this by JSON-decoding the id (so `"a"` and `"a"` both yield `s:a`). The numeric branch, however, keys on the **raw `json.Number` literal**:

```go
var n json.Number
dec := json.NewDecoder(bytes.NewReader(raw))
dec.UseNumber()
if err := dec.Decode(&n); err == nil && dec.Decode(new(json.RawMessage)) == io.EOF {
    return "n:" + string(n)
}
```

`json.Number` is the verbatim source text, not a canonical numeric form. The comment claims this keys numbers "by the canonical json.Number literal so equal numbers share a key", but `json.Number("5")`, `json.Number("5.0")`, and `json.Number("5e0")` are distinct strings, so numerically-equal ids produce **different** keys (`n:5`, `n:5.0`, `n:5e0`). This contradicts the documented contract and is the exact failure mode the string branch was written to avoid.

**Trigger / repro:**
A server-initiated request (e.g. upstream `sampling/createMessage`) is tracked under `MsgKey(id)` with `id = 5`, keyed `n:5`. The proxy relays it to the host; the host returns its response with the same id re-serialized as `5.0` (or `5e0`). `MsgKey` of the response is `n:5.0`, which misses the pending-request map and the upstream request hangs. Unit-level repro: `MsgKey(RawJSON("5")) != MsgKey(RawJSON("5.0"))` even though both denote the integer 5.

Real-world likelihood is low — conforming JSON-RPC hosts echo ids verbatim and the proxy typically generates integer ids — which is why this is filed as low severity. But the behavior diverges from the stated guarantee, so a host that normalizes numbers (or a float id) silently breaks correlation.

**Suggested fix:**
Canonicalize the numeric literal before keying, e.g. parse to a normalized form (`big.Rat`/`big.Float`, or `int64` when integral with a float fallback) and emit a canonical string, so `5`, `5.0`, and `5e0` collapse to one key. Alternatively, update the comment to state explicitly that numeric ids are matched by raw literal and are NOT canonicalized, so the contract is not overstated.

**Session:** rev-20260621T0219Z-2f5e1f

eunolabs:
This batch fixes every open review-session bug filed today (sessions `rev-20260621T0010Z`–`T0219Z`), one commit per issue, with regression tests and docs in the same PR. All changes preserve the load-bearing fail-closed invariants.

## Bugs addressed

- Closes #1026 — **Dead `ErrCodeInvalidRequest` constant.** Removed the unused, unmapped symbolic denial code from `pkg/capability/errors.go` (no producer, no wire-code mapping; would have silently mapped to CAPABILITY_DENIED).
- Closes #1027 — **`init` emitted `upstreamUrl` unquoted.** Now `%q`-quoted like every sibling field, so an operator URL with a YAML-significant character round-trips verbatim. Added a round-trip regression test.
- Closes #1028 — **Remote 202 to an enforced request → bogus allow.** `DoMCPHTTP` now fails closed (transport error) when a `202 Accepted` with no body answers a request (`msg.ID != nil`); a `202` to a notification stays a valid ack. Prevents a false `RecordAllow` + a malformed result-less host response.
- Closes #1029 — **`allowedValues` glob load-validation diverged from the runtime.** Factored `enforcement.ValidateValueGlob` (the same `**`-aware segment decomposition `MatchValueGlob` uses) and validate through it, so a `"**"`-bearing pattern that whole-string `path.Match` wrongly accepts (e.g. `[a/b]/**`) is now caught at load instead of shipping a silently dead deny-all policy.
- Closes #1030 — **Audit `rotate()` fsync ordering.** Moved `s.f.Sync()` of the rotated file to **before** `os.OpenFile` of the new log, closing the crash window that could leave an empty `logPath` + an unsynced rotated tail (→ `tail_hmac_mismatch` record loss). The fallback path is unchanged (`s.f` is still open and now already synced).
- Closes #1031 — **Upstream/kill-switch error strings leaked to clients.** The `/mcp` 500 path and the `/control/kill` handlers now log the detail to stderr and return a generic client message, instead of echoing `exec.LookPath`/dial/TLS/Redis errors (which reveal command paths, internal IPs, infra).
- Closes #1032 — **DNS-rebinding guard validated only the first `Origin` header.** `checkOrigin` now rejects any request carrying more than one `Origin` header (RFC 6454 §7.1 forbids it; a header-smuggling vector), and records every value. Added a regression test.
- Closes #1033 — **`serverVersion` pin syntax undocumented; doc example unsatisfiable.** Documented the dot/`*` grammar in the manifest guide, fixed the FM-4 log example (`>=2.0.0` → `1.4.*`), and reject range/comparison operators (`>=`, `~`, `^`, `||`, …) at manifest load so the permanent-blackout footgun surfaces up front.
- Closes #1034 — **`conditionArgumentNames` could silently drop a new condition type.** Replaced the closed type switch with a `capability.ArgumentNamer` interface implemented by every argument-carrying condition; added compile-time assertions plus a reflection guard test so a future type with an `Argument` field that forgets the method fails the build/test rather than silently bypassing FM-3.
- Closes #1035 — **JWT `resource:https://api/*` claims denied every request with a query string.** A query-less http(s) resource claim is now a path-only wildcard that accepts any target query; a query-bearing claim still pins it exactly. Documented the query semantics. Added table tests.
- Closes #1036 — **Path-confinement guard normalized backslashes but not percent-encoding.** `MatchValueGlob` and `handleAllowedExtensions` now share `decodePathForConfinement` (fold `\`, percent-decode once, fail closed on a malformed escape) so `..%2f..` / `%2e%2e` traversal is surfaced and rejected, in lock-step. Documented in the guide; added allow+deny table tests (incl. mixed-case, `%5c`, double-encoding).
- Closes #1037 — **PDP-layer deny responses lacked `RequestID`/`DecidedAt`.** `denyResponse` now stamps both (clock-threaded where available, mirroring `newAllowResponse`), so library callers get complete `EnforceResponse`s on the deny path. Added a regression test.
- Closes #1038 — **`SAMPLING_DENIED` was a bare string literal.** Exported as `capability.ErrCodeSamplingDenied`, replaced both call sites, and added an explicit `denialToJSONRPCCode` case (→ -32002) rather than relying on the default arm.
- Closes #1039 — **`TokenCache.Put` ran an O(n) write-locked sweep on every full-live insert.** Gated the fallback sweep on `phase1Scanned`, so in the steady state (cache full of live entries) the expensive scan is skipped and stays off the write lock the `Get` hot path contends on. Behavior (eviction correctness) is unchanged.
- Closes #1040 — **Duplicated phrase in a `jwt.go` comment.** Deduplicated.
- Closes #1041 — **`MsgKey` did not canonicalize numerically-equal JSON-RPC ids.** Numeric ids are now canonicalized (int64 fast path, `big.Rat` fallback) so `5`, `5.0`, `5e0` share a key — a host re-serializing an id can no longer hang a server-initiated request. Updated the contract comment; added tests.

## Testing

- `make build`, `make check-fmt`, `make check-license`, `make check-notice` — clean (no new third-party modules; only stdlib `math/big`/`net/url`/`strconv` added).
- `make lint` (`go vet` + golangci-lint) — 0 issues.
- New/updated regression tests accompany each behavior change (MsgKey canonicalization, token-cache, percent-encoding traversal allow/deny tables, JWT query-wildcard, ArgumentNamer reflection guard, serverVersion load rejection, multiple-Origin rejection, 202-fail-closed, deny-response stamping, init URL round-trip).

https://claude.ai/code/session_01Td1AsoCmgrG15b8qKKbtrH

---
_Generated by [Claude Code](https://claude.ai/code/session_01Td1AsoCmgrG15b8qKKbtrH)_


internal/transport/http_remote.go
```diff
@@ -335,6 +337,17 @@ func DoMCPHTTP(ctx context.Context, client *http.Client, endpoint string, msg mc
...
 	if resp.StatusCode == http.StatusAccepted {
+		// A 202 with no body is a valid ack ONLY for an input carrying no request
+		// that expects a response — a notification (or a response). For an enforced
+		// REQUEST (msg.ID != nil) the MCP Streamable HTTP spec requires a 200 with a
+		// JSON or SSE body; a 202 here is non-compliant (or from a hostile upstream).
+		// Without this guard DoMCPHTTP would return an empty RPCMsg with nil error,
+		// which the forward core records as an allow and stamps into a malformed
+		// result-and-error-less response to the host. Fail closed with a transport
+		// error so the request denies instead.
+		if msg.ID != nil {
```
eunolabs: **`msg.ID != nil` is broader than this fix's own stated intent — consider `msg.IsRequest()`.**

The comment just above says a 202 is *"a valid ack ONLY for an input carrying no request that expects a response — a notification (or a response)."* But `msg.ID != nil` is true for both requests **and** responses (`IsResponse()` is `ID != nil && Method == ""`), so this guard also turns a 202-ack to a forwarded **response** into a transport error — the opposite of what the comment says should happen.

It isn't exploitable today: host responses to server-initiated requests are only written to subprocess upstreams via `upWriter`, never POSTed to a remote upstream — see `http_routing.go:228` and its comment, *"Remote-mode sessions issue no server-initiated requests… upWriter is never used there."* So on this path `DoMCPHTTP` only ever sends requests or notifications.

But it's a latent landmine: the day a remote upstream gains server-initiated request support (e.g. remote sampling), the host's response POSTed back would legitimately receive a 202 and this would convert the spec-correct ack into a spurious transport error. `msg.IsRequest()` (`ID != nil && Method != ""`) is functionally identical today and exactly matches the comment's "a request expecting a response" wording:

```go
if msg.IsRequest() {
    return mcp.RPCMsg{}, resp.Header, fmt.Errorf("upstream returned 202 Accepted with no body to a request expecting a response; ...")
}
```

Not blocking — just tightening the guard to its intent.

---

eunolabs (commented):
## Review — batch fix #1026–#1041

Reviewed all 16 commits (27 files, +756/-135). This is a clean, high-quality batch: one commit per bug, each with a focused regression test and a docs update where behavior or grammar changed, and the fail-closed invariants are preserved throughout. All 41 CI checks are green. I traced the non-obvious paths against the actual code rather than taking the descriptions at face value, and verified:

- **#1026** — `ErrCodeInvalidRequest` / `INVALID_REQUEST` has **zero** references anywhere in the tree; removal is safe.
- **#1030** — the `s.f.Sync()` is now genuinely *before* `os.OpenFile`, and the reopen-failure fallback no longer closes `s.f`, so the rotated fd is still valid (and now synced). The crash window is closed.
- **#1034** — there are exactly 5 `Argument`-bearing condition types, and all 5 implement `ArgumentNamer` (value receivers ⇒ both value and pointer forms satisfy it), backed by compile-time assertions **and** the reflection guard test. The closed-switch FM-3 bypass trap is genuinely closed.
- **#1037** — `clockNow` (nil → wall clock) and the `uuid` import are present; deny responses now stamp `RequestID`/`DecidedAt` like the allow path, honoring a frozen test clock.
- **#1041** — `MsgKey` int64-fast-path + `big.Rat` fallback correctly collapses `5`/`5.0`/`5e0`/`0.5e1` and the >int64 case, with the raw-bytes fallback preserved for unparseable shapes.
- **#1029 / #1036** — load validation and the runtime matcher now share the *same* structural decomposition (`ValidateValueGlob`) / confinement decode (`decodePathForConfinement`), so they can't diverge; malformed percent-escapes fail closed; the double-encoding (`%252e` → `%2e`, not `.`) behavior matches a single-decode upstream and is pinned by tests.
- **#1032** — multiple `Origin` headers are now rejected outright (RFC 6454 §7.1) with all values recorded; good catch on the header-smuggling vector.

### One thing worth tightening (non-blocking)

- **#1028** — the `if msg.ID != nil` guard in `DoMCPHTTP` is broader than the fix's own comment, which says a 202 is a valid ack for *"a notification (or a response)."* `msg.ID != nil` also matches a JSON-RPC **response**, so the guard would error on a 202-ack to a forwarded response. Not exploitable today (host responses are only routed to subprocess upstreams via `upWriter`, never POSTed to a remote upstream), but `msg.IsRequest()` matches the stated intent exactly and future-proofs remote sampling. Inline comment on the line.

### Minor / optional

- **#1038** — nice to export `capability.ErrCodeSamplingDenied` and map it explicitly in `denialToJSONRPCCode`. One residual bare `"SAMPLING_DENIED"` literal remains in the `condPDP` test mock (`internal/pdp/pdp_test.go`) — cosmetic, could use the constant for consistency.
- **#1035** — the query-less-claim → path-only-wildcard semantics also broaden an **exact-path** claim: `resource:https://api/data` (no `*`, no query) now grants `https://api/data?anything`. That's documented and is the sensible REST model, just flagging it as a deliberate widening to confirm it's intended for non-wildcard claims too.

Nothing here blocks merge. The #1028 guard is the only change I'd consider tightening before or shortly after merge; everything else looks correct and well-tested.

---

eunolabs:
Thanks for the thorough review. Addressed in `71af503`:

- **#1028** — tightened the guard to `msg.IsRequest()` (id present **and** a method), exactly matching the comment's "a request expecting a response" wording. A 202-ack to a forwarded *response* now stays a valid no-body ack, future-proofing remote server-initiated requests. Functionally identical today (`DoMCPHTTP` only sends requests/notifications on this path) but no longer a latent landmine. Existing `TestDoMCPHTTP_202ToRequestFailsClosed` still passes.
- **#1038** — replaced the residual `"SAMPLING_DENIED"` literals in the pdp test mock/assertions with `capability.ErrCodeSamplingDenied` for consistency.

On **#1035** — yes, the query-less → any-query semantics applying to an exact-path claim (`resource:https://api/data` granting `https://api/data?anything`) is intentional. Option A from the issue treats *any* query-less claim as a path-only wildcard over the query component, since there is no glob syntax for the query and a query-bearing claim still pins it exactly. It's the sensible REST model and is documented in the manifest guide. A caller who needs to pin the exact query writes it into the claim.

---

PR #1047, opened 2026-06-21
sec!: default --require-audit to strict (fail closed on audit loss)

eunolabs:
## What & why

Release-audit finding **S1** (+ **N2**). The runtime audit gate was opt-in: a dropped or write-failed audit record was counted and warned, but the proxy kept forwarding privileged calls unless an operator passed `--require-audit=strict`. That left the **default posture fail-open on the allow path** (threat model §5.4.1) — the one item touching the fail-closed invariant.

Per the maintainer decision, this makes **strict the default**.

## Changes

- **`--require-audit` now defaults to `strict`** (`cmd/eunox-mcp/main.go`). A proxy started without the flag refuses to start if it cannot open the audit log/key, and once the audit trail degrades it denies every enforced call and `*/list` enumeration with `AUDIT_UNAVAILABLE` instead of forwarding an unaudited call.
- Flag mapping is **monotonic**: a bare `--require-audit` and the truthy spellings (`=true`/`=1`/`=yes`) select `strict`, so the flag never weakens the omitted default. The intermediate startup-fatal-only level is `=on`; the prior best-effort behavior is `=off`. A typo still fails closed at parse time (now leaving the value at strict, asserted by test).
- **N2:** corrected the `--audit` help text and added a threat-model note — in audit/observe mode `*/list` returns the **full upstream catalog unfiltered** (the kill switch is the only check that still hard-blocks). The old help text said "filtered", which was inaccurate for wiretap mode.
- Docs updated for the new default: `CHANGELOG.md` (BREAKING entry + migration), `docs/threat-model-mcp.md` §5.4.1, `docs/architecture.md`, `docs/conformance.md`.

## Migration

Deployments that prefer availability over audit completeness, or that run without a writable audit-log path, must pass `--require-audit=off` (warn + run unaudited) or `--require-audit=on` (startup-fatal only, no runtime gate), or set an audit path. Documented in the CHANGELOG.

## Testing

`go build ./...`, `go vet`, `gofmt -l`, and `go test -race ./cmd/eunox-mcp/` all pass. `TestAuditRequirementFlag` updated for the new default/truthy mapping and a new assertion that a rejected value cannot silently weaken the posture.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KPxfEQgR5KjtXGoEiXaTpn)_


eunolabs (commented):
## Review: `sec!: default --require-audit to strict`

The core change is correct and the fail-closed posture is preserved end to end:

- The default is set (`*requireAudit = auditRequireStrict`) **before** `fs.Var`, so `-help` reports `strict` as the default and the omitted-flag default is strict.
- `Set` leaves `*a` untouched on an invalid value, so a typo cannot weaken the posture — it stays at the strict default (and the new `main_test.go` assertion pins this).
- The bare-flag / truthy mapping is monotonic toward strict, and `required()` / `strict()` flow correctly into `openConfiguredAuditSink(...required())` and `proxyFlags.requireAuditStrict`.
- `go build ./...`, `gofmt -l`, and `go test -race -run TestAuditRequirementFlag ./cmd/eunox-mcp/` all pass locally.

Two items to address before merge — both are outside the changed hunks, so I'm raising them here rather than inline.

### 1. `README.md` (line 9) still describes the old opt-in behavior

The PR description lists docs updated (CHANGELOG, threat-model, architecture, conformance) but the README was missed. It currently reads:

> use `--require-audit` to fail fast at startup if the log cannot be opened, or `--require-audit=strict` to additionally deny forwards fail-closed once the trail degrades

That frames both behaviors as opt-in, which is now inaccurate — strict is the default. CLAUDE.md requires behavior changes to ship their docs update in the same PR, and this is the most user-facing surface of all. Suggest something like: "the proxy fails closed on audit loss by default (`--require-audit=strict`); pass `--require-audit=off` to run unaudited or `=on` for startup-fatal-only."

### 2. Fatal message in `openConfiguredAuditSink` (`cmd/eunox-mcp/main.go:600`) is now misleading

```go
fmt.Fprintf(os.Stderr, "[eunox-mcp] Fatal: --require-audit is set and audit sink could not be opened: %v\n", err)
```

With strict as the default, this fires for operators who **never set the flag**, so "`--require-audit` is set" reads as false, and the message gives no recovery path. Since this is the exact failure a BREAKING default change will surface for deployments without a writable audit path, it should name the default and the escape hatch — e.g. "audit sink could not be opened and `--require-audit` defaults to strict; set a writable audit path or pass `--require-audit=off` to run unaudited: %v". This directly serves the migration documented in the CHANGELOG.

### Note (no action required)

A side effect of `required()` now defaulting true is that `--audit` (wiretap) mode also refuses to start when the audit log can't be opened. That's defensible — an audit/wiretap tool that can't record has no value — and the default path (`~/.eunox/audit.jsonl`) is normally writable, so low risk. Worth a one-line mention in the CHANGELOG migration note if you want to be thorough.

---

PR #1057, opened 2026-06-21
sec: make a constraint's multiple maxCalls limits atomic across windows

eunolabs:
## What & why

Release-audit finding **N1**. A constraint that layered two or more `maxCalls` conditions (e.g. `30/minute` AND `500/hour`) was admitted in two phases — a non-committing check that every bucket had headroom, then a separate per-bucket commit. That left a narrow check→commit window in which two concurrent same-session requests could both pass the check and both commit, **over-admitting by up to the number of racing requests** (for `maxCalls: 1`, two simultaneous requests could both be admitted). A *single* `maxCalls` was already atomic and is unaffected.

## Changes

- **New `CallCounter.IncrementIfAllBelow`** (`pkg/capability/enforce.go`, mirrored in `pkg/callcounter`'s `LimiterStore`) — records a slot in **every** bucket only if all have headroom, otherwise records nothing and reports the blocking bucket:
  - **InMemory** (`memory.go`): one mutex held across the whole check-and-commit, with an all-or-nothing `maxKeys` pre-check and the same prune/retry-after semantics as `IncrementIfBelow`.
  - **Redis** (`redis.go`): one Lua `EVAL` over every bucket (the multi-key analogue of the single-bucket script), refreshing TTLs on the denied path and reporting the blocking bucket's `retryAfter`. The binary wires a single-node client, which evaluates the multi-key script directly.
- **Engine** (`engine.go`, `handlers.go`): a constraint with more than one deferred (`maxCalls`) condition is admitted through `commitDeferredAtomic`, which derives every bucket via a shared `maxCallsBucket` guard (extracted from `handleMaxCalls`, so single- and multi-condition paths key identically) and commits atomically. The single-condition path keeps the handler's own atomic `IncrementIfBelow`, so the now-obsolete non-committing check-only mode (`withMaxCallsCheckOnly`) is removed.

## Testing

- Backend tests: all-or-nothing admission, sibling-bucket **anti-burn** (a denied batch charges no other bucket), and input validation — for **both** InMemory and Redis (miniredis).
- `-race` concurrency proofs: 64 concurrent callers against two buckets of limit 1 admit **exactly one**, both at the backend and **end-to-end through the engine** (`TestEngine_MultiMaxCalls_AtomicUnderConcurrency`).
- Existing multi-maxCalls tests (`TestMaxCalls_MultiWindow_DenialDoesNotBurnSiblingQuota`, `…CheckOnlyDenialReportsTightRetryAfter`) still pass — same observable behavior via the atomic path.
- Full `make test` (`-race`), `make lint` (0 issues), `make check-license`, `make check-notice` (8 modules, no new deps), `gofmt`, `go vet` all green. `pkg/` coverage 95.2% (new method 98% / 95.8%).

Docs + CHANGELOG updated: the multi-`maxCalls` TOCTOU is now **closed**, not a documented residual risk (`docs/threat-model-mcp.md` §3, `docs/capability-manifest-guide.md`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01KPxfEQgR5KjtXGoEiXaTpn)_


pkg/callcounter/redis.go
```diff
@@ -316,6 +316,145 @@ func parseIncrIfBelowReply(res interface{}) (count int64, admitted bool, retryAf
...
+  redis.call('ZADD', KEYS[i], now, ARGV[base+2])
+  redis.call('EXPIRE', KEYS[i], ARGV[base+4])
+end
+return {1, 0, 0, 0}
+`)
+
+// IncrementIfAllBelow admits a call against several maxCalls buckets atomically
+// (see the capability.CallCounter contract and incrIfAllBelowScript). The keys,
+// windowSecs, and limits slices are parallel and must share one non-zero length.
+func (r *Redis) IncrementIfAllBelow(ctx context.Context, keys []string, windowSecs []int, limits []int64) (admitted bool, deniedIndex int, count int64, retryAfter time.Duration, err error) {
```
eunolabs, line 393: The multi-key `EVAL` is correct for the single-node client that the binary wires today (the doc comment above calls this out). Worth confirming the failure mode for anyone who later points this at a Redis Cluster: the buckets are `callcounter:<key>:<windowSec>` with the same `<key>` but different `<windowSec>` suffixes, so they hash to different slots and the `EVAL` returns a `CROSSSLOT` error. `IncrementIfAllBelow` propagates that as `err`, and `commitDeferredAtomic` maps any counter error to a deny — so a misconfigured cluster fails **closed** (every multi-`maxCalls` constraint denies) rather than silently over-admitting. That's the safe direction; just flagging it so the cluster limitation is a loud deny rather than a subtle correctness gap if cluster wiring is ever added. A one-line note in the Redis backend's exported docs (or a `{ ... hash tag ... }` keying scheme) would make it explicit.

---

pkg/callcounter/memory.go
```diff
@@ -363,6 +363,108 @@ func (m *InMemory) IncrementIfBelow(_ context.Context, key string, windowSec int
...
+				}
+			}
+			valid = compactTimestamps(valid)
+			e.timestamps = valid
+			st.e = e
+			st.valid = valid
+			st.cur = int64(len(valid))
+		}
+		states[i] = st
+		if blocked < 0 && st.cur >= limits[i] {
```
eunolabs, line 438: Minor / behavioral note (not a blocker): `blocked` (and on Redis, `denied`) is the **first** bucket by slice order that is at/over its limit, and the reported `retryAfter` is that bucket's. If two windows are simultaneously full (e.g. 30/minute *and* 500/hour), the caller is advised to retry after the minute bucket frees even though the hour bucket still blocks — they'll just get denied again with the hour's `retryAfter` on the next attempt. This matches the single-condition semantics and is harmless (fail-closed, eventually consistent advice), so I'd only suggest a one-line comment noting "first blocking bucket, not the longest-retry one" so the choice reads as deliberate rather than incidental.

---

eunolabs (commented):
## Review — `sec: make a constraint's multiple maxCalls limits atomic across windows`

Reviewed the full diff, built (`go build ./...`), and ran `go test -race -count=1 ./pkg/callcounter/ ./pkg/enforcement/` — both green. This is a clean, well-scoped security fix that does what it claims.

### What it does
Replaces the two-phase (non-committing check pass → per-bucket commit) admission for constraints carrying >1 `maxCalls` with a single atomic backend call, `IncrementIfAllBelow`, closing the documented check→commit TOCTOU where two racing same-session requests could each observe headroom and both commit. Single-`maxCalls` constraints keep their already-atomic `IncrementIfBelow` path, and the now-obsolete `withMaxCallsCheckOnly` peek mode is removed.

### Strengths
- **The atomicity argument holds at both backends.** InMemory takes `m.mu` across the whole check-and-commit; Redis does it in one Lua `EVAL`. The `-race` proofs (64 callers, two limit-1 buckets, exactly one admitted — both at the backend and end-to-end via `TestEngine_MultiMaxCalls_AtomicUnderConcurrency`) are the right tests for this class of bug.
- **Good factoring.** Extracting `maxCallsBucket` so the single- and multi-condition paths key *identically* (same fail-closed guards: nil counter, empty `sessionId`, empty target) is the load-bearing detail that makes the atomic path safe, and it's tested via `keyCapturingCounter`. `denyFromConditionError` de-duplicates the deny-response shape cleanly.
- **Fail-closed posture preserved.** Slice-length/empty-bucket validation, per-bucket window/limit validation before any state mutation, all-or-nothing `maxKeys` pre-check in memory, defensive `parseIncrIfAllBelowReply` type-checks, and counter errors mapped to deny in `commitDeferredAtomic`. The new multi-key Lua mirrors the single-key script's prune / denied-path TTL-refresh / retry-pivot logic faithfully.
- **Conventions met.** Docs (`threat-model-mcp.md` §3, `capability-manifest-guide.md`), CHANGELOG, and tests for both backends + the engine all updated in-PR; no new modules; `sec:` commit type is appropriate.

### Notes (none blocking)
- Two inline comments below: (1) the Redis multi-key `EVAL` is single-node-only — it fails *closed* on a cluster, which is the safe direction, but worth making explicit; (2) the reported `retryAfter` is the first-by-index blocking bucket, not the longest — harmless but worth a one-line comment.
- **Mergeability:** GitHub currently reports this branch as `dirty` (conflicts with base `claude/lucid-lovelace-e9w978`). Needs a rebase/merge before it can land.

Overall: LGTM pending the rebase. The security claim is sound and the test coverage matches the failure mode it closes.

---

Issue #1063, opened 2026-06-21
Engineering review: SSE teardown hang, hot-path allocation, remote-upstream response drop


eunolabs:
## Summary

Engineering excellence review of the proxy codebase. Three actionable findings, ranked by impact. The codebase is in good shape overall; these are the changes worth making.

---

### A1 — SSE client disconnect silently hangs upstream server-initiated requests [High / S effort]

**Location**: `internal/transport/http_routing.go:298–312` + `internal/transport/http_session.go:703–718`

`handleMCPGet`'s SSE loop exits via `<-r.Context().Done()` without draining `ch`. But `broadcastServerRequest` may have already written a server-initiated request (e.g., `sampling/createMessage`, `roots/list`) into `ch` and registered its ID in `serverReqs`. The upstream goroutine is now blocked waiting for a response the proxy will never deliver — bounded only by the idle reaper (worst case: several minutes). A host that opens and closes SSE streams rapidly while upstream servers are sampling can accumulate hung goroutines.

**Fix**: Replace `defer sess.removeSub(ch)` with a combined defer that stops delivery and then drains any buffered messages, sending upstream error replies for each tracked ID:

```go
defer func() {
    sess.removeSub(ch)
    for {
        select {
        case msg := <-ch:
            if msg.ID != nil && sess.serverReqs.take(mcp.MsgKey(msg.ID)) && sess.upWriter != nil {
                _ = sess.upWriter.Write(mcp.ErrorResponse(msg.ID,
                    capability.JSONRPCCodeEnforcementError, "client disconnected"))
            }
        default:
            return
        }
    }
}()
```

`removeSub` stops new deliveries; the drain loop handles exactly the items that raced into the buffer before the subscription was removed.

---

### A2 — Remote-upstream mode silently drops host responses to server-initiated requests [Medium / M effort]

**Location**: `internal/transport/http_routing.go:226–229`

In remote-upstream mode `sess.upWriter` is nil. When `handleMCPPost` receives a JSON-RPC response whose ID matches a tracked server-initiated request, the write-back is silently skipped. The remote upstream is left blocked waiting for the host's response to (e.g.) `sampling/createMessage`. The comment acknowledges the gap but provides no mitigation.

**Immediate mitigation**: Untrack the ID and log instead of silently eating the response:

```go
if sess.serverReqs.take(mcp.MsgKey(msg.ID)) {
    fmt.Fprintf(os.Stderr,
[...]

PR #1073, opened 2026-06-21
fix: drain SSE buffer on teardown, surface remote-upstream response drops

eunolabs:
## Summary

Addresses the engineering-review findings in #1063 on the HTTP transport — the three correctness/risk items plus the two polish comments. The third polish item (B3) was already documented in the current code, so it's left untouched (noted below).

### A1 — SSE client disconnect no longer hangs upstream server-initiated requests *(correctness)*
`handleMCPGet`'s SSE loop exited on client disconnect (`r.Context().Done()`) without draining its subscriber channel. A server-initiated request (`sampling/createMessage`, `roots/list`, …) that `broadcastServerRequest` had already buffered into that channel — with its ID still tracked — was then left undelivered, blocking the upstream until the idle reaper fired (worst case: minutes). A host that opens/closes SSE streams while upstreams are sampling could accumulate hung goroutines.

New `httpSession.removeSubAndDrain`, deferred by `handleMCPGet`, unsubscribes first (stopping new deliveries; both `removeSub` and `deliverToOne` take `notifMu`, so once it returns no concurrent delivery can target the channel) and then drains exactly the requests that raced into the buffer, replying a JSON-RPC error upstream and untracking each ID so the upstream unblocks immediately.

### A2 — Remote-upstream response drop is now observable *(risk)*
In remote-upstream mode `sess.upWriter` is nil, so a host response to a tracked server-initiated request was silently dropped. The ID is now untracked (so the tracker can't leak / misroute a duplicate) and a warning is logged instead of silently eating the response. The complete fix — routing the response back through the remote upstream's pending HTTP/SSE connection — remains a separate design question for the remote bridge.

### B1 — `MsgWriter.Write` no longer allocates via `fmt.Fprintf` *(perf)*
`fmt.Fprintf(mw.w, "%s\n", data)` invoked the format machinery (format-string parse, printer pool, `%s` verb dispatch) on the single highest-frequency allocation in the proxy hot path. Replaced with an in-place `append(data, '\n')` (the slice is freshly allocated by `json.Marshal` and owned by the call) + a direct `mw.w.Write`. Framing behavior is identical (single newline-terminated line, single `Write`).

### B2 / C1 — Polish comments
- `dispatchList`: documented that the triple `msg.Method` passed to `strictAuditDenial` (audit id / method / denial target) is intentional — a `*/list` request addresses no sub-target, so the method *is* the target.
- `sourceIP`: explained why the **right-most** `X-Forwarded-For` entry is the trusted one under `--trust-forwarded-for` (the trusted edge appends the connecting IP; earlier entries are client-forgeable).

### B3 — already done
The review asked to explain `audit.Sink.Close()`'s write-lock-after-channel-close ordering. The current code already documents this (the comments at both write-lock acquisitions explain that taking the write lock waits for all in-flight read-locked `record()` senders to finish before closing the channel / reading `dropped`), so no change was needed.

## Tests
- `TestHTTPRemoveSubAndDrain_RepliesErrorForBufferedServerRequest` — a buffered server-initiated request gets an error reply routed upstream and its ID untracked; the subscription is removed.
- `TestHTTPHandleMCPPost_RemoteModeServerResponseWarnsAndUntracks` — a remote-mode (nil `upWriter`) host response is untracked and warns on stderr.
- Existing `TestMsgWriter_WriteSuccess` / `TestMsgWriter_WriteError` cover the B1 framing/error behavior.

`go test -race -count=1 ./...` passes; `go vet`, `golangci-lint` (0 issues), `gofmt`, and license-header checks are clean.

https://claude.ai/code/session_01CTdNvREorgCPEMs5AqH2VF

---
_Generated by [Claude Code](https://claude.ai/code/session_01CTdNvREorgCPEMs5AqH2VF)_


internal/transport/http_routing.go
```diff
@@ -281,7 +294,13 @@ func (p *HTTPProxy) handleMCPGet(w http.ResponseWriter, r *http.Request, route *
...
 		return
 	}
-	defer sess.removeSub(ch)
+	// On stream exit (client disconnect via r.Context().Done(), or session teardown
+	// via sess.done) unsubscribe AND drain: broadcastServerRequest may have buffered a
+	// server-initiated request into ch that the loop below never read, leaving its
+	// upstream blocked on a response this departed client will never send. The drain
+	// replies an error upstream for each so it unblocks immediately instead of hanging
+	// until the idle reaper fires.
+	defer sess.removeSubAndDrain(ch)
```
eunolabs, line 303: The concurrency design here is sound: `removeSub` removes `ch` from `notifSubs` under `notifMu`, and `deliverToOne` holds `notifMu` across its `case ch <- msg` send, so after `removeSub` returns no further delivery can target `ch` and the buffer is the complete missed set. The atomic `serverReqs.take()` then guarantees each ID is handled exactly once across the drain, `broadcastServerRequest`'s fail-closed path, and the POST response handler — nice.

One residual gap worth noting (not necessarily for this PR): the drain only recovers requests still **buffered** in `ch`. A server-initiated request that the SSE loop already read off `ch` and wrote to the wire (`fmt.Fprintf(w, "data: %s\n\n", ...)` at line ~324) but the client then disconnected without POSTing a response is *not* in the buffer — its ID stays tracked and its upstream stays wedged until the idle reaper fires. That's exactly the original failure mode this PR set out to fix, just narrowed to the "delivered-but-unanswered" window rather than fully closed. Given a delivered request fundamentally can't be auto-answered, the idle reaper is the right backstop here — but the CHANGELOG/comment wording ("no longer leaves the upstream blocked") slightly oversells it. Consider noting that the delivered-but-unanswered case still relies on the reaper.

---

internal/transport/http_session.go
```diff
@@ -772,3 +772,36 @@ func (s *httpSession) removeSub(ch chan mcp.RPCMsg) {
...
+// notifications (no ID) are dropped, which is the correct best-effort behavior.
+// Remote-upstream sessions have a nil upWriter and issue no server-initiated
+// requests, so the take/Write arms are inert there.
+func (s *httpSession) removeSubAndDrain(ch chan mcp.RPCMsg) {
+	s.removeSub(ch)
+	for {
+		select {
+		case msg := <-ch:
+			if msg.ID != nil && s.serverReqs.take(mcp.MsgKey(msg.ID)) && s.upWriter != nil {
+				_ = s.upWriter.Write(mcp.ErrorResponse(msg.ID, capability.JSONRPCCodeEnforcementError,
```
eunolabs, line 805: Minor/nit (non-blocking): the upstream receives `JSONRPCCodeEnforcementError` for what is really a transport/delivery failure (the client went away), not a policy denial. This mirrors `broadcastServerRequest`'s existing no-delivery path verbatim ("no client stream available..."), so consistency-wise it's the right call and I wouldn't change it in isolation — just flagging that the upstream can't distinguish "the proxy's policy blocked me" from "my client vanished" from the error code alone. If a more delivery-specific code exists it'd be a clearer signal, but matching the established path is reasonable.

---

internal/mcp/jsonrpc.go
```diff
@@ -247,9 +247,15 @@ func (mw *MsgWriter) Write(msg RPCMsg) error {
 	if err != nil {
 		return fmt.Errorf("marshalling message: %w", err)
 	}
+	// Frame with a trailing newline by appending in place. data was just allocated
+	// by json.Marshal and is owned solely by this call, so the append is safe and,
+	// unlike fmt.Fprintf("%s\n", data), avoids the format-machinery allocation
+	// (format-string parse, printer pool, %s verb dispatch) on this hot path — every
+	// proxied JSON-RPC message is written through here.
+	data = append(data, '\n')
```
eunolabs, line 255: B1 is correct and safe: `json.Marshal` returns a freshly allocated slice owned solely by this call (it ends with `append([]byte(nil), e.Bytes()...)`), so `append(data, '\n')` either fills spare capacity or reallocates a private backing array — no aliasing risk, unlike `json.Encoder` which reuses a buffer. Moving the `append` before `mw.mu.Lock()` also shaves a hair of work out of the critical section. Framing is byte-for-byte identical. Good change.

---

eunolabs (commented):
## Review — fix: drain SSE buffer on teardown, surface remote-upstream response drops

Reviewed the full diff against the surrounding transport/session/jsonrpc code on the head branch. This is a focused, well-reasoned set of fixes and the engineering quality is high — the comments are unusually thorough and accurate, and every behavioral change has a dedicated test. I checked the concurrency reasoning, the `append` safety claim, and the logging conventions against the existing code; all hold up. Overall: looks good to merge, with a couple of small notes inline.

### What I verified
- **A1 (`removeSubAndDrain`)** — Correct. `removeSub` and `deliverToOne` both serialize on `notifMu`, so once the unsubscribe returns the channel buffer is the complete set of missed messages, and `serverReqs.take()` (mutex-guarded, atomic) makes ID handling exactly-once across the drain, `broadcastServerRequest`'s fail-closed arm, and the POST handler. No double-reply, no lost-wakeup. Buffered notifications (no ID) are correctly dropped.
- **A2 (remote-mode drop)** — Correct and consistent. `take()` untracks before the warn so the tracker can't leak/misroute; the `fmt.Fprintf(os.Stderr, "[eunox-mcp] WARNING: ...")` form matches the established convention in `broadcast`, `track`, etc.
- **B1 (`MsgWriter.Write`)** — Safe; see inline. Ownership of the `json.Marshal` slice makes the in-place `append` correct, framing is identical.
- **B2 / C1 / B3** — Comment-only clarifications, all accurate.

### Notes (none blocking)
1. **Residual gap in A1** (inline on `http_routing.go:303`): the drain recovers *buffered-but-unread* requests; a request already streamed to the client that then disconnects without answering still hangs until the idle reaper. That's inherent, but the CHANGELOG wording slightly oversells the fix — consider noting the reaper is still the backstop for the delivered-but-unanswered case.
2. **Error-code semantics** (inline on `http_session.go:800`): `JSONRPCCodeEnforcementError` is reused for a delivery failure. Matches the existing no-delivery path, so fine to keep — just flagging.

### Conventions / process
- License headers, no-emoji, conventional-commit title, CHANGELOG entry under `[Unreleased] > Fixed` — all present and correct.
- Test coverage is good: `TestHTTPRemoveSubAndDrain_RepliesErrorForBufferedServerRequest` and `TestHTTPHandleMCPPost_RemoteModeServerResponseWarnsAndUntracks` pin the two new paths, and the stderr-mutating test correctly opts out of `t.Parallel()`. (I didn't run the suite in this review; PR states `go test -race`, vet, lint, fmt, license all clean.)

Nice work — fails-closed posture and observability are both preserved.

---

internal/transport/http_routing.go
eunolabs: Good catch — tightened both the CHANGELOG entry and the `removeSubAndDrain` doc comment in 4e29d52 to scope the fix to buffered-but-unread requests and explicitly note that a request already streamed to the client (then abandoned) still relies on the idle reaper as backstop, since a delivered request can't be auto-answered.

---

PR #1091, opened 2026-06-21
Fix session rev-20260621 review bugs (stdio handshake fail-closed)

### LINKED ISSUE #1087: [rev-20260621T2010Z-7df36c] stdio live-introspection handshake skips initialize-result validation, accepting a null/malformed handshake the HTTP path rejects
**Location:** `cmd/eunox-mcp/live_upstream.go:211-222` (`runStdioHandshake`)

**Reference:** https://github.com/eunolabs/eunox/blob/6142761e5cf1b3d9ca4be9724560f5e4b745991b/cmd/eunox-mcp/live_upstream.go#L211-L222

**Severity:** Medium (fail-closed / cross-transport consistency defect; CLI introspection path, not the enforcement hot path)

**Description:**

Every other `initialize` handshake path in the codebase funnels the response through `applyInitializeResult` / `ApplyInitializeResult`, which calls `validateInitializeResultFields` to **fail closed** on a structurally invalid result — most importantly a JSON `null` result (which unmarshals into the struct without error but leaves every field zero), or a result missing the mandatory `protocolVersion` / `capabilities` / `serverInfo` fields:

- proxy subprocess upstream: `internal/transport/stdio.go:401`
- proxy HTTP session upstream: `internal/transport/http_session.go:549`
- proxy remote-HTTP bridge: `internal/transport/http_remote.go:260`
- CLI **HTTP** live probe: `cmd/eunox-mcp/live_upstream.go:71` (`fetchLiveTools`)

The CLI **stdio** live probe (`runStdioHandshake`) is the lone exception. Instead of the shared checker it hand-rolls the parse and ignores every failure:

```go
if initResp.Error != nil {
    return LiveUpstreamInfo{}, fmt.Errorf("initialize: server error %d: %s", initResp.Error.Code, initResp.Error.Message)
}
var serverVersion string
if initResp.Result != nil {
    var initResult mcp.InitResult
    if json.Unmarshal(initResp.Result, &initResult) == nil {   // error silently ignored
        if sv, ok := initResult.ServerInfo["version"].(string); ok {
            serverVersion = sv
        }
    }
}
```

A `result: null` (or `result: {}`, or any result missing the required MCP fields) is silently accepted as a successful handshake: the code proceeds to send `notifications/initialized` and `tools/list` against an upstream whose handshake was never confirmed. This directly contradicts the comment on the sibling HTTP path a few lines up, which states the shared checker exists *"so the CLI and proxy cannot diverge on what counts as a valid initialize result (e.g. a `null` result that unmarshals to all-zero fields must be rejected, not silently accepted with an empty server version)."* The stdio CLI path is exactly that divergence, and it violates the repo's fail-closed invariant ("On any ambiguity ... deny").

**Trigger / repro:**

Point `eunox-mcp init`, `eunox-mcp validate --live --config` (stdio upstream), or `eunox-mcp doctor` at a stdio MCP subprocess whose `initialize` reply is `{"jsonrpc":"2.0","id":"_init","result":null}` (or omits `protocolVersion`/`capabilities`/`serverInfo`). The HTTP equivalent returns `initialize: upstream initialize result missing required 'protocolVersion' ...`; the stdio path instead continues the handshake and reports `ServerVersion: ""` plus whatever `tools/list` returns — masking a broken/half-speaking upstream and scaffolding a manifest from an unvalidated session.

**Why earlier review/tests missed it:** the shared `applyInitializeResult` and its fail-closed behavior are well covered (`internal/transport/jsonrpc_test.go`, `coverage_extra_test.go`), but those tests exercise the transport function directly and the CLI **HTTP** probe; no test drives `runStdioHandshake` with a `null`/malformed initialize result, so the stdio path's hand-rolled parse was never asserted to fail closed.

**Suggested fix:** route the stdio handshake through the same shared checker as the HTTP probe, replacing the manual block:

```go
if initResp.Error != nil {
    return LiveUpstreamInfo{}, fmt.Errorf("initialize: server error %d: %s", initResp.Error.Code, initResp.Error.Message)
}
_, serverVersion, _, err := transport.ApplyInitializeResult(initResp)
if err != nil {
    return LiveUpstreamInfo{}, fmt.Errorf("initialize: %w", err)
}
```

Add a `runStdioHandshake` test case with a `null` and a field-missing initialize result asserting it errors, mirroring the existing transport fail-closed tests.

**Session:** rev-20260621T2010Z-7df36c

eunolabs:
## Bugs addressed (session `rev-20260621*`)

This batches the remaining open automated-review issues whose titles carry a `rev-20260621*` session id. The earlier batch (merged #1085) handled #1075–#1084; #1087 was filed afterward and is the only one of the three remaining that is a genuine defect against current HEAD.

- **Closes #1087** — The stdio CLI live-introspection handshake (`runStdioHandshake` in `cmd/eunox-mcp/live_upstream.go`) hand-rolled its `initialize` result parse and silently ignored every failure, so a JSON `null` result (or one missing the mandatory `protocolVersion`/`capabilities`/`serverInfo` fields) was accepted as a successful handshake — diverging from every other handshake path, all of which funnel through `applyInitializeResult`/`ApplyInitializeResult` to fail closed. This violated the fail-closed invariant and let `init`/`validate --live`/`doctor` scaffold a manifest from an unconfirmed session. **Changed:** the stdio path now routes through `transport.ApplyInitializeResult`, the same checker the HTTP probe uses, so the CLI and proxy cannot diverge on what counts as a valid handshake.

### Not fixed — false positives (independently re-verified against current HEAD)

These two were already assessed as false positives in #1085 and left open; I re-verified them against current HEAD and confirm no code change is warranted:

- **#1079** — Claims `Compile()` for `ipRange` sits in the value switch case. On current HEAD (`internal/config/manifest.go`) it is correctly in the `*capability.IPRangeCondition` pointer case (the form unmarshal produces), and `TestLoadManifest_PreCompilesIPRangeCIDRs` already proves `Networks()` returns `(non-nil, true)` after `LoadManifest`. No change needed.
- **#1077** — Assumes `VerifyToken` parses a multi-signature JWS into multiple `tok.Headers`. It uses `jwt.ParseSigned` (compact serialization → exactly one signature), so `tok.Headers` is always length 1 and the multi-kid / `requireKID`-on-`headers[0]` concerns are unreachable. No change needed.

(Left open for a maintainer to close; this PR does not use closing keywords for them.)

## Testing

- New regression test `TestRunStdioHandshake_InvalidInitializeResultFailsClosed` — table-driven `null`/empty/missing-`protocolVersion`/missing-`capabilities`/missing-`serverInfo` cases, each asserting the handshake fails closed.
- Completed the stdio mock fixtures (`live_upstream_test.go`, `coverage_extra_test.go`) that previously omitted the now-required `capabilities` object so they form valid handshakes.
- `make check-fmt`, `make check-license`, `go vet ./...`, and `go build ./...` — all clean.

https://claude.ai/code/session_01SbbnXRFAggntZf2iyB6WKg

---
_Generated by [Claude Code](https://claude.ai/code/session_01SbbnXRFAggntZf2iyB6WKg)_


cmd/eunox-mcp/live_upstream.go
```diff
@@ -208,17 +208,16 @@ func runStdioHandshake(ctx context.Context, w *mcp.MsgWriter, r *mcp.MsgReader)
...
-				serverVersion = sv
-			}
-		}
+	// Validate the handshake through the same checker the proxy and the HTTP probe
+	// (fetchLiveTools) use, so the stdio CLI path cannot diverge on what counts as a
+	// valid initialize result (e.g. a `null` result that unmarshals to all-zero
+	// fields, or one missing protocolVersion/capabilities/serverInfo, must be
+	// rejected, not silently accepted with an empty server version). ApplyInitializeResult
+	// also surfaces an upstream-reported error, so the explicit Error check is folded in.
+	// Fail closed.
```
eunolabs: Per CLAUDE.md ("no multi-line comment blocks — one short line max"), this 7-line block should collapse. The WHY is genuine — without it a reader might wonder where the explicit `Error` check went — but `applyInitializeResult`'s own godoc in `jsonrpc.go` already documents that behaviour, so the explanation doesn't need to live here too. Something like:

```go
// Route through ApplyInitializeResult so null/missing-field results fail closed, same as the HTTP probe.
```

covers the constraint without restating the implementation.

---

cmd/eunox-mcp/live_upstream_test.go
```diff
@@ -543,3 +545,60 @@ func TestRunStdioHandshake_UnexpectedResponseID(t *testing.T) {
...
 }
+
+// TestRunStdioHandshake_InvalidInitializeResultFailsClosed asserts the stdio
+// introspection handshake routes its initialize result through the same shared
+// checker (transport.ApplyInitializeResult) the proxy and HTTP probe use, so a
+// structurally invalid result — a JSON `null`, or one missing a mandatory MCP
+// field — is rejected (fail closed) rather than silently accepted with an empty
+// server version. Without that check the handshake would proceed to send
+// notifications/initialized and tools/list against an upstream whose handshake
+// was never confirmed.
```
eunolabs: Same convention applies in tests. The function name already encodes "fail closed" and "invalid initialize result", so this 8-line preamble is more than needed. Could be dropped entirely, or trimmed to one line, e.g.:

```go
// Structurally invalid initialize results (null, missing mandatory field) must all be rejected.
```

The detail about what would happen without the check belongs in the PR description, not the test file.

---

eunolabs (commented):
**Correctness: fix is right.**

`ApplyInitializeResult` → `applyInitializeResult` → `validateInitializeResultFields` covers every gap the old hand-rolled code missed:
- `resp.Error != nil` → now explicit error (was: caught)
- `resp.Result == nil` (neither result nor error) → now error (was: silent success, empty version)
- JSON unmarshal failure → now error (was: silent success)
- JSON `null` result (unmarshals to zero struct) → now caught by the `protocolVersion == ""` check (was: silent success)
- Missing `capabilities` or `serverInfo` → now errors (was: not checked)

The five test cases map cleanly to these cases and the existing fixture updates (adding `Capabilities: map[string]interface{}{}` to five mock servers) are necessary — the tests were previously emitting invalid handshakes that the new stricter check would have rejected.

Two style nits on comment length (see inline); no issues with logic or coverage.

---

cmd/eunox-mcp/live_upstream.go
eunolabs: Trimmed in 701dd31. One note: I couldn't find the cited "no multi-line comment blocks — one short line max" rule in CLAUDE.md or CONTRIBUTING.md, and the file's house style is explanatory multi-line comments (e.g. the sibling HTTP-probe comment ~15 lines up, and `applyInitializeResult`'s godoc). The real redundancy was that my block duplicated that sibling comment, so I cut it to 4 lines matching it rather than a single line — keeps the "where did the Error check go" WHY local while dropping the repetition.

---

cmd/eunox-mcp/live_upstream_test.go
eunolabs: Trimmed to 3 lines in 701dd31 — dropped the "what would happen without the check" detail (that lives in the PR description) and kept a one-line statement of what the cases assert.

---

PR #1102, opened 2026-06-22
Fix session 1cby6q bugs

### LINKED ISSUE #1092: [rev-20260621T2110Z-b49001] doctor's redactURL leaves URL query-string credentials intact, leaking ?api_key=/?token= secrets into the support bundle
**Location:** `cmd/eunox-mcp/doctor.go:295-315` (`redactURL`), reached from `redactConfigValue` for the `upstreamUrl` key (`doctor.go:264-266`)

**Reference:** https://github.com/eunolabs/eunox/blob/a8c2a0937c1fdc86298ef46bcbaf1aea540172c0/cmd/eunox-mcp/doctor.go#L295-L315

**Severity:** Low (secret disclosure in an operator-shared artifact; requires a credential hardcoded in the config's `upstreamUrl` query string)

**Description:**

`doctor` builds a "support bundle" whose explicit purpose is to redact secrets so the operator can safely paste it into a bug report. Section 2 ("Config (redacted)") redacts the named secret fields (`authToken`, `upstreamAuthHeader`) to a length-only placeholder, and strips **userinfo** from any `upstreamUrl` via `redactURL`:

```go
func redactURL(s string) string {
	if s == "" {
		return s
	}
	u, err := url.Parse(s)
	if err != nil {
		return redactURLFallback(s)
	}
	if u.User == nil {
		return s
	}
	u.User = url.UserPassword("REDACTED", "REDACTED")
	return u.String()
}
```

The redaction is deliberately limited to the `user:pass@` userinfo component — the comment states it leaves "scheme/host/path/**query** intact." But the **query string** is an equally common place for an upstream credential: HTTP MCP servers (and many APIs they proxy) authenticate with `?api_key=...`, `?token=...`, `?access_token=...`. A config that hardcodes such a URL:

```yaml
upstreams:
  - name: weather
    transport: http
    upstreamUrl: https://api.example.com/mcp?api_key=sk-live-9f3c2a...
```

is emitted **verbatim** (credential and all) into the bundle, because `u.User == nil` so `redactURL` returns the string unchanged.

This is internally inconsistent with the rest of `doctor`: sections 3 (Manifests) and 5 (Live) carry explicit `NOTE:` comments (`doctor.go:363-365`, `568-569`) declaring that `UpstreamURL` "carries live secrets" and must **never** be emitted there — yet section 2 prints it with only userinfo stripped. The tool takes responsibility for URL-credential redaction (it handles userinfo) but implements it incompletely, defeating its own contract for the query-string case. The "skim before sharing" footer is a mitigation, not a fix; the same argument would justify not redacting userinfo either.

(Note: the `${ENV}`-reference case is safe — section 2 reads the config raw, so `upstreamUrl: ${UPSTREAM_URL}` stays literal. The leak is specific to a credential written directly into the config file's `upstreamUrl`.)

**Trigger / repro:**

1. Write a config whose `upstreamUrl` embeds a credential in the query string (example above), no `user:pass@` userinfo.
2. Run `eunox-mcp doctor --config eunox.yaml`.
3. Section 2 prints `upstreamUrl: https://api.example.com/mcp?api_key=sk-live-9f3c2a...` verbatim — the API key is in the bundle the operator is told is "redacted."

**Suggested fix:**

Redact sensitive query parameters (or, more conservatively, redact the entire query string to a placeholder while keeping scheme/host/path for diagnostics) in `redactURL`, and mirror it in `redactURLFallback`. For example, replace every query value with a length-tagged placeholder:

```go
if u.User != nil {
	u.User = url.UserPassword("REDACTED", "REDACTED")
}
if q := u.RawQuery; q != "" {
	vals := u.Query()
	for k, vs := range vals {
		for i, v := range vs {
			if v != "" {
				vs[i] = fmt.Sprintf("<redacted len=%d>", len(v))
			}
		}
		vals[k] = vs
	}
	u.RawQuery = vals.Encode()
}
return u.String()
```

Add a `redactURL` test case asserting a query-string credential is scrubbed, and one asserting the fallback path scrubs it too.

**Session:** rev-20260621T2110Z-b49001

### LINKED ISSUE #1093: [rev-20260621T2106Z-a75b92] IncrementIfAllBelow returns count=0 on admitted path, violating LimiterStore contract
## Bug

`IncrementIfAllBelow` always returns `count=0` on the admitted path in both the in-memory and Redis backends, violating the `LimiterStore` interface contract.

## Contract

`callcounter.go` documents the `LimiterStore.IncrementIfAllBelow` signature:

> `count` is the resulting in-window total (after the record when admitted, the current total when rejected).

The counterpart `IncrementIfBelow` correctly returns the post-admission count in both backends.

## Locations

**In-memory backend** (`pkg/callcounter/memory.go`, admitted-path return):
```go
// line 471
return true, 0, 0, 0, nil   // count is hardcoded 0
```

**Redis backend** (`pkg/callcounter/redis.go`, `incrIfAllBelowScript` Lua):
```lua
-- admitted path:
return {1, 0, 0, 0}   -- element [2] (count) is hardcoded 0
```

`parseIncrIfAllBelowReply` propagates the zero without adjustment:
```go
return true, 0, 0, 0, nil
```

Compare with `IncrementIfBelow` (single-bucket), which correctly returns the post-admission count in both backends.

## Impact

The current sole caller (`commitDeferredAtomic` in `pkg/enforcement/engine.go`) only reads `count` on the **denied** path:

```go
admitted, deniedIndex, count, retryAfter, err := e.counter.IncrementIfAllBelow(...)
if !admitted {
    return denyFromConditionError(maxCallsRateLimited(mcs[deniedIndex], count, ...), ...)
}
return nil   // count discarded
```

There is therefore no active runtime defect today. However, the contract violation is latent: any future caller that logs, exposes, or enforces on the admitted-path count will silently receive 0 instead of the true in-window total. The interface documentation makes an explicit promise that both implementations break.

## Fix

**In-memory** (`pkg/callcounter/memory.go`): On the admitted path, return the actual post-increment count for the binding bucket (or the maximum across all buckets, consistent with the single-bucket semantics). The valid-entries slice is already computed before the return.

**Redis** (`pkg/callcounter/redis.go`, `incrIfAllBelowScript`): After writing all buckets, read back the ZCARD of the bucket that hit the limit (or the first bucket, by convention) and return it as element [2]:
```lua
local count = redis.call('ZCARD', KEYS[binding_bucket])
return {1, count, 0, 0}
```

Update `parseIncrIfAllBelowReply` to propagate the returned count rather than hardcoding 0.

Add an admitted-path assertion in `callcounter_test.go` and `redis_test.go` that verifies `count > 0` after a successful `IncrementIfAllBelow` call to prevent regression.

## Reviewed files

- `pkg/callcounter/memory.go`
- `pkg/callcounter/redis.go`
- `pkg/callcounter/callcounter.go`
- `pkg/callcounter/callcounter_test.go`
- `pkg/enforcement/engine.go`

### LINKED ISSUE #1094: [rev-20260621T2124Z-0fed6c] cloneAndBound skips per-element cap for []string audit Detail values, violating the auditDetailValueCap invariant
**Location:** `internal/audit/audit.go` — `cloneAndBound`, the `case []string` branch

**Contract:**
`cloneAndBound` is the per-value deep-clone-and-bound function applied to every entry in `EnforceResponse.Details` before the audit record is written. Its documented invariant is that no individual string value in the cloned map may exceed `auditDetailValueCap` (512 KiB). The `string` case enforces this:

```go
case string:
    if len(t) > auditDetailValueCap {
        return fmt.Sprintf("[eunox: omitted %d-byte value exceeding the %d-byte audit detail cap]", len(t), auditDetailValueCap)
    }
    return t
```

**Bug:**
The `[]string` case copies all elements verbatim without applying the per-element cap:

```go
case []string:
    // []string values (allowedCIDRs, allowedOperations,...) are immutable
    // manifest slices today, but the deep-clone contract requires we never
    // share backing storage with the caller regardless of the value's origin.
    return append([]string(nil), t...)
```

An element of a `[]string` that exceeds 512 KiB is written to the audit log untruncated. The 1 MiB total-Details cap (`auditDetailsTotalCap`) provides a coarse backstop — a single element between 512 KiB and 1 MiB slips through untruncated even though the contract says it must not.

**Impact:**
Today all `[]string` Detail values (`allowedCIDRs`, `allowedOperations`, `allowedExtensions`, `allowedValues`) are populated from manifest-defined slices whose individual elements are short strings. There is no current code path that produces a `[]string` Detail element exceeding 512 KiB. The impact is therefore latent rather than immediately exploitable.

However the invariant exists to bound audit log size and to prevent audit records from becoming an unintentional exfiltration channel for large argument values. A future condition handler or a direct `cloneAndBound` caller that stores a `[]string` Detail from user-supplied input would silently violate the bound.

**Suggested fix:**
Apply the same per-element cap inside the `[]string` branch:

```go
case []string:
    out := make([]string, len(t))
    for i, s := range t {
        if len(s) > auditDetailValueCap {
            out[i] = fmt.Sprintf("[eunox: omitted %d-byte value exceeding the %d-byte audit detail cap]", len(s), auditDetailValueCap)
        } else {
            out[i] = s
        }
    }
    return out
```

A companion test should assert that a `[]string` Detail whose element exceeds `auditDetailValueCap` is replaced with the placeholder string in the cloned output.

**Session:** rev-20260621T2124Z-0fed6c

### LINKED ISSUE #1095: [rev-20260621T2210Z-023311] levenshtein() in suggest.go computes edit distance over bytes, not runes — wrong/suppressed "did you mean?" hints for non-ASCII typos
**Location:** `pkg/capability/suggest.go:27-52` (`levenshtein`), reached from `NearestString` (`suggest.go:13-24`)

**Reference:** https://github.com/eunolabs/eunox/blob/565454098a59c8cf3b39889d5c1a0660c0ce41d5/pkg/capability/suggest.go#L26-L52

**Severity:** Low (UX/correctness; a diagnostic hint only — no enforcement or security impact)

**Description:**

`levenshtein` indexes its inputs as bytes throughout — `la, lb := len(a), len(b)` and `a[i-1] == b[j-1]` — so it measures edit distance over UTF-8 *bytes*, not Unicode code points:

```go
func levenshtein(a, b string) int {
	la, lb := len(a), len(b)   // byte length
	...
	for i := 1; i <= la; i++ {
		...
		for j := 1; j <= lb; j++ {
			cost := 1
			if a[i-1] == b[j-1] {   // byte comparison
				cost = 0
			}
			...
		}
	}
}
```

For any input containing a multi-byte rune this over-counts the distance. A single mistyped accented character (e.g. `é`, 2 bytes in UTF-8) counts as 2 byte-edits rather than 1, and a substitution of one multi-byte rune for another can count as up to 4 byte-edits. `NearestString` only returns a candidate when the distance is `< 3` (`bestDist := 3`), so a Unicode typo that is one *rune* away from a valid key/condition-type can exceed the byte threshold and yield **no suggestion**, or be mis-ranked against a closer ASCII candidate.

This powers the "did you mean …?" hints for unrecognized manifest keys (`internal/config/manifest.go:1171,1323`) and unknown condition types (`pkg/capability/conditions.go:242`). The candidate lists are ASCII, but the *user-supplied* typo (`s`) routinely is not under copy-paste/IME input, so the byte/rune mismatch is reachable from ordinary config files.

The codebase otherwise reaches for rune/`utf8`-aware handling where character semantics matter (`internal/audit/audit.go`, `pkg/enforcement/{engine,schema}.go`), so the byte-indexed distance here is an inconsistency rather than a deliberate choice. It is also a pure correctness bug independent of impact: a function named `levenshtein` that claims to "return the edit distance between a and b" should measure character edits.

**Trigger / repro:**

Add a manifest key or condition type that is one *rune* away from a valid name but where that rune is multi-byte, e.g. a condition type `tïmeWindow` (one accented char vs `timeWindow`). Byte distance is 2 (the `ï` differs from `i` as a 2-byte vs 1-byte sequence → an insert+substitute), inflating the score and weakening/suppressing the hint that a pure ASCII single-character typo (`timeWindov`, distance 1) would get.

**Suggested fix:**

Convert to runes once at entry and index the slices:

```go
func levenshtein(a, b string) int {
	ra, rb := []rune(a), []rune(b)
	la, lb := len(ra), len(rb)
	...
	if ra[i-1] == rb[j-1] {
		cost = 0
	}
	...
}
```

`NearestString` already lower-cases via `strings.ToLower`, which is rune-aware, so only the distance core needs the change. Add a `TestNearestString` case in `pkg/capability/suggest_internal_test.go` with a multi-byte single-rune typo asserting the expected candidate is still suggested.

**Session:** rev-20260621T2210Z-023311

### LINKED ISSUE #1096: [rev-20260621T2210Z-023311] tailAuditLines reallocates the tail ring on every line (append(ring[1:], ...)) — O(N×n) allocations scanning the audit log
**Location:** `cmd/eunox-mcp/doctor.go:511-533` (`tailAuditLines`)

**Reference:** https://github.com/eunolabs/eunox/blob/565454098a59c8cf3b39889d5c1a0660c0ce41d5/cmd/eunox-mcp/doctor.go#L511-L533

**Severity:** Low (performance; `doctor` is an operator-invoked diagnostic, not a hot path — result is correct, only inefficient)

**Description:**

`tailAuditLines` keeps the last `n` audit lines in a fixed-capacity slice (`make([]string, 0, n)`) but, once the buffer is full, advances it with:

```go
ring = append(ring[1:], line)
```

`ring` has `len == cap == n` at that point. `ring[1:]` re-slices to `len n-1, cap n-1` (the capacity shrinks because the low bound moved), so the `append` of one element needs `len n > cap n-1` and **allocates a brand-new `n`-element backing array on every iteration** for the rest of the file. This is the classic "shift via `append(s[1:], x)`" trap — correct, but it defeats the intent of a fixed ring and turns the tail into `O((N − n) × n)` allocations/copies, where `N` is the total number of lines in the log.

`doctor` already scans the entire audit file twice (once in `computeAuditStats`, then again here after `f.Seek(0, …)`), and the function's own comment advertises "a fixed-size ring so an enormous log does not blow up memory" — but the per-line reallocation means a mult-million-line log drives millions of `n`-element slice allocations rather than the constant-memory single buffer the comment implies.

**Trigger / repro:**

Run `eunox-mcp doctor --audit-log <big.jsonl>` (or `--config …` which resolves the default `~/.eunox/audit.jsonl`) against a large audit log (e.g. several million records) with the default `--audit-tail=50`. The tail computation allocates ~`(N − 50)` 50-element string slices instead of reusing one buffer. Output is correct; the cost is wasted allocation/GC pressure proportional to log size.

**Suggested fix:**

Shift in place so the backing array is reused (keeps `len == n`, no realloc):

```go
if len(ring) < n {
	ring = append(ring, line)
	continue
}
copy(ring, ring[1:])   // drop oldest, shift left
ring[n-1] = line       // append newest
```

Or use a true circular buffer with a write index and reorder once at the end. Either keeps the documented constant-memory behavior. (`internal/audit`/`stats` may have a sibling tail/ring helper worth checking for the same `append(s[1:], …)` pattern.)

**Session:** rev-20260621T2210Z-023311

### LINKED ISSUE #1097: [rev-20260622T0110Z-b1b23c] marshalCondition panics on a typed-nil condition pointer (guard missing vs. sibling directives.go)
**Location:** `pkg/capability/conditions.go:46-183` (every pointer case in `marshalCondition`, e.g. lines 54-59, 66-71, 78-83, …, 174-179)

**Reference:** https://github.com/eunolabs/eunox/blob/565454098a59c8cf3b39889d5c1a0660c0ce41d5/pkg/capability/conditions.go#L46-L183

**Severity:** Medium (latent nil-pointer panic; would be High where reachable — `Digest()` runs at startup and stamps `policy_sha256` into every audit record)

**Description:**
`ConditionWrapper.MarshalJSON` (conditions.go:22-28) guards only against an *untyped* nil interface:

```go
func (w ConditionWrapper) MarshalJSON() ([]byte, error) {
    if w.Condition == nil { // does NOT catch a typed-nil pointer
        return []byte("null"), nil
    }
    return marshalCondition(w.Condition)
}
```

A *typed-nil* pointer such as `(*MaxCallsCondition)(nil)` stored in the interface is `!= nil`, so it passes this check and reaches `marshalCondition`. Every pointer case there calls `typed.ConditionType()` with no nil check, e.g.:

```go
case *MaxCallsCondition:
    type alias MaxCallsCondition
    return json.Marshal(struct {
        conditionEnvelope
        *alias
    }{conditionEnvelope{Type: typed.ConditionType()}, (*alias)(typed)})
```

All `ConditionType()` methods have **value** receivers (`pkg/capability/condition.go:173-203`), so `typed.ConditionType()` is `(*typed).ConditionType()` — dereferencing the nil pointer to make the receiver copy — which **panics** with a nil-pointer dereference, crashing the request/marshal goroutine.

This is the exact "typed-nil interface trap" that the parallel `DirectiveWrapper`/`marshalDirective` code **already guards against**, with an explicit comment (`pkg/capability/directives.go:52-58`):

```go
case *RedactFieldsDirective:
    // A typed-nil pointer ... is not caught by the nil-interface guard in MarshalJSON,
    // so guard it here: DirectiveType() has a value receiver and would dereference nil and panic.
    if typed == nil {
        return []byte("null"), nil
    }
```

`directives.go` is documented as written "parallel to ConditionWrapper for conditions," yet this hardening fix was applied only to the directive wrapper and never back-ported to `conditions.go` — an asymmetric fix that left the conditions path exposed.

**Trigger / repro:**
```go
var c *capability.MaxCallsCondition           // typed-nil
w := capability.ConditionWrapper{Condition: c}
_, _ = json.Marshal(w)                        // panics: nil pointer dereference
```
Equivalently, a `Constraint` carrying a typed-nil condition pointer panics on marshal, because `Constraint.MarshalJSON` (constraint.go:301-302) wraps each `Conditions` element verbatim into a `ConditionWrapper{Condition: condition}`. That marshal path includes `LocalManifest.Digest()` (`internal/config/manifest_policy.go:57`), which serializes the whole manifest at load time.

**Suggested fix:**
Add the same nil guard to every pointer case in `marshalCondition`, mirroring `directives.go`:
```go
case *MaxCallsCondition:
    if typed == nil {
        return []byte("null"), nil
    }
    ...
```
(Apply to all eleven `*XxxCondition` cases.) Alternatively, guard once at the top of `marshalCondition` via reflection, or fail closed with an error rather than `"null"` — but matching the directive sibling's behavior keeps the two polymorphic wrappers consistent.

**Why earlier review/tests missed it:** `directive_internal_test.go` has an explicit `TestRedactFieldsDirective_Pointer_RoundTrip`/typed-nil case driving the directive guard, but the condition round-trip tests only exercise non-nil pointers (the loader's `newCondition` always returns `&T{}`), so no test constructs a typed-nil condition pointer. The fix landed on directives without a corresponding conditions test.

**Session:** rev-20260622T0110Z-b1b23c

### LINKED ISSUE #1098: [rev-20260622T0310Z-81d0a6] handleMCP resolves the upstream route before JWT validation, leaking configured route names (404 vs 401) to unauthenticated callers in JWT-only gateway mode
**Location:** `internal/transport/http_routing.go:22-52` (`handleMCP`)

**Reference:** https://github.com/eunolabs/eunox/blob/565454098a59c8cf3b39889d5c1a0660c0ce41d5/internal/transport/http_routing.go#L22-L52

**Severity:** Low (pre-authentication information disclosure / authentication-ordering gap; no policy bypass — the request is still denied, only the *existence* of a route name leaks)

**Description:**

`handleMCP` performs its checks in this order:

```go
func (p *HTTPProxy) handleMCP(w http.ResponseWriter, r *http.Request) {
	if !p.checkAuth(w, r) {          // 1. static bearer token — NO-OP when p.authToken == ""
		return
	}
	route := p.routes[r.PathValue("upstream")]   // 2. route lookup
	if route == nil {
		http.Error(w, "unknown upstream route", http.StatusNotFound)   // 404, NOT audited
		return
	}
	if p.jwtPDP != nil {             // 3. JWT validation — 401 on failure
		...
		if err != nil {
			... p.sink.RecordDeny(... "JWT_INVALID" ...)
			http.Error(w, "Unauthorized: "+err.Error(), http.StatusUnauthorized)
			return
		}
	}
	...
}
```

The route is resolved (step 2) **before** the JWT is validated (step 3). When the proxy is configured for **JWT-only auth** — `--jwks-uri` set with no static `--auth-token` — `checkAuth` returns `true` immediately (`p.authToken == ""`, `http_security.go:58-60`), so step 1 is a no-op. An unauthenticated request (no/invalid Bearer token, and no `Origin` header so `checkOrigin` passes freely) therefore reaches the route lookup with no credential:

- A request to a **configured** route `/mcp/<known>` → reaches step 3 → `401 Unauthorized` (`JWT_INVALID`).
- A request to a **non-existent** route `/mcp/<guess>` → step 2 returns `404 unknown upstream route`, **before** JWT validation.

This 404-vs-401 difference is an oracle that lets an unauthenticated attacker enumerate the gateway's configured upstream route names (the `/mcp/<name>` set built by `BuildRoutes`) without presenting any valid token — the classic username-enumeration pattern applied to route names. Route names routinely identify internal backends (e.g. `internal-db`, `prod-payments`), so the leak narrows an attacker's target map for the post-authentication phase.

Secondary issue on the same path: the unknown-route `404` is **not** written to the audit sink, while the `JWT_INVALID` denial on a known route is. An attacker sweeping route names therefore generates no audit trail for the probing, whereas every other rejection on this handler (`JWT_INVALID`, `ORIGIN_REJECTED`) is recorded — so the enumeration is both possible and unobservable.

This only affects JWT-only deployments: when a static `--auth-token` is also set, `checkAuth` (step 1) gates the route lookup behind the bearer token, so unknown/known routes are indistinguishable to an unauthenticated caller. Single-upstream (`/mcp`, route key `""`) deployments are unaffected since there is only one route to enumerate; the leak is specific to **gateway mode** with multiple named routes.

**Trigger / repro:**

1. Run the proxy in gateway mode (`transport: http`) with two upstreams (e.g. `alpha`, `beta`) and `--jwks-uri` configured but **no** `--auth-token`.
2. With no Authorization header:
   - `curl -s -o /dev/null -w '%{http_code}' http://127.0.0.1:PORT/mcp/alpha` → `401`
   - `curl -s -o /dev/null -w '%{http_code}' http://127.0.0.1:PORT/mcp/zzz`   → `404`
3. The 401-vs-404 split reveals that `alpha` is a real route and `zzz` is not, with no valid token and no audit record of the probe.

**Suggested fix:**

Validate the JWT (when `p.jwtPDP != nil`) **before** resolving the route, so an unauthenticated caller receives a uniform `401` regardless of whether the route exists — JWT validation is authentication, not policy, and should gate route disclosure exactly as `checkAuth` already does. The JWT block does not depend on `route`, so it can move above the route lookup:

```go
func (p *HTTPProxy) handleMCP(w http.ResponseWriter, r *http.Request) {
	if !p.checkAuth(w, r) {
		return
	}
	if p.jwtPDP != nil {
		authHeader := r.Header.Get("Authorization")
		ctx, err := p.jwtPDP.ValidateToken(r.Context(), authHeader)
		if err != nil {
			... // record JWT_INVALID, 401
			return
		}
		r = r.WithContext(ctx)
	}
	route := p.routes[r.PathValue("upstream")]
	if route == nil {
		http.Error(w, "unknown upstream route", http.StatusNotFound)
		return
	}
	...
}
```

(The downstream comment "Must happen before routing so that all subsequent handlers see the claims" already argues JWT validation belongs ahead of routing; moving it above the route lookup also closes the oracle.) Optionally also `RecordDeny` the unknown-route case so route-probing leaves an audit trail consistent with the handler's other rejections.

**Why earlier review/tests missed it:** the JWT block carries a comment explaining it must precede *handler dispatch* so claims are attached, and it does sit before the `switch r.Method`; the subtlety is that it sits *after* the `route == nil` 404, so existing tests assert the positive paths (valid JWT → dispatch) and the 401-on-bad-JWT path for a known route, but no test asserts that an unknown route returns the **same** status as a known route under an invalid token in JWT-only mode.

**Session:** rev-20260622T0310Z-81d0a6

eunolabs:
## Bugs addressed (session 1cby6q)

This batch fixes every open automated-review bug whose title carries a review-session id. One commit per bug, each referencing its issue.

- Closes #1097 — `marshalCondition` panicked on a typed-nil condition pointer (e.g. `(*MaxCallsCondition)(nil)`): it slipped past the nil-interface guard in `ConditionWrapper.MarshalJSON`, then the value-receiver `ConditionType()` dereferenced the nil pointer. Added a single reflection-based typed-nil guard at the top of `marshalCondition`, mirroring the sibling guard in `directives.go`.
- Closes #1096 — `tailAuditLines` advanced its full ring with `append(ring[1:], line)`, reallocating a fresh `n`-element array on every line — `O((N-n)*n)` allocations. Now shifts in place with `copy`.
- Closes #1095 — `levenshtein` measured edit distance over UTF-8 bytes, over-counting multi-byte rune typos and suppressing/mis-ranking "did you mean …?" hints. Now indexes `[]rune`.
- Closes #1094 — `cloneAndBound` copied `[]string` audit Detail values verbatim, letting an element over `auditDetailValueCap` (512 KiB) reach the log untruncated. Now applies the per-element cap, matching the scalar `string` case.
- Closes #1093 — `IncrementIfAllBelow` hardcoded `count=0` on the admitted path in both the in-memory and Redis backends, violating the `LimiterStore` contract. Both backends now return the maximum post-admission in-window total across buckets.
- Closes #1092 — `doctor`'s `redactURL` stripped only userinfo, leaving query-string credentials (`?api_key=`, `?token=`, …) in the "redacted" support bundle. Now scrubs every non-empty query value (length-tagged placeholder) in both the parse-success path and the `url.Parse`-failure fallback (`redactRawQuery`).
- Closes #1098 — In JWT-only gateway mode `handleMCP` resolved the upstream route before validating the JWT, so an unauthenticated caller got 404 for an unknown route vs 401 for a known one — a route-name enumeration oracle. JWT validation now runs before the route lookup, yielding a uniform 401 regardless of route existence.

### Not changed (verified false positives, already triaged by the owner)
- #1079 — `IPRangeCondition.Compile()` is already called in the pointer case (the form unmarshal produces); `TestLoadManifest_PreCompilesIPRangeCIDRs` already asserts pre-compilation. No code change needed.
- #1077 — `VerifyToken` parses via `jwt.ParseSigned` → compact JWS only (exactly one signature/header), so the multi-signature `headers[0]` selection path is unreachable. No code change needed.

## Testing

- New regression tests per fix:
  - `TestConditionWrapper_TypedNilPointer_MarshalsToNull` (covers every condition pointer case)
  - `TestNearestString` multi-byte single-rune case
  - `TestCopyDetailValue_BoundsStringSliceElements`
  - `TestInMemory_IncrementIfAllBelow_AdmittedReturnsCount`, `TestRedis_IncrementIfAllBelow_AdmittedReturnsCount`
  - `TestRedactURL_*` query cases + `TestRedactURLFallback_Cases` query case
  - `TestHTTPProxy_JWTMode_UnknownRouteReturns401NotOracle`
- `make check-fmt`, `make check-license`, `make lint` (go vet + golangci-lint) — clean.

https://claude.ai/code/session_01MuBm3p4J7KTsZTZk7KSENw

---
_Generated by [Claude Code](https://claude.ai/code/session_01MuBm3p4J7KTsZTZk7KSENw)_


cmd/eunox-mcp/doctor.go
eunolabs, line 1: `len(val)` in the query-value redaction placeholder counts URL-encoded bytes, not decoded bytes or runes. `u.RawQuery` is kept raw by `url.Parse`, so `val := p[eq+1:]` captures the percent-encoded string verbatim. For `?api_key=%F0%9F%90%B9` (one emoji, 4 decoded bytes) the placeholder would read `<redacted len=12>`, while `redactConfigValue` (which works on plain strings) would report the true byte length. The reported length is non-zero and the value is correctly suppressed, so this is only a cosmetic inconsistency, but it diverges from the `redactConfigValue` form the comment cites. Consider `url.QueryUnescape(val)` to get the decoded length, or document the "raw byte count" intention explicitly in the comment.

---

eunolabs, line 1: `redactRawQuery` replaces the entire query component with `?<redacted query>`, losing all parameter names. The parse-success path in `redactURL` preserves names (`?api_key=<redacted len=N>&mode=<redacted len=4>`), so operators diagnosing an unparseable URL in a support bundle get less context than they would for a parseable one. This is the correct tradeoff on the fallback path — you can't safely tokenize a query you couldn't parse — but it's worth a brief comment on `redactRawQuery` explaining why the names are dropped wholesale rather than preserved: without a reliable `=` / `&` parse, naive splitting could still expose a credential (e.g. a value containing `%26` that splits differently than the server would see it). A one-liner like `// Can't safely tokenize a malformed query; scrub the whole component.` would make the asymmetry intentional rather than apparent.

---

internal/audit/audit.go
eunolabs, line 1: The oversized-element placeholder in the new `[]string` case reads `"[eunox: omitted %d-byte value exceeding the %d-byte audit detail cap]"`. The scalar `string` case uses a different sentinel (not touched in this diff). If the two messages diverge, `audit-verify` output and any tooling that parses audit Detail fields will see two distinct formats for what is semantically the same truncation event. Worth aligning to a single format — or at minimum confirming the scalar message already matches — so grep and log parsers only need one pattern.

---

pkg/callcounter/redis.go
eunolabs, line 1: The `counts[i] + 1` expression relies on the just-added member always being a new ZADD entry (i.e., unique per call). The comment says as much, but it's load-bearing: if two concurrent callers race on the same key, both could read `counts[i]` before either ZADD completes and both would compute the same post-admission count, but the actual sorted-set cardinality would be higher. In Lua (single-threaded, atomic per script invocation) this race can't happen — the whole script runs atomically — so `counts[i] + 1` is exact. The comment should explicitly credit Lua atomicity rather than just noting uniqueness; "unique per call" is a property of the member, not the reason the count is safe to compute this way.

---

internal/transport/http_routing.go
eunolabs, line 1: The fix is correct and the oracle is closed: after the reorder, an unauthenticated caller always hits the JWT gate before the route lookup, so there's no 404-vs-401 split to probe.

One subtlety worth confirming: `checkAuth` at the top of the function handles static-token auth (the existing guard, not touched here). In a configuration that has both a static auth token and a JWT PDP, an unauthenticated caller will still hit `checkAuth` first and get a 401 there — which is fine, because the 401 is the right response regardless. The oracle only existed in JWT-only mode where `checkAuth` is a no-op; the comment already says this, but a brief mention that the fix is scoped to JWT-only mode (no static token) would make the reasoning self-contained for a future reader who doesn't have the issue context.

---

eunolabs (commented):
## Review

Seven focused bug fixes, each with a regression test. All fixes are mechanically correct. No blocking issues; the inline comments are suggestions and one correctness note.

### What's good

- **#1098 (JWT oracle)** is the most impactful fix. Moving JWT validation before the route lookup closes the 404-vs-401 side-channel cleanly, and the test covers both the known-route and unknown-route paths without leaking through the actual upstream.
- **#1096 (`tailAuditLines`)** — `copy(ring, ring[1:])` is safe for overlapping slices per the Go spec ("The source and destination may overlap") and correctly avoids the `O((N-n)·n)` reallocation. Good catch.
- **#1097 (typed-nil panic)** — the reflection guard mirrors `directives.go` exactly; `TestConditionWrapper_TypedNilPointer_MarshalsToNull` exercises every pointer case through `knownConditionTypes`, which is the right exhaustive approach.
- **#1093 (hardcoded 0 count)** — both backends now return `maxCount`/`cnt` on the admit path, satisfying the `LimiterStore` contract. The Redis Lua change is atomic by nature so the count is exact (see inline note on the comment text).
- **#1095 (rune vs byte Levenshtein)** — correct and the test case pinning the specific multi-byte failure is exactly what's needed.

### Suggestions (not blocking)

See inline comments for details:

1. **`redactURL` — `len(val)` is URL-encoded byte count**, not decoded length. Cosmetically inconsistent with `redactConfigValue`; worth either decoding first or documenting explicitly.
2. **`redactRawQuery` — param names are dropped wholesale** in the fallback. Intentionally safer than attempting to tokenize a malformed query, but the function has no comment explaining why. A one-liner prevents this looking like an oversight.
3. **`cloneAndBound` placeholder format** — the new `[]string` sentinel differs from the scalar `string` case. Aligning them prevents grep / log-parser fragility.
4. **Redis Lua comment** — attributes correctness to member uniqueness when the real guarantee is Lua script atomicity. Minor, but the latter is the load-bearing property.

---

cmd/eunox-mcp/doctor.go
eunolabs: Fixed in f540128. The placeholder now reports the decoded byte count via `url.QueryUnescape` (falling back to the raw length on a malformed escape), so it matches `redactConfigValue` for the same secret. Added a test case `?api_key=%F0%9F%90%B9` asserting `<redacted len=4>` (decoded), not 12 (encoded).

---

eunolabs: Added the rationale to `redactRawQuery` in f540128: we only reach the fallback because the URL would not parse, so we cannot safely tokenize the query on `=`/`&` (an encoded `%26` could split differently than the server reads it and expose a credential) — wholesale redaction is the safe tradeoff, at the cost of the parameter-name context the parseable path keeps.

---

internal/audit/audit.go
eunolabs: Confirmed already aligned: the two placeholders are byte-identical format strings — `"[eunox: omitted %d-byte value exceeding the %d-byte audit detail cap]"` — at audit.go:931 (scalar `string`) and audit.go:956 (new `[]string` element). Grep/log parsers need only one pattern. No change required.

---

pkg/callcounter/redis.go
eunolabs: Reworded the comment in f540128 to credit Lua atomicity: `counts[i] + 1` is exact because Redis evaluates the whole script single-threaded with no interleaving, so no concurrent caller can ZADD between the ZCARD that produced `counts[i]` and this ZADD. Member uniqueness is now noted as the separate guarantee that this ZADD always inserts (never a no-op overwrite).

---

internal/transport/http_routing.go
eunolabs: Expanded the comment in f540128 to make the scoping self-contained: the oracle is reachable only in JWT-only mode (no static auth token), where `checkAuth` is a no-op; when a static token is set, `checkAuth` already gated route disclosure behind the bearer token, so unknown/known routes were indistinguishable and the 401 there is correct either way.

---

PR #1111, opened 2026-06-22
sec: make tools/list probe failure fatal under --strict-drift

### LINKED ISSUE #1105: [rev-20260622T0545Z-2317a163] tools/list probe failure voids --strict-drift for manifests without descriptionHash pins (FM-1/FM-2/FM-4 unevaluated)
**Location:** `internal/drift/drift.go:588` (`MakeDriftCheck` / `driftProbeUnavailable`)

**Severity:** High (a malicious or broken upstream that withholds `tools/list` at session start bypasses `--strict-drift`)

## Description

On a `tools/list` probe failure the session-start hook returns early, before `strict` is ever consulted:

```go
if probeErr != nil {
    return driftProbeUnavailable(manifest, probeErr)
}
```

`driftProbeUnavailable` gates fatality solely on whether the manifest pins `descriptionHash`:

```go
if manifestHasPinnedDescriptions(manifest) {
    return fmt.Errorf(...)   // fatal
}
// ...
return nil                   // logged skip
```

The `strict` flag passed into `MakeDriftCheck` is only threaded into `evaluateDrift`, which is unreachable on a probe failure (the early return at drift.go:588-589 precedes it). So for any manifest **without** description pins, a probe failure is a non-fatal skip — even under `--strict-drift`.

## Trigger / repro

Manifest with glob entries (e.g. `tool:read_*`) and/or a `serverVersion` pin but **no** `descriptionHash`. Start the proxy with `--strict-drift` against an upstream that errors (or returns empty) for `tools/list` at session start. The hook logs a WARN and returns `nil`; the session proceeds with FM-1 (new/over-permitted tools), FM-2, and FM-4 (server-version) drift entirely unevaluated.

## Impact

`--strict-drift`'s fatal-on-drift guarantee is silently void whenever the upstream withholds `tools/list`. An upstream that has drifted can avoid drift detection simply by failing the probe.

## Suggested fix

When `strict` is set, treat a probe failure as fatal regardless of whether the manifest pins descriptions — under strict mode, a probe we cannot complete is indistinguishable from drift. Thread `strict` into `driftProbeUnavailable`. Add a test: strict mode + unpinned manifest + failing probe → fatal.

eunolabs:
Fixes #1105.

## Problem

On a `tools/list` probe failure at session start, the drift hook returned early through `driftProbeUnavailable` **before** the `strict` flag was ever consulted:

```go
if probeErr != nil {
    return driftProbeUnavailable(manifest, probeErr) // strict never threaded in
}
```

`driftProbeUnavailable` gated fatality solely on whether the manifest pinned a `descriptionHash`. So for any manifest **without** description pins, a probe failure was a non-fatal skip — even under `--strict-drift`. An upstream that withheld (or returned an unparseable) `tools/list` could therefore silently void `--strict-drift`'s fatal-on-drift guarantee, leaving FM-1 (new/over-permitted tools), FM-2, and FM-4 (server-version) drift entirely unevaluated. A drifted or malicious upstream could dodge drift detection simply by failing the probe.

## Fix

- Thread `strict` into `driftProbeUnavailable` and make a probe (or parse) failure fatal whenever `strict` is set, regardless of whether the manifest pins descriptions. Under strict mode a probe we cannot complete is indistinguishable from drift.
- The pinned-`descriptionHash` branch is checked first, so the existing fail-closed message and behavior for pinned manifests are unchanged; the new strict branch only adds fatality for the previously-skipped unpinned case.

## Tests

- **Unit (`internal/drift`)**: new `TestMakeDriftCheck_ProbeError_StrictNoPinsFatal` (the exact repro from the issue: glob entry + `serverVersion` pin, no `descriptionHash`, strict, failing probe → fatal), plus strict parse-error and direct `driftProbeUnavailable` strict variants.
- **Integration (`internal/transport`)**: new stdio + HTTP strict-fatal tests that drive the real `fetchUpstreamToolsRaw` probe.
- Corrected `TestRunStdioDriftCheck_SkippedWhenNoPins` and `TestRunHTTPDriftCheck_SkippedWhenNoPins`, which asserted the skip path while passing `strict=true` — an argument that was inconsequential under the old bug and encoded exactly the vulnerable behavior. They now exercise the skip in non-strict mode (the only mode where skip-on-no-pins is correct).

## Docs

- Added a "When the `tools/list` probe itself fails" subsection to the capability-manifest guide documenting the fail-closed outcome (fatal under `strictDrift`; fatal without it only when `descriptionHash` pins exist; otherwise a logged best-effort skip).

Full suite (`go test -race ./...`), `go vet`, `gofmt`, license-header check, and `golangci-lint` on the touched packages all pass.

https://claude.ai/code/session_013Fr5aFxZFTyRs6eNk38JLm

---
_Generated by [Claude Code](https://claude.ai/code/session_013Fr5aFxZFTyRs6eNk38JLm)_


internal/drift/drift.go
```diff
@@ -578,32 +578,43 @@ func ParseToolsListResult(raw json.RawMessage) ([]UpstreamTool, error) {
...
 		if probeErr != nil {
-			return driftProbeUnavailable(manifest, probeErr)
+			return driftProbeUnavailable(manifest, strict, probeErr)
 		}
 		tools, err := ParseToolsListResult(rawToolsListResult)
 		if err != nil {
-			return driftProbeUnavailable(manifest, err)
+			return driftProbeUnavailable(manifest, strict, err)
 		}
 		return evaluateDrift(manifest, tools, serverVersion, strict)
```
eunolabs, line 597: Threading `strict` through `driftProbeUnavailable` cleanly closes the probe-error and parse-error variants of the bypass. One adjacent variant it doesn't cover, worth deciding on explicitly:

An upstream can also "withhold" `tools/list` by returning a **well-formed but empty** result — `result: null`, an absent `result`, or `{"tools":[]}`. None of those is a probe error or a parse error, so they flow past both `driftProbeUnavailable` branches into this `evaluateDrift` call with an empty `[]UpstreamTool`:

- `fetchUpstreamToolsRaw` returns `(resp.Result, nil)` for any non-error response (`drift_probe.go:33`/`:83`).
- `FetchAllToolPages` marshals an empty/`null` first page to `{"tools":null}` with a nil error (`drift.go:546-551`).
- `ParseToolsListResult` then yields a zero-length slice + nil error → `evaluateDrift(manifest, []UpstreamTool{}, ...)`.

Against an empty tool list, FM-1/FM-2 produce no findings, so under `--strict-drift` this is non-fatal — the same outcome the PR makes fatal for the error/garbage cases.

Two caveats that make this lower-severity (hence a question, not a blocker):
- FM-4 still fires if `serverVersion` is populated (it's passed independently of the tool list), so only the tools-derived modes are dodged.
- Per-call manifest enforcement still gates every `tools/call` regardless of drift, so an unlisted tool isn't automatically reachable.

Is treating an empty/`null` result as "legitimately zero tools" intentional? If so, a one-line note in the new docs section ("an empty or `null` result is trusted as a genuine empty tool set, not a withheld probe") would make the boundary explicit; if not, it'd want the same strict-fatal treatment.

---

eunolabs (commented):
## Review: make `tools/list` probe failure fatal under `--strict-drift`

**What it does:** Threads the `strict` flag into `driftProbeUnavailable` so a `tools/list` probe (or parse) failure aborts startup under `--strict-drift`, not only when the manifest pins a `descriptionHash`. Previously the probe-failure path returned before `strict` was ever consulted, so any pin-less manifest silently downgraded a failed probe to a best-effort skip — letting a drifted or malicious upstream dodge FM-1/FM-2/FM-4 by withholding `tools/list`.

**Assessment — correct and well-scoped:**
- The fix is symmetric with the success path: `evaluateDrift` already gated FM-1/2/4 fatality on `strict`; this brings the failure path in line. Pins-first ordering preserves the existing fail-closed message/behavior for pinned manifests.
- Error wrapping (`%w`) is preserved and asserted via `errors.Is`.
- `driftProbeUnavailable` is unexported with exactly the two call sites in `MakeDriftCheck`; both are updated — no caller missed.
- Good catch on the two `*_SkippedWhenNoPins` tests: they asserted the skip path while passing `strict=true`, which encoded precisely the buggy behavior. Flipping them to `strict=false` (plus the new strict-fatal counterparts) is the right correction, and the fact that they passed before is itself evidence of the bug.
- Coverage is solid: unit (`MakeDriftCheck` + `driftProbeUnavailable`, probe + parse, strict + pins) and integration (stdio + HTTP driving the real `fetchUpstreamToolsRaw`). Docs updated. All 33 CI checks are green.

**One question (non-blocking)** — see the inline comment on `internal/drift/drift.go`: a well-formed-but-empty result (`null` / absent / `{"tools":[]}`) is neither a probe error nor a parse error, so it bypasses the new strict-fatal branches and is evaluated as an empty (no-drift) tool list. Worth a one-line note in the docs on whether that's intentionally trusted as "genuinely zero tools," since it's a third way to withhold the list.

**Nit:** the two new integration tests are named inconsistently — `TestRunHTTPDriftCheck_StrictFatalWhenNoPins` vs `TestRunStdioDriftCheck_StrictFatalWhenNoToolsList`. Both are the "no pins" case; worth aligning the suffix.

Overall a clean, minimal security hardening with strong tests — just the empty-result question to consider before merge.

---

internal/drift/drift.go
eunolabs: Resolved as intentional in e9f2a13: an empty/`null` result is treated as a genuine zero-tool upstream, not a withheld probe, with a one-line note added to the drift section making the boundary explicit.

Rationale for documenting rather than failing closed:
- An empty live set has nothing for FM-1/FM-2 to flag; FM-4 (server version, from `initialize`) is still evaluated; and per-call manifest enforcement still gates every `tools/call`. So nothing is actually bypassed.
- Failing closed on an empty list would false-positive on legitimately tool-less servers (e.g. tools gated behind later init).
- `FetchAllToolPages` collapses `null`/absent/`{"tools":[]}` into the same `{"tools":null}` shape, so a "malformed null" can't be distinguished from a "genuine empty" downstream without reworking the shared probe path — out of scope for this fix.

The test-naming nit is addressed in 85d899e.

---

PR #1112, opened 2026-06-22
sec: harden audit file modes, kill-switch SSE coverage, strict-audit on initialize

eunolabs:
## Summary

Three **Low**-severity hardening fixes surfaced by a whole-codebase security review. No Critical/High/Medium issues were found — the codebase is well-hardened; these close defense-in-depth gaps. Each fix has a focused test and is documented in `docs/threat-model-mcp.md`.

### 1. Audit log & HMAC key file modes are re-tightened on load (not just on create)
`internal/audit/audit.go`, `internal/audit/rotate.go`

`os.OpenFile(..., O_CREATE, 0o600)` applies the restrictive mode **only when it creates the file**. A pre-existing `audit.jsonl` or `audit.key` left group/world-readable (a looser umask on a prior run, a backup restore, or a deliberately pre-created path) was opened with that mode unchanged — unlike the control-token path, which already `chmod`s explicitly. A readable HMAC key lets any local user **forge audit records**, so this is the highest-value fix. The proxy now `chmod`s a pre-existing log/key back to `0600` (key directory to `0700`) on load, and the active log to `0600` on rotation. Best-effort with a startup warning on the rare chmod-less backend (FAT/exFAT, some CIFS/NFS mounts — the same ones the key-publish path already tolerates), so startup is never broken.

### 2. The SSE GET notification stream now consults the kill switch
`internal/transport/http_routing.go`

The enforced POST methods gate on the kill switch (the `Decide*` paths embed it; `dispatchList` and the session-creating `initialize` check it explicitly), but the standalone SSE **GET** stream (`handleMCPGet`) did not — and `/control/kill` only writes to the kill store, it does not actively reap sessions. So an emergency-killed session could open/hold an SSE stream and keep receiving upstream notifications until the idle reaper or upstream exit closed it. `handleMCPGet` now denies a killed (or globally killed) session with **HTTP 403** and records the deny, making the emergency stop total.

### 3. `--require-audit=strict` now gates the HTTP session-creating `initialize`
`internal/transport/http_routing.go`, `internal/transport/http_handlers.go`

The strict-audit gate was applied to the enforced-forward, `*/list`, and sampling paths, but not to the session-creating `initialize`. Creating a session spawns/contacts an upstream — a privileged side effect — so under a degraded audit trail an `initialize` flood could still create upstream sessions whose creation and subsequent traffic went partly unrecorded. The HTTP `initialize` branch now refuses new sessions with `AUDIT_UNAVAILABLE` once the trail degrades, mirroring the other privileged paths. (stdio's `initialize` answers from cached capabilities and starts its single upstream before any traffic, so it has no equivalent per-call side effect to gate.)

## Tests
- `internal/audit/permissions_test.go` — pre-existing loose log/key/dir re-tightened to `0600`/`0700`; post-rotation active log is `0600`.
- `internal/transport/http_test.go` — killed session's SSE GET is refused (403) while a live session still opens the stream; strict-audit + degraded sink refuses session-creating `initialize` with `AUDIT_UNAVAILABLE` and creates no session.

## Verification
- `make check-fmt`, `make check-license`, `make lint` (go vet + golangci-lint) — clean

---
_Generated by [Claude Code](https://claude.ai/code/session_01B1eNgoSDYmdF2zTKJNs81M)_


internal/transport/http_routing.go
```diff
@@ -286,6 +306,28 @@ func (p *HTTPProxy) handleMCPGet(w http.ResponseWriter, r *http.Request, route *
...
+	// it. A kill-store error fails closed via CheckKill. There is no JSON-RPC
+	// envelope on a GET, so the refusal is an HTTP 403; the deny is still recorded
+	// for the audit tape, mirroring dispatchList.
+	//
+	// route is always non-nil in production (every served stream resolves to a
+	// wired route, whose PDP — wiretap mode included — carries the kill switch);
+	// the nil guard only covers the minimal test-only construction that wires no
+	// route, which has no PDP to consult, mirroring killCheckIfPDP's nil tolerance.
+	if route != nil {
+		if deny := killCheckIfPDP(r.Context(), route.pdp, sessionID); deny != nil {
```
eunolabs, line 325: Good catch on the un-gated GET. One residual worth being precise about: this refuses *opening* a stream after a kill, but it doesn't tear down a stream that was already open when the kill landed. `handleKill` only writes to the kill store — it doesn't signal `sess.done` or evict subscribers — and a session holding an open SSE stream is spared the normal idle reaper, so it's only hard-reaped after the hard idle ceiling (`hardIdleMultiplier × idle window`). The loop below selects on `ch` / `sess.done` / `r.Context().Done()` but not on kill state, so during that whole window a session killed *while streaming* keeps receiving upstream→client notifications — which is exactly the scenario the new threat-model paragraph says is closed ("cannot keep an open channel receiving upstream notifications until the idle reaper ... closes it").

To make the emergency stop actually total for SSE you'd need to evict open streams on kill (e.g. close the per-session subscriber channels from `handleKill`, or add a kill signal to the select loop). If that's out of scope here, consider softening the doc to "cannot *open or re-open*" so the wording doesn't over-claim. Not blocking — but the doc and the code should agree.

---

internal/audit/audit.go
```diff
@@ -1634,12 +1682,24 @@ func (s *Sink) Close() error {
...
 func LoadOrCreateKeys(keyPath string) ([][]byte, error) {
-	if err := os.MkdirAll(filepath.Dir(keyPath), 0o700); err != nil { //nolint:gosec // G304: keyPath is user-configured via --audit-key-path; taint is intentional
+	keyDir := filepath.Dir(keyPath)
+	if err := os.MkdirAll(keyDir, 0o700); err != nil { //nolint:gosec // G304: keyPath is user-configured via --audit-key-path; taint is intentional
 		return nil, err
 	}
+	// MkdirAll is a no-op (and does not chmod) when the directory already exists, so
+	// drop any group/world access explicitly to honor the restrictive-directory
+	// guarantee even for a pre-existing looser-mode directory holding the HMAC key.
+	tightenPathMode(keyDir, "audit key directory")
```
eunolabs: Re-tightening the key *file* to `0600` (just below) is the load-bearing fix. Forcing the *directory* to `0700` on every load is broader than the stated threat: a `0600` file is already unreadable by other users regardless of the directory's mode — directory perms gate traversal/listing, not the contents of a mode-`0600` file.

The collateral: an operator who points `--audit-key-path` at a file inside a *shared* directory (this flag is explicitly aimed at containerized / multi-instance deployments, where the key often lives in a managed or read-only secret mount) now has that whole directory silently narrowed to `0700` — potentially locking out other consumers of sibling files, or, on a read-only mount, failing the chmod and emitting a WARNING on every startup. The threat-model frames chmod-failure as a "rare backend (FAT/exFAT/CIFS/NFS)" case, but read-only secret mounts aren't rare for exactly this flag.

Worth considering whether the directory tighten should be limited (e.g. only when eunox created the dir) or dropped, keeping the essential file-mode tighten.

---

internal/audit/audit.go
```diff
@@ -1634,12 +1682,24 @@ func (s *Sink) Close() error {
...
 
 	data, err := os.ReadFile(keyPath) //nolint:gosec // G304: path is user-configured audit key location (--audit-key-path or EUNOX_AUDIT_KEY_PATH)
 	if err == nil {
+		// Drop any group/world access from a pre-existing key file: a readable HMAC
+		// key lets any local user forge audit records, so a key left readable by a
+		// looser umask or a restore is re-secured on load. A freshly generated key is
+		// written 0600 by generateAndPersistAuditKey's temp file, so this only affects
+		// keys created outside that path. The successful ReadFile above proves owner
+		// read, which clearing only the group/world bits preserves.
+		tightenPathMode(keyPath, "audit key file")
```
eunolabs: A pre-existing group/world-readable HMAC key isn't only a mode to fix — it's a possible prior-exposure indicator: anyone who could read it may already have copied it and can forge records that pass verification. As written, the loose→tight transition is silent (these helpers warn only on stat/chmod *failure*), so the operator never learns the key was exposed and never knows to rotate it.

Given how loudly the rest of the codebase surfaces security-relevant posture (the AUDIT/NOTICE/SECURITY stderr lines, the all-zero-key rejection), consider emitting a one-line WARNING when a loose *key* is actually found-and-tightened — the `perm&0o077 != 0` branch in `tightenPathMode` is precisely that signal. (Less important for the log, which is HMAC-integrity-protected; most valuable for the key.)

---

internal/audit/audit.go
```diff
@@ -1624,6 +1630,48 @@ func (s *Sink) Close() error {
...
+			fmt.Fprintf(os.Stderr, "[eunox-mcp] WARNING: could not tighten audit log %q (drop group/world access): %v\n", logPath, cerr)
+		}
+	}
+}
+
+// tightenPathMode drops any group/world access bits from a pre-existing audit key
+// file or its directory, preserving the owner bits (so the owner keeps the read,
+// and — for a directory — the execute/traverse, it needs). It is the path-based
+// counterpart of tightenLogMode; see that function for the rationale and the
+// best-effort/warning posture. what is a short noun for the warning message.
```
eunolabs: Leftover editing note in the doc comment: "best-effort/warning posture. what is a short noun for the warning message." The trailing sentence reads as a note-to-self rather than documentation — looks like it was meant to document the `what` parameter, e.g.:

```go
// best-effort/warning posture. `what` is a short noun naming the path in the
// warning message (e.g. "audit key file", "audit key directory").
```

---

eunolabs (commented):
Reviewed all three hardening fixes and the threat-model updates against the surrounding code. Overall this is careful, well-scoped defense-in-depth: the fail-closed posture is preserved throughout, each fix has a focused test, and the reasoning in the comments/threat-model is unusually thorough. A few things I specifically checked and liked:

- Mode tightening uses the fd-based `f.Chmod` for the already-open log (no path TOCTOU), and correctly preserves owner bits — so an operator's deliberately `0200` log isn't silently made readable, and its tail read still fails closed into chain-resume rather than being re-opened for read.
- The typed-nil interface trap is handled in `initStrictAuditDenial` (`route.sink != nil` before assigning into `auditRecorder`), consistent with the existing pattern, and `routeSink`'s nil-receiver guards make the GET deny-record path panic-safe even for the minimal test construction.
- The strict-audit `initialize` gate is correctly scoped to HTTP only, with a sound rationale for why stdio's `initialize` has no equivalent per-call side effect.

Nothing here looks like a correctness regression. I left four inline notes; only the first is substantive.

**Substantive (doc-vs-code)**
- **SSE kill gate covers stream-open, not already-open streams.** The new check refuses *opening* a stream after a kill, but a stream opened *before* the kill keeps delivering upstream notifications until the hard idle ceiling, because `/control/kill` doesn't reap sessions. The threat-model paragraph reads as if open streams are stopped — either reap them on kill, or soften the wording. Details inline on `handleMCPGet`.

**Minor / non-blocking**
- **Directory re-tighten collateral** — forcing the key's *parent directory* to `0700` is broader than the threat and can bite shared / read-only secret mounts (the containerized case `--audit-key-path` targets). Inline.
- **Silent loose-key re-tighten** — a pre-existing world-readable key is a possible-exposure signal worth a WARNING on the found-and-tightened path, not just on chmod failure. Inline.
- **Doc-comment artifact** in `tightenPathMode` — `what is a short noun for the warning message.` is a leftover note. Inline.
- **`sseStreamMethod = "GET /mcp"`** lands in the audit record's `method` field, where `deriveTargetFields` treats it as an unrecognized post-dispatch method (empty `target_type`), bucketing it with unmapped-method denials like `tools/execute`. Harmless for `suggest`/`stats` (no target to act on) and you documented the choice, but it's a transport action in a field whose convention elsewhere is either a real MCP method or `""` (cf. the `JWT_INVALID` record). Worth a sanity check that a space-containing pseudo-method in OCSF `method` is intended vs. `""`.
- **Test gaps**: the SSE tests cover a per-session kill; a global-kill (kill-all) GET case would mirror `TestHTTPInitialize_GlobalKill_RefusesSessionCreation` and pin the "(or globally killed)" claim. There's also no test asserting the documented "owner-write-only `0200` log is *not* loosened" behavior, which is a nice invariant to lock down.

No blocking issues from me — the substantive item is really doc/code alignment plus an optional follow-up on actively reaping killed SSE streams.

---

PR #1113, opened 2026-06-22
fix(config): reject conflicting serverVersion/schemaVersion when merging policy files

### LINKED ISSUE #1108: [rev-20260622T0545Z-2317a163] MergeManifests silently discards later --policy files' top-level metadata (serverVersion pin, etc.) with no agreement check
**Location:** `internal/config/manifest.go:158-171` (`MergeManifests`)

**Severity:** Medium (weakens FM-4 drift coverage when combining policy files; silent)

## Description

`MergeManifests` copies top-level metadata (`SchemaVersion`, `Name`, `Version`, `Description`, `Audience`, `DefaultTTL`) from `ms[0]` only, and carries `ServerVersion` from the first **non-empty** pin:

```go
merged.SchemaVersion = ms[0].SchemaVersion
merged.Name          = ms[0].Name
// ... ServerVersion: first non-empty wins (manifest.go:163, 169-171)
```

There is no agreement/conflict check. Two `--policy` files that declare *different* values for any of these top-level fields collapse to one, silently dropping the others. This is distinct from closed #745 (the operator-supplied `expectVersion` gateway-config pin path); it is the general silent-metadata-discard in `MergeManifests` itself.

## Trigger / repro

- `a.yaml`: `serverVersion: "1.*"`
- `b.yaml`: `serverVersion: "2.*"`

`MergeManifests([a, b])` keeps `1.*` and silently drops `2.*` — no error. (`TestMergeManifests_InheritsFirstManifestMetadata` already asserts this discard with `9.9.9` vs `2.0.0`.)

## Impact

An operator who expects both files' `serverVersion` pins to contribute to FM-4 drift detection gets only the first; the second constraint is silently inoperative.

## Suggested fix

Reject the merge (clear error) when two files declare conflicting non-empty values for a top-level field that can only hold one value (`serverVersion`, `schemaVersion`, …), or document and test the precedence explicitly. At minimum, warn on a discarded non-empty `serverVersion`.

eunolabs:
Fixes #1108.

## Problem

`MergeManifests` (`internal/config/manifest.go`) inherited top-level metadata from the first manifest and carried `serverVersion` from the first **non-empty** pin, with no agreement check. Two `--policy` files declaring *different* non-empty values for a single-value top-level field collapsed to one, silently dropping the rest.

Concretely, given `a.yaml: serverVersion: "1.*"` and `b.yaml: serverVersion: "2.*"`, `MergeManifests([a, b])` kept `1.*` and silently discarded `2.*` — no error. An operator who split policy across files and expected each file's `serverVersion` pin to contribute to FM-4 drift detection got only the first; the second was silently inoperative.

## Fix

Fold the single-value fields that actually drive enforcement/drift — `serverVersion` (FM-4) and `schemaVersion` (grammar) — through a new `mergeSingleValueField` conflict check:

- **First non-empty wins** — a pin set in any *one* of several `--policy` files still survives the merge, preserving the original split-policy guarantee (and `TestMergeManifests_PreservesServerVersion`).
- **Conflicting non-empty values are rejected** with a clear, fail-closed error naming the field and both values, instead of a silent discard — consistent with the repo's load-time validators that turn a silently-wrong policy into an up-front config error.

Pure-metadata fields that only *label* the policy (`name`, `version`, `description`, `defaultTtl`, `audience`) remain inherited from the first file — none feeds enforcement or drift, and they naturally differ per file — with that precedence now documented and asserted explicitly. (`version` pinning is handled separately by `expectVersion`, which the gateway already refuses to combine with multiple policy files.)

## Scope / compatibility

No real config is affected today: no demo/example config pins `serverVersion`, and only schema `0.1` is loadable, so a `schemaVersion` conflict cannot arise through `LoadManifest` (that guard is forward-looking/defensive). All `MergeManifests` callers (`route.go`, the `validate` subcommand, `doctor.go`) already surface the `error` return.

## Tests / docs

- Rewrote `TestMergeManifests_InheritsFirstManifestMetadata` (which previously asserted the silent `serverVersion` discard) to assert the inherit-from-first precedence on the metadata fields with a non-conflicting pin.
- Added `TestMergeManifests_RejectsConflictingServerVersion` and `TestMergeManifests_RejectsConflictingSchemaVersion`.
- Documented the multi-file merge rule in `docs/capability-manifest-guide.md` (`serverVersion` section).

`go test -race ./...`, `go vet`, and `gofmt -l` all clean.

https://claude.ai/code/session_01BEQRLdkD2Z1nbAGQ2MZyuV

---
_Generated by [Claude Code](https://claude.ai/code/session_01BEQRLdkD2Z1nbAGQ2MZyuV)_


internal/config/manifest.go
```diff
@@ -166,8 +179,14 @@ func MergeManifests(ms []*LocalManifest) (*LocalManifest, error) {
 	}
 	for _, m := range ms {
 		merged.Capabilities = append(merged.Capabilities, m.Capabilities...)
-		if merged.ServerVersion == "" && m.ServerVersion != "" {
-			merged.ServerVersion = m.ServerVersion
+		// Fold the single-value enforcement/drift fields with a conflict check
+		// (first non-empty wins; a second, disagreeing non-empty value is an
+		// error) instead of silently keeping the first manifest's value.
```
eunolabs, line 185: Nit (non-blocking): `ServerVersion` and `SchemaVersion` are now derived twice. The `merged` struct literal above still pre-seeds them from `ms[0]` (`SchemaVersion: ms[0].SchemaVersion`, `ServerVersion: ms[0].ServerVersion`), and then this loop re-folds them across *all* manifests — including `ms[0]` itself on the first iteration. The first-iteration fold of `ms[0]` against the pre-seeded value is a harmless no-op (equal values), so this is correct, just redundant.

Since the loop now fully derives both fields (first-non-empty across every file), the two pre-seeds in the struct literal are dead. Dropping `ServerVersion`/`SchemaVersion` from the literal — leaving only the genuinely inherited-from-first metadata (`Name`, `Version`, `Description`, `DefaultTTL`, `Audience`) — would make the code structure match the doc-comment's "inherited vs folded" split 1:1 and give each field a single source of truth. Behavior is identical either way.

---

eunolabs (commented):
## Review

Turns a silent policy-data-loss bug into an up-front, fail-closed error when several `--policy` files disagree on a single-value field. The reasoning is sound and the change is consistent with the repo's "fail closed / surface a silently-wrong policy as a load-time error" philosophy. Nicely scoped: capabilities still union, the split-policy guarantee (a pin set in any one file survives) is preserved, and only the two fields that actually drive behavior are folded with a conflict check.

### What's good
- **Correct fold semantics.** `mergeSingleValueField` is first-non-empty-wins with a conflict on disagreement. Detection is order-independent (any two distinct non-empty values collide regardless of file order), and identical values across files merge cleanly — matching the "or make the values match" escape hatch in the error text.
- **The `ms[0]` short-circuit isn't a hole.** `len(ms) == 1` returns early, so the conflict path only runs for genuine multi-file merges; a lone file can't conflict with itself.
- **Field classification holds up.** I checked the fields the PR labels "pure metadata, not enforced":
  - `Audience` (`LocalManifest.Audience`) is write-only — the only non-test reader is the merge copy itself; JWT `aud` pinning comes from a separate CLI/gateway-config field, not the manifest. So inheriting only the first file's value is safe.
  - `DefaultTTL` is documented "informational only — not enforced by the proxy" (`docs/capability-manifest-guide.md`) and is likewise only touched by the merge.
  - That means `serverVersion` (FM-4) and `schemaVersion` are genuinely the only single-value enforcement/drift fields — the PR folds exactly those two and no others. Good.
- **Error handling surface unchanged.** `MergeManifests` already returned `error` (via `validateLocalManifest`), and all callers already handle it — no new propagation work.
- **Tests + docs.** `RejectsConflictingServerVersion` asserts the error names the field and both values; `PreservesServerVersion` already covers a pin in a *later* file surviving; docs updated in the same PR per convention.

### Minor / optional
1. **Redundant double-derivation of `ServerVersion`/`SchemaVersion`** — see the inline note on `internal/config/manifest.go`. Non-blocking; behavior is identical.
2. **Consider one more test:** two files declaring the *same* non-empty `serverVersion` merging cleanly. The equal-value (no-conflict) branch of `mergeSingleValueField` is currently only exercised implicitly via the `ms[0]` self-comparison; a dedicated assertion would lock the documented "make the values match" path against regressions.
3. **`schemaVersion` conflict is defensive/unreachable today** (only `0.1` loads, so every manifest through `LoadManifest` carries `schemaVersion: "0.1"`). The PR body already calls this out as forward-looking. No objection — folding it symmetrically with `serverVersion` is the right instinct for when a second grammar version ships; just flagging that it's dead code until then so it's a conscious choice rather than assumed-live.

### CI
Build / Lint / gofmt / vuln scans green; `Test` and `Analyze (go)` were still in progress at review time — worth a glance before merge, though the change is low-risk.

Nothing here blocks merge — the subst8antive points are confirmations, and the rest are nits.

---

internal/config/manifest.go
eunolabs: Addressed in 4e6abea: dropped `ServerVersion`/`SchemaVersion` from the struct literal (with a short comment so they don't get re-added), leaving the loop as the single source of truth for the two folded fields. Also added `TestMergeManifests_AllowsMatchingServerVersion` to lock the equal-value (no-conflict) branch that was previously only exercised implicitly. `go test -race ./internal/config/`, `go vet`, and `gofmt -l` clean.

---

Issue #1050, opened 2026-06-21
[rev-20260621T0610Z-83a06b] allowedValues glob: percent-encoded slash (%2f) crosses a single-`*` confinement boundary, defeating single-segment path scoping


eunolabs:
**Location:** `pkg/enforcement/glob.go:119-147` (the path-confinement branch + the non-`**` fast path; the load-bearing decision is the "match literally" comment at `glob.go:137-141`)

**Reference:** https://github.com/eunolabs/eunox/blob/fb69bbd43ed85a250c432853d353cdb7234e71cc/pkg/enforcement/glob.go#L119-L147

**Severity:** High (authorization-scope / path-confinement bypass on the `allowedValues` value matcher; reachable from `handlers.go` `MatchAllowedValue` and the JWT PDP capability shorthand in `internal/pdp/jwt.go`)

**Description:**
`MatchValueGlob` documents (`glob.go:88-90`) that a single `*` "matches a run of non-`/` characters and does NOT cross `/`", which is the mechanism an operator uses to confine a value to a single path segment — e.g. `allowedValues: ["/reports/*"]` to permit only direct children of `/reports`, not files in subdirectories.

The path-confinement branch (`glob.go:119-142`) percent-decodes the value via `decodePathForConfinement` **only** to scan for `.`/`..` segments, then throws the decoded form away and runs the actual `path.Match` on the **non-decoded** value (line 141 folds backslashes only; lines 146/162 never decode). The comment at `glob.go:137-141` defends this as intentional: "leaving a legitimately percent-encoded value to match its pattern literally."

That decision is internally contradictory and reopens a confinement bypass. The confinement scan decodes `%2f`→`/` **because it assumes the upstream decodes `%2f` once** (the same single-decode model documented at `glob.go:16-22`). But the match then treats `%2f` as three literal characters that a single `*` happily spans. So under the very upstream model the scan is built for, a value the operator scoped to one segment reaches a deeper path.

This is distinct from the `..` traversal-out class (which the scan still correctly blocks): full traversal via `..%2f..%2f` is rejected, but **depth/segment scoping** via a single `*` is silently defeated.

**Trigger / repro:** (verified against this SHA by calling the exported `MatchValueGlob`)

```
MatchValueGlob("/reports/*", "/reports/sub%2fsecret")     = true   // upstream resolves /reports/sub/secret
MatchValueGlob("/reports/*", "/reports/sub/secret")       = false  // the literal form is correctly excluded
[...]

PR #1114, opened 2026-06-22
sec: enforce encoded-separator confinement on single-* segments in mixed globs

### LINKED ISSUE #1107: [rev-20260622T0545Z-2317a163] Incomplete fix of #1050: mixed `**`+`*` glob pattern skips the encoded-separator-count guard, so a single-`*` segment still spans %2f
**Location:** `pkg/enforcement/glob.go:187`

**Severity:** Medium (single-segment path-scoping bypass for upstreams that percent-decode paths; follow-up to closed #1050)

## Description

The fix for #1050 added an encoded-separator-count guard so a single-`*` segment cannot span a percent-encoded `/`:

```go
if !strings.Contains(pattern, "**") && strings.Count(scan, "/") != strings.Count(value, "/") {
    return false
}
```

But the guard is gated on `!strings.Contains(pattern, "**")`, so it is **skipped for any pattern containing `**`**. In a *mixed* pattern like `/a/*/**`, the encoded separator lands in the single-`*` segment (which `**` does not cover), so the hole #1050 closed for pure single-`*` patterns reopens whenever `**` co-occurs with `*`.

## Trigger / repro (runtime-confirmed at current HEAD)

- `MatchValueGlob("/a/*", "/a/x%2fy")` → `false`  (guard fires — correct, the #1050 fix)
- `MatchValueGlob("/a/*/**", "/a/x%2fy/z")` → `true`  (guard skipped — the `*` segment matches `x%2fy` literally)

An upstream that percent-decodes resolves `x%2fy` as two path segments `x/y`, crossing the single-segment boundary the operator scoped with `*`. (No `..` traversal escape — the `.`/`..` confinement scan still holds; this is single-segment *depth* confinement only.)

## Impact

A capability scoped with a mixed `*`+`**` path pattern admits values an operator intended to confine to one segment. Reopens #1050 for the mixed-pattern case.

## Suggested fix

Enforce the per-segment decoded-vs-literal `/`-count agreement on the non-`**` portion of the pattern instead of disabling the whole guard when `**` is present. Add a table-driven case `MatchValueGlob("/a/*/**", "/a/x%2fy/z")` expecting `false`.

Related: closed #1050 (pure single-`*` case), closed #1036 (`..`-out via encoded dots).

eunolabs:
Fixes #1107.

## Problem

The `allowedValues` path-glob encoded-separator guard (the #1050 fix) was gated on `!strings.Contains(pattern, "**")`, so it was **skipped for any pattern containing `**`**:

```go
if !strings.Contains(pattern, "**") && strings.Count(scan, "/") != strings.Count(value, "/") {
    return false
}
```

In a **mixed** pattern like `/a/*/**`, the encoded separator lands in the single-`*` segment (which `**` does not cover), so the hole #1050 closed for pure single-`*` patterns reopened whenever `**` co-occurred with `*`.

Runtime-confirmed before the fix:

- `MatchValueGlob("/a/*", "/a/x%2fy")` → `false`  (guard fires — correct, the #1050 fix)
- `MatchValueGlob("/a/*/**", "/a/x%2fy/z")` → `true`  (guard skipped — the bug; the `*` segment matches `x%2fy` literally)

An upstream that percent-decodes resolves `x%2fy` as two segments `x/y`, crossing the single-segment boundary the operator scoped with `*`.

## Fix

Enforce the guard **per single-segment element** rather than disabling it wholesale when `**` is present:

- The **non-`**` fast path** keeps its existing total-`/`-count guard (every segment is single-`*`, so any encoded separator anywhere is illegal).
- For a **`**` pattern**, flag each value segment that decodes to span a `/` (carries an encoded separator) and let `matchGlobSegments` forbid a single-segment pattern element (`*`/`?`/`[…]`) from matching it. A genuine `**` segment may still absorb such a value, so legitimate subtree matches (`/reports/**` vs `/reports/a%2fb`) are unaffected.

This keeps the failure posture fail-closed for the ambiguous case while preserving the documented `**`-spans-separators behavior. Note that literal `\` separators are already folded to `/` before the segment split, so only percent-encoded separators (`%2f`, `%5c`) needed the new per-segment check.

## Tests

Added table-driven cases in `pkg/enforcement/glob_test.go`:

- deny: `/a/*/**` vs `/a/x%2fy/z` (the issue's repro), the `%5c` variant, and the trailing-`*` variant `/a/**/*` vs `/a/b/x%2fy`
- allow: `/a/*/**` vs `/a/b/x%2fy` and a deeper variant — encoded separator in the `**` portion is still admitted (no single-segment widening)

`go test -race -count=1 ./...` passes repo-wide; `glob.go` keeps 100% function coverage.

## Docs

Updated `docs/capability-manifest-guide.md` (the encoded-separator paragraph now covers the mixed-pattern case) and added a curated **Security** entry to `CHANGELOG.md`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01713ySpgGftdK1XRz5TMY91)_


pkg/enforcement/glob.go
```diff
@@ -252,8 +280,14 @@ func matchGlobSegments(pat, val []string) bool {
 			}
 			continue
 		}
-		// A non-"**" segment must match exactly one value segment.
+		// A non-"**" segment must match exactly one value segment -- and one that
+		// does not itself decode to span a '/', so a single-segment element never
+		// absorbs an encoded separator the operator scoped out (a "**" segment,
+		// handled above, still may).
 		for j := n - 1; j >= 0; j-- {
+			if valSpansSep != nil && valSpansSep[j] {
```
eunolabs: **Blocker — this fails CI lint on the current head commit.** `golangci-lint` (gocritic `weakCond`) rejects the build here:

```
pkg/enforcement/glob.go:288:7: weakCond: suspicious `valSpansSep != nil && valSpansSep[j]`; nil check may not be enough, check for len (gocritic)
```

The nil guard is semantically fine — `valSpansSep` is only ever `nil` or exactly `len(val)` long, and `j` is bounded by `n-1` — but gocritic wants the index access guarded by a length check. Swapping to a bounds check both silences the linter and is strictly safer (it's an explicit guarantee rather than an invariant the reader has to reconstruct):

```go
if len(valSpansSep) > j && valSpansSep[j] {
```

When `valSpansSep` is nil this is `len == 0 > j` → false, identical to the current "no confinement" behavior. Worth a quick `make lint` locally before pushing.

---

pkg/enforcement/glob.go
```diff
@@ -172,20 +179,34 @@ func MatchValueGlob(pattern, value string) bool {
...
+			if strings.Count(scan, "/") != strings.Count(value, "/") {
+				return false
+			}
+		} else {
+			segs := strings.Split(value, "/")
+			valSpansSep = make([]bool, len(segs))
+			for i, seg := range segs {
+				dec, derr := decodePathForConfinement(seg)
+				valSpansSep[i] = derr != nil || strings.Contains(dec, "/")
+			}
```
eunolabs, line 217: Minor / optional (not blocking): this builds `valSpansSep` — an `O(n)` allocation plus a `url.PathUnescape` per value segment — *before* `matchGlobSegments` applies the `n > maxGlobSegments` cap. The value segment count `n` is attacker-controlled (the same threat the cap exists for), so a value with millions of `/` segments does this whole pass and allocates the slice only to be rejected a moment later.

It isn't new exposure — the `"."/".."` scan above already does an unbounded `strings.Split` over the decoded value — so the marginal cost is just the extra per-segment unescape. But since you're touching this path, it might be worth applying the segment-count cap up front (before both scans) so the two O(n) passes can't be driven past the documented bound. Your call given it mirrors pre-existing behavior.

Separately: `derr != nil` here is effectively unreachable — by this point the whole value has already decoded cleanly in the `"."/".."` scan (a `%XX` can't straddle a `/`, so no `/`-free segment can fail to decode if the whole value didn't). It's harmless fail-closed defense-in-depth, just noting it in case the redundancy is worth a comment.

---

eunolabs (commented):
## Review — encoded-separator confinement on single-`*` segments in mixed globs

Solid, well-scoped security fix. The diagnosis matches #1107 exactly, and the approach — enforce the encoded-separator guard *per single-segment element* instead of disabling it wholesale when `**` is present — is the right shape. I traced the DP by hand on the key cases and the behavior is correct:

- `MatchValueGlob("/a/*/**", "/a/x%2fy/z")` → **false** — the `x%2fy` segment is flagged spanning, the `*` element can't match it, and `**` can't reach it without leaving `*` unsatisfied. ✅
- `MatchValueGlob("/a/*/**", "/a/b/x%2fy")` → **true** — `*` matches the clean `b`, `**` absorbs the spanning segment. ✅
- `/a/**/*` vs `/a/b/x%2fy` → **false** — the trailing `*` is forced onto the spanning last segment. ✅

The `.`/`..` traversal scan still runs on the fully-decoded value, so a decoded `..` inside a `**` subtree is independently caught — the two guards compose cleanly. The generalization from "`*`" to "any single-segment element (`*`/`?`/`[…]`)" via the `valSpansSep[j]` skip in `matchGlobSegments` is the correct way to express it, and `valSpansSep` staying nil for non-path / non-`**` shapes keeps the fast paths byte-for-byte unchanged.

### Must fix before merge
- **CI Lint is red on the current head** (`gocritic weakCond` at `glob.go:288`). Trivial fix inlined — `len(valSpansSep) > j && valSpansSep[j]`. The newer in-progress lint run is on the same commit, so it will fail identically.

### Minor / non-blocking
- `valSpansSep` is built (alloc + per-segment `url.PathUnescape`) before the `maxGlobSegments` cap; attacker-controlled `n`. Pre-existing scan already has the same unbounded shape, so marginal — details inline.
- Behavioral note worth pinning with a test: this also denies some values that are in-scope under *both* decode interpretations, e.g. `/**/*` vs `/x%2fy` — both a decoding (`/x/y`) and non-decoding (`/x%2fy`) upstream would resolve it within `/**/*`, but the trailing `*` is forced onto the spanning segment so it's now denied. That's a safe over-deny consistent with the fail-closed posture and the existing pure-`*` guard, and it's a change from prior behavior (where `**` patterns skipped the guard entirely). A table case locking it in would make the boundary intentional rather than incidental.

### Tests / docs
Test coverage is good — deny cases for `%2f`, `%5c`, and the trailing-`*` variant, plus allow cases confirming the `**` portion still admits encoded separators. Note that `?`/`[…]` can't actually match a multi-char spanning segment regardless of the guard (length-limited), so `*` is the only element that meaningfully exercises it — the existing `*` cases are the right ones. Docs (`capability-manifest-guide.md`) and the `### Security` CHANGELOG entry are both updated, matching the house convention. Comments are excellent and explain the *why*, not just the *what*.

Once lint is green this is good to go.

---

PR #1115, opened 2026-06-22
sec: reject conflicting capabilities when merging policy files

### LINKED ISSUE #1104: [rev-20260622T0545Z-2317a163] MergeManifests has no duplicate-target detection: a first-listed audit/permissive capability shadows a later restrictive one for the same target (policy tightening silently widens)
**Location:** `internal/config/manifest.go:167` (`MergeManifests`), interacting with `pkg/enforcement/engine.go:786` (`findMatchingCapability` tie-break)

**Severity:** High (fail-open — an operator combining `--policy` files to tighten policy can silently widen it; the effective policy is file-order-dependent)

## Description

`MergeManifests` concatenates the capability lists of multiple `--policy` files with **no duplicate-target or conflict detection**:

```go
for _, m := range ms {
    merged.Capabilities = append(merged.Capabilities, m.Capabilities...)
    // ...
}
```

The enforcement engine breaks ties between two equal-specificity, same-principal-class capabilities by **first-in-declaration-order** (`engine.go:786`):

```go
if score > bestScore || (score == bestScore && hasPrincipal && !bestPrincipal) {
```

An incumbent is replaced only on a strictly-greater score or a principal upgrade — so the **first** matching capability wins (documented at engine.go:779-785, tested at `engine_test.go:4959`). Combined, a permissive entry in the first file shadows a restrictive entry for the same target in a later file.

## Trigger / repro

`--policy a.yaml --policy b.yaml` where:
- `a.yaml`: `{target: tool:read_file, actions: [call], enforcement: audit}` (observe — allow & log, never block)
- `b.yaml`: `{target: tool:read_file, actions: [call], conditions: [{type: allowedValues, argument: path, values: [/reports/*]}]}`

`MergeManifests` yields `[A, B]`; `findMatchingCapability` scores both identically and selects `A`. `B`'s `allowedValues` restriction never runs and denials are downgraded to allow-and-log. Swap the file order and the result flips.

## Impact

An operator who adds `b.yaml` to **tighten** policy instead **widens** it, with no error or warning. The effective policy silently depends on `--policy` argument order.

## Suggested fix

Detect duplicate `(target, actions, principal-class)` tuples across merged files and either reject the merge with a clear error, or define and document deterministic precedence (e.g. most-restrictive-wins) instead of relying on the engine's incidental first-in-order tie-break. Add a test asserting the chosen precedence with two same-target capabilities of differing restrictiveness across two files.

eunolabs:
Fixes #1104.

## Problem

`MergeManifests` concatenated the capability lists of multiple policy files with **no conflict detection**:

```go
for _, m := range ms {
    merged.Capabilities = append(merged.Capabilities, m.Capabilities...)
    // ...
}
```

The enforcement engine breaks equal-specificity, same-principal-class ties by **first-in-declaration order** (`engine.go:786`) — the incumbent is replaced only on a strictly-greater score or a principal upgrade. Combined, a permissive or `enforcement: audit` entry in an earlier file **silently shadows** a restrictive entry for the same target in a later file.

The result is fail-open: an operator who adds `b.yaml` to **tighten** policy instead **widens** it, and the effective policy depends on the order the files happen to be listed (a route's `policy: [...]` list, or `eunox-mcp validate a.yaml b.yaml`), with no error or warning.

## Fix

Detect, at merge time, two capabilities from **different files** that target the same resource with the **same principal scope** and **overlapping actions**, and reject the merge with a clear error naming the target. This matches the codebase's pervasive "reject ambiguity at load, fail closed" posture and is the issue's first suggested option.

Scope chosen to flag only ties the engine would actually take (no false positives):

- **Cross-file only.** Within a single file the first-in-order tie-break is documented, tested behavior (`engine_test.go:4959`, guide § principal tie-breaking) and the operator sees both entries in one place — it is left untouched.
- **Principal scope compared by exact, canonicalized signature.** A general entry refined by a principal-scoped one for the same target does **not** conflict (the engine prefers the scoped entry deterministically, regardless of order); two entries scoped to **different** principals do **not** conflict (a request satisfies at most one). Only same-target + same-principal-signature + overlapping-actions pairs are rejected.
- **Actions compared by overlap** (`*` overlaps everything).

Error surfaces through every merge caller: gateway route construction (`BuildRoutes` / `LoadUpstreamPDP`), `eunox-mcp validate`, and `doctor`.

## Tests

- `TestMergeManifests_RejectsCrossFileConflict` — the issue repro (audit shadows restrictive `tool:read_file`) asserted in **both** file orders (proving the outcome no longer depends on ordering), plus wildcard-action overlap, identical-principal scoped duplicates, and a reordered-pattern principal (pins the signature's order-independence).
- `TestMergeManifests_AllowsNonConflicting` — distinct targets, distinct principals, general-refined-by-principal, and a within-file duplicate (confirms no over-rejection).

## Docs

Updated the capability-manifest guide's tie-breaking section to document the cross-file rejection (when it fires, and what does **not** conflict).

`make check-fmt check-license lint` clean; `go test -race ./...` green.

https://claude.ai/code/session_01N29kua5EVyT1AgKsZDfrGL

---
_Generated by [Claude Code](https://claude.ai/code/session_01N29kua5EVyT1AgKsZDfrGL)_


internal/config/manifest.go
```diff
@@ -181,6 +196,109 @@ func MergeManifests(ms []*LocalManifest) (*LocalManifest, error) {
...
+// shadowing.
+//
+// Principal scope is compared by exact, canonicalized signature, so the check
+// flags only ties the engine would actually take:
+//   - A general (unscoped) entry and a principal-scoped entry for the same
+//     target do NOT conflict — the engine prefers the scoped entry when its
+//     principal matches and the general one otherwise, deterministically and
+//     regardless of order.
+//   - Two entries scoped to DIFFERENT principals do NOT conflict — a request
+//     can satisfy at most one, so there is no tie to resolve by order.
```
eunolabs: This premise is unsound, and it's the root of a residual fail-open. `PrincipalMatches` requires all of an entry's *own* claims (AND within an entry), but two entries keyed on **different** claims can both be satisfied by a single token — so a request can satisfy more than one.

Repro: `{sub: [alice]}` in a.yaml (`enforcement: audit`) and `{iss: [acme]}` in b.yaml (restrictive). A token carrying both `sub=alice` and `iss=acme` matches both. Same target, equal specificity, both `hasPrincipal=true` → the engine keeps the first in file order (engine.go:786) and silently drops b.yaml's restriction. Their `principalSignature`s differ, so `detectMergeConflicts` lets it through — exactly the cross-file, order-dependent shadowing this function is meant to catch.

(Disjoint values on the *same* claim, like `agent_id: ci-bot` vs `release-bot`, genuinely can't co-match and are correctly allowed — that's what your allow-test covers.)

Suggest either flagging cross-file same-target/overlapping-action pairs where both are principal-scoped and the principals are co-satisfiable (no shared claim constrained to disjoint values), or correcting this rationale and documenting the residual. Full write-up in the review summary.

---

internal/config/manifest.go
```diff
@@ -181,6 +196,109 @@ func MergeManifests(ms []*LocalManifest) (*LocalManifest, error) {
...
+	// compared against every earlier one sharing its key.
+	type prior struct {
+		manifestIdx int
+		actions     []string
+	}
+	seen := make(map[string][]prior)
+	for mi := range ms {
+		for ci := range ms[mi].Capabilities {
+			c := &ms[mi].Capabilities[ci]
+			key := c.Target + "\x00" + principalSignature(c.Principal)
```
eunolabs: Keying on the exact `Target` string is sound for avoiding false positives (identical target strings always tie in specificity, and `validateLocalManifest` guarantees every target is namespace-prefixed before merge, so the raw key is canonical). But note it covers identical target strings, not all engine ties: two different *glob* targets can tie for some request and still be resolved by file order. e.g. `tool:read_*` and `tool:*_file` both score 49 (5 literal runes − 1 wildcard in `resourceSpecificity`) and both match `read_file`. Across files the engine breaks that tie positionally too.

Fully closing this needs glob-intersection analysis and is probably out of scope — but worth a one-line acknowledgement that the check targets identical target strings rather than every equal-specificity tie.

---

docs/capability-manifest-guide.md
```diff
@@ -765,6 +765,22 @@ Semantics:
...
+  a target can instead leave the new entry inert, or an earlier
+  `enforcement: audit` entry can downgrade a later restrictive one to
+  allow-and-log. So eunox **fails closed at load**: if two files declare the same
+  target with overlapping actions and the **same** `principal` scope, the merge is
+  rejected with an error naming the target. Combine the entries into one file
+  (where the documented first-wins applies), or scope them to distinct principals.
+  A general entry in one file and a principal-scoped entry for the same target in
+  another do **not** conflict — the engine resolves that pair deterministically
+  (refinement beats order) — and neither do two entries scoped to different
+  principals.
```
eunolabs: This operator-facing guidance is partly unsound. "Scope them to distinct principals" is not a reliable remedy, and "neither do two entries scoped to different principals [conflict]" is incorrect when the two entries key on **different claims**.

Principal matching is AND *within* an entry, not across entries, so a single token can satisfy two different-claim scopes at once: `{sub: [alice]}` in one file (audit) and `{iss: [acme]}` in another (restrictive) — a token with both claims matches both, ties at equal specificity, and the engine keeps whichever file was listed first. That's the same order-dependent shadowing this section says is rejected, but the merge check passes it (different `principalSignature`).

Distinct values on the *same* claim (`agent_id: ci-bot` vs `release-bot`) are genuinely safe — those can't co-match. Recommend narrowing the claim here to that case and noting the different-claim residual. Repro in the review summary.

---

eunolabs (commented):
## Review: reject conflicting capabilities when merging policy files

Well-targeted fix for the reported fail-open, and most of the scoping reasoning holds up. I checked the model against the engine: `findMatchingCapability` scores on `resourceSpecificity * resourceScoreWeight` only, with `hasPrincipal` as a binary tiebreaker and equal-class ties resolved first-in-order (engine.go:770-790, tie-break at 786). One real residual fail-open, plus a couple of smaller notes.

### What's correct
- **Exact-`Target` keying has no false positives.** Identical target strings always tie in specificity for any request that matches both, and `validateLocalManifest` rejects unprefixed targets and empty `actions` before merge (manifest.go:217-223), so the raw-string key is canonical.
- **`actionsOverlap` empty-list handling is moot** — empty `actions` can't survive load, and the engine treats them as a non-match anyway.
- **The general-vs-principal-scoped exemption is right** — the engine prefers the scoped entry when its principal matches and the general one otherwise, deterministically and order-independently.
- **Asserting both file orderings** in the test is exactly the right way to prove order-independence.

### Main concern — the "different principals" exemption is unsound (residual fail-open)
The rationale (manifest.go:225-226 and the guide) says two entries scoped to *different* principals "do NOT conflict — a request can satisfy at most one." That premise doesn't hold. Principal matching is AND *within* an entry, so two entries that key on **different claims** can both be satisfied by one token:

```yaml
# a.yaml (listed first)
- target: tool:deploy
  actions: [call]
  principal: { sub: [alice] }
  enforcement: audit            # observe-only, never blocks
```
```yaml
# b.yaml (listed second)
- target: tool:deploy
  actions: [call]
  principal: { iss: [https://idp.acme] }
  conditions: [ { type: allowedValues, argument: env, values: [staging] } ]
```

A token with `sub=alice` **and** `iss=https://idp.acme` satisfies both. Same target, equal specificity, both principal-scoped → the engine keeps the first in file order (a.yaml's audit entry) and silently drops b.yaml's `env: staging` restriction. Swap the file order and the behavior flips — the exact order-dependence this PR sets out to eliminate. `principalSignature` differs between the two, so `detectMergeConflicts` passes it.

Note the allow-test "same target, distinct principals" uses `agent_id: ci-bot` vs `agent_id: release-bot` — same claim, disjoint values, so genuinely un-co-satisfiable and correctly allowed. The gap is specifically *different claims* (or overlapping patterns on the same claim). Suggested options:
- Tighten detection to also flag cross-file same-target/overlapping-action pairs where **both** are principal-scoped and the principals are *co-satisfiable* (i.e. no shared claim constrained to disjoint values); or
- At minimum, correct the "at most one" claim in the comment and guide and document the residual order-dependence, so distinct-claim principals aren't recommended as a remedy that doesn't actually hold.

### Secondary (lower severity, inherent to a load-time check)
Exact-target keying also misses different *glob* targets that tie for some request — `tool:read_*` and `tool:*_file` both score 49 and both match `read_file`, so the engine resolves that pair by file order too. Fully closing it needs glob-intersection analysis; documenting it as a known residual is the pragmatic call.

### Nits
- The conflict error is a single ~90-word sentence — informative, but a short leading clause ("conflicting capabilities for target %q across policy files: …") would read better in a terminal.
- Rejecting byte-identical duplicate entries across files (a plausible "shared base policy" composition) is a behavior change worth a sentence in the docs — defensible under fail-closed, just call it out.

Overall a clear improvement that closes the reported case cleanly. I'd want the different-principal rationale corrected (and ideally detection broadened or the gap explicitly documented) before it reads as fully closing order-dependent cross-file shadowing.

---

PR #1117, opened 2026-06-22
sec: audit-verify across the whole rotated set, not just the base log

### LINKED ISSUE #1106: [rev-20260622T0545Z-2317a163] audit-verify reads only the current base log, so deletion of an entire interior rotated file is undetectable (cross-file HMAC chain never walked)
**Location:** `cmd/eunox-mcp/main.go:1862` (single-file `VerifyLog`) + `internal/audit/verify.go:326` (`updateChain` first-record branch)

**Severity:** High (tamper-evidence gap — the audit chain is advertised as spanning rotation, but no verifier walks it across files)

## Description

The chain is designed to continue across rotation: `rotate()` carries `seq`/`prevHMAC` in memory so each new file's head record chains to the previous file's tail. But two gaps mean that link is never checked:

**1. `audit-verify` opens only the single current base log file:**

```go
logPath, _ := audit.ResolveLogPath(*auditLogPath)
f := openAuditLogOrExit("audit-verify", logPath)   // os.Open(logPath) — one file
res, err := audit.VerifyLog(f, verifier, ...)
```

Rotated siblings (`rotatedSiblings`/`newestRotatedSibling` exist but are used only by pruning/startup-resume) are never gathered or concatenated, and nothing compares a file head's `prev_hmac` to the prior file's tail HMAC.

**2. `updateChain` validates the genesis sentinel only when `Seq == 1`:**

```go
if rec.Seq == 1 && rec.PrevHMAC != auditGenesisPrev && rec.PrevHMAC != "" {
```

A rotated/post-rotation file whose head record has `Seq > 1` gets no chain check — only an informational `FirstSeq > 1` note that `OK()` ignores (`OK()` checks only `Invalid == 0 && ChainBreaks == 0`).

## Trigger / repro

Chain spans three files: `file1[seq 1..100] -> file2[seq 101..200] -> current[seq 201..250]`. Delete the entire interior `file2`. Verify each remaining file: `current` starts at seq 201 (benign `FirstSeq>1` note, PASS), `file1` ends clean (PASS), `file2` is gone. The cross-file link (`current`-head.`prev_hmac` == `file2`-tail.`_hmac`, and seq contiguity) is never compared — `audit-verify` exits 0.

## Impact

An attacker who can delete a whole rotated segment removes those records with no tamper evidence — even though the rotated siblings exist on disk and *could* be chained. Docs acknowledge trailing/leading truncation as undetectable from a single file (verify.go:152-154), but not interior whole-file deletion. Distinct from the related closed issues #418 (genesis sentinel on non-rotated logs), #868/#661 (post-wrap seq-0 false INVALID), and #448 (chain reset after rotation-fallback restart) — this is specifically the single-file verify *scope*.

## Suggested fix

Have `audit-verify` discover and verify the full rotated set in order (oldest → current), threading `prevHMAC`/`seq` across file boundaries and failing on a head whose `prev_hmac` doesn't match the prior file's tail HMAC or whose `seq` isn't contiguous. At minimum, **fail** (not just note) when a verified file's head has `Seq > 1` and no preceding file is supplied.

eunolabs:
## Problem

Fixes #1106.

The tamper-evident audit chain is designed to span rotation — `rotate()` carries `seq`/`prev_hmac` in memory so each new file's head record links to the previous file's tail. But `audit-verify` only ever opened the **single current base log**, so deletion of an entire interior rotated file was undetectable: each surviving file verified on its own, and nothing compared a file head's `prev_hmac`/`seq` to the prior file's tail.

Repro from the issue: a chain spanning `file1[1..100] -> file2[101..200] -> current[201..250]`. Delete the whole interior `file2`. `current` starts at seq 201 (benign `FirstSeq>1` note, PASS), `file1` ends clean (PASS) — `audit-verify` exits `0` even though a whole segment was removed.

## Fix

Verify the full rotated set as **one chain** instead of a single file. This reuses the existing, heavily-invariant-checked chain logic rather than duplicating state-threading — the cross-file links simply become ordinary consecutive-record checks once the files are joined in order.

- **`audit.LogChainFiles(logPath)`** — discovers the ordered chain files (every genuine rotated sibling oldest-first via the existing `rotatedSiblings`/`rotatedAuditRe`, then the current base log). Excludes unrelated siblings (`.bak`, `.lock`); handles the just-rotated-away case (absent base, siblings present) and the fresh-install case (nothing on disk → empty slice → first-run hint).
- **`audit.VerifyLogFiles(paths, ...)`** — opens them in order and runs the existing `VerifyLog` over the newline-separated concatenation. An interior whole-file deletion now trips **both** a `CHAIN BREAK` and a `SEQ GAP` at the boundary. A chain file that can't be opened **fails closed** (can't certify what it can't read).
- **`cmdAuditVerify`** — gathers and verifies the set, prints how many files it walked as one chain (when >1), and reframes the `FirstSeq>1` note around leading records / whole leading rotated files removed or pruned by retention.

The genesis/`seq 1` check now applies to the oldest file's head. Trailing truncation and leading-record / leading-file removal (which is also what `--audit-retain` prunes) remain the only cases unprovable from the local files without an external high-water mark — unchanged and documented.

## Tests

- `TestVerifyLogFiles_DetectsInteriorWholeFileDeletion` — the regression: a 3-file chain verifies clean across boundaries; deleting the interior file is detected (`CHAIN BREAK` + `SEQ GAP`, non-zero verdict).
- `TestLogChainFiles_OrdersChainAndExcludesUnrelated` — ordering, exclusion of unrelated siblings, base-absent, and fresh-install cases.
- `TestVerifyLogFiles_MissingFileFailsClosed` — an unreadable chain file errors rather than being skipped.
- `TestCmdAuditVerify_VerifiesRotatedSet` — CLI walks the rotated set as one chain and reports both files / all records.

Full suite passes with `-race`; `go vet`, `golangci-lint`, gofmt, license-header, and NOTICE checks all clean.

## Docs

Threat model §3.4, `conformance.md`, `architecture.md`, and `CHANGELOG.md` updated to state that `audit-verify` walks the whole rotated set and now catches interior whole-file deletion.

https://claude.ai/code/session_01Ghii1g4dLJGiLbr3Be2hwQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ghii1g4dLJGiLbr3Be2hwQ)_


internal/audit/verify.go
```diff
@@ -168,6 +169,49 @@ func VerifyLog(r io.Reader, verifier *Sink, requestID string, since time.Time, o
...
+// read cannot be certified, so audit-verify fails closed.
+func VerifyLogFiles(paths []string, verifier *Sink, requestID string, since time.Time, out io.Writer) (VerifyResult, error) {
+	files := make([]*os.File, 0, len(paths))
+	defer func() {
+		for _, f := range files {
+			_ = f.Close()
+		}
+	}()
+	readers := make([]io.Reader, 0, len(paths)*2)
+	for i, p := range paths {
```
eunolabs, line 201: **Open-FD exhaustion under the keep-all retention default.** This loop opens *every* chain file up front and holds them all open until the function returns (the deferred close loop), then hands them to `io.MultiReader`. Under the default `--audit-retain 0` (keep every sidecar) the number of rotated siblings is unbounded — a long-lived log at the default 100 MiB rotation size can accumulate enough sidecars to pass the open-file ulimit, at which point `os.Open` here returns `EMFILE` and `audit-verify` can no longer verify the very logs the keep-all default was meant to preserve.

`io.MultiReader` reads its readers strictly sequentially, so we never actually need more than one open at a time. Consider opening each file lazily (a tiny `io.Reader` that `os.Open`s on first `Read` and closes at EOF) so the open-FD count stays ~1 regardless of sibling count; the fail-closed-on-open-error behavior is still easy to preserve there.

Not a correctness bug at typical sizes — flagging because keep-all is exactly the case that grows without bound.

---

internal/audit/rotate.go
```diff
@@ -157,6 +157,55 @@ func newestRotatedSibling(logPath string) string {
...
+	// such a file into the chain walk would either inject a spurious record or
+	// (worse) seed the cross-file link from the wrong source. The [:0:0] reslice
+	// reuses none of siblings' backing array, so the first append allocates fresh.
+	files := siblings[:0:0]
+	for _, m := range siblings {
+		if rotatedAuditRe.MatchString(m[len(logPath):]) {
+			files = append(files, m)
+		}
+	}
+	sort.Slice(files, func(i, j int) bool {
```
eunolabs, line 197: **Minor / known-limitation note.** Ordering here is by the filename timestamp (`rotatedOrderLess`), which equals the chain's true write order only while the rotation clock is monotonic. If the wall clock steps backward between two rotations (e.g. an NTP correction), a newer sidecar can receive an earlier stamp, sort before its predecessor, and then `VerifyLogFiles` would report a spurious cross-file `CHAIN BREAK`/`SEQ GAP`.

This matches the assumption `newestRotatedSibling`/`pruneRotated` already make, and it errs in the safe direction (a *false positive* in a tamper-evidence tool, never a false negative), so I think it's fine to ship — just noting it. If we ever want it bulletproof, ordering siblings by each file's first-record `seq` instead of the filename would remove the clock dependency.

---

eunolabs (commented):
## Review (self-review)

**What it does:** Makes `audit-verify` verify the whole rotated set as one tamper-evident chain (`LogChainFiles` discovers `rotated siblings oldest-first + current base`; `VerifyLogFiles` concatenates them and runs the existing `VerifyLog`), so deleting an entire interior rotated file is now caught at the cross-file boundary instead of slipping past single-file verification (#1106).

**What's good:**
- The concatenation approach reuses the existing, invariant-heavy chain logic (genesis/seq-1 check, forged-seq-0/legacy guards, post-wrap handling) rather than duplicating cross-file state-threading — the boundary checks fall out for free as ordinary consecutive-record links. Lower risk than a parallel code path.
- Discovery correctly reuses `rotatedSiblings`/`rotatedAuditRe`/`rotatedOrderLess`, filters out unrelated siblings (`.bak`/`.lock`), and the base log is positioned last *unconditionally* (not by timestamp), which is robust against the rotation reopen-fallback state.
- Fail-closed on an unreadable chain file; preserves the first-run hint when nothing exists.
- Test coverage is solid: the interior-deletion regression (3-file chain, both `CHAIN BREAK` and `SEQ GAP` asserted), discovery ordering/exclusion/base-absent/fresh-install, fail-closed-on-missing, and a CLI end-to-end multi-file pass. Docs (threat model §3.4, conformance, architecture) + CHANGELOG updated in-PR.

**Findings (2, both non-blocking):**
1. **`VerifyLogFiles` opens all chain files eagerly** and holds the FDs open for the whole pass — unbounded under the keep-all retention default. Inline suggestion to open lazily (details on `verify.go`). This is the one I'd most like a second opinion on.
2. **Sibling ordering is by filename timestamp**, so a backward clock step during rotation could yield a spurious cross-file break. Errs safe (false positive, not false negative) and matches existing assumptions — noted as a known limitation on `rotate.go`.

No correctness blockers found in the happy path; the chain math and seq accounting check out. I'd like to resolve finding #1 (or consciously accept it) before merge.

---

internal/audit/verify.go
eunolabs: Addressed in ad4a3ac. `VerifyLogFiles` now opens each file lazily via a `lazyFileReader` (opens on first `Read`, closes at EOF); since `io.MultiReader` reads sequentially, at most one audit-log file is open at a time regardless of sibling count. Fail-closed on an open error is preserved — it surfaces through `VerifyLog` as a read error — and a deferred backstop closes any file left open if `VerifyLog` exits early. Added a mid-stream open-error test (readable first file + unreadable second) to confirm a non-first unreadable file still fails closed.

Leaving the sibling-ordering note (other thread) as-is: it errs safe (false-positive break, never false-negative) and matches the existing `newestRotatedSibling`/`pruneRotated` assumption.

---

Issue #1109, opened 2026-06-22
[rev-20260622T0545Z-2317a163] Code-review follow-ups: lower-severity correctness + cleanup findings (tracking)


eunolabs:
Umbrella tracking issue for the lower-severity findings from a full-codebase review. The 5 substantive security findings are filed as separate issues. The items below are real but lower-severity (documented limitations, hot-path efficiency, maintainability) — grouped here for triage.

> Note: the JWKS `headers[0].KeyID` single-header finding from this review is already tracked by **#1077** (native capability-token path) and **#1103** (IdP-JWT gateway path) — not repeated here.

## Correctness (lower severity / documented limitations)

- [ ] **JWT `op=` SQL heuristic authorizes uninspected content** — `internal/pdp/jwt.go:1500`. In scan-all-args mode a call is allowed when any string argument's first word is a granted SQL verb, regardless of what the binding argument contains (e.g. grant `op=SELECT`, args `{cmd: "rm -rf /", note: "SELECT 1"}` → allowed; `rm` isn't a SQL verb so it isn't hard-denied). Documented best-effort heuristic (the `sqlVerbs` godoc says to pair with a least-privilege DB role); surfaced as the one path where a JWT grant authorizes content it never inspected.
- [ ] **Decision dropped after `Close()` leaves no in-chain audit marker** — `internal/audit/audit.go:894`. A request authorized during graceful shutdown (after `s.closed`, drainer exited) increments `dropped` but writes no `AUDIT_RECORDS_DROPPED` chain marker; the on-disk chain shows no gap. Acknowledged tradeoff, but "every decision is logged" is violated for the shutdown-race window with no in-log evidence.
- [ ] **Explicit `null` property subschema treated as "any"** — `pkg/enforcement/schema.go:107`. The loader accepts `properties: {x: null}`, yielding a nil subschema that passes any value/type for the declared property. (`{}` is correct JSON-Schema "any"; only the `null` form is anomalous.) Structural footgun, not a constraint bypass.

## Efficiency (per-request hot path)

- [ ] **`handleTimeWindow` re-parses RFC3339 strings every call** — `pkg/enforcement/handlers.go:196`. Two `time.Parse` scans per matching request over constraint fields fixed at manifest load. Pre-parse once at load (cf. `IPRangeCondition.Compile()`).
[...]

Issue #841, opened 2026-06-18
Track migration off gopkg.in/yaml.v3 (upstream archived/unmaintained)


eunolabs:
**Severity (v0.1.0 release audit): NICE-TO-HAVE (no current CVE)**

## Summary
`gopkg.in/yaml.v3 v3.0.1` parses the attacker-influenceable capability manifest and gateway config. The upstream `go-yaml/yaml` repo has been archived/sunset, so a future YAML-parser CVE would not get a v3 patch.

## Where
- `go.mod` — `gopkg.in/yaml.v3 v3.0.1`.

## Impact
No current advisory at v3.0.1 (the known yaml.v3 DoS CVEs were fixed before v3.0.1), and parsing is fail-closed, so this is not a v0.1.0 blocker. It's a forward-looking maintenance risk on a security-relevant parse path.

## Suggested action
No action for v0.1.0. Track a future migration to a maintained YAML library (e.g. `goccy/go-yaml`); the existing govulncheck CI gate will flag any new v3 advisory in the meantime.

_Found during a v0.1.0 release-readiness audit._

PR #1124, opened 2026-06-22
Fix session ri5k4k bugs

### LINKED ISSUE #1077: [rev-20260621T1605Z-747890] VerifyToken uses only headers[0].KeyID for key selection, silently failing multi-signature JWS tokens

### LINKED ISSUE #1103: [rev-20260622T0510Z-4a3673] JWTPDP.ValidateToken selects the signing key from only tok.Headers[0].KeyID — same single-header bug as #1077, in the IdP-JWT gateway path (uncovered by #1077's fix)
**Location:** `internal/pdp/jwt.go:423-426` (`JWTPDP.ValidateToken`)

**Reference:** https://github.com/eunolabs/eunox/blob/a75e21f99769832e5245bb6ab7f092b193d44729/internal/pdp/jwt.go#L423-L426

**Severity:** Low-Medium (correctness; practically limited to JWS JSON Serialization tokens, but this is the production IdP-JWT gateway auth path)

**Description:**

This is the sibling occurrence of the bug filed as #1077. That issue is scoped to the **native capability-token** path (`pkg/capability/jwks.go`, `JWKSClient.VerifyToken`), and its suggested fix only edits that file. The **IdP-JWT path** — the one the hosted gateway actually runs for every `Authorization: Bearer` request when `--jwks-uri` is configured — has the identical flaw and is left unfixed:

```go
tok, err := jwt.ParseSigned(tokenStr, capability.JWKSAlgorithms)
if err != nil {
    return ctx, fmt.Errorf("invalid JWT: %w", err)
}
if len(tok.Headers) == 0 {
    return ctx, fmt.Errorf("JWT has no headers")
}
kid := tok.Headers[0].KeyID   // <-- only the FIRST header's kid is consulted
```

`go-jose`'s `ParseSigned` accepts the JWS JSON Serialization form (`{"payload":..., "signatures":[{"protected":...}, {"protected":...}]}`), which carries one entry in `tok.Headers` per signature. Only `headers[0].KeyID` is passed into the shared `capability.VerifyWithKeyRotation`, so a multi-signature token whose valid signing key's `kid` lives in `headers[1]` (or later) fails key selection — a forced JWKS refresh still finds no match and the request is rejected, even though a later header identifies a key that would verify.

Because `VerifyWithKeyRotation` takes a single `kid` argument, the fix #1077 proposes for `jwks.go` does not propagate here; both call sites must be corrected independently. The two verification paths are deliberately kept separate (see the comments in both files about "the two cannot drift"), so fixing one and not the other re-introduces exactly the divergence those comments warn about.

**Trigger / repro:**

Configure the gateway with `--jwks-uri` (IdP-JWT PDP). Present a JWS JSON Serialization bearer token where:
- `signatures[0].protected.kid` = `"unknown-kid"` (not in the JWKS)
- `signatures[1].protected.kid` = `"valid-kid"` (in the JWKS, valid signature)

Expected: the token verifies against `valid-kid`. Actual: `ValidateToken` returns `invalid JWT` / `no matching key for kid "unknown-kid"` (HTTP 401) after a wasted forced JWKS refresh.

Note: standard compact bearer tokens (`header.payload.signature`) always have exactly one header, so the common path is unaffected — same practical-impact caveat as #1077.

**Suggested fix:**

Mirror whatever resolution #1077 lands for `jwks.go`: collect the distinct non-empty `kid`s across all `tok.Headers` and attempt verification against each (stopping at first success), or push the multi-header handling down into `capability.VerifyWithKeyRotation` so both call sites share it and cannot drift. Add a table-driven test under `internal/pdp/` with a two-signature JWS token whose valid `kid` is in `headers[1]`.

**Session:** rev-20260622T0510Z-4a3673

### LINKED ISSUE #1116: [rev-20260622T0610Z-f1be52] allowedExtensions NUL-truncation guard re-bypassed when a malformed percent-escape co-occurs (incomplete fix of #1060)
**Location:** `pkg/enforcement/handlers.go:772-789` (the `decodeErr` switch in `handleAllowedExtensions`, specifically the `errPathMalformedEscape` case at lines 784-789) and `pkg/enforcement/glob.go:38-49` (`decodePathForConfinement`)

**Reference:** https://github.com/eunolabs/eunox/blob/79a82778ea0fe4d058f7286bfbd7a2c2622b96de/pkg/enforcement/handlers.go#L772-L789

**Severity:** medium (security — extension-allowlist bypass / fail-open; same impact class as the now-closed #1060, exploitability depends on the upstream resolving NUL-truncated paths)

**Description:**

Issue #1060 closed the clean `evil.exe%00.csv` NUL-truncation bypass by having `decodePathForConfinement` reject an embedded NUL with `errPathNUL`. But that NUL check runs **only after a successful `url.PathUnescape`**:

```go
func decodePathForConfinement(value string) (string, error) {
	folded := strings.ReplaceAll(value, "\\", "/")
	decoded, err := url.PathUnescape(folded)
	if err != nil {
		return "", errPathMalformedEscape   // <-- returns BEFORE the NUL check below
	}
	decoded = strings.ReplaceAll(decoded, "\\", "/")
	if strings.IndexByte(decoded, 0) >= 0 {
		return "", errPathNUL
	}
	return decoded, nil
}
```

`handleAllowedExtensions` treats `errPathMalformedEscape` leniently — a literal `%` is a valid filename character for an upstream that does not percent-decode — and falls back to the **raw literal** form without re-checking for NUL:

```go
case errors.Is(decodeErr, errPathMalformedEscape):
	// Literal '%' that is not a valid escape: match the extension on the
	// separator-folded literal form (no percent-decode) ...
	normalizedPath = strings.ReplaceAll(filePath, "\\", "/")   // NUL, if present, survives here
}
```

So an attacker who includes **both** a malformed `%` escape **and** a NUL byte sidesteps the guard the handler's own comment promises ("An embedded NUL fails closed (deny)"). `url.PathUnescape` fails on the malformed escape → `errPathMalformedEscape` → the literal (NUL-bearing) path flows into `path.Ext`/`path.Base`/`strings.HasSuffix`, which treat NUL as an ordinary byte and match the **post**-NUL extension. A NUL-truncating upstream opens the **pre**-NUL file.

The NUL byte is reachable from a JSON-RPC argument via `""`, which decodes to a literal NUL in the Go string.

`MatchValueGlob` is **not** affected: it treats any error from `decodePathForConfinement` (including `errPathMalformedEscape`) as no-match → deny. The gap is specific to `handleAllowedExtensions`' lenient malformed-escape branch.

**Trigger / repro:** (decode + handler logic reproduced standalone at this SHA)

```
condition: allowedExtensions{argument: "path", extensions: [".csv"]}

arguments: {"path": "/data/secrets.dbx%.csv"}   => ALLOWED   (should deny)
```

Walkthrough: `url.PathUnescape` fails on the trailing `%.` → `errPathMalformedEscape` → `normalizedPath = "/data/secrets.db\x00x%.csv"` (NUL intact) → `path.Ext` = `.csv`, `path.Base` ends with `.csv` → admitted. A NUL-truncating upstream resolves `/data/secrets.db` (`.db`, not in the allowlist). The same holds for a trailing bare `%` or any other invalid escape paired with a NUL.

**Suggested fix:**

Check for NUL on the raw input *before* attempting `url.PathUnescape`, so an embedded NUL fails closed regardless of whether the escape decodes — and both callers stay in lock-step:

```go
func decodePathForConfinement(value string) (string, error) {
	folded := strings.ReplaceAll(value, "\\", "/")
	if strings.IndexByte(folded, 0) >= 0 {
		return "", errPathNUL          // NUL never legitimate; reject before any decode branch
	}
	decoded, err := url.PathUnescape(folded)
	if err != nil {
		return "", errPathMalformedEscape
	}
	decoded = strings.ReplaceAll(decoded, "\\", "/")
	if strings.IndexByte(decoded, 0) >= 0 {   // still catch %00 -> NUL
		return "", errPathNUL
	}
	return decoded, nil
}
```

This makes `errPathNUL` win over `errPathMalformedEscape` whenever a literal NUL is present, so `handleAllowedExtensions`' NUL branch fires and denies. (Equivalently, re-check `strings.IndexByte(normalizedPath, 0) >= 0` inside the handler's `errPathMalformedEscape` case, but fixing it at the shared decode keeps the two callers aligned.) Add a deny-test case pairing `"/data/secrets.dbx%.csv"` (and a literal `evil.exe\x00.csv%`) with allowlist `[.csv]`.

**Regression note:** introduced by the fix for #1060 (commit `4b8e986`, 2026-06-21), which added the lenient `errPathMalformedEscape` literal-fallback branch alongside the new NUL check — the fallback path bypasses the very check the same commit added. Existing tests cover the clean `%00` case (now denied) but not the malformed-escape + NUL combination, so the residual gap went uncaught.

**Session:** rev-20260622T0610Z-f1be52

### LINKED ISSUE #1118: [rev-20260622T0710Z-279b44] stdio→remote-HTTP drift probe binds session-start tools/list to --upstream-timeout, can fail startup the HTTP transport deliberately avoids
**Location:** `internal/transport/drift_probe.go:60-88` (StdioProxy probe), contrasted with `internal/transport/drift_probe.go:23-35` + `internal/transport/http_handlers.go:24-27` (httpSession probe)

**Reference:** https://github.com/eunolabs/eunox/blob/e254dee340b926c1b6c10f827e93aaa977806d71/internal/transport/drift_probe.go#L60-L88

**Severity:** medium

**Description:**

The session-start drift probe (`tools/list`) is bounded differently across the two transports, and the stdio→remote-HTTP path is bound to the wrong deadline.

- The **HTTP transport** probe deliberately runs `withoutUpstreamTimeout(ctx)` and is bounded by `sessionStartTimeout` (20s, `http_routing.go:155`), *not* by `--upstream-timeout`. Its own doc comment (`drift_probe.go:16-22`) explains why:

  > Binding it to the per-request latency knob would let a tight `--upstream-timeout` fail session establishment outright when the manifest pins descriptionHash (a fetch failure is fatal then), even though the upstream answers within the session-start budget.

- The **StdioProxy** probe, on its remote-HTTP-bridge branch, does the exact thing that comment warns against:

  ```go
  if p.upHTTP != nil {
      postCtx := ctx
      if p.upstreamTimeMs > 0 {
          postCtx, cancel = context.WithTimeout(ctx, msToDuration(p.upstreamTimeMs))
          defer cancel()
      }
      p.upHTTP.postWithCtx(postCtx, req)
  }
  ```

  `Start` passes the long-lived process `ctx` (no session-start deadline — see `stdio.go:221`), so on this branch `--upstream-timeout` is the *only* bound applied to the probe. (The subprocess branch applies no timeout at all and relies on pipe EOF, so the HTTP-bridge branch is the lone outlier among all three drift-probe paths.)

The result is a behavioral inconsistency with a concrete availability impact: the same upstream, the same manifest, and the same `--upstream-timeout` will start cleanly behind an HTTP host but fail to start behind a stdio host.

**Trigger / repro:**
1. Deploy `transport: stdio` with an upstream whose own transport is `http` (the `httpUpstream` bridge).
2. Set a tight per-call latency budget, e.g. `--upstream-timeout 2000` (a normal choice to keep request latency low).
3. Use a manifest that pins `descriptionHash` (so a drift-probe failure is **fatal**, per `MakeDriftCheck` / `driftProbeUnavailable`).
4. Point at an upstream whose cold-start `tools/list` takes longer than 2s (large tool catalog, cold cache, slow init).

The probe times out at 2s, `driftCheck` receives a `probeErr`, and the proxy refuses to start. The equivalent `transport: http` deployment (bounded by the 20s session-start budget, with extra headroom when `upstreamTimeMs` exceeds it — `http_routing.go:162-166`) starts fine, and so does a stdio→subprocess deployment (unbounded probe).

**Suggested fix:**
Make the stdio HTTP-bridge probe use the same posture as the httpSession probe: do not bind the session-start `tools/list` to `--upstream-timeout`. Either post under a dedicated session-start budget (mirroring `sessionStartTimeout`, with headroom when `upstreamTimeMs` is larger), or post the probe without the per-call deadline as `withoutUpstreamTimeout` does for httpSession. Update the `drift_probe.go:49-59` doc comment, which currently documents the per-call-timeout behavior as intended, to match. A test asserting that a tight `--upstream-timeout` does not fail a stdio→remote-HTTP startup whose upstream answers within the session-start budget would pin the fix (the existing coverage exercises subprocess and httpSession paths but not this one, which is why the divergence went unnoticed).

**Session:** rev-20260622T0710Z-279b44

### LINKED ISSUE #1121: Code review: slice GC hold-off in removeSub, dead profiles.go, stale ticket reference
## Summary

Engineering review of the full codebase. The codebase is in good health — no goroutine leaks, no data races, fail-closed invariants consistently enforced, concurrency handled correctly throughout. Three actionable items found; one is a real (if low-impact) GC bug.

---

## (a) Correctness — `removeSub` backing-array channel retention

**File:** `internal/transport/http_session.go:853`

```go
// current code
s.notifSubs = append(s.notifSubs[:i], s.notifSubs[i+1:]...)
```

`append` shifts elements left in-place. After the shift, `backing_array[len(result)]` still holds a reference to the channel that was at the *end* of the slice before the removal — not `ch` itself, but the last surviving channel, which becomes double-referenced. After all 8 subscribers remove themselves the entire backing array retains references to channel structs, blocking GC until the session is reaped by the idle reaper. With `maxSubsPerSession = 8` and idle-reaper-bounded session lifetimes this is low severity, but it is a real hold-off and the fix is the established Go idiom for this pattern.

**Fix:**

```go
n := len(s.notifSubs)
copy(s.notifSubs[i:], s.notifSubs[i+1:])
s.notifSubs[n-1] = nil // zero the stale reference before reslicing
s.notifSubs = s.notifSubs[:n-1]
```

---

## (b) Simplification — `profiles.go` is dead code

**File:** `cmd/eunox-mcp/profiles.go`

The file contains only the copyright header and `package main`. No declarations, no imports, nothing referenced anywhere in the binary. Delete it.

---

## (c) Polish — internal ticket reference in source comment

**File:** `cmd/eunox-mcp/redis.go:4`

```go
// Redis client construction for the eunox-mcp proxy (T-04).
```

`T-04` is an internal ticket reference. CONTRIBUTING/CLAUDE.md is explicit: *"Never reference issue numbers, bug IDs, or internal milestones in code comments — those references rot and belong in the PR description or commit message only."*

**Fix:** remove `(T-04)` from the file-level comment.

---

## Not actionable (verified against source)

These were raised during review but confirmed to be non-issues:

| Claim | Reality |
|---|---|
| `lock_other.go` silently succeeds on unsupported platforms | Returns a descriptive error (fail closed). Lines 18–20. |
| `pingRedis` error ignored | Checked at both call sites: `main.go:552` and `main.go:1766`. |
| Callcounter unbounded per-key slice growth | Bounded by `limit` itself; denied calls record no timestamp. Documented at `memory.go:293`. |
| No timeout on upstream `initUpstream` | `sessionStartTimeout = 20 s` at `http_routing.go:155`, applied via `context.WithTimeout` at line 167. |
| SSE subscribers unbounded | Capped at `maxSubsPerSession = 8` with non-blocking sends and drop counter. |
| `expandEnvRefs` recursive expansion | Regex-substituted from `os.Getenv`; no recursive step. |

---

## Proposed scope

All three items together are under 15 lines changed. Suggested order: (a) → (b) → (c), single PR.

eunolabs:
Batch fix of open review-session bug issues, landed in one PR.

## Bugs addressed

- Closes #1116 — **allowedExtensions NUL guard re-bypassed by a co-occurring malformed percent-escape.** `decodePathForConfinement` only checked for an embedded NUL *after* a successful `url.PathUnescape`, so a path carrying both a malformed `%` escape and a literal NUL short-circuited to `errPathMalformedEscape`; `handleAllowedExtensions`' lenient malformed-escape fallback then matched the NUL-bearing literal, admitting a path a NUL-truncating upstream opens as a different (pre-NUL) file. Now NUL is checked on the raw input *before* any decode branch so `errPathNUL` wins, keeping both callers (`handleAllowedExtensions` + `MatchValueGlob`) in lock-step. Added deny test for the malformed-escape + NUL combination.

- Closes #1118 — **stdio→remote-HTTP drift probe bound to `--upstream-timeout`.** The session-start `tools/list` probe on the `StdioProxy` remote-HTTP bridge was bounded by `--upstream-timeout`, so a tight per-call budget could fail startup outright when the manifest pins `descriptionHash` (a probe failure is fatal then) — even though the upstream answers within the session-start budget. The `httpSession` probe deliberately avoids this via `withoutUpstreamTimeout`, so the same upstream/manifest/timeout started behind an HTTP host but not a stdio one. The bridge probe now uses the session-start budget (the larger of `sessionStartTimeout` and `--upstream-timeout`); `sessionStartTimeout` is promoted to a shared package const. Added a regression test.

- Closes #1121 — **slice GC hold-off in `removeSub`, dead `profiles.go`, stale ticket references.** `removeSub` now shifts-and-zeros the stale tail slot before reslicing so the last surviving channel isn't double-referenced in the backing array (held the session off GC until the idle reaper ran). Deleted the empty `cmd/eunox-mcp/profiles.go`. Removed internal ticket references (`T-04`, and — same category, found by sweep — `T-02`/`T-01`) from file-level comments in `redis.go`, `jwt.go`, and `http_remote.go`.

- Closes #1077, Closes #1103 — **JWT signing-key selected from `headers[0]` only, in both verification paths.** Both `JWKSClient.VerifyToken` and `JWTPDP.ValidateToken` hard-coded `tok.Headers[0].KeyID` and applied `requireKID` to the first header only. Replaced with a shared `capability.CandidateKIDs` (distinct kids across all headers, `requireKID` enforced across all of them) fanned through a new `capability.VerifyWithKeyRotationMultiKID`, so the two paths share multi-header handling and cannot drift, and the `headers[0]` foot-gun is gone. **Reachability note:** these issues were filed against go-jose v3, where `jwt.ParseSigned` accepted JWS JSON Serialization (multiple headers). The codebase has since moved to go-jose v4, whose `jwt.ParseSigned` is compact-only (`jose.ParseSignedCompact`), so a parsed token carries exactly one header today and the multi-signature scenario is **not reachable** through the production parse path. The change is therefore hardening + the cross-header `requireKID` correctness, behaviorally unchanged for the single-header (only reachable) case. Added a `CandidateKIDs` table test.

### Already resolved (no change)

- #1079 (misplaced `IPRangeCondition.Compile()`) is already fixed on `main` — `Compile()` is in the pointer case, and a `Networks()`-after-`LoadManifest` regression test already exists.

### Out of scope (deliberately not closed)

- #1109 is an umbrella tracking issue of lower-severity / documented-limitation items; partial fixes wouldn't close it and several entries are acknowledged tradeoffs.
- #841 (migrate off archived `yaml.v3`) explicitly says "No action for v0.1.0".
- #836 (remote-HTTP server-initiated requests) needs a product decision (support vs. document) before code.

## Testing

- `gofmt -l` clean; `go vet` clean on all touched packages; `make check-license` passes.
- New/updated tests: malformed-escape + NUL deny case (`pkg/enforcement`), stdio→remote-HTTP tight-timeout probe regression (`internal/transport`), `CandidateKIDs` contract table test (`pkg/capability`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01WLzAcC5RZY2wLr7LtK6MtU)_


internal/transport/http_remote.go
```diff
@@ -1,7 +1,7 @@
 // Copyright 2026 Eunolabs, LLC
 // SPDX-License-Identifier: Apache-2.0
 
-// Remote HTTP upstream support for the MCP proxy (T-01: --upstream-url).
+// Remote HTTP upstream support for the MCP proxy (--upstream-url).
```
eunolabs, line 4: Good cleanup dropping the `T-01` ref here, but the sweep missed the sibling test file. `internal/transport/http_remote_test.go:4` still reads:

```go
// Tests for the remote HTTP upstream mode (T-01: --upstream-url).
```

Same `(T-01: --upstream-url)` pattern you removed from this line. Worth deleting for consistency with the PR's stated sweep and the repo convention against internal ticket refs in comments (it's the only remaining `T-NN` hit in the tree). One-line fix.

---

pkg/capability/jwks.go
```diff
@@ -160,6 +160,75 @@ func (e *terminalError) Unwrap() error { return e.err }
...
+	kids []string,
+	verify func(key *jose.JSONWebKey) (result *T, err error),
+) (*T, error) {
+	if len(kids) == 0 {
+		// Caller contract violation; fail closed rather than verify against nothing.
+		return nil, fmt.Errorf("no kid candidates to verify against (fail closed)")
+	}
+	var lastErr error
+	for _, kid := range kids {
+		result, err := VerifyWithKeyRotation(ctx, cache, kid, verify)
```
eunolabs, line 228: Two low-priority notes on the new fan-out helper. Both are non-blocking — the multi-header path is unreachable today (compact-only parse ⇒ one kid ⇒ this loop runs once), exactly as the PR states.

1. **Terminal errors aren't short-circuited across kids.** `VerifyWithKeyRotation` distinguishes a `Terminal` error (signature verified, claims invalid — documented as "never retried against other keys or a refreshed set") from a plain signature failure. This wrapper collapses that distinction: on any non-success it just records `lastErr` and moves to the next kid. For a genuinely multi-signature token, a claims failure (e.g. expired) under the first kid would be retried under the rest and ultimately surface the *last* kid's signature error, masking the real terminal reason. Moot for the single reachable kid, and a multi-sig JWS shares one payload so claims validity can't change across keys — but if this ever becomes the seam for a reachable multi-sig path, consider returning immediately when `VerifyWithKeyRotation` yields a terminal error.

2. **No direct test for the fan-out.** `TestCandidateKIDs` covers kid extraction thoroughly, but `VerifyWithKeyRotationMultiKID` itself isn't exercised directly (first-kid-fails-then-second-succeeds, all-fail-returns-last-err, empty-kids-fails-closed). It's new exported API on the security-critical verify path; a small table test would lock the contract against future edits.

For what it's worth, I confirmed this wrapper can't fail open: `VerifyWithKeyRotation` never returns `(nil, nil)` (the `tryKeys` closure converts that case to an explicit error), so the fallthrough `lastErr` is always non-nil when no kid succeeds.

---

eunolabs (commented):
## Review: batch bug-fix (`ri5k4k`)

Reviewed all five fixes against `e18e0c4`. Checked the branch out locally and ran `go build ./...`, `go vet`, `gofmt -l`, and `go test -race -count=1` on the touched packages (`pkg/enforcement`, `pkg/capability`, `internal/transport`, `internal/pdp`) — all green — and CI is passing across all 41 checks (`mergeable_state: unstable` is not a failing check here).

Overall a clean, well-scoped batch. The fixes are correct, the fail-closed posture is preserved throughout, and the JWT change is a genuine de-duplication rather than a patch. One convention nit and two optional follow-ups (all inline).

### Fix-by-fix

- **#1116 — NUL guard moved before decode (`pkg/enforcement/glob.go`)** — Correct. Checking `IndexByte(folded, 0)` ahead of `url.PathUnescape` makes `errPathNUL` win over `errPathMalformedEscape`, closing the lenient malformed-escape fallback in `handleAllowedExtensions`. Backslash-folding can't add/remove a NUL, so testing `folded` is equivalent to testing the raw input; the post-decode `%00` check is correctly retained, and both callers (`handleAllowedExtensions` + `MatchValueGlob`) stay in lock-step. Deny test added and passing.

- **#1118 — drift probe no longer bound to `--upstream-timeout` (`internal/transport/drift_probe.go`)** — Correct. The session-start probe now uses `max(sessionStartTimeout, --upstream-timeout)`, matching the `httpSession` probe's `withoutUpstreamTimeout` posture so the same upstream/manifest/timeout behaves identically behind stdio and HTTP hosts. The per-page context + `defer cancel()` sit inside the `FetchAllToolPages` closure, so each page gets its own deadline with no leak. Promoting `sessionStartTimeout` to a shared package const (and dropping the local one in `http_routing.go`) is clean — `time` is still used there, so no dangling import. Regression test added.

- **#1121 — `removeSub` GC hold-off + dead file + ticket refs (`internal/transport/http_session.go`)** — The shift-zero-reslice is the textbook leak-free slice delete. `profiles.go` is gone. One missed ticket ref noted inline.

- **#1077 / #1103 — JWT signing key across all headers (`pkg/capability/jwks.go`, `internal/pdp/jwt.go`)** — Correct, and the shared `CandidateKIDs` + `VerifyWithKeyRotationMultiKID` seam means the `headers[0]` foot-gun and the first-header-only `requireKID` are both gone and the two verify paths can't drift. I verified there's no fail-open: `VerifyWithKeyRotation` can never return `(nil, nil)`, so the wrapper's fallthrough error is always non-nil. The reachability caveat (go-jose v4 is compact-only ⇒ single header) is accurate, so this is hardening + the cross-header `requireKID` correctness, behaviorally unchanged for the reachable single-kid case. Two optional follow-ups inline.

### Findings

1. **(nit) Incomplete ticket-ref sweep** — `internal/transport/http_remote_test.go:4` still carries `(T-01: --upstream-url)`, the same pattern removed from `http_remote.go`. Inline.
2. **(optional) Multi-kid terminal-error short-circuit + a direct test for `VerifyWithKeyRotationMultiKID`** — inline on `jwks.go`.

No security regressions: every path stays fail-closed and the deny/audit semantics are unchanged. The nit is the only thing I'd ask for before merge; the other two are nice-to-haves.

---

